PR #2, opened 2016-02-21
Remove now default 'sudo: false'

frewsxcv:
`sudo: false` is now default on Travis CI


tomaka:
I didn't setup a license yet, but I guess that doesn't really matter for just a one-line change?

PR #22, opened 2016-04-04
Fix linux library name

Ben-Alderson:
Currently the code searches for "libvulkan-1.so", but vulkan library (at least on my system) is called "libvulkan.so.1".


tomaka:
Thanks!

PR #29, opened 2016-04-14
Make DeviceMemory and MappedDeviceMemory generic over the pointer to …

tomaka:
…Device


tomaka:
I went with using a custom trait named `SafeDeref` to guarantee that the device was always the same.

Issue #20, opened 2016-04-03
About OomError


tomaka:
Most functions in Vulkan can return `VK_ERROR_OUT_OF_HOST_MEMORY` and `VK_ERROR_OUT_OF_DEVICE_MEMORY`. In fact most functions can only return one of these two errors.

This is reflected in vulkano by the fact that a lot of functions can return `OomError`.

However in practice it is very unlikely that an out-of-memory error actually happens, with the exception of `vkAllocateMemory`.

The API would be much more convenient to use if an out of memory error returned by Vulkan resulted in a panic instead of returning an error, like it's the case for the core Rust.
Again, with the exception of `DeviceMemory` which would still return `OomError`.

tomaka:
(this issue is for discussion about the problem, not something that is currently planned)

Mixthos:
If a the device memory is filled with cached textures/meshes and an important command returns `VK_ERROR_OUT_OF_DEVICE_MEMORY`, you would probably want to release some textures/meshes and repeat the command. Is this something that vulkano could handle?

tomaka:
@Mixthos Currently yes, but if `OomError` is turned into panicking like this issue suggests, then no.
The reasoning is that the chance that anything returns OUT_OF_DEVICE_MEMORY are very low if you except the commands that allocate memory.

For example why would creating a fence or a semaphore return out of memory? A fence or a semaphore is usually just an integer.

Mixthos:
I guess it depends on what the drivers do. Maybe we could test this for each function before changing it? Preferably in most drivers, in case one of the drivers has some weird behaviour.

tomaka:
Well, drivers are more or less non-existing today. The ones that exist don't do many optimizations. We can't assume that if it works today it will likely work tomorrow.

Mixthos:
The only added convenience is not having to call `unwrap()`, right? In that case I wouldn't want to give up the ability to recover from an error in case it does happen at some point in time. Otherwise there's no way to easily implement a patch to the application if it crashes on a new driver.

Mixthos:
Or there could be two versions of each of these functions. For Example:

```rust
fn newSafe(device: &Arc<Device>) -> Result<Arc<Fence>, OomError>;
fn new(device: &Arc<Device>) -> Arc<Fence> {
    Fence::newSafe(device).unwrap()
}
```

tomaka:
That looks like a good idea.
This would be similar to the difference between `[]` and `.get()` in the stdlib for example.

[...]

PR #32, opened 2016-04-14
Optional OomErrors

Mixthos:
(Mostly) as discussed in #20.
`SwapchainImage::from_raw` isn't changed yet.
I fixed the teapot example to work with my changes. The other examples were broken anyway.
I'm not confident about my ability to change stuff inside these huge macros without breaking them, so you should probably double-check them. ;)

### DIFF
--- a/vulkano/src/framebuffer/sys.rs
+++ b/vulkano/src/framebuffer/sys.rs
@@ -34,24 +34,7 @@ pub struct UnsafeRenderPass {
 }
 
 impl UnsafeRenderPass {
-    /// Builds a new renderpass.
-    ///
-    /// This function calls the methods of the `Layout` implementation and builds the
-    /// corresponding Vulkan object.
-    ///
-    /// # Safety
-    ///
-    /// This function doesn't check whether all the restrictions in the attachments, passes and
-    /// passes dependencies were enforced.
-    ///
-    /// See the documentation of the structs of this module for more info about these restrictions.
-    ///
-    /// # Panic
-    ///
-    /// Can panick if it detects some violations in the restrictions. Only unexpensive checks are
-    /// performed. `debug_assert!` is used, so some restrictions are only checked in debug
-    /// mode.
-    ///
+    /// See the docs of new().
     pub unsafe fn raw<Ia, Ip, Id>(device: &Arc<Device>, attachments: Ia, passes: Ip,
                                   pass_dependencies: Id)
                -> Result<UnsafeRenderPass, OomError>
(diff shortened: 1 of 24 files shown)


vulkano/src/image/sys.rs
```diff
@@ -479,6 +479,12 @@ impl UnsafeImage {
     }
 
     #[inline]
+    pub unsafe fn bind_memory(&self, memory: &DeviceMemory, offset: usize)
```
tomaka: `bind_memory` doesn't need a `raw` variant. It's an unsafe and dangerous function which isn't supposed to be used directly, so making it more convenient to use doesn't make sense.

vulkano/src/framebuffer/sys.rs
```diff
@@ -52,7 +52,7 @@ impl UnsafeRenderPass {
     /// performed. `debug_assert!` is used, so some restrictions are only checked in debug
     /// mode.
     ///
-    pub unsafe fn new<Ia, Ip, Id>(device: &Arc<Device>, attachments: Ia, passes: Ip,
+    pub unsafe fn raw<Ia, Ip, Id>(device: &Arc<Device>, attachments: Ia, passes: Ip,
```
tomaka: Since the documentation is so long, could you remove the docs from `raw` and just say `See the docs of new()`?

tomaka:
Thanks! That's very nice!

About your comments on the fact that Arcs can be omitted, I'm really not sure. It depends on how semaphores/fences are going to be pooled, whose design is currently blocked by some ambiguities in the Vulkan specs.

Did you try running `cargo test --lib` to see if all the tests compile?

Mixthos:
> Did you try running cargo test --lib to see if all the tests compile?

Oops! I knew I forgot _something_. I'll fix those now.

vulkano/src/image/sys.rs
Mixthos: I reverted it.

Mixthos:
> bind_memory doesn't need a raw variant. It's an unsafe and dangerous function which isn't supposed to be used directly, so making it more convenient to use doesn't make sense.

I reverted it in cebabe1f639c45037cb8d40d2a5ca3a88ecba722.

> Since the documentation is so long, could you remove the docs from raw and just say See the docs of new()?
> This would avoid having to maintain the same docs twice.

I replaced all docs for the "raw" functions in b8269ead29d1b30af09ed8eaddea1095b62f1183.

> Did you try running cargo test --lib to see if all the tests compile?

I fixed them in b6b073c888dbb257d2c904572aabfb5d2e69a148:

> test result: ok. 64 passed; 0 failed; 6 ignored; 0 measured

Did I forget anything?

tomaka:
Thanks!

PR #33, opened 2016-04-15
[WIP] Rework command buffers

tomaka:


tomaka:
The reason why this is blocked is that I realized that I have to delay submitting commands to the command buffer.

In other words, in the low-level code each command currently does two things:
1) Verify that the command is correct.
2) Add the command to the command buffer.

Meanwhile the high-level wrapper needs:
1) Verify that the command is correct.
2) Add the command to a queue that will be submitted layer.

Which means that I have to adjust the code of the low-level wrapper to avoid having to verify each command twice. However I'm unsure how to handle at the same time being overhead-free, allowing delaying commands, and being safe.

tomaka:
Status of this PR: since I'd like to avoiding adding too many TODOs and FIXMEs in the code, I need to fix *a lot* of stuff beforehand in regards to buffers and images.
If I included everything in a single PR, it would be too big. Instead I'm going to submit several PRs beforehand and rebase this one over them.

tomaka:
Blocked on #54 / #103

tomaka:
Continuing without #54 / #103

tomaka:
This PR is dead. It's too complicated and is heading the wrong way.
I'm leaving it open for a few days/weeks so that I can keep the code (if I close it now I'll later forget to delete the branch).

Going for #153 instead.

PR #36, opened 2016-04-15
Fix examples so that cargo test works

tomaka:


tomaka:
Travis fails since the examples are specific to Windows :-/

Issue #38, opened 2016-04-15
Provide a way to query supported instance extensions


Mixthos:
The Vulkan spec lists the function `vkEnumerateInstanceExtensionProperties` with which you can query for supported extensions. A function like `InstanceExtensions::supported()` which returns supported extensions would be useful. Should I implement it?

Mixthos:
It's probably better as `Instance::supported_extensions()` to keep Vulkan calls localized.
Also: `vkEnumerateInstanceLayerProperties` can be used for `Instance::supported_layers()`.
For clarification: Neither function needs a vulkan instance.

PR #39, opened 2016-04-16
InstanceExtensions::supported()

Mixthos:
As suggested in #38.

I wanted it for [vulkano-info](https://github.com/Mixthos/vulkano-info), so I didn't wait for a reply. Sorry. ;)

`cargo test --lib` crashed without panicking, but that's likely because I had to switch to an NVIDIA GPU. :/

### DIFF
--- a/vulkano/src/instance/extensions.rs
+++ b/vulkano/src/instance/extensions.rs
@@ -53,8 +53,8 @@ macro_rules! instance_extensions {
         }
         
         impl $sname {
-            /// See the docs of supported().
-            pub fn supported_raw() -> Result<$sname, OomError> {
+            /// See the docs of supported_by_core().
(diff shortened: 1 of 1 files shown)


tomaka:
I think the method name should reflect the fact that these are only the extensions supported by the core driver and not the .

Mixthos:
> I think the method name should reflect the fact that these are only the extensions supported by the core driver and not the .

Did you mean "not by layers"? Maybe `supported_by_core()`?

tomaka:
Yes for both.

Mixthos:
I renamed it.

There should also be ways to query the extensions supported by layers and the layers and extensions supported by a physical device. Maybe this requires a rework of the layers and extensions modules?

Issue #77, opened 2016-05-01
Spelling issue in triangle example


kazimuth:
["Only", not "Onle"](https://github.com/tomaka/vulkano/blob/master/examples/src/bin/triangle.rs#L259).

PR #80, opened 2016-05-03
Fix spelling errors in triangle example

magnontobi:
Just read through the triangle example and while I'm stupid and didn't really understand (still lacking in both rust and graphics pipeline understanding), I did find some spelling errors I thought were worth fixing.
Also, I checked issues and saw #77 at the top, which would obviously be fixed with this PR.


tomaka:
Thanks!

PR #91, opened 2016-05-08
Let user choose descriptor pool capacity

tomaka:


tomaka:
I'm not satisfied with this system.
In my opinion vulkano should provide some sort of automated way to handle descriptor pools.

PR #96, opened 2016-05-11
Add ShaderStages and AccessFlagBits

tomaka:
Will be needed later.


tomaka:
For some reason travis doesn't pick the branch. Since I tested it locally, merging.

PR #114, opened 2016-05-23
Window creation: Add support for wayland on unixes

nicokoch:
On unix, we try to get the `wayland_display` and `wayland_surface`. If this fails, we fall back to X11.

There's also some rustfmt formatting contained here, hope it doesn't bother you too much :/

Since the winit version which exposes the wayland functions has not been publieshed yet, we depend on the git version of winit. You could also publish winit before merging and I will change the `Cargo.toml` files to depend on 0.5.1.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -9,7 +9,7 @@ vulkano = { path = "../vulkano" }
 vulkano-win = { path = "../vulkano-win" }
 cgmath = "0.7.0"
 image = "0.6.1"
-winit = { git = "https://github.com/tomaka/winit"  }
+winit = "0.5.1"
 
 [build-dependencies]
 vk-sys = { path = "../vk-sys" }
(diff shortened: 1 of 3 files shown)


tomaka:
> Since the winit version which exposes the wayland functions has not been publieshed yet, we depend on the git version of winit. You could also publish winit before merging and I will change the Cargo.toml files to depend on 0.5.1.

That's what I'm going to do.
I have to leave in 5mn though. Please ping me if in 24hours if I forgot.

tomaka:
I don't really like the rustfmt formatting, but oh well :-/

nicokoch:
@tomaka I just updated this PR.

I also have some general notes on vulkan linux support, if you're interested:

As always, the open source drivers are in a pretty bad shape compared to the windows drivers.
The intel driver does not support `shaderCullDistance` and `shaderClipDistance` yet, so you cannot even run the triangle example.
Window creation is also kind of a problem, because the intel driver can only deal with wayland and xcb windows for now, but not xlib (this was the motivation for this PR in the first place).

On the plus side, at least for simple tasks, linux performance is on par with windows, which is kind of a huge deal for graphics stuff

tomaka:
> The intel driver does not support shaderCullDistance and shaderClipDistance yet, so you cannot even run the triangle example.

That's actually an annoying problem with glslang which adds these capabilities in the SPIR-V output even when they are not needed.

PR #133, opened 2016-06-12
Added StorageImage as a new descriptor for the pipeline_layout macro.

attackgoat:
Thanks - This approach will work for me. It would be nice to be able to pass in a struct of defaults I want to set for DescriptorImageDesc and make use of the Rust .. syntax to add in ::empy() values I don't care about, but I now sure how to cleanly add that here yet.


tomaka:
It's just a matter of API.

You're right that it should be possible to pass a DescriptorDesc, as it will be checked anyway.

PR #138, opened 2016-06-14
New compute example

attackgoat:
Hi again - I've built an example using your API - it does compute properly and I can pass/return CPU buffers but I am having trouble with the storage image type: the shader will call GLSL imageSize() but the results are uninitialized values and vary each execution.

I am not yet sure if this is a misuse of the API or an issue with the beta NVIDIA drivers I'm using to run Vulkan.

If you see anything simple I'm doing wrong I would love to be able to add this to your codebase - I'm continuing to debug it.

### DIFF
--- a/examples/build.rs
+++ b/examples/build.rs
@@ -3,6 +3,8 @@ extern crate vulkano_shaders;
 fn main() {
     // building the shaders used in the examples
     vulkano_shaders::build_glsl_shaders([
+        ("src/bin/compute_buffer.glsl", vulkano_shaders::ShaderType::Compute),
+        ("src/bin/compute_image.glsl", vulkano_shaders::ShaderType::Compute),
         ("src/bin/triangle_vs.glsl", vulkano_shaders::ShaderType::Vertex),
         ("src/bin/triangle_fs.glsl", vulkano_shaders::ShaderType::Fragment),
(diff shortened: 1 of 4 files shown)


tomaka:
Nice!

I guess you used an `AttachmentImage` because it was the most appropriate type. However an `AttachmentImage` isn't supposed to be used for anything else than an attachment. The unsafe code knows that and forces the layout of the image to be `COLOR_OPTIMAL`, which is an invalid use of Vulkan.
As a hack, you could try change [this line](https://github.com/tomaka/vulkano/blob/master/vulkano/src/image/attachment.rs#L183-L184) to `Layout::General` to see if it works.

Instead the proper way it should be done is by creating a new type named `StorageImage`.

attackgoat:
That's exactly what I tried doing - but I had problems with my implementation: `MutableImage`. AttachmentImage seemed close enough but no dice. Based on your comments I think I'll investigate that route more now.

tomaka:
The other problem you're probably encountering is this: https://github.com/tomaka/vulkano/blob/master/vulkano/src/descriptor/descriptor_set/pool.rs#L40-L62
You should add a `STORAGE_IMAGE` entry.

attackgoat:
Closing this and my other pull so I can work my changes into branches.

PR #141, opened 2016-06-15
Added debugging example code

attackgoat:
I'm still having trouble with your new storage image implementation (thank you!) so I've decided to get better at debugging Vulkan. The steps to get this going, while simple, took a few stabs to get correct and so I've put together a small example for others to follow.

I also added some helper and formatter code to to the instance/debug module.

### DIFF
--- a/examples/src/bin/debug.rs
+++ b/examples/src/bin/debug.rs
@@ -1,0 +1,83 @@
+// Copyright (c) 2016 The vulkano developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or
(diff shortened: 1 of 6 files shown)


attackgoat:
The last change here removed a vulkan debugger error:
I *think* offset & size are relative to the buffer and not ranges within greater memory. This fix is not optimal as it just locks the whole buffer because I don't have the 'actual' buffer offset that should be used.

tomaka:
Thanks, but could you split all these changes in multiple PRs?

> I think offset & size are relative to the buffer and not ranges within greater memory.

That's something that they recently fixed in the vulkan specs. I already fixed it in the `cb` branch which I intend to merge soon.

PR #145, opened 2016-06-16
Added debug example code and required changes to vulkano debug module

attackgoat:
Putting changes into branches; I hadn't realized how messy it had gotten with the array of things I had found.

### DIFF
--- a/examples/src/bin/debug.rs
+++ b/examples/src/bin/debug.rs
@@ -1,0 +1,109 @@
+// Copyright (c) 2016 The vulkano developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>,
+// at your option. All files in the project carrying such
+// notice may not be copied, modified, or distributed except
+// according to those terms.
+
+#[macro_use]
+extern crate vulkano;
+
+use vulkano::device::{Device, DeviceExtensions};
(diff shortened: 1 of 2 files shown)


tomaka:
I've been hesitating about this PR because I'm not sure if an example demonstrating debug callbacks is really useful.

The example assumes that the user has the SDK installed, which is not mandatory to use vulkano. I also don't want user to get the feeling that they need to have the SDK installed.

attackgoat:
I could add a note about installing the SDK in order to get the default layers; but I think Vulkan in general needs some very basic example code.

It took me time to setup these callbacks because although I read the documentation I hadn't properly setup the extensions as well; this required reading your code to convert from the regular Vulkan constant to the Vulkano version. Until I got those steps correct I had code that built but did not log; which was frustrating.

IMHO setting up a debug-ready codebase on a new platform is extremely helpful, because seasoned programmers we are all new to Vulkan.

nicokoch:
I actually really like this for the reasons @attackgoat mentioned.

tomaka:
Ok, since there are 3 people in favor of it, let's merge this.

@attackgoat Please remove the implementation of `Display` for `MessageTypes` and use `{:?}` in the println instead. Once it's done I'll merge.

attackgoat:
Completed; I'd like to get back to the compute shader example too and see how the newly baked code is doing. Thanks!

tomaka:
Also, just do `let _debug_callback = ...` which removes the warning, and remove the last line that you added in your latest commit.

attackgoat:
Awesome, didn't know that _ feature worked as a prefix in Rust.

PR #155, opened 2016-06-24
Some documentation for samplers

tomaka:
The `sampler` module is more or less finished now.

### DIFF
--- a/vulkano/src/sampler.rs
+++ b/vulkano/src/sampler.rs
@@ -9,9 +9,38 @@
 
 //! How to retreive data from an image within a shader.
 //!
+//! When you retreive data from an image, you have to pass the coordinates of the pixel you want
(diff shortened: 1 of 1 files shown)


vulkano/src/sampler.rs
```diff
@@ -9,10 +9,39 @@
 
 //! How to retreive data from an image within a shader.
 //!
+//! When you retreive data from an image, you have to pass the coordinates of the pixel you want
```
jonas-schievink: retrieve

vulkano/src/sampler.rs
```diff
@@ -9,10 +9,39 @@
...
+//! to retreive. The implementation then performs various calculations, and these operations are
```
jonas-schievink: Ditto

tomaka:
Thanks. I don't know where I got that habit of writing `retreive`.

PR #162, opened 2016-06-29
[RFC] Draft for StdCommandBufferBuilder

tomaka:
Here's the draft for the new design for command buffers.

Instead of using an opaque giant structure with several vecs and hashmaps like it is currently the case, the idea here is to have an iterator-like system where each command adds a wrapper around the previous ones.

For example you start with `StdPrimaryCommandBufferBuilder::new(...)`, which gives you a `StdPrimaryCommandBufferBuilder`. You then call `update_buffer()` on it (through the `StdCommandBufferBuilder` trait that is similar to `Iterator`), which will yield you a `StdUpdateBufferBuilder<StdPrimaryCommandBufferBuilder, ...>` (the second template parameter being the type of data). You then call `build()` to finish the construction, and it will yield you a `StdUpdateBuffer<StdPrimaryCommandBuffer, ...>`.

Of course these types are not supposed to be expressed directly. The idea is that a if you want to express the type of a command buffer builder or a command buffer, you should use a `Box<Trait>` or an `Arc<Trait>` like you would do with an iterator today. Maybe in the future Rust will provide a syntax to make this easier. Since iterators and closures have the same problem, it is likely to happen.

This new approach has several advantages:

- No allocation. Right now a command buffer builder has several Vecs and HashMaps. In the new design, all their contents will be put in regular struct members instead.
- It will be possible to use something else than `Arc`s to keep objects alive. For example in the future you should be able to pass a simple reference to a buffer to a command buffer instead of an `Arc<Buffer>`.
- Less indirections. Since everything is a regular struct member and trait objects will no longer be used, this means that more inlining and more optimizations are possible by the compiler.
- No more distinction between `PrimaryCommandBufferBuilder`, `SecondaryGraphics...` and `SecondaryCompute...`. Instead it will be done through traits.
- The "update buffer" command no longer needs to do an additional copy of the data.

The drawbacks are:

- The commands are effectively added only when you call `build`. This means that adding commands will take very few time (only validations are performed), but calling `build` will take a long time. I don't think it's that much of a problem though, as the total time remains the same.
- The `StdPrimaryCommandBufferBuilder` trait will need to be imported by the user in order to add commands.

### DIFF
--- a/vulkano/src/command_buffer/mod.rs
+++ b/vulkano/src/command_buffer/mod.rs
@@ -52,11 +52,17 @@ pub use self::outer::SecondaryGraphicsCommandBufferBuilder;
 pub use self::outer::SecondaryGraphicsCommandBuffer;
 pub use self::outer::SecondaryComputeCommandBufferBuilder;
 pub use self::outer::SecondaryComputeCommandBuffer;
+pub use self::standard::primary::StdPrimaryCommandBufferBuilder;
+//pub use self::standard::primary::StdPrimaryCommandBufferBuilder as PrimaryCommandBufferBuilder;
 
-mod inner;
-mod outer;
+use self::pool::CommandPool;
+use self::sys::UnsafeCommandBuffer;
+
+mod inner;      // TODO: remove
+mod outer;      // TODO: remove
 
 pub mod pool;
+pub mod standard;
 pub mod sys;
 
 #[repr(C)]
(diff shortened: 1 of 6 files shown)


tomaka:
Another advantage of this system is that the various implementations can be specialized to be optimized.

For example if you call:

```rust
cb.copy_buffers(buf1.slice(0 .. 1), buf2.slice(0 .. 1))
  .copy_buffers(buf1.slice(0 .. 1), buf2.slice(1 .. 2))
  .copy_buffers(buf1.slice(0 .. 1), buf2.slice(2 .. 3))
```

It can be optimized to just a single copy by specializing the `copy_buffers` method on `StdCopyBuffersBuilder`.

tomaka:
Still not satisfactory as writes-after-reads can't be handled properly with this system. Needs some tweaks.

tomaka:
Superceded by #216

Issue #163, opened 2016-07-01
libvulkan.so: cannot open shared object file: No such file or directory


JustMaximumPower:
Vulkano tries to load the vulkan system library on unix by looking up libvulkan.so. This library does not exist on Ubuntu (after installing vulkan-utils) or in the steam runtime. The library is called libvulkan.so.1 (which is a symlink to libvulkan.so.1.0.8. ). One could create a symlink libvulkan.so -> libvulkan.so.1 but I think there is no reason not to just load libvulkan.so.1

see #164

JustMaximumPower:
This was fixed by merging 164.

PR #164, opened 2016-07-01
Change vulkan library name on unix

JustMaximumPower:
libvulkan.so does not exist in ubuntu or steam for linux. There is
however a libvulkan.so.1 . Using this vulkano works on a default install.

see #163


tomaka:
In #41 I changed that to `libvulkan.so` because Android doesn't have `libvulkan.so.1`.

Could you do a little tweak so that it looks for `libvulkan.so` on Android? With `#[cfg(target_os = "android")]` and `#[cfg(all(unix, not(target_os = "android")))]`.

JustMaximumPower:
Unfortunately I have neither a windows nor an android system to test the other branches.

tomaka:
Thanks, should be good.

PR #178, opened 2016-07-20
Add UnsafeCommandBufferBuilder::clear_color_image

tomaka:


tomaka:
And `clear_depth_stencil_image` while we're at it.

PR #185, opened 2016-07-21
Use XCB surface to be compatible with Nvidia drivers

ColinKinloch:
CreateXlibSurfaceKHR doesn't seem to be implemented by the Nvidia propitiatory drivers.
I don't know if there is any downside to using CreateXcbSurfaceKHR.


tomaka:
Thanks for the PR, but I don't see how that's supposed to work.
Maybe it works thanks to a happy coincidence (as xlib uses xcb internally), but that's definitely not a robust code.

ColinKinloch:
Turns out the Khronos loader doesn't build with Xlib support by default which is why it wasn't loading the symbol, my mistake.

PR #205, opened 2016-07-31
Support for intel linux driver

### LINKED ISSUE #204: Out of the box support for linux (Intel) without wayland
So I have been digging around the whole xcb/xlib problem that currently exists with vulkano, or rather the mesa intel driver. In this issue I will summarize what I have found out so far, and propose a solution.
## The Problem
Currently, this library cannot be used on linux on Intel GPUs out of the box. This is on the one hand, because winit does not support xcb so far and, from the other perspective, because the mesa driver does not support `VK_KHR_xlib_surface`.

## Current MESA status
So I have asked the mesa people about supporting  `VK_KHR_xlib_surface` in their bug tracker (https://bugs.freedesktop.org/show_bug.cgi?id=97096). As it seems, there is no desire to implement the extension, so we need to find a workaround.

Getting support for xcb in winit seems like a bunch of work, so I looked for solutions that are easier to implement. As it turns out, there is a (pretty) simple solution.

## Workaround
Luckily, xcb and xlib are very compatible. An existing X11 `Display` can be converted to an `xcb_connection_t` and then be passed to vulkan. The corresponding documentation can be found [here](https://xcb.freedesktop.org/MixingCalls/). By doing this, vulkano no longer relies on  `VK_KHR_xlib_surface` to exist.

*Details:*
Currently, vulkano checks if a wayland surface exists and if not, it will fallback to `Surface::from_xlib`.
The proposed solution is to fall back to xcb instead. To do this, the existsing xlib Display must be converted to a XCB connection with `XGetXCBConnection(Display)`.
The we can create the xcb `Surface` like this (utilizing xlib and xcb compability):
```rust
let xcb_connection = XGetXCBConnection(win.get_xlib_display().unwrap());
Surface::from_xcb(instance,
                xcb_connection,
                win.get_xlib_window().unwrap())
```
Of course, this code would require vulkano-win to depend on x11-dl, which is not desirable, so the conversion should probably be implemented somewhere in winit (via new api function).

*TODOs*:
- Add a function to winit(?) to convert xlib display to xcb connection
- `x11-rs` does not currently wrap `XGetXCBConnection`, so send a PR to them
- Somehow fall back to xlib, if a driver does not support `VK_KHR_xcb_surface`

Is this kind of workaround solution desirable?
I have a working implementation lying around (its hacky, adding manually linking to xlib), but at least it shows that the workaround works.

I yes, what would be your strategy to implement the conversion function?

nicokoch:
fixes #204

### DIFF
--- a/vulkano/src/swapchain/surface.rs
+++ b/vulkano/src/swapchain/surface.rs
@@ -395,7 +395,7 @@ impl Surface {
                                                                modes.as_mut_ptr())
                 ));
                 modes.set_len(num as usize);
-                debug_assert!(modes.iter().find(|&&m| m == vk::PRESENT_MODE_FIFO_KHR).is_some());
+                debug_assert!(modes.iter().count() > 0);
                 SupportedPresentModes::from_list(modes.into_iter())
             };
 
--- a/vulkano/src/swapchain/surface.rs
+++ b/vulkano/src/swapchain/surface.rs
@@ -395,6 +395,9 @@ impl Surface {
                                                                modes.as_mut_ptr())
                 ));
                 modes.set_len(num as usize);
+                // TODO: Use this assertion, once mesa fixes their driver.
+                // https://bugs.freedesktop.org/show_bug.cgi?id=97153
+                // debug_assert!(modes.iter().find(|&&m| m == vk::PRESENT_MODE_FIFO_KHR).is_some());
                 debug_assert!(modes.iter().count() > 0);
                 SupportedPresentModes::from_list(modes.into_iter())
             };
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -9,7 +9,7 @@ vulkano = { path = "../vulkano" }
 vulkano-win = { path = "../vulkano-win" }
 cgmath = "0.7.0"
 image = "0.6.1"
-winit = "0.5.1"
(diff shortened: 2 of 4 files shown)


vulkano/src/swapchain/surface.rs
```diff
@@ -395,7 +395,7 @@ impl Surface {
                                                                modes.as_mut_ptr())
                 ));
                 modes.set_len(num as usize);
-                debug_assert!(modes.iter().find(|&&m| m == vk::PRESENT_MODE_FIFO_KHR).is_some());
+                debug_assert!(modes.iter().count() > 0);
```
nicokoch: So this part is kind of awkward. PRESENT_MODE_FIFO_KHR is actually required to be supported, but the intel driver does not implement it. We can change this assertion back, once they do.

I have reported this issue to mesa: https://bugs.freedesktop.org/show_bug.cgi?id=97153

tomaka: Please put the old code but commented out and with a TODO.

tomaka:
r+ after the change!

nicokoch:
Ping @tomaka

remexre:
Is this change in 0.2.0? If not, could a 0.2.1 release be made?

Issue #67, opened 2016-04-28
Creating an uninitialized CPU-accessible buffer or image should be unsafe


tomaka:
Asking on IRC, it seems that uninitialized memory really is unsafe. It is an undefined behavior to read from it, and the only reason why things work today is that the compiler doesn't realize it's uninitialized.

lfairy:
It looks like the methods in question are already deprecated (e.g. [ImmutableImage::new](https://docs.rs/vulkano/0.9.0/vulkano/image/immutable/struct.ImmutableImage.html#method.new)).

Am I correct to say this can be closed after we remove those deprecated methods?

Rua:
Since the memory is allocated by Vulkan before it's exposed to Rust, from the perspective of the Rust compiler it's the same as if the Vulkan implementation had written random bytes to the memory before handing it over to Rust. I don't think they could be considered "uninitialised" in the compiler UB sense.

PR #206, opened 2016-08-03
Added gpgpu compute example

attackgoat:
Hope this helps - I created it to determine if Vulkano would work for my next project and I think that other people may find it interesting too.

### DIFF
--- a/examples/src/bin/compute.rs
+++ b/examples/src/bin/compute.rs
@@ -1,0 +1,170 @@
+// Copyright (c) 2016 The vulkano developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>,
+// at your option. All files in the project carrying such
+// notice may not be copied, modified, or distributed except
+// according to those terms.
+
+#[macro_use]
+extern crate vulkano;
+
+use vulkano::buffer::{BufferUsage, CpuAccessibleBuffer};
+use vulkano::command_buffer;
+use vulkano::command_buffer::PrimaryCommandBufferBuilder;
+use vulkano::device::{Device, DeviceExtensions};
+use vulkano::descriptor::descriptor_set::DescriptorPool;
+use vulkano::format::Format;
+use vulkano::image::sys::Dimensions;
+use vulkano::image::StorageImage;
+use vulkano::instance::{Instance, InstanceExtensions, PhysicalDevice};
+use vulkano::pipeline::ComputePipeline;
+
+use std::time::Duration;
+
+fn main() {
+    // Vulkano Compute Shader Example Code
+    //
+    // This example code will demonstrate using the compute functions of the Vulkano api. This example is completely headless
+    // and has no windowing support/requirements.
+    //
+    // Contrived problem definition:
+    // We will start with a simple data buffer (packed_input) filled with 32bpp RGBA values. This should be similar to many
+    // common image processing problems, but simplified to show the concepts only. From the packed_input we will compute a
+    // regular vulkan texture object (uimage2d) that can be used in rendering logic which will not be shown here. Sometimes
+    // compute applications require the opposite (texture -> buffer) logic and so we'll use a second compute shader to
+    // re-create the original data buffer as a second step.
+    //
+    // For simplification, we're not using real images and will instead algorithmically generate and verify the color values. 
+
+    // Image metadata:
+    let image_width: u32 = 2;
+    let image_height: u32 = 2;
+
+    // Buffer metadata:
+    let bytes_per_pixel: u32 = 4;
+    let byte_len = (bytes_per_pixel * image_width * image_height) as usize;
+
+    // Workgroup metadata:
+    let dispatch_count = image_width * image_height;
+
+    // Create a vulkan compute queue
+    let instance = Instance::new(None, &InstanceExtensions::none(), None).expect("failed to create Vulkan instance");
+    let physical = PhysicalDevice::enumerate(&instance).next().expect("no device available");
+    let queue = physical.queue_families().find(|q| {
+        q.supports_compute()
+    }).expect("couldn't find a compute queue family");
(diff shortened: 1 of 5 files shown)


examples/src/bin/compute.rs
```diff
@@ -0,0 +1,170 @@
...
+                },
+                set1: {
+                    output_image: StorageImage
+                }
+            }
+        }
+
+        // Create and fill an input buffer: The values will 0, 1, 2, ..(etc) for our contrived example instead of real RGBA values.
+        let packed_input = {
+            let buffer = CpuAccessibleBuffer::<[u8]>::array(&device, byte_len, &BufferUsage::all(), Some(queue.family())).expect("failed to create buffer");
```
tomaka: You could probably use `from_iter` here, which is recent and is cleaner.
It should be something like `from_iter(&device, &BufferUsage::all(), Some(queue.family()), (0 .. byte_len))`

tomaka:
Please put line breaks at column 100 for readability (so that no line is longer than 100 characters).

vulkano/src/buffer/cpu_access.rs
```diff
@@ -164,19 +164,21 @@ impl<T> CpuAccessibleBuffer<[T]> {
...
     {
-        unsafe {
-            CpuAccessibleBuffer::uninitialized_array(device, len, usage, queue_families)
-        }
+        CpuAccessibleBuffer::uninitialized_array(device, len, usage, queue_families)
     }
 
     /// Builds a new buffer. Can be used for arrays.
     #[inline]
-    pub unsafe fn uninitialized_array<'a, I>(device: &Arc<Device>, len: usize, usage: &Usage,
```
attackgoat: I removed unsafe from this method because the given parameters don't provide a way for the caller to produce an unsafe situation. Calling raw with something other than a multiple of size T would be unsafe but that is an implementation detail of uninitialized_array.

attackgoat:
Done - also I removed an unsafe that should make calling this method easier.

vulkano/src/buffer/cpu_access.rs
tomaka: After a long discussion on IRC, the conclusion is that it is unsafe to have uninitialized data in Rust, even if it's only primitive types (#67). For the moment `CpuAccessibleBuffer` is the only type that correctly handles this.

attackgoat: That doesn't match with my interpretation of The Book, but I do admit I've got less experience watching the ups-and-downs of how Rust has changed. I've only been watching from 1.9+.

Here's the relevant bit:
> In addition, the following are all undefined behaviors in Rust, and must be avoided, even when writing unsafe code:
> 
> Data races
> Dereferencing a null/dangling raw pointer
> [Reads of undef (uninitialized) memory](https://doc.rust-lang.org/book/unsafe.html)
> _[more]_

_This list is in a section that says these things do not qualify for `unsafe` specifically._

The compute example uses undef memory such that it writes into the buffer before reading, which although dangerous, is correct and safe.

tomaka: Unfortunately there's no way to enforce that the user writes to the buffer before reading.

attackgoat: _(Separate opinion post on the above matter)_

Hoisting `unsafe` to also indicate 'dangerous' is not the correct use of the keyword because there is no reason to turn on the three compiler Superpowers in places where we're just trying to tell the caller to beware.

Attributes (something like `deprecated`?) are also not correct because a case like the compute example is correct and should not produce any warnings.

IMHO this is a case where the scary name (`uninitialized_array`) is the proper way of signaling what is going on.

tomaka: The definition of `unsafe` is that if you use unsafe code the wrong way you might trigger an undefined behavior, and that there's no way for the compiler to check.
This is exactly the situation we're in. If you call `uninitialized_array` then write, you're fine. If you call `uninitialized_array` then read, you're not. Unsafe just means that it's impossible for the compiler to enforce the correctness of your code.

attackgoat: Wouldn't it be better to decorate `read` (_and not `uninitialized_array`_) as `unsafe` if that is where the undefined behavior could happen?

I don't believe so because that would be an attempt to subvert a keyword meant to enable superpowers (and likely different code generation) where documentation is the correct tool for the job.

In short, adding `unsafe` does not prevent incorrect API usage:
> Rust cannot prevent all kinds of software problems. Buggy code can and will be written in Rust. These things aren’t great, but they don’t qualify as unsafe specifically.

vberger: If I may, `std::mem::uninitialized()` is unsafe, so I would expect any function providing you with uninitialized memory to be unsafe too.

attackgoat: That's a good point, and the [documentation](https://doc.rust-lang.org/std/mem/fn.uninitialized.html) includes a nice comment too:
> // DANGER ZONE: if anything panics...

I think that comment pattern would help the reader understand that compute.rs is dangerous between the points of creating the uninitialized array and filling it.

Any comments on my proposal to:
- Undo change 427c7e7 related to the unsafe block
- Add a danger start..end comment block around the unsafe lines of code

tomaka: > Add a danger start..end comment block around the unsafe lines of code

In my opinion initializing with zeroes would still be a better option. After all, if you did exactly the same operation on the CPU in Rust you would also have to initialize the output buffer with a default value as well.

`from_iter` is currently not well optimized, but with some tweaking it should be equivalent to calling `uninitialized_array` followed by memset.

However I don't have a strong opinion on this. An unsafe block with a comment would do too.

svenstaro:
Is there any further work on this? I think this example is super useful and it would be awesome to have it merged.

tomaka:
The example is fine (the code style is inconsistent with rest of the library, but that's okay-ish), but the changes in the library are not (see the discussion).

attackgoat:
The library changes were reverted; as for the style please change whatever you'd like or close this please.

rukai:
Looks like tomaka opted to include a simpler compute example, about a year after the last activity on this PR https://github.com/vulkano-rs/vulkano/blob/master/examples/src/bin/basic-compute-shader.rs
Closing.

PR #208, opened 2016-08-05
Fix a crash in nightly Rust

tomaka:
Since the switch to MIR, there has been a crash when creating a swapchain.

The crash actually appears here: https://github.com/aturon/crossbeam/blob/1afa88c5f821e572a740207e86c3b66055cae37a/src/sync/ms_queue.rs#L147
The code reaches a `ud2` opcode right after `is_data`.
I couldn't find the exact cause, but everything points towards a bug in crossbeam.

For the moment I'll replace `MsQueue` with `Mutex<Vec<>>`.

Pinging @aturon

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -15,7 +15,6 @@ use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::Ordering;
 use std::time::Duration;
-use crossbeam::sync::MsQueue;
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks to IRC, the bug has been identified in crossbeam. I'll submit a PR there, but I'll also merge this one because I don't want to wait for crossbeam to be fixed.

tomaka:
cc https://github.com/aturon/crossbeam/pull/82

PR #215, opened 2016-08-16
ApplicationInfo::from_cargo_toml now uses compiletime env

Eroc33:
It seemed like ApplicationInfo::from_cargo_toml should not panic if compiled with, but not run under, cargo, but it did. This should fix that.


tomaka:
Thanks!

PR #216, opened 2016-08-17
New command buffers system

tomaka:
New design for command buffers. For the moment I added it as a secondary system alongside with the existing system. In the end the current system will be removed in favor of this one.

Ok, so how exactly does that work?

While we're building a command buffer, each buffer or image has a state associated to them. What this state actually is depends on the implementation of the buffer/image, but it typically contains the pipeline stages, access masks, whether the resources was read/written, and for images their layout.

Whenever a buffer or an image is used in a command in the command buffer building process, two steps are involved: first we retrieve the current state of the resource in the builder (which can be empty if the resource wasn't used yet), then we ask the state to mutate itself in order to accommodate for the usage.

Retrieving the current state of the resource is done with the help of the `is_same` method that is implemented on buffers and images. Thanks to this, it is possible to have multiple objects implement the Buffer or Image trait and make them share the same state. This should allow for designs such as multiple resources sharing the same memory. If you want to split a buffer/image into multiple sub-buffers or sub-images, you can also make `is_same` return false for the different sub-buffers/sub-images even when the underlying buffer/image is the same.

Asking the state to mutate itself is done through a method on the state that takes as parameter the way the buffer/image is used by the command, and can return an optional "pipeline barrier request". The command buffer builder must take these requests into account and can merge them. The request contains the range of positions where the barrier can be put, so that we're free to reorganize things a bit. Buffers and images can also generate a final pipeline barrier, so that we can change the layout or flush caches to the HOST stage when necessary.

In order to make this system work, the command buffer builder has changed. It now uses an iterator-like API where each command wraps around the previous builder. This allows us to keep strong types and makes it possible for the compiler to aggressively inline. This also means that it will be possible to rework descriptor sets and framebuffers to keep strong typing as well.

Once we call the `build` function, the state of each buffer/image is turned into a "finished" state that is simpler than the state in the builder. When a command buffer is submitted, we query the finished state of each buffer/image in order to know which barriers to add and which semaphores to wait upon.

**For the moment, there's only one command supported (updating a buffer) and only cpu-accessible buffers implement the new tracking system. Some implementations are placeholder and unsound. The tracking system for images isn't even there yet. This PR is meant to lay the design for the new system, and more PRs will come in the future to actually implement it correctly.**

The one unresolved question is how to handle the fact that the user may want to submit to multiple queues simultaneously. If you submit two command buffers that each use the same buffer and the same image, then the order in which the `on_submit` function is called is important. I think some sort of "pending semaphore" system can be added. In last resort, we can also switch back to the current system which is that a global lock prevents parallel submissions.

### DIFF
--- a/vulkano/src/buffer/cpu_access.rs
+++ b/vulkano/src/buffer/cpu_access.rs
@@ -37,10 +37,17 @@ use buffer::sys::UnsafeBuffer;
 use buffer::sys::Usage;
 use buffer::traits::AccessRange;
 use buffer::traits::Buffer;
+use buffer::traits::CommandBufferState;
+use buffer::traits::CommandListState;
 use buffer::traits::GpuAccessResult;
+use buffer::traits::SubmitInfos;
+use buffer::traits::TrackedBuffer;
 use buffer::traits::TypedBuffer;
+use buffer::traits::PipelineBarrierRequest;
+use buffer::traits::PipelineMemoryBarrierRequest;
 use command_buffer::Submission;
 use device::Device;
+use device::Queue;
 use instance::QueueFamily;
 use memory::Content;
 use memory::CpuAccess as MemCpuAccess;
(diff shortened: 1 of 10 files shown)


tomaka:
Since I guess nobody's going to review this, I might as well merge it.

PR #221, opened 2016-08-19
Lay foundation for descriptor sets tracking system

tomaka:
Implements `dispatch` with the new design.
Since dispatch uses descriptor sets, the biggest part was to make descriptor sets work with the new design.


tomaka:
Note that while the traits necessary for descriptor sets are here, they are not yet implemented on descriptor sets themselves. Therefore you can only pass `()` for now to the new dispatch command.

Implementing on descriptor sets themselves will require more changes to descriptor sets, in order to keep strong typing.

PR #226, opened 2016-08-20
Bypass buggy driver behaviors when creating instance

tomaka:


tomaka:
With this change, the vulkano examples work with the experimental Anvil Intel driver.

PR #227, opened 2016-08-20
Some improvements to the generated assembly

tomaka:
I have taken a look at the generated assembly, and spotted some things that could be improved.

However the biggest problem I think is the fact that `SmallVec` is not very optimized.
cc https://github.com/rust-lang/rfcs/issues/618


jonas-schievink:
Ah that reminds me that I've done the hasher replacement for glium, I should open a PR :)

PR #228, opened 2016-08-21
Don't bind pipeline objects when not necessary

tomaka:


vulkano/src/command_buffer/std/dispatch.rs
```diff
@@ -117,6 +119,18 @@ unsafe impl<'a, L, Pl, S, Pc> StdCommandsList for DispatchCommand<'a, L, Pl, S,
     }
 
     #[inline]
+    fn is_compute_pipeline_binded<OPl>(&self, pipeline: &Arc<ComputePipeline<OPl>>) -> bool {
```
jonas-schievink: `bound`, not `binded`, right?

tomaka: Thanks

PR #239, opened 2016-09-01
Update glslang to `219b025` and use preinstalled `glslangValidator` if available

lummax:
This updates `glslang` to https://github.com/KhronosGroup/glslang/commit/219b025d7eae1383770c89247e6c5eb1909ac188 to fix a compile error.

In addition the `build.rs` will use a preinstalled `glslangValidator` if found. This part might not be portable to android? I unfortunately have no experience there.

### DIFF
--- a/glsl-to-spirv/build/build.rs
+++ b/glsl-to-spirv/build/build.rs
@@ -15,6 +15,12 @@ fn main() {
         // TODO: check the hash of the file to make sure that it is not altered
         Path::new("build/glslangValidator.exe").to_owned()
 
+    } else if let Ok(true) = Command::new("glslangValidator").arg("-version").status().map(|s| s.success()) {
+        // we found `glslangValidator` in $PATH and we know that we are not on Windows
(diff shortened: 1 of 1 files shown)


tomaka:
There are two reasons why I'm hesitating about this:

1) The PR doesn't include the precompiled windows binary, which I included because it can be tedious to build. But I'm hesitating about what to do with that binary.
2) `glslangValidator` is supposed to be an implementation detail, so I'm not sure it's a good idea to try to pick the program from the PATH. For example it could be annoying if shaders compiled differently depending on whether or not you have `glslangValidator` in your PATH.

lummax:
No sweat :)

It's just what I did to prevent the lengthy and erroneous `glslang` build as I (with nixos) hat it in my `$PATH`. Your second reason sound good to just deny including the `$PATH` logic.

rukai:
glsl-to-spirv is no more. https://github.com/vulkano-rs/vulkano/pull/947

PR #242, opened 2016-09-01
vk-sys/vulkano{-win}: correct type of native window

lummax:
This commit changes the definitions for `window` in `XlibSurfaceCreateInfoKHR`
and `XcbSurfaceCreateInfoKHR` with to `u64` (xlib) and `u32` (xcb) (see
definitions below).

`vk-sys` defined `XlibSurfaceCreateInfoKHR` and `XcbSurfaceCreateInfoKHR` with
`pub window: *const c_void`. This propagates through to the API in
`vulkano::swapchain::Surface::from_{xlib,xcb}()`. As the native window types
in xcb/xlib are defined as unsigned ints, this may lead to the belief that a
pointer to such an integer is required. `vulkano` will create a `Surface` but
this surface will not be supported by any queue family.

xlib: https://cgit.freedesktop.org/xorg/proto/x11proto/tree/X.h#n96 and
      https://cgit.freedesktop.org/xorg/proto/x11proto/tree/X.h#n66

      (search `Windows`)

xcb: https://xcb.freedesktop.org/manual/group__XCB____API.html

    (search `xcb_window_t`)

(`winit` defines the field in `XWindow` correcty but casts into a `*mut c_void` in `get_xlib_window()`.)

### DIFF
--- a/vk-sys/src/lib.rs
+++ b/vk-sys/src/lib.rs
@@ -17,6 +17,7 @@ use std::ffi::CString;
 use std::fmt;
 use std::os::raw::c_char;
 use std::os::raw::c_void;
+use std::os::raw::c_ulong;
 
 pub type Flags = u32;
 pub type Bool32 = u32;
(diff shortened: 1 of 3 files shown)


tomaka:
Oh, good point. Thanks for the fix.

While `xcb_window_t` is defined as `u32`, `XID` is defined as `unsigned long`, which is 32bits on 32bits Linux and 64bits on 64bits Linux.

Please use `std::os::raw::c_ulong` instead of `u64` for xlib.

tomaka:
Thanks!

PR #244, opened 2016-09-02
Rework tons of things

tomaka:
[breaking change]

You are invited to check the changes in the examples to see how to adjust your code.

Here is a non-exhaustive list of the biggest changes.
- Many functions that were taking `&Arc<_>` as parameter now take `Arc<_>` in order to reflect the fact that cloning is performed. You may need to turn `foo(&bar)` into `foo(bar.clone())`.
- Arcs for buffers and images are no longer mandatory. For example instead of taking a `Arc<T> where T: Buffer`, vulkano now takes a `T where T: Buffer`, and the `Buffer` trait is implemented on `Arc<T> where T: Buffer` but also on other types. This change is very similar to the previous one, but the error message is probably going to be less clear.
- The `pipeline_layout!` macro was removed. You no longer need to create explicit pipeline layouts, as they are automatically computed when the pipeline is created. The possibility to create an explicit pipeline layout will be restored later.
- In order to create a descriptor set, you are encouraged to use the `simple_descriptor_set!` macro. See the teapot and image examples. The descriptor pool is automatically handled for now (will be changed later to allow specifying a pool).
- Creating a render pass is now done in only one step. The `single_pass_renderpass!` and `ordered_passes_renderpass!` macros are now expressions and must be called inside of a function. You are encouraged to cast their return value into a `Arc<RenderPass>` if you want to store it in a struct.
- Command buffers got a big rework. You should use the `AutoCommandBufferBuilder` struct and the `CommandBufferBuilder` trait in order to build command buffers.
- Submitting now uses futures. This change is too long to document here, so you should check #385 and/or the docs generated by `cargo doc`..

Note that the biggest changes of this PR are in vulkano's internals. Tons of traits were changed. However I'm not mentionning these, as the traits are typically not used from external code.

These changes are not intended to be immediately published. They are more a starting point for a kind of "vulkano2" with a reworked design. Not everything is polished.


tomaka:
A small note: I'm probably breaking more things than needed, but this PR is more or less the start of "Vulkano 2.0". Therefore lots of small things will need adjustments after this PR anyway.

tomaka:
While working on this PR, I realized that there's a lot of code deduplication.
Framebuffers, descriptor sets, and command buffers each have their own code to track resources state. It's probably possible to write more generic traits or something.

tomaka:
The next short-term tasks are:

- "Flatten" the TrackdImage trait
- Remove the `extract_buffer_state` and `extract_image_state` functions, and replace that with a system where each buffer and image looks its own state in a collection passed as parameter.
- Once that's done, finish the descriptor sets rework.

RobertWHurst:
I suppose if we get the MoltenVK stuff working we should just start over upon this branch or the new head of master provided this gets merged?

This looks great by the way.

tomaka:
This branch shouldn't conflict with the moltenvk stuff.

tomaka:
I'm currently waiting for [this book](https://www.amazon.com/Vulkan-Programming-Guide-Official-Learning/dp/0134464540/) to come before I continue working on the rework, so I can read what it has to say.

tomaka:
Although it's not very third-party-reviews-friendly, I'm going to merge this branch now that tests are "passing" (read: I ignore a few tests that were failing).
The `master` branch isn't maintained anymore anyway, so it's no use having a separate branch.

Issue #254, opened 2016-09-08
Example for UniformBuffer update


Batanick:
Suppose we want to add camera rotation to our fancy teapot example. In the example we have 2 _command buffers_ with shared _uniform buffer_ and  _pipeline_layout::set0::Set_ attached to them. How can we perform an update on each frame? 

1. Do we need to create 3 uniform buffers for this? 
1. Do we need to update _uniform buffer_ directly or do some magic with _pipeline_layout::set0::Set_ ?

tomaka:
If the uniform buffer contained a large amount of data, you would need to create 3 uniform buffers and create a new descriptor set before each rendering.

There is a technique that vulkano does not yet support: you can create just one buffer that you split in three parts, and when you bind the descriptor sets you can specify an offset in the buffer for the start of the data. This technique has the advantage that you're only using one buffer and that the descriptor set stays the same.

However for small amounts of data (like here), you can just either use an `update_buffer` command in the command buffer, or use push constants (which are like uniforms, but limited to 128 bytes of data).

Batanick:
Thanks for the answer, took some time to figure out.

PR #258, opened 2016-09-11
Teapot example: rotation added

Batanick:
Had troubles with figuring out how to actually update buffers after creation (see #254). So I've added a simple rotation to the teapot example, maybe would be useful to someone

### DIFF
--- a/examples/src/bin/teapot_vs.glsl
+++ b/examples/src/bin/teapot_vs.glsl
@@ -18,11 +18,13 @@ layout(location = 1) in vec3 normal;
 layout(location = 0) out vec3 v_normal;
 
 layout(set = 0, binding = 0) uniform Data {
-    mat4 worldview;
+    mat4 world;
+    mat4 view;
     mat4 proj;
 } uniforms;
 
 void main() {
-    v_normal = transpose(inverse(mat3(uniforms.worldview))) * normal;
-    gl_Position = uniforms.proj * uniforms.worldview * vec4(position, 1.0);
+    mat3 worldview = mat3(uniforms.view * uniforms.world);
+    v_normal = transpose(inverse(worldview)) * normal;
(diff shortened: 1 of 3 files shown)


examples/src/bin/teapot_vs.glsl
```diff
@@ -18,11 +18,13 @@ layout(location = 1) in vec3 normal;
...
+    mat4 view;
     mat4 proj;
 } uniforms;
 
 void main() {
-    v_normal = transpose(inverse(mat3(uniforms.worldview))) * normal;
-    gl_Position = uniforms.proj * uniforms.worldview * vec4(position, 1.0);
+    mat3 worldview = mat3(uniforms.view * uniforms.world);
+    v_normal = transpose(inverse(worldview)) * normal;
+    gl_Position = uniforms.proj * uniforms.view * uniforms.world * vec4(position, 1.0);
```
tomaka: Please replace `uniforms.view * uniforms.world` with `worldview` since you're calculating it already.

tomaka:
Thanks!

r+ after the minor fix.

Batanick:
fixed

nicokoch:
Hey, I noticed the teapot does not rotate when I launch the executable directly like `./target/debug/teapot`.
When invoking with `cargo run --bin teapot` everything works fine.

Is this supposed to happen?

tomaka:
> Hey, I noticed the teapot does not rotate when I launch the executable directly like ./target/debug/teapot.
When invoking with cargo run --bin teapot everything works fine.

`examples/target/debug/teapot` probably refers to an old compilation.
Since vulkano uses Cargo workspaces, the *real* `target` directory is now at the root of the repo.

nicokoch:
Yep that's it, thanks. I wish `cargo clean` would delete the old .target dirs

PR #260, opened 2016-09-11
Add support for cubemap images

tomaka:
The major public change is that `image::Dimensions` has two additional enum variants: `Cubemap` and `CubemapArray`.

There are other internal changes:
- Creating an image now takes an `ImageDimensions` (which is different from `Dimensions`)
- Creating an image view takes a `ViewType`.
- The `ImageView` trait has an additional `dimensions()` method.

This API is closer to what Vulkan proposes.

### DIFF
--- a/examples/src/bin/debug.rs
+++ b/examples/src/bin/debug.rs
@@ -13,7 +13,7 @@ extern crate vulkano;
 use vulkano::device::{Device, DeviceExtensions};
 use vulkano::format::Format;
 use vulkano::image::ImmutableImage;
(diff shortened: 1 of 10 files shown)


tomaka:
If you're googling "vulkano cubemap" and find this PR:

Using a cubemap is not different than a regular texture 2D. The difference is that a cubemap has 6 layers, and a cubemap array has N*6 layers.
When you write to an image, you can choose to which layer to write to.

The order of the layers is: positive X (right), negative X (left), positive Y (top), negative Y (bottom), positive Z (front), negative Z (back).

PR #275, opened 2016-09-20
Update glslang

svenstaro:
The current one doesn't support int64 stuff and also glslang moves quickly so we should upgrade this now and then.


tomaka:
Thanks. I opened #276 instead, which also updates the windows executable.

PR #302, opened 2016-10-08
ensure macOS is excluded from unix target

RobertWHurst:
Fixes a few compile errors were linux targeted code ends up included in the macOS build.


tomaka:
Thanks

PR #304, opened 2016-10-13
Fix linking on mac

RobertWHurst:


examples/build.rs
```diff
@@ -1,6 +1,7 @@
 extern crate vulkano_shaders;
 
 fn main() {
+    println!("cargo:rustc-link-search=framework={}", "/Library/Frameworks");
```
RobertWHurst: This line allows searching the /Library/Frameworks directory for macOS frameworks

tomaka:
This is definitely very hacky, but I'll do some cleanup when it's working.

PR #308, opened 2016-10-19
WIP: Rework render passes in the same way as pipeline layouts

tomaka:
In order to have the same system for both, for consistency.


tomaka:
Superceded by #321

Issue #315, opened 2016-11-10
Make DeviceExtensions and InstanceExtensions extensible


tomaka:
Right now if you initialize the extensions with `DeviceExtensions { foo: true, bar: true }` and a new members is added, the code will break.

Unfortunately, Rust's rules doesn't allow the trick where you add a private field to a struct (see https://github.com/rust-lang/rfcs/blob/master/text/0736-privacy-respecting-fru.md).

tomaka:
Marking as "easy" and "workable", but basically I have no idea how to fix it. If someone has an idea, it should be easy to implement.

Mixthos:
Doesn't the alternative trick mentioned in the RFC work in this case?

````rust
pub struct Bar { pub a: u8, pub b: String, pub _hidden: BarHidden }
pub struct BarHidden { _cannot_construct: () }
```

tomaka:
Oh I didn't even read the RFC to the end!

Mixthos:
I'll implement it then. ;)
Should the PR go to `master` or `incoming`?

tomaka:
`incoming` doesn't touch this area so you can put that in `master`

PR #316, opened 2016-11-10
Future extensibility for extensions

Mixthos:
As discussed in #315.

### DIFF
--- a/vulkano/src/instance/extensions.rs
+++ b/vulkano/src/instance/extensions.rs
@@ -28,6 +28,9 @@ macro_rules! extensions {
             $(
                 pub $ext: bool,
             )*
+			
+			/// Helper for future extensibility.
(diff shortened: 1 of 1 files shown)


tomaka:
- The indentation is wrong.

- I think it should be better documented. Imagine you're a newcomer and you see a member documented as `Helper for future extensibility`. That wouldn't be very helpful for you.

- Also the name should be more like `Unbuildable`, since it's not actually hidden.

- Please mark the `Hidden`/`Unbuildable` type with `#[doc(hidden)]` so that it only appears in the docs of InstanceExtensions and DeviceExtensions.

Mixthos:
This should fix the problems you mentioned.

I'm not happy about the documentation, but I couldn't think of better wording.

tomaka:
Thanks!

PR #317, opened 2016-11-15
Nvglslc output

alexd2580:
Add functionality to compile shaders using the nvidia NVGLSLC shader compiler.
The rust wrapper module is still generated from the SPIR-V code which is generated by glslang.

The usage of the new feature requires enabling the `nv_glsl_shader` device feature,
and using 
```
vulkano_shaders::build_glsl_shaders_nvidia(shader_list.iter().cloned());
```
instead of 
```
vulkano_shaders::build_glsl_shaders(shader_list.iter().cloned());
```


tomaka:
I'm not sure about this PR because the `VK_NV_glsl_shader` extension is clearly a step in the wrong direction.

alexd2580:
I would argue, that this is not necessarily the case. As i understood it, this extension is only usable on NVIDIA GPUs, and the corresponding compiler (NVGLSLC) should be able to optimize the GLSL shader better, because it's designed specifically for NVIDIA architectures. I ran some tests with high-load compute shaders, which, although not perfect, performed better using NVGLSLC than GLSLANG.

tomaka:
The path forward is to move these optimizations from their GLSL compiler to their SPIRV compiler. Feeding GLSL to the driver is not only not-cross-platform but also doesn't encourage them to do that.

However to be honest I'm also interested because I have a compute shader that runs poorly.

Issue #291, opened 2016-09-26
Compile, load and run GLSL at runtime


Limeth:
I can currently transpile GLSL to SPIR-V at runtime via:
```rust
    let vs_spirv = glsl_to_spirv::compile(SHADER_VERTEX, ShaderType::Vertex)
        .expect("failed to transpile the vertex shader");
    let fs_spirv = glsl_to_spirv::compile(SHADER_FRAGMENT, ShaderType::Fragment)
        .expect("failed to transpile the fragment shader");
```
Then, I [edited](https://github.com/Limeth/vulkano/commit/c919c2250f0ee748b500a1ec885bddb20959ef77#diff-4439a959368b9ae0576a673e3685c267R200) `vulkano-shaders` so I could do this:
```rust
    let vs = vulkano_shaders::load_spirv(vs_spirv, &device);
    let fs = vulkano_shaders::load_spirv(fs_spirv, &device);
```
The last thing I need, is to get the `main_entry_point` to run the shader, if I understand this correctly. I am willing to implement it myself, so it complies with the current code. Would you mind guiding me?

tomaka:
Getting the entry point is the trickiest aspect, as it's at this moment that you must specify the input/output interfaces, the specialization constants and the pipeline layout.

For example if you call `vertex_shader_entry_point`, the arguments you must pass are the name of the entry point (usually `main`), then two objects that must implement the `ShaderInterfaceDef` trait, then one object that must implement `PipelineLayoutDesc`. Also, the `S` parameter must implement `SpecializationConstants`.

If you're looking for a quick solution, you can implement them in a dummy way (returning empty sets). Doing it properly would require a rework of `vulkano_shaders` in order to separate the part that introspects the SPIR-V from the part that emits Rust code.

tomaka:
To elaborate a bit more on this.

It should be possible to write a wrapper around `ShaderModule` that is capable of reloading the shader at runtime (and reparsing the SPIR-V every time).

I guess it should also be possible to write a wrapper around `ComputePipeline` and `GraphicsPipeline` that automatically reloads the shaders when they are modified.
The reload can fail obviously if the source code of the shader doesn't compile, but also if the shader's interface no longer matches the pipeline layout or subpass passed used to build the pipeline.

minecrawler:
I wouldn't see it as a drawback. Which situations require you to recompile the shaders? It would be tools! As for tools, it does not matter if they stutter a bit when reloading the shader, so imho it's ok :)

[...]

Issue #281, opened 2016-09-20
Allow for using a system-installed version of glslang


svenstaro:
I think it would make sense to provide a way to use a system-installed version of glslang.

tomaka:
There was already a small discussion in #239 

> glslangValidator is supposed to be an implementation detail, so I'm not sure it's a good idea to try to pick the program from the PATH. For example it could be annoying if shaders compiled differently depending on whether or not you have glslangValidator in your PATH.

rukai:
glsl-to-spirv is no more. https://github.com/vulkano-rs/vulkano/pull/947

PR #322, opened 2016-11-23
glsl-to-spirv: Search `PATH` for `glslangValidator`.

the-kenny:
This PR extends `build.rs` for `glsl-to-spirv` to search `PATH` for `glslangValidator`. This makes it much easier to use the crate on my setup ([NixOS](https://nixos.org)).

The old behavior (for non-windows) still exists and is used as a fallback in case `glslangValidator` isn't in `PATH`.

### DIFF
--- a/glsl-to-spirv/build/build.rs
+++ b/glsl-to-spirv/build/build.rs
@@ -14,13 +14,21 @@ fn main() {
     let path = if target.contains("windows") {
         // TODO: check the hash of the file to make sure that it is not altered
         Path::new("build/glslangValidator.exe").to_owned()
-
     } else {
-        // Try to initialize submodules. Don't care if it fails, since this code also runs for
-        // the crates.io package.
-        let _ = Command::new("git").arg("submodule").arg("update").arg("--init").status();
-        cmake::build("glslang");
-        Path::new(&env::var("OUT_DIR").unwrap()).join("bin").join("glslangValidator")
+        // Try to find `glslangValidator` in `PATH`
+        let paths = env::var("PATH").unwrap_or("".to_string());
+        env::split_paths(&paths)
+            .flat_map(|path| {
+                let full = path.join("glslangValidator");
+                if full.is_file() { Some(full) } else { None }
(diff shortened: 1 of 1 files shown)


tomaka:
This is essentially #281 and #239

I think that searching for glslangValidator in PATH is wrong, but my opinion is not set in stone.

the-kenny:
Oh I'm sorry - I didn't check existing PRs before implementing this (I was in a hurry to get it working).

I agree with some of your points in the other PR - it would be very annoying if users have an incompatible or outdated version in `PATH`. However, compiling it from source isn't very stable either. As mentioned in #281 it fails on NixOS because it can't find `std::isinf` (possible because NixOS uses LLVM? I'm not sure). 

Maybe it's possible to add a `feature` flag to `glsl-to-spirv` to enable this behavior as an opt-in?

aristidb:
Why not use an environment variable VULKANO_GLSLANG_VALIDATOR_EXTERNAL_PATH. If it's not set, you can fall back to the current behavior. So then nobody can _accidentally_ use the wrong version of glslangValidator.

the-kenny:
Thinking a bit more about it, I'd actually prefer a environment variable that's used at *runtime* even more. Maybe it's possible to change the behavior (or API?) so `glsl-to-spirv` can run `glslValidator` from `PATH` without copying/symlinking it at build-time?

tomaka:
@the-kenny That's #291 and it requires tons of changes.

Ralith:
@the-kenny I believe the NixOS build was fixed two months ago by 801a66e8565ef41b7bb7f06ce84d43822f69e7cd which has not yet been released on crates.io.

the-kenny:
That's good to hear :) I still think supporting out-of-tree binaries is a
good idea to simplify the build process.

Am 27.11.2016 00:08 schrieb "Benjamin Saunders" <notifications@github.com>:

rukai:
glsl-to-spirv is no more https://github.com/vulkano-rs/vulkano/pull/947

PR #327, opened 2016-12-06
vkspec.pdf update->1.0.26; added Vulkan apispec.pdf

paulirotta:
Vulkan spec update
Added API spec


tomaka:
To be honest there's no reason why vkspecs.pdf is in the repo in the first place. I put it here because "why not".

paulirotta:
Well, it is convenient documentation. Want to update it or delete?

PR #328, opened 2016-12-07
Manual implementation of Copy in vulkano-shaders structs

### LINKED ISSUE #253: Do not use #[derive] for structs generated by vulkano-shaders
Because of large arrays that do not implement `Clone`.
Since everything should implement `Copy` (including large arrays), it should be easy to write `impl Clone for OurStruct {}` manually.

wjvds:
Fixes #253

### DIFF
--- a/vulkano-shaders/src/structs.rs
+++ b/vulkano-shaders/src/structs.rs
@@ -12,6 +12,25 @@ use std::mem;
 use parse;
 use enums;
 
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks!

Could you just please move the `write_structs` function at the top? Since it represents the API of the module, it's what should appear first.

PR #333, opened 2016-12-08
Disabled broken Clone impl for unsized buffers

### LINKED ISSUE #332: vulkano-shaders: Clone instance generated for buffers with unsized arrays
vulkano-shaders generates `Clone` instances for buffer structs, although the members of that struct are not `Sized`.

This GLSL code:
```
#version 450
layout(local_size_x = 8, local_size_y = 8, local_size_z = 1) in;
layout(set = 0, binding = 0, std430) readonly buffer A { uint a[]; };
void main(void) {}
```

generates this Rust code:
```
pub struct A {
    pub a: [u32] /* offset: 0 */
} /* total_size: None */

impl Clone for A {
    fn clone(&self) -> Self {
        A {
            a: self.a
        }
    }
}
```

which does not compile:
```
    |
144 | impl Clone for A {
    |      ^^^^^ the trait `std::marker::Sized` is not implemented for `[u32]`
    |
    = note: `[u32]` does not have a constant size known at compile-time
    = note: required because it appears within the type `asd::ty::A`
    = note: required by `std::clone::Clone`
```

wjvds:
Fixes #332

### DIFF
--- a/vulkano-shaders/src/structs.rs
+++ b/vulkano-shaders/src/structs.rs
@@ -174,15 +174,20 @@ fn write_struct(doc: &parse::Spirv, struct_id: u32, members: &[u32]) -> String {
             });
         }
     }
+    
+    // We can only implement Clone if there's no unsized member in the struct.
+    let impl_text = if current_rust_offset.is_some() {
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks

Issue #337, opened 2016-12-20
Pipeline Instantiation: vertex_input: unable to infer enough type information about `_`


minecrawler:
I use the [sample code from the triangle example](https://github.com/tomaka/vulkano/blob/master/examples/src/bin/triangle.rs#L278) (though the other ones aren't too different)

I try to split up the whole initialization process and make it a bit easier to work with (since I have a specific setting in mind and only need certain configurations).

So, at the moment, I try to instantiate a new `GraphicsPipeline` object. Unfortunately, I get the error

```
error[E0282]: unable to infer enough type information about `_`
  --> src/renderer/graphics_pipeline_new.rs:70:27
   |
70 |             vertex_input: SingleBufferDefinition::new(),
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for `_`
   |
   = note: type annotations or generic parameter binding required
```

I tried to add the rest of the example code below, which solves the error, but I do not want to have the rendering loop in my graphics_pipeline module. So I tried to find out what might be the problem. I guess I need to add some `T` as in `vertex_input: SingleBufferDefinition::<T>::new(),`, but I do not understand what type I have to pass here. Could anyone please elaborate? I am sorry if this is  a stupid question...

tomaka:
> So I tried to find out what might be the problem. I guess I need to add some T as in vertex_input: SingleBufferDefinition::<T>::new(),, but I do not understand what type I have to pass here.

Hi. The type here is the type of the struct that represents the vertex. For example in the triangle example it's `Vertex`.

minecrawler:
Ah, perfect. So I just have to create a new struct and call the `impl_vertex` macro on it

```rust
struct Vertex { position: [f32; 2] }
impl_vertex!(Vertex, position);
```

and then pass it as type `SingleBufferDefinition::<Vertex>::new()`! I can work with that, thank you a lot!

PR #339, opened 2016-12-21
Clarify description (#337)

minecrawler:
Clarify manual type inferring for `SingleBufferDefinition` as explained in #337


tomaka:
Since the `Vertex` type already exists in the example, I don't think we should tell people to create a `Vertex` type. Instead of think we should just mention that in the example the type is automatically inferred to be `Vertex` but that in a real code it may be necessary to make it explicit.

PR #340, opened 2016-12-22
Fixed warning about CpuAccessibleBuffer::array being depreciated

callym:
I figured out how to do this when I was looking at the texture example, so thought it might be worth changing it to not use the depreciated `CpuAccessibleBuffer::array` function.


tomaka:
Thanks

PR #341, opened 2016-12-25
Fix off-by-one error when creating queues

dflemstr:
The user should be allowed to create exactly as many queues as are available in `queue_family.queues_count()`, but now they are only allowed to create one fewer (for example `15` if `16` are available).


tomaka:
Thanks!

PR #343, opened 2017-01-03
Examples now use min_image_count-buffering

Binero:
This fixes a bug on some devices where the hard coded values were too small, and not supported.


Binero:
Two tests failed for me, but they don't seem related to the changed code.

tomaka:
Thanks

PR #344, opened 2017-01-04
Nvidia problem fixed

tomaka:
It has been reported that it's been fixed. I'm currently downloading the latest drivers to check myself.


tomaka:
Err.. still sometimes deadlocks for me with version 376.33.

tomaka:
I'm leaving this PR open, but I'll wait until they fix the problem before merging.

tomaka:
It looks like it's actually fixed now. Let's merge.

Issue #256, opened 2016-09-09
RFC: Plan if custom derives are stabilized


tomaka:
If custom derives are stabilized, I plan to make it possible to write something like this:

``` rust
#[macro_use]
extern crate vulkano;
#[macro_use]
extern crate vulkano_macros;

shader!{r#"
#version 450

void main() {
    gl_Position = vec4(1.0, 0.0, 0.0, 1.0);
}

"#}
```

This would work by defining the macro `shader!` in vulkano like this:

``` rust
#[macro_export]
macro_rules! shader {
    ($src:expr) => (
        #[derive(ShaderCompiler)]
        #[src($src)]
        struct Shader;
    )
}
```

And by writing `vulkano_macros` something like this:

``` rust
extern crate rustc_macro;
extern crate vulkano_shaders;

use rustc_macro::TokenStream;

#[rustc_macro_derive(ShaderCompiler)]
pub fn derive_shader_compiler(input: TokenStream) -> TokenStream {
    let input = input.to_string();
    let num_lines = input.lines().count();
    let shader_src = input.lines().skip(1).take(num_lines - 3).collect::<Vec<String>>().concat();

    ... dispatch to vulkano_shaders ...
}
```

Invoking `shader!` would expand to a struct with the shader source code in an attribute, which is then picked up by the handler of `ShaderCompiler` and dispatched to `vulkano_shaders`.

The token stream returned by custom derives actually overwrites the struct, so the user wouldn't get any garbage `struct Shader` in the final code.

tomaka:
Link to tracking issue, for reference: https://github.com/rust-lang/rust/issues/35900

tomaka:
Another idea could be to do this:

```rust
#[derive(Shader)]
#[shader("some_file.glsl")]
struct Shader;
```

Which would build the content of the `Shader` struct and generate some methods.

While for shaders it might be confusing, I think this could be very nice to use for pipeline layout, descriptor sets, and structures defined in GLSL.

For example, in GLSL:

```glsl
struct SomeStruct {
    vec2 a;
    int b;
}
```

And then in Rust:

```rust
#[derive(ExtractFromShader)]
#[shader("shader.glsl")]
struct SomeStruct;
```

Which would be automatically expanded to:

```rust
SomeStruct {
    a: [f32; 2],
    b: i32,
}
```

I like this method a lot, because the user decides exactly which are the types that should be included in scope.

tomaka:
There's a problem however:

```glsl
[...]

PR #359, opened 2017-01-21
Add a small crate to allow inlining shader source codes

tomaka:
cc #256 

I don't know if I should feel proud or ashamed by this.

I'm not sure whether I want to keep this for the examples.


tomaka:
I think I'm going to leave the example as it is today, but add this new crate anyway.

tomaka:
Oh right this will need to wait for Rust 1.15 to merge.

tomaka:
I merged the branch in `incoming`.

PR #360, opened 2017-01-22
Rework command buffers again

### LINKED ISSUE #354: RFC: Yet another command buffers rework
A new idea I have had recently had about command buffers is to have a base `UnsafeCommandBuffer` type with layers around it.

In other words, you would have the following types:

- `UnsafeCommandBufferBuilder`, the most low level type that unsafely takes raw commands and doesn't check their validity. Has zero overhead.
- `StateCacheLayer<I>`, wraps around a command buffer builder and avoids duplicate state changes.
- `BufferedCommandsLayer<I>`, wraps around a command buffer builder and allows buffering and flushing commands (so that we can dynamically insert pipeline barriers before non-flushed commands).
- `AutoBarriersLayer<I>`, wraps around a command buffer builder and automatically adds pipeline barriers for safety. Uses `BufferCommandsLayer` under the hood.
- ... a few others

In the end you would manipulate for example a `StateCacheLayer<AutoBarriersLayer<UnsafeCommandBufferBuilder>>`.
A struct named `AutoCommandBufferBuilder` would hide this complexity and offer a nice API.

I had considered this design before, but there was a problem: in order to make it work, you would need a very complex trait with one of associated type and one method for each possible command that you can add to the builder. This is very cumbersome.

I have found a solution to this by creating one struct per command, each representing this single command.
In order words, you would also have the following structs:

- `CmdFillBuffer<B>`
- `CmdCopyImageToBuffer<I, B>`
- `CmdDraw<Gp, Ds, Pc, Vb, Ib>`
- ...and so on (about 30 structs)

In order to create a struct that represents a command, the command must be valid. In other words for example you can't create a command that writes 1024 bytes to 512-bytes buffer because the struct's constructor would return an error.
Note that a command can be valid by itself, but invalid in a given context. For example you can't write to a buffer when inside a render pass, and you can't draw if you're submitting the command buffer to a queue that only has transfer capabilities. Checking the context is done when adding the command to the builder (see below).

Once a struct representing a command has been constructed, it can be added to a command buffer with the `AddCommand` trait:

```rust
pub unsafe trait AddCommand<Cmd> {
    type Out;
    fn add(self, cmd: Cmd) -> Result<Self::Out, (Self, AddCommandError)>;
}
```

This makes it possible for some command buffer builders to not implement some traits. For example a hypothetical `SecondaryCommandBufferBuilder` would not implement `AddCommand<CmdExecuteCommands>` because this command is forbidden in secondary command buffers.

## Small exception

The design specifies that creating a command struct would ensure the safety of the command.

Because drawing or dispatching a compute shader requires a sequences of multiple commands, they need a small special treatment.

For dispatching:

- The `CmdUnsafeDispatch` command would be unsafe and would simply call `vkCmdDispatch` without checking that the correct descriptor sets/push constants are bound.
- The `CmdDispatch` command would be safe but would consist in a succession of several commands (binding descriptor sets, binding push constants, then actual dispatch).

The various layers would only implement `AddCommand<CmdUnsafeDispatch>` themselves, while `AddCommand<CmdDispatch>` would be automatically implemented on any `T` that implements `AddCommand<CmdUnsafeDispatch> + AddCommand<CmdBindDescriptorSets> + AddCommand<CmdPushConstants>`.

The same applies for drawing.

## Advantages

- Makes it possible for users to completely customize the behavior of command buffers recording without having to reimplement things themselves.
- Considerably reduces the complexity on the implementation side by having each layer do one precise thing.
- We can provide alternate ways to do the same thing. For example the AutoBarriersLayer can be implemented in various ways, each way having its own trade-off.
- We can easily add new "commands" that wrap around other commands (similar to `CmdDispatch` described above), by writing `impl<T> AddCommand<MyCmd> for T where T: AddCommand<Cmd1> + AddCommand<Cmd2> { ... }` for example. The user could even create their own commands (for example a command that generates mipmaps on an image) in a safe way.

tomaka:
Close #354 

New API example: https://github.com/tomaka/vulkano/blob/caab5988349fe63d083b61267f54f911e04f349d/examples/src/bin/triangle.rs#L374-L397

For example here is the generated assembly when you build the command buffer in the triangle example: https://gist.github.com/tomaka/309cf8da547f93aaf445ca942c368136


tomaka:
:shipit:

Issue #361, opened 2017-01-23
Add Instance::with_loading_function for better cross-platform handling


tomaka:
Right now the Vulkan library is loaded in [the `loader` module](https://github.com/tomaka/vulkano/blob/21da4133f43ba7d989df3398f981fe84029452cb/vulkano/src/instance/loader.rs). Depending on the operating system the path to the library is different.

When trying to add support for MoltenVK, this became a problem because MoltenVK is a static library that must be linked at compile-time. PR #301 tweaks the `loader` module to have a totally different behavior on OSX.

However it would be great to not rely on any platform-specific behavior, in case the user wants to use a custom way of accessing Vulkan. Therefore I suggest adding `Instance::with_loading_function`:

```rust
impl Instance {
    pub unsafe fn with_loading_function(f: *const c_void, ... other params ...) -> Result<Instance, ...> {
        ...
    }
}
```

The function `f` passed to `with_loading_function` must be the implementation `vkGetInstanceProcAddr`, and will be used by vulkano to load everything.

The `Instance::new` function would be modified to use `Instance::with_loading_function` under the hood.

With this change `Instance::new` would try to work transparently on all major platforms, but the user could choose a custom implementation/behavior.

tomaka:
Unfortunately what I didn't think about is that getting the list of layers and extensions also requires loading the function pointers.
Therefore we will need some sort of intermediate object I think.

Example:

```rust
let vk_library = unsafe { LoadedLibrary::from_function(...) };
let layers = vk_library.layers_supported_by_core();
let instance = Instance::with_library(vk_library).unwrap();
```

PR #362, opened 2017-01-24
Add support for the Nintendo Switch!

tomaka:


tomaka:
Note that this only adds support for creating a surface.

Since I have no information about how to load the Vulkan implementation, it will probably not work out of the box. One may have to use the future `Instance::with_loading_function` function defined in #361.

PR #378, opened 2017-02-03
Show error symbol instead of only raw int value.

Yamakaky:


tomaka:
That's oddly specific for device creation. I don't want to introduce code in vulkano that helps people write an implementation.

Yamakaky:
What do you mean? That's just a more precise error, without adding new lines of code.

tomaka:
The code you modified only happens if the Vulkan implementation returns an error that it's not supposed to return. In other words it should basically never happen. `unreachable!()` would be appropriate as well for example.

There are tons of similar codes throughout vulkano, so I don't see why specifically device creation should show the enum value and not the other places.

Yamakaky:
Ok

PR #383, opened 2017-02-09
Use cargo test --all for travis

tomaka:


tomaka:
I just realized that the reason why travis fails is that all the tests marked as `#[should_panic]` return early because they can't load Vulkan.

Some clean ways to fix that would be for example to ignore the tests at runtime (which isn't possible in Rust).

A hacky way to solve that would be to simulate a panic with the same panic message as expected if Vulkan isn't available.

Issue #303, opened 2016-10-12
All examples crash when being resized


lilianmoraru:
rustc: `rustc 1.14.0-nightly (a3bc191b5 2016-10-10)`
Last `vulkano` commit: `c8bf001ea112932d383b486fae66ae66146c5710`
Vulkan version: `1.0.8`
`vulkaninfo`: https://gist.github.com/lilianmoraru/dbe8521411e538c0232b5fec6f500aa7

After going through some `DRI 3` issues(I tried vulkano examples before but they were crashing so I left. Now I tried again and saw from `vulkano` a message about `DRI 3`, so I searched and [fixed this issue](http://askubuntu.com/a/808876/22490)).

All the examples run, but when trying to resize them, they all reliably crash:

```
$ RUST_LOG=trace RUST_BACKTRACE=1 ./teapot                                                                                     
../../../../../src/intel/vulkan/anv_device.c:414: FINISHME: Get correct values for VkPhysicalDeviceLimits                                                
Using device: Intel(R) HD Graphics 530 (Skylake GT2) (type: IntegratedGpu)                                                                               
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: OutOfDate', ../src/libcore/result.rs:788                                         
stack backtrace:                                                                                                                                         
   1:     0x55f67d2d2ff9 - std::sys::backtrace::tracing::imp::write::h00e948915d1e4c72                                                                   
   2:     0x55f67d2d63fc - std::panicking::default_hook::_{{closure}}::h7b8a142818383fb8                                                                 
   3:     0x55f67d2d56a9 - std::panicking::default_hook::h41cf296f654245d7                                                                                                                                                                                                     
   4:     0x55f67d2d5d38 - std::panicking::rust_panic_with_hook::h4cbd7ca63ce1aee9                                                                                                                                                                                             
   5:     0x55f67d2d5b92 - std::panicking::begin_panic::h93672d0313d5e8e9                                                                                                                                                                                                      
   6:     0x55f67d2d5b00 - std::panicking::begin_panic_fmt::hd0daa02942245d81                                                                                                                                                                                                  
   7:     0x55f67d2d5a81 - rust_begin_unwind                                                                                                                                                                                                                                   
   8:     0x55f67d30ff2f - core::panicking::panic_fmt::hbfc935564d134c1b                                                                                                                                                                                                       
[...]

PR #384, opened 2017-02-10
Update triangle example with swapchain recreation

mikedilger:
This is the start of a series of commits to add swapchain recreation to the examples (see #303).

Before I do the rest of them, I'd like feedback on this.

I'm also confused about something.  Without this code, when I resize the window, nothing bad happens.  Is that just the particular undefined behavior of my nvidia driver?  Or has something happened where recreating the swapchain manually is no longer necessary by users of vulkano?


tomaka:
> I'm also confused about something. Without this code, when I resize the window, nothing bad happens. Is that just the particular undefined behavior of my nvidia driver? Or has something happened where recreating the swapchain manually is no longer necessary by users of vulkano?

Resizing a window usually does not trigger a out-of-date error for the swapchain (the specs don't say why such errors happen, but in practice it doesn't happen when resizing). Instead resizing a window will make the implementation return a "SUBOPTIMAL" success code which is not yet supported by vulkano.

Even if there was an error, there's no undefined behavior when drawing to an obsolete swapchain. It just doesn't work but it's allowed.

tomaka:
As for your PR, unfortunately it doesn't work because the framebuffer object holds an `Arc` to the old swapchain image.
You need to reorganize the code to recreate the framebuffers as well.

This is probably something that the doc of swapchain should mention eventually as well.

mikedilger:
> As for your PR, unfortunately it doesn't work because the framebuffer object holds an Arc to the old swapchain image.
> You need to reorganize the code to recreate the framebuffers as well.
 
Ok.  I notice in the example that the Command Buffers are created every loop iteration. Could it be rewritten so that the command buffers are only rebuilt when the swapchain/framebuffers are rebuilt?

tomaka:
Technically there's nothing wrong in keeping command buffers alive when you can.

However I think it confuses beginners to do so.
In a real-world program it is very often impossible to keep command buffers alive between frames, and I think it's a good thing to show in the example that it's not a mistake to recreate them every time.

mikedilger:
It seems a lot of things depend on the image dimensions:  depth buffer,  projection matrix (aspect ratio),  pipeline (viewport if Fixed).  Online tutorials also rebuild the render pass.  I find when I try to rebuild all this stuff, I wedge my graphics driver (nvidia under ubuntu).. and I'm worried about all the hard-power-offs I've had to do.  Maybe I should just back away from this and let someone with more experience in Vulkan take it on.

PR #388, opened 2017-02-17
Implement the new submission system

tomaka:


tomaka:
As usual with the `incoming` branch I'm tempted to merge this quickly.

This new system is so much better than the old one that there's no debate I think.
Not that the new system is perfect, but the old one was so bad that it is a clear improvement.

In order to go further in correctness I'd need to continue working on command buffers, which would be out of scope of this PR.

PR #395, opened 2017-03-05
Implement command pool trimming.

nicokoch:
As an exercise to re-familiarize myself with the vulkano codebase, I implemented trimming of `CommandPool`.

Work is based on https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_KHR_maintenance1
and https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#vkTrimCommandPoolKHR

### DIFF
--- a/vk-sys/src/lib.rs
+++ b/vk-sys/src/lib.rs
@@ -850,6 +850,7 @@ pub type CommandPoolResetFlagBits = u32;
 pub const COMMAND_POOL_RESET_RELEASE_RESOURCES_BIT: u32 = 0x00000001;
 pub type CommandPoolResetFlags = Flags;
 
+pub type CommandPoolTrimFlagsKHR = Flags;
 
 pub type CommandBufferUsageFlagBits = u32;
 pub const COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT: u32 = 0x00000001;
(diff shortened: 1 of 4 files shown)


tomaka:
I'd take the changes in `vk-sys` and in `extension.rs`, but the rest would conflict with the incoming branch (which is where work is being done).

nicokoch:
Oh wow I didn't notice. I will be fixing that up.

Recently, Khronos released a whole bunch of new extensions and, in  turn, new functions and enums.
I will add all of them to the appropriate places in vk-sys and extension.rs, without providing higher-level functions for now. Is this fine for you?

As a side note, I noticed this comment: https://github.com/tomaka/vulkano/blob/e315c56876d51c0f1332b4043580556cc13941c6/vulkano/src/sampler.rs#L90

`VK_KHR_mirror_clamp_to_edge` is definitely a device extension, and the appropriate extension must be enabled if you want to use it. So eventually, vulkano should implement the required checks when the `MirrorClampToEdge` is used.

tomaka:
> I will add all of them to the appropriate places in vk-sys and extension.rs, without providing higher-level functions for now. Is this fine for you?

Sure, that would be nice!

> Edit: What about vendor-specific extensions?

That's still a more or less open question. I'm going to open an issue dedicated to it with my thoughts.

> VK_KHR_mirror_clamp_to_edge is definitely a device extension

Ok. I wasn't sure because there was some sort of fiasco where this extension should have been in core but ended up not being in core or something like that which I can't recall exactly.

tomaka:
Opened #396

tomaka:
> Edit: What about vendor-specific extensions?

Oh in fact I realized there's nothing wrong in adding support for vendor-specific extensions in vk_sys. So #396 is irrelevant here.

nicokoch:
Implemented in #400

PR #397, opened 2017-03-06
Khr extensions support

nicokoch:
This PR adds low-level support for all `VK_KHR_*` extensions.

I also improved error reporting on device creation, if a requested extension is not present. (Previously it would only report `unexpected error value: -7`)

Some extensions, like `VK_KHR_shader_fdraw_parameters` do not add any new symbols to vk-sys, but are added to `extensions.rs` so that users are able to enable it.

This PR is against `incoming` branch.


vulkano/src/instance/extensions.rs
```diff
@@ -156,12 +156,18 @@ instance_extensions! {
...
     nn_vi_surface => b"VK_NN_vi_surface",
     ext_swapchain_colorspace => b"VK_EXT_swapchain_colorspace",
+    khr_get_physical_device_properties2 => b"VK_KHR_get_physical_device_properties2",
 }
 
 extensions! {
     DeviceExtensions,
     khr_swapchain => b"VK_KHR_swapchain",
     khr_display_swapchain => b"VK_KHR_display_swapchain",
+    khr_maintenance1 => b"VK_KHR_maintenance1",
```
tomaka: I'd prefer to add extensions to this list only if they are supported by vulkano, and not in advance.

tomaka:
Thanks for the work

nicokoch:
@tomaka done

tomaka:
Uuhhh, you didn't remove all of them.

nicokoch:
`khr_sampler_mirror_clamp_to_edge` and `khr_shader_draw_parameters` are supported though.

`khr_sampler_mirror_clamp_to_edge` is implemented in `SamplerAddressMode`, but cannot currently be used because the user cannot enable the extension. Now they can.

`khr_shader_draw_parameters` only adds support for new spir-v features. I'm pretty sure vulkano does not have to implement anything to support this extension. If user's use these features in their shaders, they need to be able to enable this extension.

tomaka:
> I'm pretty sure vulkano does not have to implement anything to support this extension.

It does. It's undefined behavior to use a feature in your shader that is not enabled, so vulkano checks that.

nicokoch:
Oh well nvm then. Ill remove that, and `khr_get_physical_device_properties2` too I suppose?

But mirror_clamp_to_edge can be kept?

tomaka:
> and khr_get_physical_device_properties2 too I suppose?

Yes!

> I also have a follow-up PR that checks for the extension in during Sampler creation

Ideally it's this other PR that adds the extension, but I guess it doesn't really matter.

PR #399, opened 2017-03-07
Add functions to query supported device extensions

nicokoch:
Similar to `InstanceExtension`s, supported `DeviceExtension`s should be queryable.

### DIFF
--- a/vulkano/src/instance/extensions.rs
+++ b/vulkano/src/instance/extensions.rs
@@ -8,7 +8,7 @@
 // according to those terms.
 
 use std::error;
-use std::ffi::CString;
+use std::ffi::{CString, CStr};
 use std::fmt;
 use std::ptr;
 use std::str;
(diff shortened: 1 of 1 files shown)


vulkano/src/instance/extensions.rs
```diff
@@ -143,6 +146,62 @@ macro_rules! instance_extensions {
...
+                    properties
+                };
+
+                let mut extensions = $sname::none();
+                for property in properties {
+                    let name = property.extensionName;
+                    $(
+                        // TODO: this is VERY inefficient
+                        // TODO: Check specVersion?
+                        let same = {
```
tomaka: This code should probably use `CStr::from_ptr`, then `to_bytes()` instead.

vulkano/src/instance/extensions.rs
```diff
@@ -113,16 +116,11 @@ macro_rules! instance_extensions {
 
                 let mut extensions = $sname::none();
                 for property in properties {
-                    let name = property.extensionName;
+                    use std::ffi::CStr;
```
tomaka: At the top of the module please :)

nicokoch: Oops, gimme a sec :)

PR #400, opened 2017-03-12
Khr maintenance1 support

nicokoch:
This patch series implements most of the stuff introduced by `VK_KHR_maintenance1`.
https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_KHR_maintenance1

I'm not really sure how to handle `VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT_KHR` , but I think it's fine to ignore for now.

### DIFF
--- a/vulkano/src/command_buffer/pool/sys.rs
+++ b/vulkano/src/command_buffer/pool/sys.rs
@@ -10,6 +10,8 @@
 use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
+use std::error;
+use std::fmt;
 use std::sync::Arc;
 use std::vec::IntoIter as VecIntoIter;
 use smallvec::SmallVec;
@@ -21,6 +23,7 @@ use device::Device;
 use OomError;
 use VulkanObject;
 use VulkanPointers;
+use Error;
 use check_errors;
 use vk;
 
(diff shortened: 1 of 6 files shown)


nicokoch:
@tomaka did this fall through the cracks?

tomaka:
No, I just noticed some things that I need to review first.

vulkano/src/command_buffer/pool/sys.rs
```diff
@@ -106,6 +109,22 @@ impl UnsafeCommandPool {
         Ok(())
     }
 
+    /// Trims a command pool, which recycles unused internal memory from the command pool back to the system.
+    ///
+    /// Command buffers allocated from the pool are not affected by trimming.
+    #[inline]
+    pub fn trim(&self) -> Result<(), CommandPoolTrimError> {
```
tomaka: This function should be marked unsafe because of thread safety.

vulkano/src/image/sys.rs
```diff
@@ -139,6 +139,14 @@ impl UnsafeImage {
...
                 return Err(ImageCreationError::UnsupportedUsage);
             }
+            if device.loaded_extensions().khr_maintenance1 {
+                if usage.transfer_source && (features & vk::FORMAT_FEATURE_TRANSFER_SRC_BIT_KHR == 0) {
+                    return Err(ImageCreationError::UnsupportedUsage);
+                }
+                if usage.transfer_dest && (features & vk::FORMAT_FEATURE_TRANSFER_DST_BIT_KHR == 0) {
+                    return Err(ImageCreationError::UnsupportedUsage);
+                }
+            }
```
tomaka: The biggest problem is here. I feel like a `else` is missing.
If I don't enable the extension, does that mean that copy operations must be supported for all image formats? That doesn't feel right.

tomaka: Oh nevermind. If the feature bits reported by the implementation are 0, then the format is not supported. In other words, enabling the extension makes it possible for the implementation to support additional formats.

PR #401, opened 2017-03-15
Update ordered_passes_renderpass!

DaseinPhaos:
This PR reverses the current aggregation order of `AttachmentsList` inside `ordered_passes_renderpass!`.

This macro offers a handy way to aggregate an `AttachmentsList`. However, the list provided actually produces attachments in the *reverse* order of the `attachments` specified in the original macro invocation.

Quick example (boilerplate omitted):

```rust
//...
let renderpass = Arc::new(single_pass_renderpass!(
    device.clone(),
    attachments: {
        color: {  /*...*/ },
        depth: { /*...*/ }
    },
    pass: { /*...*/ }
).unwrap());

// let color_image: ImageView = ...;
// let depth_image: ImageView = ...;

let attachments = renderpass.desc().start_attachments()
    .color(color_image.clone()).depth(depth_image.clone());
let atch_list = renderpass.check_attachments_list(attachments);

// Order Reversed:
assert_eq!(
    atch_list.raw_image_view_handles(),
     vec![depth_image.inner(), color_image.inner()]
);
```

I don't know if this is intentional, but it feels anti-intuitive to use (if possible to use at all). 

For example, the macro also provides a handy way to initialize a sequence of clear values. However, these two can't be used in couple as the sequences of images they should be operating on are in reverse order.

I came across this when tinkering with the teapot example under the `incoming` branch. As no checking has actually been done by the macro yet, using them in couple actually got its way pass the compiler, resulting in an unhandled runtime error.

Reversing the aggregation order as proposed by this PR shall resolve this problem.

### DIFF
--- a/vulkano/src/framebuffer/macros.rs
+++ b/vulkano/src/framebuffer/macros.rs
@@ -420,10 +420,10 @@ macro_rules! ordered_passes_renderpass {
         }
 
         impl<$($prev_params,)* $first_param> $next<$($prev_params,)* $first_param> {
-            fn check_attachments_list(self) -> Result<($first_param, $($prev_params,)*), FramebufferCreationError> {
+            fn check_attachments_list(self) -> Result<($($prev_params,)* $first_param), FramebufferCreationError> {
                 let ($($prev_params,)*) = try!(self.prev.check_attachments_list());
                 // FIXME: check attachment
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks, that's definitely not wanted.

PR #402, opened 2017-03-16
Restore CmdDrawIndexed

DaseinPhaos:
This PR is the result of my tinkering with the teapot example under the `incoming` branch. It
- restores `CmdDrawIndexed` based on the current design of `CmdDraw`.
- adds a method `draw_indexed` to trait `CommandBufferBuilder`.
- passes through relative `AddCommand`-`impl`s to relative layers and finally to `AutoCommandBufferBuilder`.
- makes the teapot example work again, based on the new triangle example.

I'm still new to vulkano, and I'm not sure whether this is the right way to go. Comments appreciated!


tomaka:
Looks good, thanks a lot!

PR #403, opened 2017-03-17
Add IntoBuffer and IntoImage traits

tomaka:
After the latest changes in submission stuff, whenever a resource is used in a submission it automatically gets locked and is unlocked only when the resource is destroyed.

This works nicely for "pool-like" types such as a `UploadBuffer` (yet to be actually written). The buffer would be subdivided into multiple "sub-buffers", and destroying the pseudo-object that represents of these subbuffers only frees the subbuffers (and not the buffer as a whole) and allows it to be reclaimed. Combined with the submission rework, this also unlocks the subbuffer.

However this doesn't work nicely for regular buffers. You usually want to pass an `Arc<MyBufferType>` for the buffer, and with the new submission stuff you would need to destroy the buffer entirely in order to unlock it. In other words once a buffer has been used in a submission it couldn't be used anymore.

In order to make it usable, we need an additional type named `MyBufferTypeAccess` that holds a `Arc<MyBufferType>` and that represents a GPU lock of the buffer. Destroying this object unlocks the buffer. But the consequence is that you'd need to call something like `my_buffer.gpu_access()` every time you use the buffer.

Therefore this PR adds new traits: `IntoBuffer`, `IntoImage` and `IntoImageView`, similar to `IntoIterator`. Using a buffer in a descriptor set, framebuffer or GPU command only requires a `Into*` trait. You don't need to call `gpu_access` manually, as it's done through these traits.

### DIFF
--- a/vulkano/src/buffer/cpu_access.rs
+++ b/vulkano/src/buffer/cpu_access.rs
@@ -37,6 +37,7 @@ use buffer::sys::UnsafeBuffer;
 use buffer::sys::Usage;
 use buffer::traits::Buffer;
 use buffer::traits::BufferInner;
+use buffer::traits::IntoBuffer;
 use buffer::traits::TypedBuffer;
 use device::Device;
 use device::DeviceOwned;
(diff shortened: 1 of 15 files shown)


tomaka:
We probably want to rename `Buffer` to `BufferGpuAccess` and `Image` to `ImageGpuAccess` eventually, but this can be done later in a documentation pass.

PR #407, opened 2017-04-02
khr_physical_device_properties_2

nicokoch:
There are still some places left in vulkano, where the old functions get called unconditionally.
One example is the Image initialization where we never call `vkGetPhysicalDeviceFormatProperties2KHR`. These can be fixed later and are not included in this PR.

### DIFF
--- a/vk-sys/src/lib.rs
+++ b/vk-sys/src/lib.rs
@@ -1108,6 +1108,7 @@ pub struct FormatProperties {
 }
 
 #[repr(C)]
+#[derive(Clone)]
 pub struct Extent3D {
     pub width: u32,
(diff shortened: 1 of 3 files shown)


tomaka:
I'm hesitating to merge because I don't like adding `#[derive]`s to items in vk-sys.

nicokoch:
Yeah I don't like that part, too.
Any suggestions for a workaround?

tomaka:
Instead of `vec![...; num]`, something like `(0 .. num).map(|_| ...).collect::<Vec<_>>()` should work as well.

tomaka:
Closing in favor of #505

PR #412, opened 2017-04-08
fix teapot example

DaseinPhaos:
The new `IntoImageView` design of `AttachmentImage` made the teapot example unworkable. This PR fixes it.

It also fixes a snake_case warning introduced by `ordered_passes_renderpass!`.


tomaka:
Thanks

Issue #409, opened 2017-04-03
Creating command buffers per frame leading to a memory leak


dwerner:
Using Vulkano 0.3.2

I have an application that constructs a scene graph and pulls together all the geometry and per-object transformations at render time, where I am using caching `CpuAccessibleBuffer` structs in a lookup. I am using `PrimaryCommandBufferBuilder` and performing `draw_inline()` with some push constants.

I try to compile all my commands into a single buffer per scene graph and include them in a single submission. I see this showing up in any size of scene, and not particularly faster with larger scenes, so I'm thinking it's not geometry data itself that's getting duped and held on to.

Regardless of the number of buffers, and my speculation, if I generate my command buffers each frame, it looks like I'm getting a memory leak. There is a limited amount of time before an OOM is imminent (GPU stops drawing), and I'm roughly certain that I'm not holding on to and duplicating data in my application. It should be noted that if I clear my lookup Vec of it's cached `CpuAccessibleBuffer`s, I see memory use drop back to the ambient level, even if it has grown significantly first.

Here, I've highlighted the memory use pattern (scene is staying the same in structure, with dynamic transformations, etc), and circled is where I free my cache of `Arc<CpuAccessibleBuffer>` - memory use drops to normal, but then ramps up again as I am keeping those around again.

![image](https://cloud.githubusercontent.com/assets/1497784/24596358/8f9dd40a-17f3-11e7-925e-15fbfe97a710.png)

Heap profiling with visual studio in windows, and judging by the graph of memory use, it looks like vulkano is holding on to a Vec, which roughly doubles in size. This fits what Vec is doing as items are added and it needs to meet the need and increases it's capacity. The method implicated is:

`vulkano::command_buffer::inner::submit<alloc::arc::Arc<vulkano::command_buffer::pool::standard::StandardCommandPool>>`

It's my understanding that I should be able to create command buffers each frame with Vulkan, admittedly I'm new to the domain. This could well be that I'm new to Vulkan, but I'm also happy to write up test cases, share code, if need be. Also, I found #126, so perhaps this is already addressed in `incoming`?

VS Heap Profiler, snapshot showing a single large allocation
![image](https://cloud.githubusercontent.com/assets/1497784/24596206/60dfe848-17f2-11e7-96f0-8f35436bde18.png)

dwerner:
Here you can see the idle behavior while hanging on to those `CpuAccessibleBuffer`s

[...]

PR #413, opened 2017-04-09
fixing memory leak in command buffers

dwerner:
Spent a good chunk of the day porting my application to `incoming` as per #409 - but I still found there was a memory leak at the end of the day. 

 `UnsafeCommandBuffer`'s `Drop` impl should return the command buffer to the correct pool for reuse.

Please let me know if you need changes, or if I've made a mistake, but I was pretty excited when this solved my memory leak.


tomaka:
I hesitated to merge this because there may be thread safety issues and the whole commands pool thing needs a big rework.
However we might as well merge anyway and rework later.

Thanks!

PR #426, opened 2017-04-22
CRLF -> LF

corngood:
I ran `dos2unix` on the files with CRLF or mixed line endings.

I'm going to see about enforcing it in Travis.


corngood:
I got the empty tree trick from http://peter.eisentraut.org/blog/2014/11/04/checking-whitespace-with-git/

Using the git whitespace check requires also removing trailing whitespace, conflict markers, blank lines at EOF, missing LF at EOF.

Do you think those are all acceptable requirements?

tomaka:
Not that I'm against the change, but as usual with these PRs it basically breaks all active branches for no good reason.

corngood:
Was there previous discussion about this?  I couldn't find anything obvious.  Sorry if it's a dup.

As for breakage do you just mean generating unnecessary conflicts?

tomaka:
> As for breakage do you just mean generating unnecessary conflicts?

Yes that's what I mean.

corngood:
Is there any of this stuff you would like to do?  I assume those .orig files weren't meant to be there.  What about checking PR diffs for new whitespace errors?

I can also prepare fixes for public branches if that helps.

The files with mixed line endings do cause problems with some editors, so IMO there is a reason (but maybe not a very good one) to fix those.  The purely CRLF files and the other whitespace errors are only a slight annoyance for people using git w/ default settings.

nallar:
It would help to add `*.orig` to `.gitignore`

To avoid further incorrect line endings you can create a `.gitattributes` file like this: https://gist.github.com/nallar/b0d1aeed491fe97571eadc1d89a018a4

rukai:
If you were to remake this PR for latest vulkano I would merge it. However at this point you've probably lost interest. Just let me know if that's the case and I'll do it myself.

rukai:
Superceded by https://github.com/vulkano-rs/vulkano/pull/1099

PR #438, opened 2017-05-06
Rework the fence future

tomaka:
This PR implements `GpuFuture` on `Arc<FenceSignalFuture>`, and reworks the `FenceSignalFuture` to be cleaner and to fix several problems.

Most notably, this solves a corner case when submitting a present operation would succeed and submitting
a dummy command buffer with just a fence would fail.

### DIFF
--- a/vulkano/src/sync/future.rs
+++ b/vulkano/src/sync/future.rs
@@ -8,8 +8,10 @@
 // according to those terms.
 
 use std::error::Error;
+use std::mem;
 use std::sync::Arc;
 use std::sync::Mutex;
+use std::sync::MutexGuard;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


tomaka:
The reason why implementing `GpuFuture` on `Arc<FenceSignalFuture>` specifically is important is that you may want to write something like this in your main loop:

```rust
let mut a_specific_frame_future = None;
let mut previous_frame_finished: Box<GpuFuture> = Box::new(DummyFuture::new()) as Box<_>;

loop {
    // Sends a draw operation after the future passed as parameter.
    let new_frame_finished_future = draw(previous_frame_finished);

    if some_condition {
        let f = Arc::new(new_frame_finished_future);
        a_specific_frame_future = Some(f.clone());
        previous_frame_finished = Box::new(f) as Box<_>;
    } else {
        previous_frame_finished = Box::new(new_frame_finished_future) as Box<_>;
    }
}
```

PR #445, opened 2017-05-08
Update vulkano-win and examples to winit 0.6

Ralith:
The current build of vulkano times out during `swapchain.acquire_next_image` for me after the first 2 to 60 frames or so in all the examples that needed updating, so I wasn't able to test this to my satisfaction, but it's simple enough so hopefully others can.


tomaka:
Thanks!

> The current build of vulkano times out during swapchain.acquire_next_image for me after the first 2 to 60 frames or so in all the examples

That's actually a bug in Mesa, if you specify an infinite timeout then there's no timeout. I thought it would be fixed quickly, but I'm probably going to add a workaround in the examples as many people probably experience the same thing.

PR #451, opened 2017-05-11
No longer require buffer content to be 'static

tomaka:


tomaka:
Obviously that doesn't mean that you can or should put non-static things in a buffer. It's just meant to simplify the API and trait requirements.

PR #452, opened 2017-05-12
Correctly check image layouts

tomaka:
- Correctly implement the submit_sync layer.
- Add a check for image layouts in the GpuFuture API.
- The SwapchainAcquire future returns an error if the layout of the swapchain image is not undefined or present.
- Querying resources access now returns a proper error.
- The layout guess of a render pass attachment is now `Undefined` if the load OP is `Clear` or `DontCare`.
- `CommandBuffer::execute()` now returns a `Result` and can fail (instead of panicking before this PR).


tomaka:
The design when it comes to performing the initial transition is still not implemented.

The idea is basically that the way you perform this initial transition depends on the type of the image.
For example if you use an `ImmutableImage`, then the initial transition would be performed when you perform the initial upload.
If you use a `StorageImage`, then the constructor of the image would build a command buffer that performs the transition or modifies an existing command buffer builder to add the command.
In other words, it's the implementation of `Image` that chooses how to behave.

The exact details are still unknown though.

tomaka:
One interesting change I just made is that if you don't specify the initial layout of a render pass attachment, it will now default to `Undefined` if the load operation is `Clear` or `DontCare`.

In other words vulkano will now tell the Vulkan implementation that the content of the image attached to the framebuffer may be garbage. Before this PR vulkano would just guess the layout, which may force a pipeline barrier before entering the render pass.

What this means is that if vulkano happened to guess right, then this change could make things very slightly slower, and if vulkano guessed wrong then it could make things very slightly faster. Of course that is purely speculative, and on nVidia for example there's no difference at all.

tomaka:
Another problem.
When an image operation is added to a command buffer, the auto_barriers layer will ask the image for a hint about the layout it is supposed to use before entering the command buffer and perform a transition if necessary.
But on the other hand right now the submit_sync layer will instead suppose that the image is already in the layout required by the operation, without asking for the hint.

If we change its behavior to use the hint, then it would be wrong to use the submit_sync layer alone without the auto_barriers.

There are several ways to solve that, and the cleaner way would be to merge the two layers (as their behavior are interlocked). But that would mean that people can't customize the overhead of command buffers as much as they want.
Instead I'm going to add a `behavior` field to the submit_sync layer in order to be able to choose whether the layer uses the layout hint of the image.

PR #461, opened 2017-05-14
Some work on CpuAccessibleBuffer

tomaka:
Preliminary work for a correct implementation.
Further work blocked on #460


tomaka:
Superceded by #567

PR #463, opened 2017-05-17
Allow injection of winit::Window to vulkano-win

dwerner:
Hello there, my second PR in the last while, wanted to say I really appreciate all the efforts on this lib and many other's you've contributed to the community.

My problem:

I am currently building vulkano in my project into a dll that I can hot-load at runtime, and I found an issue with the `user32::RegisterClassEx()` that is done in winit (init.rs) - and fought around trying to  find a good place to do `UnregisterClass`, but in the end decided that I would just hoist ownership of the window itself to my main.rs exe code. This PR is really just intended as a discussion point - would it make sense to allow injecting the window, as prepared by `winit::WindowBuilder` externally? In my case I chose to use `Arc<Mutex<Window>>`, and that's why I've chosen to implement the trait on that type.

Let me know what you think - but I needed some kind of way to prevent my dll from registering COM components and not deregistering them. I'm happy to change my approach as well, but I couldn't see a way to get this kind of functionality with winit + vulkano.


tomaka:
The problem I have with this is that the choice of `Arc<Mutex<Window>>` is a bit arbitrary.

In my opinion a better way to do it is to turn `pub struct Window { window: winit::Window, ... }` into `pub struct Window<W = winit::Window> { window: W, ... }`.
Since the window isn't accessed ever again by `Window`, except for the getter, I think it's fine to leave the choice to the user about what represents a window.

dwerner:
I tried to implement what you suggested, learned about default type parameters in the progress of that. One thing that does stand out is that it's difficult to implement the new trait `IntoVkWindow` without access to those private functions that implement. Also, I can't constrain the type. Let me know what you think, but in my opinion what I'm doing with the impls of IntoVkWindow seems repetitive, and I'd understand if you wouldn't want that in the library.

dwerner:
For what it's worth, here's my project consuming vulkano, and should explain the reasoning behind my desire for this PR if you are up for reading - https://github.com/dwerner/sg-engine

dwerner:
This is possible with a customized shim replacing vulkano_win.

PR #474, opened 2017-05-22
added code to get to OutOfDeviceMemory error

buzmeg:
Added the code required to get to the OutOfDeviceMemory error.

Will add the rest of the initialization code when that problem is solved.


tomaka:
Thanks for the PR. You will need some platform-specific shenanigans though, as the new code should only be enabled on OSX. We should add some travis OSX builds.

tomaka:
I worked on it and opened #482 instead. Hopefully I didn't break anything.

PR #482, opened 2017-05-26
MoltenVK: Follow-up for added code to get to OutOfDeviceMemory error

tomaka:


tomaka:
Err... obviously the travis build fails because it cannot find MoltenVK.

tomaka:
I removed the travis OSX build and opened #483 instead.

PR #494, opened 2017-06-04
Support enabling statically unknown extensions

Ralith:
This allows vulkano to be used with libraries like OpenVR which require access to a vulkan instance with a set of extensions that can only be known at runtime.

OpenVR is dynamically loaded and is updated independently of applications, so it can easily come to require extensions which did not exist at the time an application was released. See [the OpenVR Vulkan docs](https://github.com/ValveSoftware/openvr/wiki/Vulkan#extensions) for details.

This is a (small) breaking change due to the removal of the `Copy` trait from the extension structs.

### DIFF
--- a/vulkano/src/device.rs
+++ b/vulkano/src/device.rs
@@ -120,7 +120,7 @@ use VulkanPointers;
 use check_errors;
 use vk;
 
-pub use instance::DeviceExtensions;
+pub use instance::{DeviceExtensions, RawDeviceExtensions};
 
 /// Represents a Vulkan context.
 pub struct Device {
@@ -161,10 +161,11 @@ impl Device {
     ///
     // TODO: return Arc<Queue> and handle synchronization in the Queue
     // TODO: should take the PhysicalDevice by value
-    pub fn new<'a, I>(phys: &'a PhysicalDevice, requested_features: &Features,
-                      extensions: &DeviceExtensions, queue_families: I)
-                      -> Result<(Arc<Device>, QueuesIter), DeviceCreationError>
-        where I: IntoIterator<Item = (QueueFamily<'a>, f32)>
+    pub fn new<'a, I, Ext>(phys: &'a PhysicalDevice, requested_features: &Features,
+                               extensions: Ext, queue_families: I)
+                           -> Result<(Arc<Device>, QueuesIter), DeviceCreationError>
+        where I: IntoIterator<Item = (QueueFamily<'a>, f32)>,
+              Ext: Into<RawDeviceExtensions>,
     {
         let queue_families = queue_families.into_iter();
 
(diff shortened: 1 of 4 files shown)


tomaka:
I think this is a bit delicate. The biggest problem I see is that we really want to avoid having a known extension in `other`.
But avoiding this means that we would have to make the field private, which means that the struct can no longer be constructed with the `InstanceExtension { foo: true, .. InstanceExtensions::none() }` syntax.

Ralith:
What problem are we trying to avoid by keeping known extensions out of `other`? Would it be solved by running a normalization pass on it before further processing anywhere that consumes an extensions struct?

tomaka:
The specific problems I see is that a user may want to enable a known extension by putting it in `other`.
We also probably want to avoid passing duplicates to the vulkan implementation, and this kind of corner cases.

Instead of adding a field named `other`, I have an other idea. Let's create another struct named `RawInstanceExtensions` that contains the unparsed list of extensions as CStrings. `InstanceExtensions` would implement the `Into<RawInstanceExtensions>` trait, and `Instance::new` would accept any `E` where `E: Into<RawInstanceExtensions>`. We could then also add other utilities to `RawInstanceExtensions`. This would also make the process more explicit.
Same for the device.

Tell me what you think.

Ralith:
That works for me. It seems slightly less conducive to querying unknown extensions, but I think the usefulness of that is limited anyway. I'll draft an implementation after I get some sleep.

tomaka:
> It seems slightly less conducive to querying unknown extensions

Similar to `InstanceExtensions::supported_by_core`, you would have `RawInstanceExtensions::supported_by_core`.
You could then call `InstanceExtensions::parse_from_raw` (or a similar name) that would return a tuple of both an `InstanceExtensions` struct and an iterator for the unrecognized extensions.

Ralith:
Implemented the scheme discussed. I opted for `impl<'a> From<&'a InstanceExtensions>` because there's no benefit to taking ownership and it dramatically reduces breakage.

Some of the raw methods are quadratic when they don't have to be, but I figured n is expected to be small enough that the complexity and overhead probably isn't worthwhile. If you feel otherwise, or that assumption proves false in the future, it should be possible to swap them out without breaking the interface.

I haven't tested this yet.

Ralith:
Went ahead with `HashSet` because at that point it was easier.

tomaka:
After thinking a bit, I think `RawDeviceExtensions` and `RawInstanceExtensions` should only be one struct. After all there's no real difference between the two. And instead of the name `RawExtensions`, I think something like `UnparsedExtensionsList` would be better.

Ralith:
I think it's useful to maintain a distinction between the two. You wouldn't want to inadvertently provide device extensions to an interface expecting instance extensions, or vis versa, and operations like intersection or comparison don't make sense. The difference is much the same as between the existing non-raw structs, we just can't statically bound the possible values.

tomaka:
Yeah you're right because of the `Into` impl.
Thanks!

PR #496, opened 2017-06-04
Remove dead data causing update_buffer to be unusable

### LINKED ISSUE #493: CommandBufferBuilder::update_buffer is unusable
[`update_buffer`](https://docs.rs/vulkano/0.4.0/vulkano/command_buffer/trait.CommandBufferBuilder.html#method.update_buffer) requires `AddCommand<CmdUpdateBuffer<B, D>>` requires `CmdUpdateBuffer<B, D>: Send + Sync + 'static`, but [`CmdUpdateBuffer`](https://docs.rs/vulkano/0.4.0/vulkano/command_buffer/commands_raw/struct.CmdUpdateBuffer.html) is never `Send` or `Sync` because [it contains a raw pointer](https://docs.rs/vulkano/0.4.0/src/vulkano/command_buffer/commands_raw/update_buffer.rs.html#44) and has no explicit implementation of those traits. I'm not sure what invariants are supposed to apply here or I'd make a PR myself.

This is the one blocking issue that I ran into updating my code from 0.3.2.

Ralith:
Fixes #493

### DIFF
--- a/vulkano/src/command_buffer/commands_raw/update_buffer.rs
+++ b/vulkano/src/command_buffer/commands_raw/update_buffer.rs
@@ -10,8 +10,6 @@
 use std::error;
 use std::fmt;
 use std::sync::Arc;
-use std::os::raw::c_void;
-use std::ptr;
 
 use buffer::Buffer;
 use buffer::BufferAccess;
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks

PR #501, opened 2017-06-05
Allow specifying a filename for VulkanoShader src

YeyaSwizaw:
This is a small change that allows specifying

```rust
#[src = "filename"]
```

when deriving a shader, instead of placing the shader code directly in the attribute.

It checks if the string is a valid file, and if it is it attempts to read the source from the file. Otherwise, it treats the string as the shader source as before.

### DIFF
--- a/vulkano-shader-derive/src/lib.rs
+++ b/vulkano-shader-derive/src/lib.rs
@@ -36,7 +36,7 @@ pub fn derive(input: TokenStream) -> TokenStream {
         let source = iter.next().expect("No source attribute given ; put #[src = \"...\"] or #[path = \"...\"]");
 
         if iter.next().is_some() {
-            panic!("Multilpe src or path attributes given ; please provide only one");
+            panic!("Multiple src or path attributes given ; please provide only one");
         }
 
         match source {
--- a/vulkano-shader-derive/src/lib.rs
+++ b/vulkano-shader-derive/src/lib.rs
@@ -9,29 +9,51 @@ use std::io::Read;
 
 use proc_macro::TokenStream;
 
-#[proc_macro_derive(VulkanoShader, attributes(src, ty))]
+enum SourceKind {
+    Src(String),
+    Path(String),
+}
+
+#[proc_macro_derive(VulkanoShader, attributes(src, path, ty))]
 pub fn derive(input: TokenStream) -> TokenStream {
     let syn_item = syn::parse_macro_input(&input.to_string()).unwrap();
 
     let src = {
-        let src = syn_item.attrs.iter().filter_map(|attr| {
+        let mut iter = syn_item.attrs.iter().filter_map(|attr| {
             match attr.value {
                 syn::MetaItem::NameValue(ref i, syn::Lit::Str(ref val, _)) if i == "src" => {
-                    Some(val.clone())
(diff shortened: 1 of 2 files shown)


tomaka:
If I remember correctly, I didn't implement this on purpose because it doesn't really doesn't work well.

I don't remember the exact details, but I think you have to specify the path relative to the Cargo.toml, while users would expect the path to be relative to the current directory instead.

Another problem I think is that if you run `cargo build` at the root of a Cargo workspace, then it won't find the file or something.

Maybe these have been fixed now?

YeyaSwizaw:
Hmm, I was aware of the first problem however didn't see it as a major issue, especially if it is documented. I had neglected to test in a workspace however, and yeah, that does appear to be a problem unfortunately.

The easy fix would be some way of getting the current file inside a proc_macro, however I can't find a way to do so. Is that really not possible at the moment?

YeyaSwizaw:
I found the conversation in the macros tracking issue mentioning `env::var("CARGO_MANIFEST_DIR")` and added it as it does solve the workspace problem.

The annoyance of it needing to be relative to that is still there, though

tomaka:
Ok, I guess it's ok then.
However I'd prefer if this didn't reuse `#[src]`. I'd prefer something like `#[path = "foo.glsl"]` instead, and the code should panic if there is both a `#[path]` and a `#[src]`.

YeyaSwizaw:
This was something I thought about but decided againist due to simplicity and because I didn't initially implement this with the intent to submit a pull request.

It's changed now though; I do think a separate attribute is probably better. I also updated the readme this time to actually mention the addition and the fact the path is relative to the project instead of the file.

vulkano-shader-derive/src/lib.rs
```diff
@@ -9,29 +9,51 @@ use std::io::Read;
...
+            SourceKind::Path(path) => {
+                let root = std::env::var("CARGO_MANIFEST_DIR").unwrap_or(".".into());
+                let path = Path::new(&root).join(&path);
+
+                if path.is_file() {
+                    let mut buf = String::new();
+                    File::open(path).and_then(|mut file| file.read_to_string(&mut buf)).expect("Unable to read source from given file");
+                    buf
+                } else {
+                    panic!("Given source file does not exist");
```
tomaka: Please put the path in the error message. Also I think we should mention here that the path is relative to the Cargo.toml, because it's really surprising in my opinion.
For example `Source file {:?} doesn't exist ; keep in mind that the path must be relative to your Cargo.toml`.

tomaka:
Should be good after the small change.

YeyaSwizaw:
It's currently showing the relative path specified in the attribute in the error: I'm not sure whether showing the full path or both would be preferable instead.

tomaka:
It's good I think, thanks!

YeyaSwizaw:
Thanks!

Mixthos:
Maybe you should require the path to begin with `/`. This would make it more obvious that the path is "absolute" from the crate root. It would also be forwards compatible if it ever becomes possible to use relative paths in attributes.

YeyaSwizaw:
I think that would make it seem too much like a true absolute path, though I agree some form of future-proofing is probably a good idea

PR #502, opened 2017-06-06
Fix actual type not being a reference

lloydmeta:
[`from_macos_moltenvk`](https://github.com/tomaka/vulkano/blob/da6faac1f32567b0b538533285cb683b4c00faff/vulkano/src/swapchain/surface.rs#L376) expects a `&Arc` instead of `Arc`.


tomaka:
Thanks!

tomaka:
`from_ios_moltenvk` should also be fixed eventually.

tomaka:
Oh wait, I misread your PR. The problem is that the function expects an `&Arc`. That shouldn't be the case.

lloydmeta:
Oh ok, was not sure what the intent was; I'll send a followup PR.

PR #503, opened 2017-06-06
Fix functions taking ref instead of value

lloydmeta:
As discussed [here](https://github.com/tomaka/vulkano/pull/502#issuecomment-306452974), it should have been the function param type that was changed as opposed to the callsite.


tomaka:
Thanks!

Issue #497, opened 2017-06-04
Discussion: command buffers design


tomaka:
# Introduction

If you're not familiar with Vulkan, a command buffer is a list of commands that can be executed by the GPU.
The purpose of this issue is to find a convenient way for the user to be able to create a struct that represents a command buffer, and this struct must implement [the `CommandBuffer` trait](https://github.com/tomaka/vulkano/blob/3b20aed99007ac71bd295506492e25f4489085e6/vulkano/src/command_buffer/traits.rs#L39-L143).

A brief description of this trait:

- There are only four methods that need to be implemented.
- The `inner()` method is an implementation detail that makes the library work. Nothing special to say about it.
- The `prepare_submit()` must check whether the command buffer is able to be executed at a given moment on a given queue. The implementation is supposed to make sure that all the resources it accesses are available from that queue and at this given moment, and that the command buffer itself is allowed to be executed on the given queue.
- The `check_buffer_access()` and `check_image_access()` methods exist to tell the caller whether the command buffer accesses or modifies a given resource during its execution, and if so at which pipeline stage.
- The `// FIXME: lots of other methods` comment refers to several safety details and to secondary command buffers, which isn't the priority here.

As you can see this trait is pretty loose and can be implemented in various ways as long as you go through an `UnsafeCommandBuffer` (as this is required by `inner()`).

The `UnsafeCommandBuffer` type is, as its name tells, the lowest level object that vulkano provides and that can be used to create command buffers.

# Safety challenges

Here is an exhaustive list of what an implementation of a command buffer must do in order to correctly use an UnsafeCommandBufferBuilder:

- It must keep alive all the buffers, images, framebuffers, render passes, pipeline objects, and descriptors that it uses during its execution. Since the command buffer object itself is going to be kept alive during this execution, all the command buffer has to do is hold ownership of these resources.
- It must make sure that the device of the buffers/images/etc. is the same as the device of the command buffer.
- It must make sure that the commands are supported by the queue family of the command buffer (each command buffer has a precise queue family and can only be executed on this family).
- Some commands must be executed in a precise context. For example the commands that modify the state must only do so after a pipeline object has been bound and if this pipeline object has the corresponding states marked as dynamic. Similarly if you add a draw command for example, then the currently bound vertex buffers must match the currently bound pipeline. In summary you can't just accept any command in any order.
[...]

Issue #385, opened 2017-02-11
RFC: Submissions design rework


tomaka:
# Summary

Under this new proposal you would not just simply submit a command buffer, but submit a
command buffer *after* a previously submitted command buffer. In other words, you wouldn't just
call `cb.submit()` but `cb.exec_after(previous_submission)`.

Of course it would still be possible to submit a command buffer alone, because when a program
submits its first ever command buffer there is no previous submission. However that would be
the exception rather than the rule.

Handling resources synchronization would be done in two steps:

- First you acquire a kind of "lock" on a resource in order to start using it from the GPU, and
  pass it to either to the first command buffer of a chain or to all the command buffers of a chain.
- Then you can contine using this resource lock in all the command buffers of that chain.

# Futures

This proposal introduces a trait named `GpuFuture` which represents an event that will happen in
the future. The `Gpu` prefix is here because the GPU is usually involved, but it's not always the
case.

We'll see later the actual definition of the `GpuFuture` trait.

So which structs implement `GpuFuture`?

- When you submit a command buffer to a queue, you obtain a `CommandBufferExecFuture` struct
  that represents the moment when the queue finishes executing it.
- You can ask the GPU to signal a semaphore, in which case you obtain a `SemaphoreSignalFuture`
  struct that represents the moment when the semaphore is going to be signaled.
- You can also signal multiple semaphores at once, in which case you obtain a
  `MultiSemaphoreSignalFuture`.
- You can ask the GPU to signal a fence, in which case you obtain a `FenceSignalFuture` struct
  that represents the moment when the fence is going to be signaled.
- When you acquire an image from a swapchain, you obtain a `SwapchainAcquireFuture` that represents
  the moment when the GPU acquires the lock on the swapchain image.
- You can join multiple futures into one with a `JoinFuture`.

As explained in the summary, the idea behind this design is that most of the time you will ask the
GPU to execute things after a given point in time represented by a future.

For example if you call `cb.exec_after(sem_sig)` where `sem_sig` is a future that represents a
semaphore being signaled, then the command buffer will only start executing after the semaphore is
signaled. Behind the scenes the `exec_after` function adds the semaphore as a waiting semaphore
when it submits the command buffer to the Vulkan implementation.

# Submitting

Calling `command_buffer.exec_after(...)` would return a future that represents the moment when the
command buffer has finished executing, but doesn't actually submit the execution immediately.
Similarly you can add actions to execute after a future (eg. execute another command buffer or
[...]

PR #508, opened 2017-06-09
Set default final color attachment layout to PresentSrc

Ralith:
I'm not certain whether this change is correct in a bigger-picture sense, so please review carefully, but it silences a validation error. For more advanced uses (e.g. deferred rendering) this may not be the correct setting, but those cases seem more likely to use custom render passes regardless, or perhaps more customization points are called for in this macro.

### DIFF
--- a/vulkano/src/framebuffer/macros.rs
+++ b/vulkano/src/framebuffer/macros.rs
@@ -276,7 +276,7 @@ macro_rules! ordered_passes_renderpass {
                             if initial_layout.is_none() {
                                 initial_layout = Some(ImageLayout::ColorAttachmentOptimal);
                             }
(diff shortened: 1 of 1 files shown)


tomaka:
That's not a correct fix. Image layouts are a complicated problem that I've been tackling with #497 and #385.

Ralith:
Suspected as much. Is there a way for an application targeting current vulkano to not incur undefined behavior, in the mean time?

tomaka:
When you declare the render pass with the macro, you can specify `initial_layout` and `final_layout` for each attachment, so you can do the same thing as your fix.

When it comes to `ImmutableImage`, unfortunately no. But the immutable image is a "separate" problem, in the sense that it's not blocked by any API design but because it wasn't switched to the new design yet.

Issue #381, opened 2017-02-08
Use the validation layers in the tests


tomaka:
If the validation layers are present on the user's system, the tests should enable them and panic if an error is reported.

The problem is that right now each test does:

```rust
#[test]
fn test() {
    let (device, queue) = gfx_dev_and_queue!();
    ...
}
```

But since the debug callback is an RAII object, it has to be kept alive somehow.
If we mem::forget this object, then the instance will never be destroyed.

PR #509, opened 2017-06-09
Enable validation layers in the examples

Ralith:
This reveals two apparent bugs in vulkano:

> ERROR: DS: Cannot submit cmd buffer using image (0xc) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout VK_IMAGE_LAYOUT_UNDEFINED when first use is VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL.

Appears to arise from the use of `copy_buffer_to_image` targeting an `ImmutableImage`. It's unclear to me how the layouts of `ImmutableImage`s are meant to be managed, but the error suggests that the barriers necessary to transition them aren't being properly configured or installed. I'd be happy to fix this if I can work that out.
> ERROR: DS: Images passed to present must be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR but is in VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL. For more information refer to Vulkan Spec Section '30.6. WSI Swapchain' which states 'Any given element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout at the time the operation is executed on a VkDevice' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/xhtml/vkspec.html#VkPresentInfoKHR)

Arises from presenting. #508 silences this error.

### DIFF
--- a/examples/src/bin/image.rs
+++ b/examples/src/bin/image.rs
@@ -20,6 +20,7 @@ extern crate vulkano_win;
 use vulkano_win::VkSurfaceBuild;
 use vulkano::command_buffer::CommandBufferBuilder;
 use vulkano::sync::GpuFuture;
+use vulkano::instance::debug::{self, DebugCallback};
 
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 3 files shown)


tomaka:
The two errors are known and expected unfortunately.

As for enabling the validation layers in the examples, not only does this complicate the examples (which should be simple, since they are examples) but I also don't want to force people to install the SDK.
Instead I'd prefer doing #381.

Note that if you want to enable the validation layers, you can also set the `VK_INSTANCE_LAYERS` environment variable to a list of layers separated with `;`.

Ralith:
Do the validation layers not ship with the loader on windows?

tomaka:
No, they are in the SDK.

PR #513, opened 2017-06-10
Add a GraphicsPipelineBuilder

tomaka:
Should make the API more ergonomic and removes the necessity to import all these small structs.

Not a breaking change, as the old API is only deprecated and not removed.

### DIFF
--- a/examples/src/bin/image.rs
+++ b/examples/src/bin/image.rs
@@ -125,32 +125,19 @@ fn main() {
                                                  vulkano::sampler::SamplerAddressMode::Repeat,
                                                  0.0, 1.0, 0.0, 0.0).unwrap();
 
-    let pipeline = Arc::new(vulkano::pipeline::GraphicsPipeline::new(device.clone(), vulkano::pipeline::GraphicsPipelineParams {
-        vertex_input: vulkano::pipeline::vertex::SingleBufferDefinition::new(),
-        vertex_shader: vs.main_entry_point(),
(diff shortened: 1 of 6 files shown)


tomaka:
Should add specialization constants before merging, otherwise we are going to need to break the API.

dylanede:
How will this interact with pipeline derivatives? What about pipeline caches?

tomaka:
These are good points.

There are two aspects when it comes to pipeline derivates:
- You can create a pipeline that is a derivate of a previously-created pipeline. This can easily be added with a method on the builder.
- You can pass multiple `VkGraphicsPipelineCreateInfo ` at once to one call to `vkCreateGraphicsPipelines`, and indicate that some of them are derivates of others. However this one is much, much harder because you typically want to share the same intermediate structs (`VkPipelineShaderStageCreateInfo`, `VkPipelineVertexInputStateCreateInfo`, and so on) between multiple create infos.

Unfortunately I have really no idea for now how to do the latter properly. My opinion is to just leave it for later, without caring about an API break at this point.

When it comes to pipeline caches, since pipeline caches are part of the call to `vkCreateGraphicsPipelines` and not to the `VkGraphicsPipelineCreateInfo`, I think this should be added as an alternative to `build()`. In other words, add a method `build_with_cache()` for example.

tomaka:
These two things would need to be added later though. They are not supported by the current system, and for now the builder is just a wrapper around the current system.

PR #520, opened 2017-06-12
Support for taking differences of extension and feature sets

### LINKED ISSUE #518: Support `difference` operations on extensions and features structs
The main use of this would be for error reporting - it would make it easy to report which extensions/features are not supported by the device, (e.g. by subtracting the supported features from the required features to get the unsupported required features). At the moment we just have intersection operations supported.

dylanede:
Fixes #518


tomaka:
Thanks

PR #529, opened 2017-06-20
Fix warnings during build

GuildMasterInfinite:
Back with another pull request :smile:

I noticed there are a lot of warnings issued by the compiler during normal / Travis CI build. It clutters up the log and *probably* makes it harder to notice *real* errors. 

Short summary of the changes:
- removed unused imports / allowed dead code in the examples.
- added various #[cfg(...)] attributes to the imports in the `vulkano-win` crate (e.g. some `use` statements were used on Windows, but not on Linux).
- updated the versions of the libraries used by the `example`s. The `color-rs` crate was causing a lot of warnings in the log.
- replace the use of a deprecated function in the `swapchain` module.

<hr/>

Note, however, that there is one little problem: the old code used the [`surface.get_capabilities()`](https://docs.rs/vulkano/0.4.1/vulkano/swapchain/struct.Surface.html#method.get_capabilities) function, which panicked on errors. 

I've replaced it with the new [`surface.capabilities()`](https://docs.rs/vulkano/0.4.1/vulkano/swapchain/struct.Surface.html#method.capabilities) function, which returns a `CapabilitiesError`. 

However, because I didn't want to change the return type of `Swapchain::new_inner` I pass through `OomError`s and panic on `CapabilitiesError`, as before.

A decision should be made: should we change the public API of `Swapchain::new_inner`, and if so, which error type to use?

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -7,7 +7,7 @@ authors = ["Pierre Krieger <pierre.krieger1708@gmail.com>"]
 vulkano = { path = "../vulkano" }
 vulkano-shader-derive = { path = "../vulkano-shader-derive" }
 vulkano-win = { path = "../vulkano-win" }
-cgmath = "0.12.0"
-image = "0.6.1"
+cgmath = "0.14.1"
+image = "0.14.0"
 winit = "0.6.4"
-time = "0.1.35"
+time = "0.1.37"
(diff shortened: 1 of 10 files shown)


tomaka:
I think it's better to leave swapchain.rs as it is right now, instead of fixing it this way.
The other changes are good though.

GuildMasterInfinite:
Done. I reverted the changes for this pull request and [opened an  issue]() to track the `swapchain.rs` change.

tomaka:
Thanks

PR #538, opened 2017-06-21
Add SubmitBindSparseBuilder

tomaka:


tomaka:
This PR adds the low-level code that allows submitting a sparse binding operation to a queue.
Non-opaque binding to images has been left unimplemented because its API remains to be decided. It's most likely going to be the same API as image operations in command buffers, which will need to be decided sooner.

PR #542, opened 2017-06-23
Rework command buffers

### LINKED ISSUE #497: Discussion: command buffers design

tomaka:
Close #497 

See #497 for an explanation of what it does.


tomaka:
A side-effect of this change is that the `ImmutableBuffer::from_buffer_with_builder` constructor disappears, which is because the `CommandBufferBuilder` trait no longer exists.

tomaka:
Let's yolo-merge this. There's no point in leaving the PR sitting idle.

PR #543, opened 2017-06-23
replace occurences of `CpuAccessBuffer` with `CpuAccessibleBuffer`

Limeth:
replace occurences of `CpuAccessBuffer` with `CpuAccessibleBuffer` in docs.


tomaka:
Thanks

PR #544, opened 2017-06-23
Publish 0.4.4

tomaka:
I'm publishing this versions because from now on PRs will introduce breaking changes, which is the road to 0.5.

Changelog (minor): #538 #540


tomaka:
Also #541

PR #558, opened 2017-06-26
Add BufferAccess::unlock() and ImageAccess::unlock()

### LINKED ISSUE #519: Presenting a swapchain image that has not been written to results in invalid API usage
In other words, if you have an "empty" render loop that just presents images that have never been rendered to, you get error messages like this from the debug layers:

```text
MEM warning: vkQueuePresentKHR(): Cannot read invalid swapchain image 0x2775c9be030, please fill the memory before using.
DS error: Images passed to present must be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR or VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR but is in VK_IMAGE_LAYOUT_UNDEFINED. For more information refer to Vulkan Spec Section '30.8. WSI Swapchain' which states 'Any given element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR or VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR layout at the time the operation is executed on a VkDevice' (https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkPresentInfoKHR)
```

tomaka:
Fix #519 

When this BufferAccess/ImageAccess thing was designed, the idea was that each access object would represent a pending lock on a resource. When you called `try_gpu_lock()`, the lock would become active and only the destructor of the access would unlock the resource.

However this has caused troubles because command buffers need to be able to run multiple times, and rebuilding the access objects every time you run a command buffer is not possible because we can't store a `Box<Buffer>` or `Box<Image>` as they have an associated type.
In addition to this, there was also a problem with the fact that framebuffers and descriptor sets already store access objects, and there was no way to unlock an access stored in a framebuffer or descriptor set once it was locked.

Therefore this PR removes the fact that the destructor is supposed to unlock the resource, and adds a manual `unlock()` method that does this job. The `SyncCommandBuffer` calls this method in its destructor.

A better solution would be to return some sort of lock whose destructor would perform the unlocking, but it would have the same problem as `Buffer`/`Image` in that we wouldn't be able to store a `Box<BufferAccess>` or `Box<ImageAccess>` anymore.

Right now command buffers can only be submitted once (submitting them multiple times will most likely runtime error), and in the future the API of command buffers will need to be slightly modified to allow running a command buffer multiple times.


tomaka:
A consequence is that the `Buffer` and `Image` traits no longer serve a purpose.
They are more or less equivalents to `IntoBufferAccess` and `IntoImageAccess`, so that one could pass an implementation of `Buffer` and obtain a lock (ie. an access) whose destructor would perform the unlocking.

Since the destructor no longer plays any role in this, the `BufferAccess`/`ImageAccess` traits could be directly implemented on buffers and images.

PR #560, opened 2017-06-27
Run rustfmt on the code

tomaka:
Adds a `.rustfmt.toml` config file that tries to match with the existing style, and runs rustfmt on the code.
This didn't touch `vk-sys` and `examples`.


tomaka:
Merging quickly so that I don't have any conflict to resolve for further work.

PR #579, opened 2017-07-02
 dst/dest inconsistency

### LINKED ISSUE #51: dst/dest inconsistency
Some parts of the library (eg. struct members or function names) use `dst` and some others use `dest` as a shorthand for "destination".

I suggest that they should just all be renamed to "destination", and "src" renamed to "source".

tyoc213:
This will fix  #51 dst/dest inconsistency 

Things that called my attention but didnt touch

```rust
enum SourceKind {
  Src(String),
  Path(String)
}
```

And things like
```rust
.map(|(src, de, dz)|{
}
// or

if let Some((src_queue, dest_queue))
```

Also I was not sure to touch (and didnt touch)

```rust
dual_src_blend=>dualSrcBlend,

// or

max_fragment_dual_src_attachments:u32=>maxFragmentDualSrcAttachments,
```

Should I change `UnsafeCommandBufferBuilderPipelineBarrier` ? because I haven't

```rust
pub struct UnsafeCommandBufferBuilderPipelineBarrier {
    src_stage_mask: vk::PipelineStageFlags,
    dst_stage_mask: vk::PipelineStageFlags,
    dependency_flags: vk::DependencyFlags,
    memory_barriers: SmallVec<[vk::MemoryBarrier; 2]>,
    buffer_barriers: SmallVec<[vk::BufferMemoryBarrier; 8]>,
    image_barriers: SmallVec<[vk::ImageMemoryBarrier; 8]>,
}
```

I also have run `cargo build` and `cargo test` inside vulkano folder and it goes OK.

I found other shorthands like `de`, `sr` or `src_queue`, `dest_queue` but didnt touch them.


tomaka:
Thanks, looks fine.

The issue was mostly about the public API of vulkano, so no need to modify the pipeline builder or the other things you mentionned.
For the feature names, it's just like enums, I think it would be confusing to change them.

PR #587, opened 2017-07-04
Publish 0.5.0

tomaka:
Also [vulkano.rs](http://vulkano.rs).


tomaka:
PRs merged since version 0.4.4: #541 #545 #542 #546 #547 #556 #554 #557 #558 #552 #551 #560 #561 #562 #563 #564 #565 #566 #568 #567 #570 #571 #573 #575 #576 #578 #579 #581 #580 #582 #583 #586

tomaka:
Most important: #542 #558 #567 #573 

In addition to this, there are several breaking changes like functions that take an `Arc` instead of an `&Arc` and similar things which should be easy to fix.

PR #594, opened 2017-07-04
Implement VulkanObject for Image types

Ralith:
Seems like these were missing. Needed for OpenVR.


Ralith:
Should these be `SynchronizedVulkanObject`?

tomaka:
I'm not sure if that's useful, as you can just call `my_image.inner().image.internal_object()`.

Ralith:
Whoops, missed that somehow, thanks!

PR #601, opened 2017-07-06
Additional extension helpers

Ralith:
Just some simple conveniences.

### DIFF
--- a/vulkano/src/instance/extensions.rs
+++ b/vulkano/src/instance/extensions.rs
@@ -13,6 +13,7 @@ use std::ffi::{CStr, CString};
 use std::fmt;
 use std::ptr;
 use std::str;
+use std::iter::FromIterator;
 
 use Error;
 use OomError;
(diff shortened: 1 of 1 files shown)


vulkano/src/instance/extensions.rs
```diff
@@ -136,6 +141,14 @@ macro_rules! extensions {
             }
         }
 
+        impl<I> From<I> for $rawname
```
tomaka: If you want to do that, you should implement the `FromIterator` trait instead.

Ralith: Whoops, fixed!

vulkano/src/instance/extensions.rs
```diff
@@ -136,6 +142,14 @@ macro_rules! extensions {
             }
         }
 
+        impl iter::FromIterator<CString> for $rawname {
```
tomaka: Minor style: use `FromIterator` instead of `iter::FromIterator`.

tomaka:
Will merge after nit.

Ralith:
Tweaked as requested.

Issue #512, opened 2017-06-10
Add DynamicDescriptorSets


tomaka:
Could get a better name, but I couldn't find one.

The idea is that a `DynamicDescriptorSets` would allocate 3, 4, or more descriptor sets (in a lazy way) that all have the same layout.
At each frame you could then call `dyn_desc_sets.build(...)`, which would lock a descriptor set from that list and write the descriptors to it.
Once a descriptor set is no longer in use by the GPU, it would be returned to the list.

This would be appropriate for situations where you need to change the bound descriptor at each frame, and avoids having to perform a costly allocation every time.

dylanede:
Sounds more like something that should have "pool" in the name, IMO.

PR #603, opened 2017-07-06
Add PersistentDescriptorSet

tomaka:
Replaces `simple_descriptor_set!`.

While using names for the descriptor fields is nice, it is too much of a hack.
This PR deprecates `simple_descriptor_set!` and replaces it with `PersistentDescriptorSet`.

Further things to do afterwards:
- Remove the `PipelineLayoutDescNames` trait (only once `simple_descriptor_set!` no longer exists).
- Add a wrapper around `PersistentDescriptorSet` that uses a special descriptor pool that allows reclaming descriptor sets that are "destroyed" by the user (#512).


tomaka:
Before merging this, all methods should add checks and return `Result`s. Otherwise we would need a breaking change later.

tomaka:
Should also add binding buffer views and samplers while we're at it.

PR #604, opened 2017-07-06
Allow query inheritance in UnsafeCommandBufferBuilder

tomaka:
Of course that doesn't mean that queries are actually usable. It's a first step.


tomaka:
Technically a breaking change, but since it's about an obscure internal part of the library I don't think it's a problem to break here.

PR #610, opened 2017-07-07
Fix #609

### LINKED ISSUE #609: vulkano-shaders: Using #[derive(VulkanoShader)] with geometry shaders gives compile errors
When I tried to compile this code:
```rust
mod color_gs {
    #[derive(VulkanoShader)]
    #[ty = "geometry"]
    #[path = "assets/color.geom"]
    struct _Dummy;
}
```
I got this compile time error:
```
error[E0061]: this function takes 5 parameters but 4 parameters were supplied
   --> src\main.rs:571:14
    |
571 |     #[derive(VulkanoShader)]
    |              ^^^^^^^^^^^^^ expected 5 parameters

error: aborting due to previous error(s)
```

Grzechovsky:
Just a suggestion how to fix #609 issue

### DIFF
--- a/vulkano-shaders/src/entry_point.rs
+++ b/vulkano-shaders/src/entry_point.rs
@@ -89,26 +89,26 @@ pub fn write_entry_point(doc: &parse::Spirv, instruction: &parse::Instruction) -
         },
 
         enums::ExecutionModel::ExecutionModelGeometry => {
-            let mut execution_mode = "Points"; // fallback value
-
+            let mut execution_mode = None;
+            
             for instruction in doc.instructions.iter() {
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks! Looks good.

However I'd prefer if there was no fallback value. In other words, `execution_mode` should be an option, and when writing the value we unwrap it.

tomaka:
Thanks, r+. Will merge when CI passes.

Grzechovsky:
I was thinking to do that back then, but I just haven't seen any unwrap in those source files, so I did wanted to introduce potentially dangerous code :D

PR #613, opened 2017-07-07
Fix resources being unlocked when they shouldn't be

tomaka:
Before this PR a command buffer would unlock its resources in its destructor, which is obviously a mistake because if the command buffer wasn't submitted then the resources weren't locked.

This change also makes it possible to submit the same command buffer multiple times, although you have to do that manually for now.


tomaka:
Just like some other PRs, this is technically a breaking change but since it touches a part of the library that users aren't supposed to touch then it's ok to break.

Issue #608, opened 2017-07-07
About buffer content in descriptor sets


tomaka:
Right now the buffer content of shaders isn't strong-typed. You can for example bind a buffer that contains a `Foo` while the shader reads a `Bar` from it.

I initially planned to describe the layout of buffers and compare the expected and actual layouts are runtime, but that would be cumbersome and expensive.

Instead I think a better approach is to just accept mismatches, provided that it isn't unsafe. This means two things:

- The buffer bounds aren't enforced, so the robustBufferAccess feature must be enabled.
- The content of the buffer must not have any invalid state. In other words, no reference, no str, no bools, no enums.

The second point could be handled by requiring a trait on the buffer content. This trait would be declared as `impl NoInvalidValue for .. {}` with some `!impl` definitions on str, bool, etc. Of course that can only be done once that syntax is stable in Rust (https://github.com/rust-lang/rust/issues/13231).

However there's the problem of signalling NaNs. Technically `f32` does have an invalid state, which are signalling NaNs. However I think we have to ignore this problem for now because forbidding `f32` would make the library too cumbersome.

tomaka:
The same problem arises for `fill_buffer`. We could accept any buffer whose content implements the `NoInvalidValue` trait, but it would be too cumbersome to forbid f32s.

tomaka:
TL;DR:

- Right now you can trigger undefined behavior by creating a buffer that contains bools, enums, strs, etc. and writing to it with vulkano.
- This could be improved by creating an OIBIT trait, but these aren't stable yet (https://github.com/rust-lang/rust/issues/13231). Also to be pedantic we would have to forbid f32s and f64s, which would be too cumbersome.
- In the end the proper fix is to have strongly-typed descriptor sets, which needs better integration of shaders (like #549).

fst3a:
In case of strongly-typed descriptor sets, there is currently no easy way to work with certain GLSL types (e.g. vec3) in compute shaders because, in case of, say, vec3, vec3 has alignment of 16 bytes but itself is only 12 bytes so user has to manually pad and align the structure containing vec3.  I would assume, this would make strongly typed descriptor set builder to complain.

PR #630, opened 2017-07-09
Remove DescriptorBufferContentDesc

tomaka:
See #608 

Initially the idea was to make sure that the layout of content of the buffer expected by the shader was the same as the actual layout.
This is now abandoned because too cumbersome and too "weird".


tomaka:
Technically a breaking change, but it's ok since it's an obscure part of the library.

tomaka:
We may wish to wait for vulkano 0.6 to merge this, as it requires an update of simultaneously `vulkano` and `vulkano-shaders`. In other words vulkano 0.5.3 wouldn't be compatible with vulkano-shaders 0.5.2.

PR #635, opened 2017-07-11
Keep track of allocation count

### LINKED ISSUE #588: Count the number of allocations for maxMemoryAllocationCount
I thought the driver would return a `VK_ERROR_TOO_MANY_OBJECTS`, but the validity section of `vkAllocateMemory` says that you must make sure you don't exceed `maxMemoryAllocationCount`.

nicokoch:
Fixes #588

### DIFF
--- a/vulkano/src/device.rs
+++ b/vulkano/src/device.rs
@@ -133,6 +133,7 @@ pub struct Device {
         Mutex<HashMap<u32, Weak<StandardCommandPool>, BuildHasherDefault<FnvHasher>>>,
     features: Features,
     extensions: DeviceExtensions,
+    allocation_count: Mutex<u32>,
 }
 
 // The `StandardCommandPool` type doesn't implement Send/Sync, so we have to manually reimplement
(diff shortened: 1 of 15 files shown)


vulkano/src/device.rs
```diff
@@ -411,6 +413,10 @@ impl Device {
             },
         }
     }
+
+    pub(crate) fn allocation_count(&self) -> &Mutex<u32> {
```
tomaka: Needs some documentation here as well, explaining that it is specific for memory allocations.

nicokoch:
Letting `alloc` return a new error type will have great implication for other parts of the library though (Returned error type of the `memory::pool` stuff will change, as well as other parts such as `ImageCreationError`, since all these can now result in TooManyObjects). Is this fine for you?

Additionally, how do I handle `Error::MemoryMapFailed` in `DeviceMemory::alloc_and_map`? Should it also be part of the new DeviceMemoryAllocError, or another layer on top (DeviceMemoryMapError)?

tomaka:
> Is this fine for you?

Yes!

> Additionally, how do I handle Error::MemoryMapFailed in DeviceMemory::alloc_and_map? Should it also be part of the new DeviceMemoryAllocError, or another layer on top (DeviceMemoryMapError)?

Looks like they sneaked this error in. I guess that it could be part of `DeviceMemoryAllocError` so that we only one error type for all memory allocations.

nicokoch:
ping @tomaka

tomaka:
Thanks!

PR #646, opened 2017-07-15
Make it possible to create and run secondary command buffers

tomaka:


tomaka:
Note that this is implemented in a "hacky" way because synchronization is going to be hard to implement I think.
I'm experimenting with something, which is that `execute_commands` is marked as `unsafe` until the safety checks are implemented. When it is ready the `unsafe` will be removed, which shouldn't be a breaking change.

PR #655, opened 2017-07-19
Use supports_compute() in compute example.

jasondavies:
Not sure if this is important, but `supports_compute()` seems more appropriate for this example?


tomaka:
Right, thanks!

PR #659, opened 2017-07-19
Fixes to the CpuBufferPool

tomaka:
- The code didn't take into account the offset of the memory that is bound to the buffer. In other words
it would always write your data at offset 0 of the memory instead at the correct location.

- When calling `next` or `chunk`, the CpuBufferPool will now correctly align the data to fulfil the requirements for uniform buffers or storage buffers if the corresponding usage was enabled in the constructor. Some debug asserts have been added as well.


tomaka:
Forgot to mention that this PR modifies the teapot example to use the CpuBufferPool, as it's the most appropriate buffer for continuously-uploaded data.

PR #667, opened 2017-07-20
Example showing how to manually load SPIR-V shaders from disk

fst3a:
This PR adds one simple example demonstrating how to write structures that describe shader layout, normally generated automatically by VulkanoShader derive macro and then load the shader from external source (disk). 

There are valid use cases when user wants to load shader files from external sources and this example aims to demonstrate how to do so.  It shows things normally done by VulkanoShader macro and explains why certain things are done.


fst3a:
Fixed, sorry for delay.

tomaka:
Thanks

PR #669, opened 2017-07-20
Correctly synchronize resources in descriptor sets

tomaka:
Before this PR the buffers and images used in descriptor sets weren't taken into account at all for pipeline barriers and image transitions. This fixes it.

This PR also prevents pipeline barriers from happening within a render pass. The Vulkan specs allow them under very specific conditions, but that will be for another day.

I basically went "full C" in this PR. I'm not sure that everything is correct, but it looks like it works and doesn't trigger any validation error.

### DIFF
--- a/vulkano/src/command_buffer/synced.rs
+++ b/vulkano/src/command_buffer/synced.rs
@@ -2040,7 +2040,7 @@ impl<'b, P> SyncCommandBufferBuilderBindDescriptorSets<'b, P> {
                     fn image(&self, mut num: usize) -> &ImageAccess {
                         for set in self.0.iter() {
                             if let Some(img) = set.image(num) {
-                                return img.0;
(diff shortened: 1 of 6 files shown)


tomaka:
In theory, after this PR vulkano should be able to detect when you bind the same buffer/image twice in the same descriptor set with one of the two bindings being writable.

tomaka:
I had to add a hack because of input attachments.
If you put an input attachment in a descriptor set, vulkano will detect that it is the same image as one of the attachments of the framebuffer that is currently bound and think that a pipeline barrier is needed (which is not the case).
Since pipeline barriers are kind-of forbidden within render passes, vulkano will return a conflict error.

In order to keep input attachments usable, they are ignored in descriptor sets right now for synchronization purposes.

PR #671, opened 2017-07-20
Added window resizing to examples

Elzair:
I added the ability to resize windows to all three graphical examples: triangle, image, and teapot. I tested them under X11, and resizing, maximizing, and moving across displays all work. 

The only weird thing I experienced is that moving the window too quickly across displays will sometimes lead to a tiny view of the scene in the upper-left corner and garbage in the rest of the window.


tomaka:
Thanks for the PR!

One things that bothers me is that the code for creating the framebuffers is duplicated.
Could you change it to use an `Option` instead? Ie. `framebuffers` gets the value `None` at initialization, and is created inside the loop if necessary.
What I mean is that when `recreate_swapchain` is true you set `framebuffers` to `None`, and below that you add `if framebuffers.is_none() { framebuffers = ... }`.

Elzair:
When I made the changes, the Rust compiler started complaining about "use of moved values"

```
error[E0382]: use of moved value: `framebuffers`
   --> src/bin/triangle.rs:371:12
    |
371 |         if framebuffers.is_none() {
    |            ^^^^^^^^^^^^ value used here after move
...
415 |             .begin_render_pass(framebuffers.unwrap()[image_num].clone(), false,
    |                                ------------ value moved here
    |
    = note: move occurs because `framebuffers` has type `std::option::Option<std::vec::Vec<std::sync::Arc<vulkano::framebuffer::Framebuffer<std::sync::Arc<vulkano::framebuffer::RenderPass<main::scope::CustomRenderPassDesc>>, ((), std::sync::Arc<vulkano::image::SwapchainImage>)>>>>`, which does not implement the `Copy` trait

error[E0382]: use of moved value: `framebuffers`
   --> src/bin/triangle.rs:377:31
    |
377 |             mem::replace(&mut framebuffers, new_framebuffers);
    |                               ^^^^^^^^^^^^ value used here after move
...
415 |             .begin_render_pass(framebuffers.unwrap()[image_num].clone(), false,
    |                                ------------ value moved here
    |
    = note: move occurs because `framebuffers` has type `std::option::Option<std::vec::Vec<std::sync::Arc<vulkano::framebuffer::Framebuffer<std::sync::Arc<vulkano::framebuffer::RenderPass<main::scope::CustomRenderPassDesc>>, ((), std::sync::Arc<vulkano::image::SwapchainImage>)>>>>`, which does not implement the `Copy` trait

error[E0382]: use of moved value: `framebuffers`
   --> src/bin/triangle.rs:415:32
    |
415 |             .begin_render_pass(framebuffers.unwrap()[image_num].clone(), false,
    |                                ^^^^^^^^^^^^ value moved here in previous iteration of loop
    |
    = note: move occurs because `framebuffers` has type `std::option::Option<std::vec::Vec<std::sync::Arc<vulkano::framebuffer::Framebuffer<std::sync::Arc<vulkano::framebuffer::RenderPass<main::scope::CustomRenderPassDesc>>, ((), std::sync::Arc<vulkano::image::SwapchainImage>)>>>>`, which does not implement the `Copy` trait
```

tomaka:
Try using `as_ref().unwrap()` instead of `unwrap()`

Elzair:
Thanks. Here are the changes I have made so far to the triangle example:

```
diff --git a/examples/src/bin/triangle.rs b/examples/src/bin/triangle.rs
index 824f823b..efdfddbe 100644
--- a/examples/src/bin/triangle.rs
+++ b/examples/src/bin/triangle.rs
@@ -306,11 +306,13 @@ void main() {
     //
     // Since we need to draw to multiple images, we are going to create a different framebuffer for
     // each image.
-    let mut framebuffers = images.iter().map(|image| {
-        Arc::new(Framebuffer::start(render_pass.clone())
-            .add(image.clone()).unwrap()
-            .build().unwrap())
-    }).collect::<Vec<_>>();
+    let mut framebuffers: Option<Vec<Arc<vulkano::framebuffer::FramebufferAbstract + Send + Sync>>> = None;
+
+    //let mut framebuffers = images.iter().map(|image| {
+    //    Arc::new(Framebuffer::start(render_pass.clone())
+    //        .add(image.clone()).unwrap()
+    //        .build().unwrap())
+    //}).collect::<Vec<_>>();
 
     // Initialization is finally finished!
 
@@ -361,17 +363,20 @@ void main() {
             mem::replace(&mut swapchain, new_swapchain);
             mem::replace(&mut images, new_images);
 
+
+            recreate_swapchain = false;
+        }
+
+        if framebuffers.as_ref().is_none() {
             // Because framebuffers contains an Arc on the old swapchain, we need to
             // recreate framebuffers as well.  This is just like we did before, and so you
             // might want to make a function for this in your code.
-            let new_framebuffers = images.iter().map(|image| {
+            let new_framebuffers = Some(images.iter().map(|image| {
                 Arc::new(Framebuffer::start(render_pass.clone())
                          .add(image.clone()).unwrap()
                          .build().unwrap())
-            }).collect::<Vec<_>>();
+            }).collect::<Vec<_>>());
             mem::replace(&mut framebuffers, new_framebuffers);
-
-            recreate_swapchain = false;
         }
 
         // Before we can draw on the output, we have to *acquire* an image from the swapchain. If
@@ -408,7 +413,7 @@ void main() {
             // The third parameter builds the list of values to clear the attachments with. The API
             // is similar to the list of attachments when building the framebuffers, except that
             // only the attachments that use `load: Clear` appear in the list.
-            .begin_render_pass(framebuffers[image_num].clone(), false,
+            .begin_render_pass(framebuffers.as_ref().unwrap()[image_num].clone(), false,
                                vec![[0.0, 0.0, 1.0, 1.0].into()])
             .unwrap()
```

However, it does not like the type of framebuffers.

```
error[E0308]: mismatched types
   --> src/bin/triangle.rs:379:45
    |
379 |             mem::replace(&mut framebuffers, new_framebuffers);
    |                                             ^^^^^^^^^^^^^^^^ expected trait vulkano::framebuffer::FramebufferAbstract, found struct `vulkano::framebuffer::Framebuffer`
    |
    = note: expected type `std::option::Option<std::vec::Vec<std::sync::Arc<vulkano::framebuffer::FramebufferAbstract + std::marker::Send + std::marker::Sync>>>`
               found type `std::option::Option<std::vec::Vec<std::sync::Arc<vulkano::framebuffer::Framebuffer<std::sync::Arc<vulkano::framebuffer::RenderPass<main::scope::CustomRenderPassDesc>>, ((), std::sync::Arc<vulkano::image::SwapchainImage>)>>>>`
```

I tried replacing the type of `framebuffers` with the one specified, but then it complained about not finding `main::scope::CustomRenderPassDesc`.

Elzair:
Well I got it to compile by replacing `vulkano::framebuffer::FramebufferAbstract + Send + Sync` with `vulkano::framebuffer::Framebuffer<_,_>`, but now it crashes when I try to resize.

Elzair:
Alright it is fixed. I forgot to set framebuffers to None in the recreate_swapchain block.

tomaka:
There's a merge conflict in case you didn't notice.

Elzair:
Okay. I will try to fix it.

tomaka:
Thanks!
I won't nit-pick because the examples have accumulated tons of nit-picking material over time, and it's not worth bothering with that.

PR #672, opened 2017-07-21
Re-export errors for AutoCommandBufferBuilder

### LINKED ISSUE #595: AutoCommandBufferBuilder errors are inaccessible
The module containing them [is private](https://github.com/tomaka/vulkano/blob/master/vulkano/src/command_buffer/mod.rs#L94) and they are not re-exported, which makes it difficult to incorporate them into your own errors in downstream code, at least if you don't want to write `.chain_err(|| ...)` after every single vulkano call. More broadly, I understand that private types in public interfaces is bad style and perhaps even not intended to work.

Guvante:
Went through all of the pub functions in the auto module and re-exported them from the command_buffer module.

Resolves #595

### DIFF
--- a/vulkano/src/command_buffer/mod.rs
+++ b/vulkano/src/command_buffer/mod.rs
@@ -75,6 +75,20 @@
 
 pub use self::auto::AutoCommandBuffer;
 pub use self::auto::AutoCommandBufferBuilder;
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks

PR #679, opened 2017-07-22
Fix final image transition not updating state

tomaka:
cc @Ralith


Ralith:
This fixes the issue I observed.

PR #680, opened 2017-07-22
ImmutableImage refactor

Ralith:
This implements basic support for correct use of `ImmutableImage` modeled after the existing `ImmutableBuffer` interface.

Areas of particular interest for code review:
- Choice of `u32` return type for the new `Dimensions::texels` method--I have to cast this to `usize` everywhere I use it
- The `Pixel` trait, which serves as a sanity check and as documentation but is not strictly needed
- Design of `AcceptsPixels`, especially concerning its `rate` method
- Panics (as opposed to new errors) in `ImmutableImage::from_buffer` compatibility checks
- Implementation of `AcceptsPixels` for `Format`

### DIFF
--- a/vulkano/src/format.rs
+++ b/vulkano/src/format.rs
@@ -103,6 +103,10 @@
 //!
 
 use std::vec::IntoIter as VecIntoIter;
+use std::{mem, error, fmt};
+
+use half::f16;
+
 use vk;
 
 // TODO: add enumerations for color, depth, stencil and depthstencil formats
@@ -130,6 +134,39 @@ unsafe impl Data for u8 {
     fn ty() -> Format {
         Format::R8Uint
     }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct IncompatiblePixelsType;
+
+impl error::Error for IncompatiblePixelsType {
+    #[inline]
+    fn description(&self) -> &str { "supplied pixels' type is incompatible with this format" }
+}
+
+impl fmt::Display for IncompatiblePixelsType {
+    #[inline]
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(fmt, "{}", error::Error::description(self))
+    }
+}
+
+pub unsafe trait AcceptsPixels<T> {
+    /// Returns an error if `T` cannot be used as a source of pixels for `Self`.
+    fn ensure_accepts(&self) -> Result<(), IncompatiblePixelsType>;
+
+    /// The number of `T`s which make up a single pixel.
+    ///
+    /// ```
+    /// use vulkano::format::{AcceptsPixels, R8G8B8A8Srgb};
+    /// assert_eq!(<R8G8B8A8Srgb as AcceptsPixels<[u8; 4]>>::rate(&R8G8B8A8Srgb), 1);
+    /// assert_eq!(<R8G8B8A8Srgb as AcceptsPixels<u8>>::rate(&R8G8B8A8Srgb), 4);
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// May panic if `ensure_accepts` would not return `Ok(())`.
+    fn rate(&self) -> u32 { 1 }
 }
 
 macro_rules! formats {
@@ -357,12 +394,26 @@ macro_rules! formats {
     (__inner_ty__ $name:ident compressed=$f:tt) => { FormatTy::Compressed };
 
 
+    (__inner_strongstorage__ $name:ident [$ty:ty; $dim:expr]) => {
+        formats!(__inner_strongstorage_common__ $name [$ty; $dim]);
+        unsafe impl AcceptsPixels<$ty> for $name {
+            fn ensure_accepts(&self) -> Result<(), IncompatiblePixelsType> { Ok(()) }
+            fn rate(&self) -> u32 { $dim }
+        }
+    };
     (__inner_strongstorage__ $name:ident $ty:ty) => {
+        formats!(__inner_strongstorage_common__ $name $ty);
+    };
+    (__inner_strongstorage__ $name:ident ) => {};
+
+    (__inner_strongstorage_common__ $name:ident $ty:ty) => {
         unsafe impl StrongStorage for $name {
             type Pixel = $ty;
         }
+        unsafe impl AcceptsPixels<$ty> for $name {
+            fn ensure_accepts(&self) -> Result<(), IncompatiblePixelsType> { Ok(()) }
+        }
     };
-    (__inner_strongstorage__ $name:ident ) => {};
 }
 
 formats! {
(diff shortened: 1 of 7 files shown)


Ralith:
I also chose not to declare `ImmutableImage::uninitialized` unsafe, because I'm not aware of an safety issues associated with reading uninitialized data as an image.

fst3a:
@Ralith, ImmutableImage::uninitialized should be unsafe because, clearly, the contents of uninitialized image are undefined and undefined behavior is not considered safe?

Ralith:
Reading an undefined value is not usually undefined behavior.

Ralith:
This currently does not permit creation of `ImmutableImage`s involving f16 data, which I need. Would it make sense to pull in the `half` crate and use its type? This would make it very convenient to e.g. interoperate with the `openexr` crate, though placing other crates' types in the public API can be bad for stability. An alternative would be to define our own stub f16 type with methods to convert to/from u16, or to expose u16 directly.

vulkano/src/image/mod.rs
```diff
@@ -313,6 +313,11 @@ impl Dimensions {
             Dimensions::CubemapArray { .. } => ViewType::CubemapArray, 
         }
     }
+
+    #[inline]
+    pub fn texels(&self) -> u32 {
```
tomaka: Should be named `num_texels` or something like that.

vulkano/src/format.rs
```diff
@@ -132,6 +134,44 @@ unsafe impl Data for u8 {
...
+}
+
+impl_pixel! {
+    u8; i8; u16; i16; u32; i32; u64; i64; f16; f32; f64;
+}
+
+pub unsafe trait AcceptsPixels<T>
+    where T: Pixel
+{
+    fn accepts(&self) -> bool;
```
tomaka: Please change the signature to `ensure_accepts(&self) -> Result<(), IncompatiblePixelsType>`, or alternatively add a TODO if you don't want to bother.

vulkano/src/format.rs
```diff
@@ -132,6 +134,44 @@ unsafe impl Data for u8 {
...
+    fn rate(&self) -> u32 { 1 }
```
tomaka: Needs some documentation about what this is.

vulkano/src/format.rs
```diff
@@ -132,6 +134,44 @@ unsafe impl Data for u8 {
...
 
+/// Stub half float type
+#[derive(Debug, Copy, Clone)]
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub struct f16(u16);
+
+impl f16 {
+    fn from_bits(bits: u16) -> Self { f16(bits) }
+    fn as_bits(&self) -> u16 { self.0 }
```
tomaka: `to_bits` instead of `as_bits`, because f32 and f64 already have `to_bits`.

vulkano/src/format.rs
```diff
@@ -132,6 +134,44 @@ unsafe impl Data for u8 {
...
+}
+
+pub trait Pixel {}
```
tomaka: I don't think this trait is necessary at all.
I suppose you added it because of `impl<T: Pixel> AcceptsPixels<T> for Format`, but I'd prefer to add several individual implementations (eg. `impl AcceptsPixels<u8> for Format { .. }`, `impl AcceptsPixel<u16> for Format { .. }`, and so on).

vulkano/src/image/immutable.rs
```diff
@@ -28,55 +41,41 @@ use instance::QueueFamily;
...
-        where F: FormatDesc,
-              I: IntoIterator<Item = QueueFamily<'a>>
-    {
-        ImmutableImage::with_mipmaps(device,
-                                     dimensions,
-                                     format,
-                                     MipmapsCount::One,
-                                     queue_families)
-    }
+pub struct ImmutableImageInitialization<F, A = StdMemoryPoolAlloc> {
```
tomaka: Please add a note about the fact that this must not implement `Clone` because of the `used` variable.

vulkano/src/image/immutable.rs
```diff
@@ -130,19 +129,83 @@ impl<F> ImmutableImage<F> {
...
+    /// TODO: Support mipmaps
+    pub fn from_buffer<'a, B, P, I>(source: B, dimensions: Dimensions, format: F, usage: ImageUsage, layout: ImageLayout, queue_families: I, queue: Arc<Queue>)
+                                    -> Result<(Arc<Self>, CommandBufferExecFuture<NowFuture, AutoCommandBuffer>),
+                                              ImageCreationError>
+        where B: BufferAccess + TypedBufferAccess<Content = [P]> + 'static + Clone + Send + Sync,
+              P: Pixel + Send + Sync + Clone + 'static,
+              F: FormatDesc + AcceptsPixels<P> + 'static + Send + Sync,
+              I: IntoIterator<Item = QueueFamily<'a>>,
+    {
+        let usage = ImageUsage { transfer_destination: true, ..usage }; // Ensure we can actually initialize the image
```
tomaka: Please add a TODO that says that this code should be reworked in order to not panic.

tomaka:
I think it could be a good idea to use the `half` crate, provided that we reexport all its content (ie. put `pub use half::f16;` and `pub use half::consts;` somewhere in vulkano).

vulkano/src/image/immutable.rs
Ralith: If you feel these are appropriate for `ImageCreationError`, I'm happy to go directly to that. Or perhaps a new `ImageInitializationError`?

tomaka: The errors should be returned by `copy_buffer_to_image`, and this function should simply wrap around.

Ralith:
Applied requested changes, including reexport of the `half` crate and TODOing the initialization error checking in lieu of rich `copy_buffer_to_image` error checking.

I stared at `formats` for a while but I haven't yet come up with a clever way to make the `Format` impl of `AcceptsPixels` perfectly match the static impls without repeating a bunch of data.

Ralith:
I tweaked `ImmutableImage::from_buffer` to use `copy_buffer_to_image_dimensions` instead of `copy_buffer_to_image` so that it can properly transfer multi-layer images (e.g. cubemaps).

Ralith:
This makes the validation layers substantially happier, but I'm still seeing a suspicious error:
> you cannot transition the layout of aspect 1 from VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL when current layout is VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL.

The layout I'm specifying is `ShaderReadOnlyOptimal` and afaict I'm not doing anything strange that would require transitioning *away* from it--there should only be a single transition from `TransferDst` to it, then no more. This suggests to me that either I've got something subtly wrong with the initial/final layout requirements or there's still an issue in vulkano's implementation of them.

On another subject, would it be worth adding functions implementing the old API and marking them deprecated? Then perhaps this could be merged without a major version bump, but things're iterating pretty fast so maybe that's not worth the effort.

tomaka:
> On another subject, would it be worth adding functions implementing the old API and marking them deprecated? Then perhaps this could be merged without a major version bump, but things're iterating pretty fast so maybe that's not worth the effort.

That would be nice.

> The layout I'm specifying is ShaderReadOnlyOptimal and afaict I'm not doing anything strange that would require transitioning away from it--there should only be a single transition from TransferDst to it, then no more. This suggests to me that either I've got something subtly wrong with the initial/final layout requirements or there's still an issue in vulkano's implementation of them.

I think let's just merge this PR and I'll fix it the problem afterwards.

Ralith:
Dropped in what I believe are sufficiently close approximations of the old methods.

vulkano/src/image/immutable.rs
```diff
@@ -130,19 +165,84 @@ impl<F> ImmutableImage<F> {
...
+        };
+
+        Ok((image, init))
+    }
+
+    /// Construct an ImmutableImage from the contents of `iter`.
+    ///
+    /// TODO: Support mipmaps
+    #[inline]
+    pub fn from_iter<'a, P, I, J>(iter: I, dimensions: Dimensions, format: F, usage: ImageUsage, layout: ImageLayout,
```
tomaka: After thinking a bit about it, I changed my mind. I think the `ImageLayout` should simply be hard-coded to `ShaderReadOnlyOptimal` for now. Keep it as a field in the struct, but just don't expose it in as a parameter of the constructor.

vulkano/src/image/immutable.rs
```diff
@@ -130,19 +165,84 @@ impl<F> ImmutableImage<F> {
...
+                                                    BufferUsage::transfer_source(),
+                                                    iter::once(queue.family()),
+                                                    iter)?;
+        ImmutableImage::from_buffer(source, dimensions, format, usage, layout, queue_families, queue)
+    }
+
+    /// Construct an ImmutableImage containing a copy of the data in `source`.
+    ///
+    /// TODO: Support mipmaps
+    pub fn from_buffer<'a, B, P, I>(source: B, dimensions: Dimensions, format: F, usage: ImageUsage, layout: ImageLayout, queue_families: I, queue: Arc<Queue>)
```
tomaka: Same here.

PR #684, opened 2017-07-24
fixes #645 Returning `Result<(), OomError>` for `reset` and `multi_reset`

### LINKED ISSUE #645: Fence::reset and Fence::multi_reset do not return OomError
The vulkan spec states that vkResetFences can result in OUT_OF_*_MEMORY

tyoc213:
fixes #645 Returning `Result<(), OomError>` for `reset` and `multi_reset`

Im not sure if that is the correct way to handle the else in line 314 in `multi_reset` or if there is another way more `idiomatic`.

### DIFF
--- a/vulkano/src/sync/fence.rs
+++ b/vulkano/src/sync/fence.rs
@@ -263,15 +263,10 @@ impl<D> Fence<D>
     #[inline]
     pub fn reset(&mut self) -> Result<(), OomError> {
         unsafe {
-            let vk = self.device.pointers();
-
-            match check_errors(vk.ResetFences(self.device.internal_object(), 1, &self.fence)) {
-                Err(why) => Err(OomError::from(why)),
-                _ => {
-                    self.signaled.store(false, Ordering::Relaxed);
-                    Ok(())
-                }
-            }
+             let vk = self.device.pointers();
+             check_errors(vk.ResetFences(self.device.internal_object(), 1, &self.fence))?;
+             self.signaled.store(false, Ordering::Relaxed);
+             Ok(())
         }
     }
 
@@ -305,18 +300,12 @@ impl<D> Fence<D>
         if let Some(device) = device {
             unsafe {
                 let vk = device.pointers();
-                match check_errors(vk.ResetFences(device.internal_object(),
+                check_errors(vk.ResetFences(device.internal_object(),
                                fences.len() as u32,
-                               fences.as_ptr())) {
-                    Err(why) => Err(OomError::from(why)),
-                    _ => {
-                        Ok(())
-                    }
-                }
+                               fences.as_ptr()))?;
             }
-        } else {
-            Ok(())
         }
+        Ok(())
     }
 }
 
@@ -441,8 +430,9 @@ mod tests {
         let (device, _) = gfx_dev_and_queue!();
 
         let mut fence = Fence::alloc_signaled(device.clone()).unwrap();
-        fence.reset();
-        assert!(!fence.ready().unwrap());
+        if (fence.reset().is_ok()) {
+            assert!(!fence.ready().unwrap());
+        }
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


vulkano/src/sync/fence.rs
```diff
@@ -261,11 +261,15 @@ impl<D> Fence<D>
...
-    pub fn reset(&mut self) {
+    pub fn reset(&mut self) -> Result<(), OomError> {
         unsafe {
             let vk = self.device.pointers();
-            vk.ResetFences(self.device.internal_object(), 1, &self.fence);
-            self.signaled.store(false, Ordering::Relaxed);
+            if check_errors(vk.ResetFences(self.device.internal_object(), 1, &self.fence)).is_ok() {
+                Ok(self.signaled.store(false, Ordering::Relaxed))
+            }else {
+                Err(OomError::OutOfDeviceMemory)
```
nicokoch: This is a bit problematic. You are hardcoding the error type. Technically, we can also have `OomError::OutOfHostMemory` here.

You can make your life a lot easier by using the question mark operator (which is just syntax sugar for try!).

```
...
check_errors(vk.ResetFences(self.device.internal_object(), 1, &self.fence))?;
self.signaled.store(false, Ordering::Relaxed);
...
```

You don't even need the if.

vulkano/src/sync/fence.rs
```diff
@@ -299,10 +303,16 @@ impl<D> Fence<D>
...
             unsafe {
                 let vk = device.pointers();
-                vk.ResetFences(device.internal_object(),
+                if check_errors(vk.ResetFences(device.internal_object(),
                                fences.len() as u32,
-                               fences.as_ptr());
+                               fences.as_ptr())).is_ok() {
+                    Ok(())
+                } else {
+                    Err(OomError::OutOfDeviceMemory)
```
nicokoch: Same as above, you can avoid the `if`and propagate the error by using the question mark operator.

vulkano/src/sync/fence.rs
tyoc213: Well, trying that I get

```rust
error[E0308]: mismatched types
   --> src\sync\fence.rs:265:9
    |
265 | /         unsafe {
266 | |             let vk = self.device.pointers();
267 | |             check_errors(vk.ResetFences(self.device.internal_object(), 1, &self.fence))?;
268 | |             self.signaled.store(false, Ordering::Relaxed);
269 | |         }
    | |_________^ expected enum `std::result::Result`, found ()
    |
    = note: expected type `std::result::Result<(), OomError>`
               found type `()`
```

Im trying to see how I get the correct behaviour with `?`

tyoc213:
I have made it with a match and converting the errors to OomErrors because it seems that the others are private or something like that?

Also I keeped the `else` and returned `Òk(())` because now the function has a return type diferent than `()` and I choosed `Ok(())` because you say the only way to enter the else is if there is no error.

nicokoch:
With `?`, this will give you the correct behaviour:
unsafe {
             let vk = self.device.pointers();
             check_errors(vk.ResetFences(self.device.internal_object(), 1, &self.fence))?;
             self.signaled.store(false, Ordering::Relaxed);
             Ok(())
}

It will also remove some branches from the code, making it much easier to read.

tyoc213:
You where right, I was getting the error for not returning `Ok(())` int he correct place...

vulkano/src/sync/fence.rs
```diff
@@ -441,8 +430,9 @@ mod tests {
         let (device, _) = gfx_dev_and_queue!();
 
         let mut fence = Fence::alloc_signaled(device.clone()).unwrap();
-        fence.reset();
-        assert!(!fence.ready().unwrap());
+        if (fence.reset().is_ok()) {
```
tomaka: This should be `fence.reset().unwrap()`. No need for error handling in tests.

tomaka:
> If you want, I could squash the commits xD... lol

Github lets me squash commits, don't worry about that.

tomaka:
`fence.reset().unwrap()`, not just `fence.reset()`. Otherwise you have a warning.

tyoc213:
So, it should be alrihgt now.

tomaka:
Thanks

PR #687, opened 2017-07-24
Cache vertex buffers

tomaka:
Untested. Do not merge.


tomaka:
It was indeed buggy, but I think it's working fine now.

PR #690, opened 2017-07-25
Add Instance::with_loader

### LINKED ISSUE #361: Add Instance::with_loading_function for better cross-platform handling

tomaka:
Fix #361 

Reworks the loading system.
- Adds a `Loader` trait with only one method that is the equivalent of `vkGetInstanceProcAddr`.
- Provides two implementations of `Loader`: `DynamicLibraryLoader` and `statically_linked_vulkan_loader!`.
- The loader can now be owned by the `Instance` instead of always being stored in a `lazy_static!`.
- Adds various methods to allow using a custom loader.

It would be nice if someone on OSX could try if it still works.

### DIFF
--- a/vulkano/src/instance/extensions.rs
+++ b/vulkano/src/instance/extensions.rs
@@ -182,7 +182,7 @@ macro_rules! instance_extensions {
         impl $rawname {
             /// See the docs of supported_by_core().
             pub fn supported_by_core_raw() -> Result<Self, SupportedExtensionsError> {
-                let entry_points = try!(loader::entry_points());
+                let entry_points = try!(loader::default_function_pointers()).entry_points();
 
                 let properties: Vec<vk::ExtensionProperties> = unsafe {
                     let mut num = 0;
@@ -211,7 +211,7 @@ macro_rules! instance_extensions {
         impl $sname {
             /// See the docs of supported_by_core().
             pub fn supported_by_core_raw() -> Result<Self, SupportedExtensionsError> {
-                let entry_points = try!(loader::entry_points());
+                let entry_points = try!(loader::default_function_pointers()).entry_points();
 
                 let properties: Vec<vk::ExtensionProperties> = unsafe {
                     let mut num = 0;
--- a/vulkano/src/instance/extensions.rs
+++ b/vulkano/src/instance/extensions.rs
@@ -182,7 +182,15 @@ macro_rules! instance_extensions {
         impl $rawname {
             /// See the docs of supported_by_core().
             pub fn supported_by_core_raw() -> Result<Self, SupportedExtensionsError> {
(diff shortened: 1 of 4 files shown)


tomaka:
To go into some details, this allows customizing which Vulkan implementation to use or where to find it.

For example if you're writing a custom implementation of Vulkan, you can now use vulkano with it (provided that your implementation is compliant).

This also means that vulkano is guaranteed to work on all platforms without any tweak in its source code, which can be useful for the Nintendo Switch or the Fuchsia OS for example.

nicokoch:
> It would be nice if someone on OSX could try if it still works.

I can do this, but probably not before the weekend

nicokoch:
I just tested this on macos. Everythings fine!

nicokoch:
However, I noticed another test failed that was added in a recent PR:
```
command_buffer::pool::standard::tests::reuse_command_buffers stdout ----
	thread 'command_buffer::pool::standard::tests::reuse_command_buffers' panicked at 'assertion failed: `(left == right)` (left: `140193622435904`, right: `140193621430240`)', src/command_buffer/pool/standard.rs:299
```

Not sure if this is vulkano's or MoltenVk's fault.

tomaka:
Thanks!

I'm not sure about this test. When I wrote it I noticed that it was failing, but since I was still tweaking some code I thought I had made some mistake. So I started the tests again multiple times, and then it wasn't failing anymore.

Basically the test can be summed up as:

```rust
let mut container = crossbeam::MsQueue::new();
container.push(foo);
assert_eq!(container.pop(), Some(foo));   // this fails
```

Since the `MsQueue` is a lock-free queue, I'm not sure whether this is a normal behaviour or if there's some sort of problem.

Anyway this is off-topic here.

PR #691, opened 2017-07-26
Implemented dynamic TwoBuffersDefinition decode

Guvante:
Ran into an `unimplemented!()` while translating the teapot example to use abstracts, made an implementation.

Modeled the length calculation after the tuple version and the dynamic nature after SingleBufferDefinition.

The SingleBufferDefinition had a note about safety and I wasn't sure how to improve the safety so copied the FIXME.


tomaka:
Thanks!

> The SingleBufferDefinition had a note about safety and I wasn't sure how to improve the safety so copied the FIXME.

Vertex sources in general need a rework :-/

PR #703, opened 2017-07-30
Deleting simple.rs and still need to update the doc string fixes #631

### LINKED ISSUE #631: Remove PipelineLayoutDescNames
Requires an API breaking change. Should be done before version 0.6 is released.

tyoc213:
fixes #631

The only missing thing would be the example to replace the doc string?

`cargo test` and `cargo build` ok.


tomaka:
Thanks

PR #716, opened 2017-08-04
Update Cargo.toml metadata

tomaka:


tomaka:
For some reason travis doesn't pick this PR.

PR #729, opened 2017-08-06
Support for VK_KHR_incremental_present

nicokoch:
The code should be pretty self-explanatory.

I'm unsure about the #[derive(Debug)] in vk_sys.

### DIFF
--- a/vk-sys/src/lib.rs
+++ b/vk-sys/src/lib.rs
@@ -2630,7 +2630,6 @@ pub struct RectLayerKHR {
 }
 
 #[repr(C)]
-#[derive(Debug)]
 pub struct PresentRegionKHR {
     pub rectangleCount: u32,
     pub pRectangles: *const RectLayerKHR,
--- a/vk-sys/src/lib.rs
+++ b/vk-sys/src/lib.rs
@@ -181,6 +181,7 @@ pub const STRUCTURE_TYPE_SPARSE_IMAGE_FORMAT_PROPERTIES_2_KHR: u32 = 1000059007;
 pub const STRUCTURE_TYPE_PHYSICAL_DEVICE_SPARSE_IMAGE_FORMAT_INFO_2_KHR: u32 = 1000059008;
 pub const STRUCTURE_TYPE_VI_SURFACE_CREATE_INFO_NN: u32 = 1000062000;
 pub const STRUCTURE_TYPE_PHYSICAL_DEVICE_PUSH_DESCRIPTOR_PROPERTIES_KHR: u32 = 1000080000;
+pub const STRUCTURE_TYPE_PRESENT_REGIONS_KHR: u32 = 1000084000;
 pub const STRUCTURE_TYPE_DESCRIPTOR_UPDATE_TEMPLATE_CREATE_INFO_KHR: u32 = 1000085000;
 pub const STRUCTURE_TYPE_MEMORY_DEDICATED_REQUIREMENTS_KHR: u32 = 1000127000;
 pub const STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO_KHR: u32 = 1000127001;
@@ -2619,6 +2620,28 @@ pub struct MemoryRequirements2KHR {
     pub sType: StructureType,
     pub pNext: *mut c_void,
     pub memoryRequirements: MemoryRequirements,
+}
+
+#[repr(C)]
+pub struct RectLayerKHR {
+    pub offset: Offset2D,
+    pub extent: Extent2D,
+    pub layer: u32,
+}
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct PresentRegionKHR {
+    pub rectangleCount: u32,
+    pub pRectangles: *const RectLayerKHR,
+}
+
+#[repr(C)]
+pub struct PresentRegionsKHR {
+    pub sType: StructureType,
+    pub pNext: *const c_void,
+    pub swapchainCount: u32,
+    pub pRegions: *const PresentRegionKHR,
 }
 
 macro_rules! ptrs {
--- a/vulkano/src/command_buffer/submit/queue_present.rs
+++ b/vulkano/src/command_buffer/submit/queue_present.rs
@@ -14,8 +14,10 @@ use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
 
+use device::DeviceOwned;
 use device::Queue;
 use swapchain::Swapchain;
+use swapchain::PresentRegion;
 use sync::Semaphore;
 
 use Error;
@@ -32,6 +34,7 @@ pub struct SubmitPresentBuilder<'a> {
     wait_semaphores: SmallVec<[vk::Semaphore; 8]>,
     swapchains: SmallVec<[vk::SwapchainKHR; 4]>,
     image_indices: SmallVec<[u32; 4]>,
+    present_regions: SmallVec<[vk::PresentRegionKHR; 4]>,
     marker: PhantomData<&'a ()>,
 }
 
@@ -43,6 +46,7 @@ impl<'a> SubmitPresentBuilder<'a> {
             wait_semaphores: SmallVec::new(),
             swapchains: SmallVec::new(),
             image_indices: SmallVec::new(),
+            present_regions: SmallVec::new(),
             marker: PhantomData,
         }
     }
@@ -71,6 +75,9 @@ impl<'a> SubmitPresentBuilder<'a> {
 
     /// Adds an image of a swapchain to be presented.
     ///
+    /// Allows to specify a present region (VK_KHR_incremental_present).
+    /// Areas outside the present region *can* be ignored by the vulkan implementation for
+    /// optimizations purposes.
     /// # Safety
     ///
     /// - If you submit this builder, the swapchain must be kept alive until you are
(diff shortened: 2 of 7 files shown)


vulkano/src/command_buffer/submit/queue_present.rs
```diff
@@ -97,14 +125,28 @@ impl<'a> SubmitPresentBuilder<'a> {
             assert!(!self.swapchains.is_empty(),
                     "Tried to submit a present command without any swapchain");
 
-            let vk = queue.device().pointers();
-            let queue = queue.internal_object_guard();
+            let mut p_next = ptr::null();
+            let present_regions;
```
tomaka: About the style, I'd prefer `let present_regions = if ... { Some(vk::PresentRegionsKHR { ... }) } else { None };`.
And then `pNext: present_regions.as_ref().map(|pr| &pr as *const _).unwrap_or(ptr::null())`.

Similar example: https://github.com/vulkano-rs/vulkano/blob/ae51cd154f83c2ef50b482e375fb2058d0195d7e/vulkano/src/pipeline/graphics_pipeline/mod.rs#L802-L830 and https://github.com/vulkano-rs/vulkano/blob/ae51cd154f83c2ef50b482e375fb2058d0195d7e/vulkano/src/pipeline/graphics_pipeline/mod.rs#L1194-L1197

vulkano/src/command_buffer/submit/queue_present.rs
```diff
@@ -71,6 +75,9 @@ impl<'a> SubmitPresentBuilder<'a> {
 
     /// Adds an image of a swapchain to be presented.
     ///
+    /// Allows to specify a present region (VK_KHR_incremental_present).
+    /// Areas outside the present region *can* be ignored by the vulkan implementation for
```
tomaka: `Vulkan` is spelled with a capital letter and needs a new line before `# Safety`.
Also needs a text saying that the present region is ignored if the extension isn't enabled.

vk-sys/src/lib.rs
```diff
@@ -2621,6 +2622,28 @@ pub struct MemoryRequirements2KHR {
...
 
+#[repr(C)]
+pub struct RectLayerKHR {
+    pub offset: Offset2D,
+    pub extent: Extent2D,
+    pub layer: u32,
+}
+
+#[repr(C)]
+#[derive(Debug)]
```
tomaka: Yeah I don't like putting traits in vk-sys.
It's not a strong opinion, but here it would be the first ever trait impl in vk-sys.

vulkano/src/swapchain/present_region.rs
```diff
@@ -0,0 +1,80 @@
...
+        // FIXME negative offset is not disallowed by spec, but semantically should not be possible
+        debug_assert!(self.offset[0] >= 0);
+        debug_assert!(self.offset[1] >= 0);
+        self.offset[0] as u32 + self.extent[0] <= swapchain.dimensions()[0] &&
+        self.offset[1] as u32 + self.extent[1] <= swapchain.dimensions()[1] &&
+        self.layer < swapchain.layers()
+    }
+}
+
+impl<'a> Into<vk::RectLayerKHR> for &'a RectangleLayer {
```
tomaka: I'd prefer a `pub(crate)` method directly on `RectangleLayer`, in order to be consistent with the rest of the library.

vulkano/src/command_buffer/submit/queue_present.rs
```diff
@@ -79,12 +86,33 @@ impl<'a> SubmitPresentBuilder<'a> {
     /// - The swapchains and semaphores must all belong to the same device.
     ///
     #[inline]
-    pub unsafe fn add_swapchain(&mut self, swapchain: &'a Swapchain, image_num: u32) {
+    pub unsafe fn add_swapchain(&mut self, swapchain: &'a Swapchain, image_num: u32,
+                                 present_region: Option<&'a PresentRegion>) {
```
tomaka: The `PresentRegion` doesn't need the `'a` lifetime. Contrary to the swapchain we don't need to keep it alive.

vulkano/src/swapchain/present_region.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.rectangles.iter().all(|rect| rect.is_compatible_with(swapchain))
+    }
+}
+
+impl<'a> Into<vk::PresentRegionKHR> for &'a PresentRegion<'a> {
+    fn into(self) -> vk::PresentRegionKHR {
+        let vk_rects = self.rectangles.iter().map(|rect| rect.into()).collect::<SmallVec<[vk::RectLayerKHR; 4]>>();
+        vk::PresentRegionKHR {
+            rectangleCount: self.rectangles.len() as u32,
+            pRectangles: vk_rects.as_ptr(),
```
tomaka: That's unsound! Dangling pointer.

tomaka:
In addition to the changes, I think we shouldn't try to be smart (for now) about references. The present region and rectangles should be taken by value everywhere, except for `add_swapchain` in queue_present.rs.

nicokoch:
Thanks for the review. Will rework as soon as possible (which is probably later this week)

vulkano/src/swapchain/present_region.rs
nicokoch: How do you propose to solve this?

tomaka: Just take everything by value. The problem is that you're implementing `Into` on a reference.
The `vkPresentRegionKHR` objects should be built just before the call to `vkQueuePresent`, and not in a separate function.

nicokoch: Well, this means we have to store `present_regions: SmallVec<[Option<&'b PresentRegion>; 4]>` in `PresentSubmitBuilder` (and not `vk::PresentRegion`), because we can only convert them in `submit`. Which means adding the new lifetime 'b to that struct (Or should it have the lifetime 'a?)

We could also take PresentRegion by value in `add_swapchain`, but this would introduce a bunch of clones which I would like to avoid.

Do you agree?

tomaka: Ah yes I forgot that it was about the `PresentSubmitBuilder`.

Basically the builder should keep a `SmallVec<[&'a vk::RectLayerKHR]>` and a `SmallVec<[&'a vk::PresentRegion]>`. Do not store the `PresentRegion` objects (the non-Vulkan ones). When you call `add_swapchain`, it adds entries to the two smallvecs. The pointer to the rects inside the `vk::PresentRegion` objects is set to null, since the rects storage can still change.

Then in `submit`, the pointers are modified to point to the `SmallVec` of the rects.

nicokoch: Thanks, this helps a lot. I hate dealing with these types of lifetime issues. 

There is another problem though. `vk::PresentRegion` can hold any amount of `vk::RectLayerKHR`. In submit, we have no idea which indices of `SmallVec<[&'a vk::RectLayerKHR]> ` correspond to which `vk::PresentRegion`.

nicokoch: Wait, nevermind. We do know because of rectangleCount.

tomaka: I re-read my comment and I realized I screwed up. It should read `SmallVec<[vk::RectLayerKHR]>` and `SmallVec<[vk::PresentRegionKHR]>` (no `&'a`). The `add_swapchain` function builds the `vk::*` structs that correspond to its parameter and inserts them in the smallvecs.

tomaka: Though it could be possible to use `SmallVec<[&'a vk::RectLayerKHR]>` and `SmallVec<[&'a vk::PresentRegionKHR]>` if `RectLayer` and `PresentRegion` had exactly the same memory layout, but since it's not guaranteed to be the case you have to perform a conversion somewhere anyway.

nicokoch: Thanks for the input.

Note that your suggested changes require `submit` to take `mut self` instead of `self`.
Unless I'm missing something.

tomaka: Turning `self` into `mut self` is not an API change, so it's not a problem.

nicokoch:
@tomaka I fixed the code. Only compile-tested for now

vulkano/src/command_buffer/submit/queue_present.rs
```diff
@@ -125,6 +189,16 @@ impl<'a> SubmitPresentBuilder<'a> {
     }
 }
 
+impl<'a> fmt::Debug for SubmitPresentBuilder<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f,
```
tomaka: Please use `f.debug_struct()` instead: https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct

vulkano/src/command_buffer/submit/queue_present.rs
```diff
@@ -79,32 +90,85 @@ impl<'a> SubmitPresentBuilder<'a> {
...
 
+            let vk = queue.device().pointers();
+            let queue = queue.internal_object_guard();
+
             let infos = vk::PresentInfoKHR {
                 sType: vk::STRUCTURE_TYPE_PRESENT_INFO_KHR,
-                pNext: ptr::null(),
+                pNext: present_regions
+                    .as_ref()
+                    .map(|pr| pr as *const _ as *const _)
```
tomaka: I'd prefer `pr as *const vk::PresentRegionsKHR` because these casts to pointers can be a bit hazardous.

vulkano/src/command_buffer/submit/queue_present.rs
```diff
@@ -71,6 +76,12 @@ impl<'a> SubmitPresentBuilder<'a> {
 
     /// Adds an image of a swapchain to be presented.
     ///
+    /// Allows to specify a present region.
+    /// Areas outside the present region *can* be ignored by the Vulkan implementation for
+    /// optimizations purposes.
+    ///
+    /// If `VK_KHR_incremental_present` is not enabled, the parameter is ignored.
```
tomaka: Should be "the `present_region` parameter", otherwise "the parameter" alone is ambiguous.

tomaka:
Thanks

PR #731, opened 2017-08-07
Finish implementing dedicated allocation

### LINKED ISSUE #640: Add support for VK_KHR_dedicated_allocation
Has just been made "stable" in Vulkan 1.0.54.

tomaka:
Close #640 

Breaking change.

Modifies the API of `MemoryPool`. Its sole required method is renamed to `alloc_generic`, while `alloc` is turned into a provided method and takes an additional parameter which is the dedicated allocation.

`alloc` now also returns a `PotentialDedicatedAllocation<A>` instead of a `A`.

### DIFF
--- a/vulkano/src/memory/mod.rs
+++ b/vulkano/src/memory/mod.rs
@@ -118,8 +118,8 @@ pub struct MemoryRequirements {
     pub memory_type_bits: u32,
 
     /// True if the implementation prefers to use dedicated allocations (in other words, allocate
-    /// a whole block of memory dedicated to this resource alone). If the implementation doesn't
-    /// support dedicated allocations, this will be false.
+    /// a whole block of memory dedicated to this resource alone). If the
+    /// `khr_get_memory_requirements2` extension isn't enabled, then this will be false.
     ///
     /// > **Note**: As its name says, using a dedicated allocation is an optimization and not a
     /// > requirement.
--- a/vulkano/src/memory/mod.rs
+++ b/vulkano/src/memory/mod.rs
@@ -138,10 +138,20 @@ impl MemoryRequirements {
     }
 }
 
+/// Indicates whether we want to allocate memory for a specific resource, or in a generic way.
+///
+/// Using dedicated allocations can yield faster performances, but requires the
+/// `VK_KHR_dedicated_allocation` extension to be enabled on the device.
+///
+/// If a dedicated allocation is performed, it must only be bound to any resource other than the
+/// one that was passed with the enumeration.
 #[derive(Debug, Copy, Clone)]
 pub enum DedicatedAlloc<'a> {
+    /// Generic allocation.
     None,
+    /// Allocation dedicated to a buffer.
     Buffer(&'a UnsafeBuffer),
+    /// Allocation dedicated to an image.
     Image(&'a UnsafeImage),
 }
 
--- a/vulkano/src/memory/pool/mod.rs
+++ b/vulkano/src/memory/pool/mod.rs
@@ -109,14 +109,14 @@ pub unsafe trait MemoryPool: DeviceOwned {
 
         match map {
             MappingRequirement::Map => {
-                let mem = DeviceMemory::dedicated_alloc(self.device().clone(), ty, size,
-                                                        dedicated)?;
-                Ok(PotentialDedicatedAllocation::Dedicated(mem))
-            },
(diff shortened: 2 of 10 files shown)


tomaka:
Unfortunately I can't test whether it works. I thought the latest nVidia beta driver implemented the extension, but they don't.

vulkano/src/memory/pool/mod.rs
```diff
@@ -45,7 +83,43 @@ pub unsafe trait MemoryPool {
...
-             map: MappingRequirement) -> Result<Self::Alloc, DeviceMemoryAllocError>;
+             map: MappingRequirement, dedicated: DedicatedAlloc)
+             -> Result<PotentialDedicatedAllocation<Self::Alloc>, DeviceMemoryAllocError>
+    {
+        if !self.device().loaded_extensions().khr_dedicated_allocation {
+            let alloc = self.alloc_generic(ty, size, alignment, layout, map)?;
+            return Ok(alloc.into());
+        }
+
+        let use_dedicated = match dedicated {
```
tomaka: Oh right, I totally forgot about this thing!

My code was based on this blogpost: https://developer.nvidia.com/what%E2%80%99s-your-vulkan-memory-type (the "Do's" section near the bottom), but I guess `prefersDedicatedAllocation` is meant for that and didn't exist when the blogpost was written.

tomaka:
Following the comment about `prefer_dedicated`, I had to make some changes.

`prefer_dedicated` is obtained from the `MemoryRequirements` struct. If we added an additional parameter `prefer_dedicated` to `alloc`, then it would be more or less a duplicate of the `DedicatedAlloc` parameter. I think it would make more sense to call `alloc` with `DedicatedAlloc::None` when `prefer_dedicated` is false`.

However I think that it would be too cumbersome for the user to have to adjust `DedicatedAlloc` based on `prefer_dedicated`. Passing the `MemoryRequirements` struct would make more sense, but then it could conflict with `memory_type`. I really dislike when you can call a function with an invalid combination of parameters.

Therefore I added a brand new function `alloc_from_requirements` that chooses the memory type from the memory requirements, then performs an allocation.

Then I noticed that having both `alloc_from_requirements` and `alloc` could be confusing, because the `dedicated` parameter of `alloc` means that the function *must* perform a dedicated allocation, while the `dedicated` parameter of `alloc_from_requirements` is just a hint and can be ignored. Therefore I totally removed `alloc`.

tomaka:
Oh, a nightly-only failure on travis:

> error: Error during translation/LLVM phase.
> 
> thread 'rustc' panicked at 'Box<Any>', /checkout/src/librustc/session/mod.rs:236:8
> 
> thread 'rustc' panicked at 'couldn't compile the test', /checkout/src/librustdoc/test.rs:273:12

tomaka:
Restarting the travis build fixed it.

vulkano/src/memory/mod.rs
```diff
@@ -118,8 +118,8 @@ pub struct MemoryRequirements {
     pub memory_type_bits: u32,
 
     /// True if the implementation prefers to use dedicated allocations (in other words, allocate
-    /// a whole block of memory dedicated to this resource alone). If the implementation doesn't
-    /// support dedicated allocations, this will be false.
+    /// a whole block of memory dedicated to this resource alone). If the
+    /// `khr_get_memory_requirements2` extension isn't enabled, then this will be false.
```
nicokoch: At the moment it is actually always false, because nowhere `VkMemoryDedicatedRequirementsKHR` is queried

tomaka: That's not true, it is queried here: https://github.com/vulkano-rs/vulkano/blob/ae51cd154f83c2ef50b482e375fb2058d0195d7e/vulkano/src/buffer/sys.rs#L129-L171 and here: https://github.com/vulkano-rs/vulkano/blob/ae51cd154f83c2ef50b482e375fb2058d0195d7e/vulkano/src/image/sys.rs#L522-L562

nicokoch: Oh, sorry. I totally didn't look there.
I was confused because it's hardcoded as `false` here: https://github.com/vulkano-rs/vulkano/blob/d94adf0f992691b278410a52d0017e15910ee6e0/vulkano/src/memory/mod.rs#L131

tomaka:
Let's merge this I guess, thanks for the review!

PR #733, opened 2017-08-08
Correct recovery in SyncCommandBuffer::lock_submit

tomaka:


tomaka:
Before this PR, `lock_submit` would lock each buffer and image one by one. If at any point an error is detected, the function would immediately return an keep the resources locked. This fixes it.

As to why we don't use RAII instead, if we did we would have to allocate a `Vec` that holds the locks, which is clearly unnecessary.

PR #740, opened 2017-08-14
draw_subset and draw_indexed_subset

fst3a:
This PR adds two methods to `AutoCommandBufferBuilder` struct that lets user insert draw commands that only use a subset of data provided in vertex or index buffer and corresponding validation functions.  This is useful e.g. when user wants to have different LODs of model stored in one big monolithic buffer as this allows to make buffer rebinds less frequent.  It re-uses DrawIndirectCommand and DrawIndexedIndirectCommand structures.

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -21,6 +21,7 @@ use buffer::BufferAccess;
 use buffer::TypedBufferAccess;
 use command_buffer::CommandBuffer;
 use command_buffer::CommandBufferExecError;
+use command_buffer::DrawIndexedIndirectCommand;
 use command_buffer::DrawIndirectCommand;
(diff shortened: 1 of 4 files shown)


tomaka:
Thanks for the PR, but right now you can already do that by passing slices of buffers to the draw commands.
Eg:

```rust
let vert_input = vec![my_vertex_buffer.into_buffer_slice().slice(1024 .. 2048).unwrap()];

cb_builder.draw(.., .., vert_input, ..);
```

This doesn't allow for setting a base instance though.

fst3a:
Oh, I didn't know that, sorry.

fst3a:
Should the PR be closed now?

tomaka:
I guess so. I didn't close it in case I was missing something.

PR #746, opened 2017-08-16
Remove the list of queue families in most places when creating a buffer or image

### LINKED ISSUE #577: Remove the list of queue families from buffers and images creation
I think it's too cumbersome to have to do that. Buffers and images should probably default to shared sharing mode for all the queue families of the device.
Of course you could still customize the sharing mode to use, but with another constructor.

And after this change is made, I think the API for the sharing mode should be more explicit instead of just an iterator to a list of queue families. I went with the iterator approach because it is less cumbersome, but I think that it feels too magic.

tomaka:
Fix #577 

Makes the API much less cumbersome to use.

Obviously a breaking change. In theory we could deprecate existing functions and introduce new ones instead, but we'd run out of names pretty quickly if doing so.

### DIFF
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -112,7 +112,7 @@ void main() {
         // Iterator that produces the data.
         let data_iter = (0 .. 65536u32).map(|n| n);
         // Builds the buffer and fills it with this iterator.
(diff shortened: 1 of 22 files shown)


nicokoch:
Love this change. Will be even nicer when impl trait arrives!

PR #751, opened 2017-08-19
Add a link to the README in the Cargo.toml

tomaka:


tomaka:
The point of this PR is to show the content of the README on crates.io, but I think putting `../` for the path will not work.

tomaka:
For the version 0.6.1 I'll merge this PR just to see if it works.
If it doesn't, I'll open a cargo issue.

tomaka:
Turns out it *does* work.

PR #754, opened 2017-08-20
Prevent creating a new swapchain from a stale one

### LINKED ISSUE #735: Bug in swapchain staleness?
Hi. This is just a remark from browsing the source.

I notice that `Swapchain::stale` is never set to `true`. The field comment suggests that it's supposed to be set when image acquisition from the old chain is made illegal by creating a new descendant chain. So I suppose the `false` in [swapchain.rs:314](https://github.com/vulkano-rs/vulkano/blob/a067f6e89dacef68e965428a388ba432943153fb/vulkano/src/swapchain/swapchain.rs#L314) ought to be `true`?

GabrielMajeri:
This pull request prevents creating a swapchain from an old one, if the old one has already been used to recreate another one.

It does so by checking the `swapchain.stale` property, and sets it correctly. Also added a new `SwapchainCreationError::OldSwapchainAlreadyUsed` error to report this situation.

Fixes #735.

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -212,10 +212,6 @@ impl Swapchain {
     /// swapchain. The order in which the images are returned is important for the
     /// `acquire_next_image` and `present` functions.
     ///
-    /// According to the specification, if the `old_swapchain` is not `None`,
-    /// > any images not acquired by the application may be freed by the implementation,
-    /// > which may occur even if creation of the new swapchain fails.
-    ///
     /// # Panic
     ///
     /// - Panics if the device and the surface don't belong to the same instance.
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -213,8 +213,8 @@ impl Swapchain {
     /// `acquire_next_image` and `present` functions.
     ///
     /// According to the specification, if the `old_swapchain` is not `None`,
-    /// > "any images not acquired by the application may be freed by the implementation,
-    /// > which may occur even if creation of the new swapchain fails."
+    /// > any images not acquired by the application may be freed by the implementation,
+    /// > which may occur even if creation of the new swapchain fails.
     ///
     /// # Panic
     ///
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks for the PR.
Please add a comment saying something like this:

> According to the documentation of `VkSwapchainCreateInfoKHR`:
> > Upon calling vkCreateSwapchainKHR with a oldSwapchain that is not VK_NULL_HANDLE, any images not acquired by the application may be freed by the implementation, which may occur even if creation of the new swapchain fails.
> Therefore we set `stale` to `true` and keep it to true even if `vkCreateSwapchain` fails.

GabrielMajeri:
Done, added the comment and also updated the docs with the information.

vulkano/src/swapchain/swapchain.rs
```diff
@@ -212,6 +212,10 @@ impl Swapchain {
     /// swapchain. The order in which the images are returned is important for the
     /// `acquire_next_image` and `present` functions.
     ///
+    /// If the `old_swapchain` is not `None`, the old swapchain **cannot** be used after the call,
+    /// except for presenting already acquired images. Even if the call fails, the implementation
```
tomaka: That's not true. For example if the call fails because the request capabilities are not supported, then the old swapchain can still be used.

vulkano/src/swapchain/swapchain.rs
```diff
@@ -350,6 +354,13 @@ impl Swapchain {
             if *stale {
                 return Err(SwapchainCreationError::OldSwapchainAlreadyUsed);
             } else {
+                // According to the documentation of VkSwapchainCreateInfoKHR:
+                //
+                //  "Upon calling vkCreateSwapchainKHR with a oldSwapchain that is not VK_NULL_HANDLE,
```
tomaka: Please use blockquotes by putting `> ` at the start of each line instead

vulkano/src/swapchain/swapchain.rs
GabrielMajeri: Sorry, I must've misunderstood the spec. I've updated the docs now.

vulkano/src/swapchain/swapchain.rs
GabrielMajeri: Didn't know that was possible, fixed now, thanks.

vulkano/src/swapchain/swapchain.rs
tomaka: What I mean is that if we don't even call any Vulkan function then the old swapchain remains valid.

vulkano/src/swapchain/swapchain.rs
tomaka: Also remove the `"` please

vulkano/src/swapchain/swapchain.rs
```diff
@@ -212,6 +212,10 @@ impl Swapchain {
     /// swapchain. The order in which the images are returned is important for the
     /// `acquire_next_image` and `present` functions.
     ///
+    /// According to the specification, if the `old_swapchain` is not `None`,
```
tomaka: I'm not sure what you're trying to say with this comment. I think the simplest is to just remove it altogether.

GabrielMajeri: I've removed the comment. Just noticed the line below

> TODO: remove `old_swapchain` parameter and add another function `with_old_swapchain`.

meaning that it's pointless to document this here anyway, since this will be included in the `with_old_swapchain`'s function doc.

PR #757, opened 2017-08-22
Check pipeline layout limits

### LINKED ISSUE #585: Check limits of shaders
When a shader module is loaded, its limits should be tested against the physical device's limits.

For example: the compute shader local size.

tomaka:
Fix #585

### DIFF
--- a/vulkano/src/descriptor/pipeline_layout/limits_check.rs
+++ b/vulkano/src/descriptor/pipeline_layout/limits_check.rs
@@ -36,45 +36,37 @@ pub fn check_desc_against_limits<D>(device: &Device, desc: &D)
     for set in 0 .. desc.num_sets() {
         for binding in 0 .. desc.num_bindings_in_set(set).unwrap() {
             let descriptor = desc.descriptor(set, binding).unwrap();
+            num_resources.increment(descriptor.array_count, &descriptor.stages);
 
             match descriptor.ty.ty().expect("Not implemented yet") {        // TODO:
                 DescriptorType::Sampler => {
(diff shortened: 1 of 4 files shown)


vulkano/src/descriptor/pipeline_layout/limits_check.rs
```diff
@@ -0,0 +1,448 @@
...
+    let mut num_storage_images = Counter::default();
+    let mut num_input_attachments = Counter::default();
+
+    for set in 0 .. desc.num_sets() {
+        for binding in 0 .. desc.num_bindings_in_set(set).unwrap() {
+            let descriptor = desc.descriptor(set, binding).unwrap();
+
+            match descriptor.ty.ty().expect("Not implemented yet") {        // TODO:
+                DescriptorType::Sampler => {
+                    num_resources.increment(descriptor.array_count, &descriptor.stages);
```
nicokoch: This line in each match arm can be removed and added once before the match

tomaka:
I wonder if `maxUniformBufferRange` and `maxStorageBufferRange` should also be checked.
The specs say that these limits must be respected when binding a buffer. In theory you could bind only a certain range of a uniform buffer if you know that only this range will be accessed by the shader, in which case what you're doing is valid.

tomaka:
Note that vulkano already checks maxUniformBufferRange and maxStorageBufferRange when binding the buffer.

nicokoch:
IMO they should be checked when binding the buffer, not when creating the layout, as the specs say and as vulkano already does. So everything is fine like it is

PR #774, opened 2017-08-25
Reflect specialization constants in vulkano-shaders

tomaka:
`vulkano-shaders` now generates a `SpecializationConstants` struct for each shader module, that implements the `SpecializationConstants` trait of vulkano. Also implements `Default` for when you want to use the defaults that were specified in the SPIR-V code.

If a shader module doesn't contain any spec const the specialization constants will be `()` instead, because of backward compatibility.

This makes specialization constants usable for compute pipelines, but not for graphics pipelines yet because they are not implemented in vulkano itself yet.

### DIFF
--- a/vulkano-shaders/examples/example.rs
+++ b/vulkano-shaders/examples/example.rs
@@ -14,6 +14,8 @@ fn main() {
     let shader = r#"
 #version 450
 
+layout(constant_id = 5) const int index = 2;
+
 struct S {
     vec3 val1;
     bool val2[5];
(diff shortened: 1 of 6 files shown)


tomaka:
A more general explanation:

*Specialization constants* are constants that you put in your GLSL code (or SPIR-V code in general) and whose values can be modified when you create the compute or graphics pipeline.

Reference for GLSL: https://github.com/KhronosGroup/Vulkan-Docs/blob/1.0/doc/specs/misc/GL_KHR_vulkan_glsl.txt#L188

After this PR, you will be able to do something like this:

```rust
mod cs {
    #[derive(VulkanoShader)]
    #[ty = "compute"]
    #[src = "
#version 450

layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;

layout(constant_id= 0) const int my_constant = 2;
layout(constant_id= 1) const bool my_other_constant = false;

void main() {
    ...
}"
    ]
    struct Dummy;
}

let shader = cs::Shader::load(device.clone())
    .expect("failed to create shader module");
let spec_consts = cs::SpecializationConstants {
    my_constant: 12,
    .. Default::default()   // sets `my_other_constant` to false
};
let compute_pipeline = ComputePipeline::new(device.clone(), &shader.main_entry_point(), &spec_consts)
    .expect("failed to create compute pipeline");
```

PR #776, opened 2017-08-25
Fix `ApplicationInfo::from_cargo_toml`

### LINKED ISSUE #500: `ApplicationInfo::from_cargo_toml()` doesn't work as intended
Example ([relevant code](https://github.com/tomaka/vulkano/blob/ed0b85a48476cfec075fbf5cf0ffbba69fecff7a/vulkano/src/instance/instance.rs#L367)):
```
extern crate vulkano;

use vulkano::instance::{Instance, InstanceExtensions, ApplicationInfo};
use vulkano::instance::Version;

fn main() {
    println!("{:?}", ApplicationInfo::from_cargo_toml());

    let version = Version {
        major: env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap(),
        minor: env!("CARGO_PKG_VERSION_MINOR").parse().unwrap(),
        patch: env!("CARGO_PKG_VERSION_PATCH").parse().unwrap(),
    };

    let name = env!("CARGO_PKG_NAME");

    println!("{:?}", ApplicationInfo {
        application_name: Some(name.into()),
        application_version: Some(version),
        engine_name: None,
        engine_version: None,
    });
}
```
Output:
```
ApplicationInfo { application_name: Some("vulkano"), application_version: Some(0.4.0), engine_name: None, engine_version: None }
ApplicationInfo { application_name: Some("my_crate_name"), application_version: Some(0.1.0), engine_name: None, engine_version: None }
```
`ApplicationInfo::from_cargo_toml()` always returns the cargo metadata of vulkano instead of the actual application (which I suspect was the intention).

GabrielMajeri:
This pull request fixes #500.

Until now, `from_cargo_toml` always filled the `ApplicationInfo` structure with the name and version of the `vulkano` crate, instead of the calling crate.

To fix this, `from_cargo_toml` is now a macro, which becomes an expression which gathers the required environment variables. It is meant to be invoked when the user's crate is built, not when `vulkano` is built.

### DIFF
--- a/vulkano/src/instance/instance.rs
+++ b/vulkano/src/instance/instance.rs
@@ -507,6 +507,33 @@ pub struct ApplicationInfo<'a> {
     pub engine_version: Option<Version>,
 }
 
+impl<'a> ApplicationInfo<'a> {
+    /// Builds an `ApplicationInfo` from the information gathered by Cargo.
+    ///
+    /// # Panic
+    ///
+    /// - Panics if the required environment variables are missing, which happens if the project
+    ///   wasn't built by Cargo.
+    ///
+    #[deprecated(note="Please use the `from_cargo_toml!` macro instead")]
+    pub fn from_cargo_toml() -> ApplicationInfo<'a> {
+        let version = Version {
+            major: env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap(),
+            minor: env!("CARGO_PKG_VERSION_MINOR").parse().unwrap(),
+            patch: env!("CARGO_PKG_VERSION_PATCH").parse().unwrap(),
+        };
+
+        let name = env!("CARGO_PKG_NAME");
+
+        ApplicationInfo {
+            application_name: Some(name.into()),
+            application_version: Some(version),
+            engine_name: None,
+            engine_version: None,
+        }
+    }
+}
+
 /// Builds an `ApplicationInfo` from the information gathered by Cargo.
 ///
 /// # Panic
--- a/vulkano/src/instance/instance.rs
+++ b/vulkano/src/instance/instance.rs
@@ -538,8 +538,12 @@ macro_rules! from_cargo_toml {
 
 impl<'a> Default for ApplicationInfo<'a> {
     fn default() -> ApplicationInfo<'a> {
-        // Use vulkano's version and name as default.
-        from_cargo_toml!()
+        ApplicationInfo {
+            application_name: None,
+            application_version: None,
+            engine_name: None,
+            engine_version: None,
+        }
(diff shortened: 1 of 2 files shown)


vulkano/src/instance/instance.rs
```diff
@@ -507,35 +507,39 @@ pub struct ApplicationInfo<'a> {
...
 
-impl<'a> ApplicationInfo<'a> {
-    /// Builds an `ApplicationInfo` from the information gathered by Cargo.
-    ///
-    /// # Panic
-    ///
-    /// - Panics if the required environment variables are missing, which happens if the project
-    ///   wasn't built by Cargo.
-    ///
-    pub fn from_cargo_toml() -> ApplicationInfo<'a> {
```
tomaka: Please leave this method but put a `#[deprecated]` attribute to it instead.

vulkano/src/instance/instance.rs
```diff
@@ -507,35 +507,39 @@ pub struct ApplicationInfo<'a> {
...
+            }
         }
     }
 }
 
 impl<'a> Default for ApplicationInfo<'a> {
     fn default() -> ApplicationInfo<'a> {
-        ApplicationInfo::from_cargo_toml()
+        // Use vulkano's version and name as default.
+        from_cargo_toml!()
```
tomaka: I think this should simply be `None` everywhere instead.

vulkano/src/instance/mod.rs
```diff
@@ -81,13 +81,14 @@
 //! behavior for your application alone through a control panel.
 //!
 //! ```no_run
+//! # #[macro_use] extern crate vulkano;
```
tomaka: If you add an `extern crate` statement in a doctest, rustdoc will think that you are outside of the `main` function (whereas if you don't put one, it will think you are inside). Therefore you have to add `# fn main() {` and `# }` as well around the rest of the code.

vulkano/src/instance/instance.rs
```diff
@@ -533,9 +534,43 @@ impl<'a> ApplicationInfo<'a> {
...
 
+/// Builds an `ApplicationInfo` from the information gathered by Cargo.
+///
+/// # Panic
+///
+/// - Panics if the required environment variables are missing, which happens if the project
+///   wasn't built by Cargo.
+///
+#[macro_export]
+macro_rules! from_cargo_toml {
```
tomaka: One more nit: since macros aren't tied to a module, I think the macro should be named `app_info_from_cargo_toml` instead. Otherwise it's not obvious that it's related to `ApplicationInfo`.

Issue #320, opened 2016-11-19
Allow the user to configure the limits


tomaka:
When you use Vulkan you have to explicitly enable the extensions and features that you are going to use at initialization. If you don't enable a given feature and try to use it, you will get an error. This means that you can check at initialization whether any given hardware supports the features&extensions that your application needs to work.

It would be nice if there was a similar thing for limits. Right now it is possible that your code works on a given hardware but doesn't on another hardware because its limits are inferior to what your code requires. For example you're trying to push 150 bytes of push constants while hardwares only support 128 bytes.

I think it should be possible to specify hand-written limits when you create the device, and vulkano will check your code against these hand-written limits instead of your hardware's limits. Then when your code runs on a different hardware, it will check at initialization whether the hardware supports the required limits.

PR #785, opened 2017-08-27
Extract the Limits struct in its own module

tomaka:


tomaka:
This is some preliminary work for #320

PR #789, opened 2017-08-28
Implement `device::QueuesIter::device(&self)`.

Binero:
This way people can get access to the device without first popping a queue from the iterator. 

To me it seems silly now that `Device::new` returns a device and a `QueuesIter`, even though the former is easily accessible through the latter. This might be a point of future refactoring?

### DIFF
--- a/vulkano/src/device.rs
+++ b/vulkano/src/device.rs
@@ -527,6 +527,13 @@ pub struct QueuesIter {
     families_and_ids: SmallVec<[(u32, u32); 8]>,
 }
 
+impl QueuesIter {
+    /// Returns the device the queues were produced for.
+    pub fn device(&self) -> &Arc<Device> {
+        &self.device
+    }
+}
+
 impl Iterator for QueuesIter {
     type Item = Arc<Queue>;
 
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks for the PR. However I'd prefer if the `DeviceOwned` trait was implemented instead: https://docs.rs/vulkano/0.6.1/vulkano/device/trait.DeviceOwned.html

Binero:
I noticed that trait, but figured as `Queue` doesn't implement that
`QueuesIter` should probably not either. Do you want me to change over
both, or just `QueuesIter`. If we change over `Queue` as well, that would
be a breaking change.

tomaka:
Using the trait is the right method. But instead of removing the `device()` method on `Queue`, just deprecate it instead.

tomaka:
Oh sorry, I forgot that deprecating `device()` would have consequences on the rest of the code.
Please don't deprecate `device()` then. Let's just have both `DeviceOwned` and `device()`.

Binero:
Undid the deprecation changes in the last commit and squashed it into it.

PR #792, opened 2017-08-31
Fixes check_blit_image

### LINKED ISSUE #791: Blit resulting in SourceFormatNotSupported despite having transfer_source set to true
The docs mention "Can be used as a source for transfers. Includes blits." on transfer_source in ImageUsage, but when actually using an image with this set as a source for blitting this error occurs.

LaylConway:
Fixes inverted checks and a check on the wrong structure in `check_blit_image`. (Fixes #791)


tomaka:
Err... that explains it.
The funniest part is that I quickly read this code after seeing your issue but didn't even spot anything wrong.

Thanks!

Issue #793, opened 2017-08-31
Missing checks in check_blit_image


nicokoch:
While reading over the `check_blit_image` code, I noticed here: https://github.com/vulkano-rs/vulkano/blob/5270ac0029be3efd333516fbdb291ae4d75cb574/vulkano/src/command_buffer/validity/blit_image.rs#L75 and here: https://github.com/vulkano-rs/vulkano/blob/5270ac0029be3efd333516fbdb291ae4d75cb574/vulkano/src/command_buffer/validity/blit_image.rs#L82

the code does not check for `destination.format().ty() == FormatTy::Sint && source.format().ty() != FormatTy::Sint`. The specs say that both must be checked: ```If *either* of srcImage or dstImage was created with a signed integer VkFormat, the other must also have been created with a signed integer VkFormat```

I currently don't have time for a PR, but I figured I'll create an issue so we don't forget.

PR #794, opened 2017-08-31
Fixes blitting validation

LaylConway:
Some extra fixes to blitting:
- Fixes image layout check on source, this was checking for TransferDstOptimal instead of TransferSrcOptimal.
- Fixes bottom_right dimensions check, this should be 1 instead of 0 on dimensions beyond the ones applicable to the image.
- Fixes incorrect array_layers bounds check, which needed to be `>` rather than `>=`.

Added on request:
- Fixes format type validation


nicokoch:
While you're at it, would you mind including fixes for #793 ?

tomaka:
Thanks!

PR #817, opened 2017-09-09
Fix layer count check in swapchain constructor

Systemcluster:
I assume this is the correct behavior since the existing check is impossible to trigger.


tomaka:
Thanks!
Could you add an entry in `CHANGELOG.md` as well?

Systemcluster:
Sure, done!

tomaka:
Yes, thanks

PR #818, opened 2017-09-11
Provide 32-bit word constructor for ShaderModule

udoprog:
This provides the alternative constructor `ShaderModule::from_words(..)`, which uses `&[u32]` for the SPIR-V code.

`ShaderModule` currently uses an argument of type `&[u8]`, which causes interoperability issues with libraries like [shaderc][shaderc] and [rspirv][rspirv].

This is spec-compliant since the byte-code of the shader module is defined as `[..] a series of 32-bit words in host endianness.`

[shaderc]: https://docs.rs/shaderc/0.2.0/shaderc/struct.CompilationArtifact.html
[rspirv]: https://docs.rs/rspirv/0.4.0/rspirv/binary/trait.Assemble.html#tymethod.assemble

### DIFF
--- a/vulkano/src/pipeline/shader.rs
+++ b/vulkano/src/pipeline/shader.rs
@@ -53,7 +53,7 @@ pub struct ShaderModule {
 }
 
 impl ShaderModule {
-    /// Builds a new shader module from SPIR-V.
+    /// Builds a new shader module from SPIR-V bytes.
     ///
     /// # Safety
     ///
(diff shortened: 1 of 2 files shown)


vulkano/src/pipeline/shader.rs
```diff
@@ -63,14 +63,37 @@ impl ShaderModule {
...
+
+    /// Builds a new shader module from SPIR-V.
+    ///
+    /// # Safety
+    ///
+    /// - The SPIR-V code is not validated.
+    /// - The SPIR-V code may require some features that are not enabled. This isn't checked by
+    ///   this function either.
+    ///
+    pub unsafe fn from_ptr(device: Arc<Device>, spirv: *const u32, spirv_len: usize) -> Result<Arc<ShaderModule>, OomError> {
```
tomaka: I'd prefer if `from_ptr` wasn't public for now. We can always add it later.

tomaka:
Thanks for the PR.
Could you also please add a line to CHANGELOG.md?

udoprog:
Thanks for the quick feedback.

Added an entry to CHANGELOG.md, and `from_ptr` is no longer public.

PR #819, opened 2017-09-12
Bufferless vertex definition

Ralith:
This makes it convenient to perform a draw without any buffers at all, which is useful when you have a pipeline that comes up with vertex data on its own by some mechanism. For example, this can be used to [simplify full-screen passes](https://www.saschawillems.de/?page_id=2122).

It seems to unfortunately be necessary to implement the inappropriate trait `VertexSource<Vec<T>>` for this to be usable in a graphics pipeline, and it is probably not currently usable through a boxed abstract pipeline for related reasons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Changed `CpuBufferPool::next()` and `chunk()` to return a `Result` in case of an error when
   allocating or mapping memory.
 - Fixed `layers` argument validation in `Swapchain::new_inner`.
+- Added `vulkano::pipeline::vertex::BufferlessDefinition` and `BufferlessVertices` to enable
+  bufferless drawing, which is useful when drawing with shaders that compute their vertices
+  internally.
 
 # Version 0.6.2 (2017-09-06)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@
   allocating or mapping memory.
 - Fixed `layers` argument validation in `Swapchain::new_inner`.
 - Added `vulkano::pipeline::vertex::BufferlessDefinition` and `BufferlessVertices` to enable
-  bufferless drawing, which is useful when drawing with shaders that compute their vertices
-  internally.
+  bufferless drawing.
 
 # Version 0.6.2 (2017-09-06)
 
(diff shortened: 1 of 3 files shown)


vulkano/src/pipeline/vertex/bufferless.rs
```diff
@@ -0,0 +1,43 @@
...
+use pipeline::vertex::VertexSource;
+use buffer::BufferAccess;
+
+/// Implementation of `VertexDefinition` for drawing with no buffers at all.
+///
+/// This is only useful if your shaders come up with vertex data on their own, e.g. by inspecting
+/// `gl_VertexIndex`
+pub struct BufferlessDefinition;
+
+unsafe impl VertexSource<usize> for BufferlessDefinition {
```
tomaka: I think this should not be `usize`, but instead a struct that contains both the number of vertices and instances.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - Changed `CpuBufferPool::next()` and `chunk()` to return a `Result` in case of an error when
   allocating or mapping memory.
 - Fixed `layers` argument validation in `Swapchain::new_inner`.
+- Added `vulkano::pipeline::vertex::BufferlessDefinition` and `BufferlessVertices` to enable
+  bufferless drawing, which is useful when drawing with shaders that compute their vertices
```
tomaka: I don't think it's necessary to add what's after the comma.

vulkano/src/pipeline/vertex/bufferless.rs
```diff
@@ -0,0 +1,52 @@
...
+/// This is only useful if your shaders come up with vertex data on their own, e.g. by inspecting
+/// `gl_VertexIndex`
+pub struct BufferlessDefinition;
+
+/// Value to be passed as the vertex source for bufferless draw commands.
+///
+/// Note that the concrete type of the graphics pipeline using `BufferlessDefinition` must be
+/// visible to the command buffer builder for this to be usable.
+pub struct BufferlessVertices {
+    vertices: usize,
```
tomaka: These fields need to be public.

tomaka:
Thanks

Issue #738, opened 2017-08-13
The triangle example crashes with UnsupportedDimensions on MacOS 10.12 + MoltenVK


golddranks:
I'm using a 2017 MacBook Pro with Intel Iris Graphics 550 1536 MB and a 2560 x 1600 display.

The example builds successfully, and when I run it, it starts fine, printing some messages, including:
`[mvk-info] You require a license for the MoltenVK Vulkan Core feature set. Reverting to evaluation mode`

After that, it panics when creating a new `Swapchain`. The error is `UnsupportedDimensions`. Debug-printing the dimensions passed to the `new` method, the dimensions turn out to be `[2048, 1536]`. The device capabilities are 

```
Capabilities {
	min_image_count: 2,
	max_image_count: Some(2),
	current_extent: Some([1024, 768]),
	min_image_extent: [1024, 768],
	max_image_extent: [1024, 768],
	max_image_array_layers: 1, 
	supported_transforms: SupportedSurfaceTransforms {
		identity: true, 
		rotate90: false, 
		rotate180: false, 
		rotate270: false, 
		horizontal_mirror: false, 
		horizontal_mirror_rotate90: false, 
		horizontal_mirror_rotate180: false, 
		horizontal_mirror_rotate270: false, 
		inherit: false
	}, 
	current_transform: Identity, 
	supported_composite_alpha: SupportedCompositeAlpha {
		opaque: true, 
		pre_multiplied: true, 
		post_multiplied: true, 
		inherit: false
	}, 
	supported_usage_flags: ImageUsage {
		transfer_source: true, 
		transfer_destination: true, 
		sampled: true, 
		storage: true, 
		color_attachment: true, 
		depth_stencil_attachment: false, 
		transient_attachment: false, 
		input_attachment: false
	}, 
	supported_formats: [
		(B8G8R8A8Unorm, SrgbNonLinear),
		(B8G8R8A8Srgb, SrgbNonLinear),
		(R16G16B16A16Sfloat, SrgbNonLinear)
	], 
	present_modes: SupportedPresentModes {
		immediate: false, 
		mailbox: false, 
		fifo: true, 
		relaxed: false
	}
}
```

So it seems that it fails because of the max/min image extent checks.

I'm not knowledgeable enough to determine which of these values are wrong, but if I understand the meaning of `max/min_image_extent` capabilities right, it seems odd to me that the device supports only that. What should I do to fix this or to diagnose the problem further?

tomaka:
[...]

PR #830, opened 2017-09-21
Fixed the layer scaling issue on macOS.

golddranks:
Fixed the layer scaling issue on macOS, as `metal-rs` published the needed API.

Also changed the triangle example to use the surface extents instead of window dimensions, as that seems to be a more robust source of correct information. (At least one user reported that on their X.org-based system the surface was created according to the outer dimensions of the window, not inner, which is contrary to macOS. I don't know about Windows.)

The triangle example now runs straight out of the box on my macOS-based system, if the MoltenVK dynamic library is installed. Also, according to the MoltenVK support forums, the next version is going to fix the excessive logging.

Fixes https://github.com/vulkano-rs/vulkano/issues/738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fixed an HiDPI scaling issue on macOS in vulkano-win.
+
 # Version 0.7.0 (2017-09-21)
 
 - Added `RuntimePipelineDesc`, an implementation of `PipelineLayoutDesc` that makes creating custom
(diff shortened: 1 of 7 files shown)


tomaka:
Could you add a line in `CHANGELOG.md` about the fix in vulkano-win?

The fix to the example should also be propagated to the vulkano-examples repository eventually (not necessarily by you).

golddranks:
I updated the changelog. I also propagated the fix to the other in-repo examples, but haven't looked at the vulkano-examples repo at all.

examples/src/bin/.vscode/tasks.json
```diff
@@ -0,0 +1,46 @@
+{
```
tomaka: Please remove

golddranks: Ugh sorry. I hate when VS Code does that.

golddranks:
I `git commit --amend` and `git push --force`d that to prevent littering the history with a unrelated stuff like that.

PR #832, opened 2017-09-23
Fixed le() and ge() for DescriptorsCount

### LINKED ISSUE #808: Possible Memory Leak with Descriptor Sets
I tried updating to 0.6.2 today, but when I started my game up the fps started dropping rather fast. Seems as if graphics memory is getting allocated constantly. Xorg will eventually freeze which also hints toward a graphics memory leak. I found that https://github.com/vulkano-rs/vulkano/commit/80e965aef51dd860822cfbc86583404a78f707ec caused the issue to start occurring.

I do make many new PersistentDescriptorSet every frame. I am not sure if there is a more proper way to do sets or not.

IcyDefiance:
Also fixes #808, because checking for space in a descriptor pool uses `ge()` at [descriptor/descriptor_set/std_pool.rs:80](https://github.com/vulkano-rs/vulkano/blob/089d0879f7460b89c68de84c37671467ec359c76/vulkano/src/descriptor/descriptor_set/std_pool.rs#L80).

It may be better to scrap the PartialOrd trait entirely and implement comparison functions that must be called explicitly, but this is the easy solution and it's less of a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Fixed `le()` and `ge()` for `DescriptorsCount`, which also fixes a potential memory leak when
+  allocating descriptor sets.
+
 # Version 0.7.0 (2017-09-21)
 
 - Added `RuntimePipelineDesc`, an implementation of `PipelineLayoutDesc` that makes creating custom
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks a lot for the investigation! I didn't really think of that.

PR #842, opened 2017-10-01
Add an unsafe method to execute an arbitrary secondary vk::CommandBuffer

Ralith:
This makes it possible to execute secondary `UnsafeCommandBuffer`s, for example.


tomaka:
I guess this doesn't warrant an entry in the changelog.

PR #847, opened 2017-10-05
Minimal command buffer debug marker support

Ralith:
Not yet tested.

Exposes debug markers for unsafe command buffers only. Extension to auto command buffers should be straightforward future work. No attempt was made to provide a good interface for setting object names/tags. That could be done well through a trait; perhaps new methods on `VulkanObject`?

This is needed to support a good experience for debugging with tools like renderdoc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Allow `impl_vertex!` to support generic structs.
 - Fixed creating a buffer view not checking the `min_texel_buffer_offset_alignment` limit.
+- Added support for loading the `VK_EXT_debug_marker` extension and adding debug markers to
+  `UnsafeCommandBufferBuilder`
 
 # Version 0.7.1 (2017-09-28)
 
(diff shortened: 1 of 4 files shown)


Ralith:
One design question that *does* need to be addressed is whether command buffer builders (unsafe or otherwise) should automatically suppress these calls when the extension isn't loaded, rather than attempting them unconditionally and panicing. The former would be more convenient, but downstream code could accomplish the same behavior manually with fewer branches.

GabrielMajeri:
Debug markers are a feature used by developers, not users, right? So we could have this behind a `#[feature(debug_markers)]`.

That way, developers can enable this extension only on development machines (after checking that the extension is supported).

Ralith:
I'm not sure why a feature flag would be useful. This PR does not introduce significant code complexity, compatibility restrictions, or new dependencies. You can choose to enable the extension or not according to whatever criteria suit you.

nicokoch:
Feature gates are nightly only as far as I'm aware.

GabrielMajeri:
@nicokoch I meant a [Cargo.toml feature](http://doc.crates.io/manifest.html#the-features-section), fixed it in the comment.

Ralith:
After some thought I'm leaning towards suppressing debug markers automatically, because it is a substantial ergonomics win and the performance cost should be trivial, especially for a debugging aid.

tomaka:
> After some thought I'm leaning towards suppressing debug markers automatically

The problem is that people may try to use these functions without enabling the extension, and then think they don't work.

Ralith:
Ah, good point. I'll back that out then.

PR #848, opened 2017-10-06
Allow Surface to reference an external window object

### LINKED ISSUE #844: Dropping Window before Swapchain can lock up program
Currently, the `Swapchain` and `Window` objects are totally unrelated. This means that it's possible to drop a `Window` while a `Swapchain` derived from it is still in use. At least under the Linux Intel driver (17.0.7), this will cause the driver to lock up in `vkDestroySwapchainKHR`, waiting on an event for a window that no longer exists.

When all of your Vulkan objects are function-local variables, this drop order is implicitly going to be "right" most of the time (the opposite order you initialized them in). When storing them in structs, though, it's easy to get wrong. The following struct will lock up when it is dropped:

```
struct VkObjects {
    previous_frame_end: Box<GpuFuture>,
    framebuffers: Option<Vec<Arc<FramebufferAbstract + Send + Sync>>>,
    render_pass: Arc<RenderPassAbstract + Send + Sync>,
    images: Vec<Arc<SwapchainImage>>,
    window: Window,
    swapchain: Arc<Swapchain>,
    queue: Arc<Queue>,
    device: Arc<Device>,
    event_loop: EventsLoop,
}
```

I'm unsure if this is a bug in Vulkano or the driver (or just in my code). I couldn't find anything in the Vulkan spec that says a native window must outlive its associated `VkSurface`, but I'm also not surprised that the driver makes this assumption.

This can *probably* be solved with a new `WindowAbstract` trait that can be `Arc`'d by the `Surface`, and another layer of indirection around window handling. I'm willing to do that work if you're interested in that approach.

branan:
This makes the `Surface` struct generic across a Window type, so that a surface can own (or reference, if W is an Arc) an external window object.

This also updates vulkano-win to take advantage of this. There is no longer a dedicated `Window` struct in vulkano-win. Instead, the Surface is simply passed ownership of the `winit::Window`.

This fixes #844


branan:
Because this changes a bunch of public fns, this would presumably need to wait for 0.8.0

tomaka:
Thanks for the PR.
However I think that a better way to do would be to make `Surface` generic. In other words, turn `Surface` into `Surface<T>`. This way the `vulkano_win::Window` struct would hold a `Surface<winit::Window>`. It would also remove the necessity for a `WindowAbstract` trait.

Also, could you fill `CHANGELOG.md` with the changes?

branan:
I've gone ahead and tried that approach. It ended up touching a lot of code - `Swapchain` holds an `Arc<Surface>`, and `Swapchain` gets used in a few places. A nice advantage of this approach is that we don't actually need `vulkano_win::Window` anymore - `build_vk-surface` can just return a Surface that owns a `winit::Window`. I've done that as well in this update.

This still needs a CHANGELOG update, and I need to make a pass on the examples to update them for the new Surface/Window API usage. I wanted to get your opinion on these updated changes before I did that work, though.

vulkano/src/swapchain/surface.rs
```diff
@@ -35,7 +35,10 @@ use vk;
 /// Represents a surface on the screen.
 ///
 /// Creating a `Surface` is platform-specific.
-pub struct Surface {
+pub struct Surface<W>
+    where W: Send + Sync
```
tomaka: This `where W: Send + Sync` shouldn't be required, and it's polluting the code everywhere.

branan:
I've updated a few more comments and rebased for the changelog conflict. I believe this is ready to go now, unless you'd like me to squash it.

tomaka:
The PR is good, but I don't want to merge it right now as I'm trying to avoid bumping the major version too often.

branan:
Commit churn is just me fixing up to use my personal instead of work email on these commits. Apologies for any noise.

PR #849, opened 2017-10-06
impl_vertex accept generic structs

porky11:
I changed the `impl_vertex!` macro to accept `ty` as first token, instead of `ident`
This way, it's possible to accept generic structs for the impl_vertex macro.
Following wasn't possible before:
```
#[derive(Copy, Clone)]
struct Vertex<T> {
    pos: T
}
impl_vertex!(Vertex<f32>, pos);
```


porky11:
A similar issue has been in glium, I think. You may look at it.

tomaka:
Could you add a line in the CHANGELOG as well?

porky11:
I updated changelog.

PR #852, opened 2017-10-07
Debug object names

Ralith:
Depends on #847.

This commit makes all vk-sys handle types strong, which is a breaking change. I'd like to get some comments on the design of the new `Handle` trait in vk-sys before this is merged.

This use of associated constants triggers a panic bug in rustc 1.20.0. The bug appears to be fixed (tests pass) as of 1.21.0-beta.3 or earlier.

### DIFF
--- a/vulkano/src/device.rs
+++ b/vulkano/src/device.rs
@@ -470,9 +470,13 @@ impl Device {
 
     /// Assigns a human-readable name to `object` for debugging purposes.
     ///
-    /// Does nothing if the `VK_EXT_debug_marker` device extension is not loaded.
-    pub fn set_object_name<T: VulkanObject>(&self, object: &T, name: &CStr) -> Result<(), OomError> {
-        if !self.extensions.ext_debug_marker { return Ok(()); }
+    /// Requires the `VK_EXT_debug_marker` device extension.
+    ///
+    /// # Panics
+    /// * If the `VK_EXT_debug_marker` device extension is not loaded.
+    /// * If `object` is not owned by this device.
+    pub fn set_object_name<T: VulkanObject + DeviceOwned>(&self, object: &T, name: &CStr) -> Result<(), OomError> {
+        assert!(object.device().internal_object() == self.internal_object());
         let info = vk::DebugMarkerObjectNameInfoEXT {
             sType: vk::STRUCTURE_TYPE_DEBUG_MARKER_OBJECT_NAME_INFO_EXT,
             pNext: ptr::null(),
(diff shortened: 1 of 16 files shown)


vulkano/src/device.rs
```diff
@@ -465,6 +467,22 @@ impl Device {
     pub(crate) fn event_pool(&self) -> &Mutex<Vec<vk::Event>> {
         &self.event_pool
     }
+
+    /// Assigns a human-readable name to `object` for debugging purposes.
+    ///
+    /// Does nothing if the `VK_EXT_debug_marker` device extension is not loaded.
+    pub fn set_object_name<T: VulkanObject>(&self, object: &T, name: &CStr) -> Result<(), OomError> {
```
tomaka: This is UB if object doesn't belong to this device (if you create multiple devices).

tomaka:
I don't really see the advantage of strong-typing the objects in vk-sys and adding a trait, compared to putting the object type on `VulkanObject`.

Ralith:
Strong-typing the vk-sys objects is a nice-to-have in its own right, IMO; now the compiler will yell at you if you confuse a `vk::Fence` with a `vk::Semaphore`. Even the C API accomplishes this (at least on 64-bit platforms).

That aside, the upside to introducing a new trait for the raw handles is that the information doesn't need to be repeated for every single high-level wrapper of the same inner object type; we have a number of cases of multiple-wrapping and I think we expect to get more as the use-cases vulkano supports expand. It also allows for the possibility of debug-type-sensitive APIs which operate directly on raw objects--in fact I think I'll add a `set_object_name_raw` method.

Ralith:
I resolved the UB you identified by requiring `VulkanObject + DeviceOwned` for `set_object_name`. This strengthens the motivation for the vk-sys changes, as now `set_object_name_raw` is *required* for unsafe wrappers which often do not implement `DeviceOwned` (e.g. `UnsafeDescriptorSet`). An alternative approach would be to add device handles to all unsafe wrappers currently lacking them and mandate `DeviceOwned` across the board, but this seems like too big a hammer (and would be tantamount to merging `DeviceOwned` into `VulkanObject`).

Ralith:
On review, a more realistic alternative approach would be to simply require `VulkanObject` for `set_object_name_raw`. I like having the option to manipulate raw handles directly as a worst-case fallback, but it's not totally mandatory, and is marginally less ergonomic when working with wrappers.

Ralith:
Closing in favor of #895, since it appears that this is too contentious to proceed.

Issue #866, opened 2017-10-15
Question about copy_image_to_buffer_dimensions()


AustinJ235:
Is there any reason this:

https://github.com/vulkano-rs/vulkano/blob/c60265c67b59d1ee59f324b63f9472b98482cd1a/vulkano/src/command_buffer/auto.rs#L733-L741

can't be this?

```rust
image_aspect: UnsafeCommandBufferBuilderImageAspect {
    color: source.has_color(),
    depth: source.has_depth(),
    stencil: source.has_stencil()
}
```

I ran into this as I was trying to read the depth buffer to get a world position of the center of the screen. I replaced it with that I suggested and it seems to work as intended. I do however do not have a stencil, so I am not sure about that part. It seems to work as intended so is there a reason for the unimplemented!()?

tomaka:
If I remember correctly, it's because there are additional validity restrictions when it comes to depth/stencil copies which vulkano would have to check.

AustinJ235:
I haven't used vulkan at all so my knowledge is very limited in that regard, but from what I read from the spec your checks seem valid. The only concern is that the source image needs to have linear tiling if I am correct. I read that you need to copy the non-linear tiling image to a temporary image with linear tiling and then copy that to a buffer.

Currently I am reading the depth from an attachment image, those seem to be created with non-linear tiling. So from what I read that shouldn't work or at least should be illegal. It does however produce the result I would expect. It works both on Linux and Windows using my amd card.

Is there a more proper way to read the depth using the current state of vulkano?

tomaka:
> I haven't used vulkan at all so my knowledge is very limited in that regard, but from what I read from the spec your checks seem valid. The only concern is that the source image needs to have linear tiling if I am correct. I read that you need to copy the non-linear tiling image to a temporary image with linear tiling and then copy that to a buffer.

Indeed it looks like the code you proposed is correct. This `unimplemented!()` was probably meaning something more like "need to check the specs before allowing that".

PR #868, opened 2017-10-16
Allow depth/stencil images to be used with AutoCommandBufferBuilder::copy_image_to_buffer()

AustinJ235:
As I suggested here: https://github.com/vulkano-rs/vulkano/issues/866


AustinJ235:
Well made a idiot mistake on the first commit... changed the wrong function. I changed it to the correct function and made sure it actually compiles this time.

tomaka:
Thanks!

PR #869, opened 2017-10-18
Clear value validation for AutoCommandBufferBuilder::begin_render_pass()

AustinJ235:
I had a go at validating clear values. Let me know your thoughts. It works as atm but I have not tested too much. I added indexes for the clear values and attachments as it would make debugging easier for the end user.


tomaka:
Thanks for the PR.
Looks good, but please add an entry in `CHANGELOG.md`, and also a `TODO` comment near your code saying that these panics should be turned into proper errors eventually.

PR #870, opened 2017-10-23
Do not try to hardlink glslangValidator

### LINKED ISSUE #95: glslang_validator not copied during build correctly
I'm on linux, and trying to compile the examples is failing because glslang_validator is not generating the output file it is supposed to. It turns out that my 
...vulkano/examples/target/debug/build/glsl-to-spirv-11036aceadb12936/out/glslang_validator application is 0 bytes. It returns success when you run (apparently), but it doesn't do anything.

If I manually copy ...vulkano/examples/target/debug/build/glsl-to-spirv-11036aceadb12936/out/bin/glslangValidator to ...vulkano/examples/target/debug/build/glsl-to-spirv-11036aceadb12936/out/glslang_validator then it builds.

I assume that is what this code in glsl-to-spirv/build/build.rs is supposed to do:

```
    if let Err(_) = fs::hard_link(&path, &out_file) {
        fs::copy(&path, &out_file).expect("failed to copy executable");
    }
```

However I was unable to figure out how to debug that code. (My attempts to print out paths didn't add any output to cargo build).

branan:
Previously, it was possible for the hardlinked glslangValidator to become truncated when the glsl-to-spirv crate was rebuilt. The file would be successfully hardlinked the first time, but on subsequent builds the hardlink attempt would fail becasue the target already exists. This would cause the build script to fall back to a copy, which truncates when source and dest are the same file.

This removes the hardlinking entirely, meaning that we always just copy our built glslangValidator to where we want it. While it would be possible to try to make the hardlinking/copying logic idempotent, the complexity (and potential fragility to similar bugs in the future) doesn't seem worth it to avoid one copy.

This fixes #95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Changed `ShaderInterfaceMismatchError` to be more verbose.
 - Allow depth/stencil images to be used with `AutoCommandBufferBuilder::copy_image_to_buffer()`
 - Clear value validation for `AutoCommandBufferBuilder::begin_render_pass()`
+- Fix occasional truncation of glslang_validator when glsl-to-spirv is rebuilt
 
 # Version 0.7.2 (2017-10-09)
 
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks!

PR #872, opened 2017-11-05
Fix `Instance` example

GabrielMajeri:
This pull request fixes an example in the documentation. Previously it didn't build because of ownership issues.

This changes the `Instance::new` / `Instance::with_loader` functions to take an `Iterator` to `Item = &'a str`, instead of `Item = &'a &'a str`.

It also replaces `unwrap` with `?`.


tomaka:
The problem is that this is a breaking change.

GabrielMajeri:
@tomaka It is, however there is probably little code out there which uses layers.

The only layers which exist right now are the validation layers, and the docs recommend enabling them using environment variables, not manually.

We're just future-proofing the code here, to make it easier to use when layers become more popular.

In other words, most code would be using `None` for layers, meaning no breakage.

P.S.: should I also update the changelog in this pull request, mentioning this change?

rukai:
This looks good to me. However, it needs an entry in the changelog before I can merge it.

rukai:
Thanks

PR #876, opened 2017-11-12
Updated Dependencies.

thelearnerofcode:


vulkano-win/Cargo.toml
```diff
@@ -9,9 +9,9 @@ categories = ["rendering::graphics-api"]
 
 [dependencies]
 vulkano = { version = "0.7.0", path = "../vulkano" }
-winit = "0.7.0"
+winit = "0.8.0"
```
tomaka: That would be a breaking change, so unfortunately this needs to be reverted.

vulkano/Cargo.toml
```diff
@@ -12,9 +12,9 @@ build = "build.rs"
...
-shared_library = "0.1.5"
-smallvec = "0.4.3"
-lazy_static = "0.2.2"
+fnv = "1.0.6"
+shared_library = "0.1.7"
+smallvec = "0.5.0"
+lazy_static = "0.2.9"
 vk-sys = { version = "0.3.2", path = "../vk-sys" }
-half = "1"
+half = "1.0.1"
```
fst3a: There is no reason to specify exact minor and patch versions, having just major is enough like it has been before.

PR #880, opened 2017-11-17
Fix linking against MoltenVK

### LINKED ISSUE #879: MoltenVK >=0.19.0 depends on IOSurface.framework
This is currently not linked, resulting in linker errors when building vulkano-examples.

jansol:
Fixes #879


tomaka:
Thanks, could you also add a line in the CHANGELOG?

jansol:
There you go.

jasondavies:
Just ran into this.  Would be nice to get a version bump with this included, if possible.  Thanks for the great work.

Issue #883, opened 2017-11-20
vkCmdCopyImage command is not exposed?


arianvp:
the more general `vkCmdBlitImage`  command seems to be exposed, but `vkCmdCopyImage` is not.

tomaka:
When a function isn't exposed, it's generally because its API hasn't been made safe yet unfortunately. I think that's the case here.

Rua:
This seems to have been implemented with #888.

PR #888, opened 2017-11-29
Expose vkCmdCopyImage

cbbowen:
Exposes vkCmdCopyImage as requested in #883 through copy_image methods on AutoCommandBufferBuilder and UnsafeCommandBufferBuilder.


tomaka:
Reading your code, I now remember that the reason why this function is not exposed is because of the compressed/uncompressed thing.
However leaving that as a TODO is fine to me.

The PR looks good, please also add a line in the `README.md` and I will merge!
Thanks!

cbbowen:
I've now included in the addition in `CHANGELOG.md`.

PR #894, opened 2017-12-12
Simplify constructing pipelines that use dynamic buffers

Ralith:
This should be feature-complete, but has not yet been tested.


Ralith:
I opted for `set_dynamic_buffers` so that other tweaks could be easily added in the future if needed. Per prior discussion we don't expect this, and it's not exported anyway, so I'll make the change..

PR #903, opened 2018-01-04
Update dependencies

GabrielMajeri:
This pull request updates dependencies for all the crates to the latest versions.


LPGhatguy:
Winit 0.11 came out since this PR was created; it causes no breaking changes with Vulkano.

tomaka:
It *does* cause a breaking change because winit is unfortunately exposed in the API.

tomaka:
Please add an entry in CHANGELOG.md and then I'll merge!

tomaka:
Releasing a version doesn't cost much, we can just release 0.9 :)

PR #904, opened 2018-01-05
Optimize SPIR-V code

GabrielMajeri:
This pull request adds an additional tool ([spirv-opt](https://www.lunarg.com/reduce-spir-v-binary-file-sizes/)) to the shader building pipeline. It is used to optimize the shader modules even further.

`glslangValidator` is good at compiling code, but it is meant to be a reference compiler, not an optimizer, therefore it lacks advanced optimizations.

The [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools) repository contains a tool called `spirv-opt` which can be used to optimize shader binaries. I haven't run any advanced benchmarks, but on the examples, this tool reduces shader code size by about 33% (by eliminating dead / unnecessary code).

This will also open the way to even more advanced optimizations as `spirv-opt` gets improved.

The pull request includes the `x64-Release` Windows binary from CI, available [here](https://github.com/KhronosGroup/SPIRV-Tools/releases). I understand if you want to remove it and add it yourself since I could've potentialy included malicious code.

In order to enable to use additional tools I had to change `glsl-to-spirv`. I refactored both the `build.rs` and the `lib.rs` files, it should be easier to chain more tools in the future.

<hr/>

To do:
- [x] ~Fix building on Linux.~ CI works now.
- [ ] Test this on more code to make sure it works. The examples seem to be fine, and I enabled no "unsafe" optimizations, but I don't know if everything will still work.
- [x] ~I tried adding the `--strip-debug` flag but it prevents shader reflection from working. @tomaka do you want to investigate or is it alright to merge without it?~ Won't enable it, unnecessary.


tomaka:
> I tried adding the --strip-debug flag but it prevents shader reflection from working. @tomaka do you want to investigate or is it alright to merge without it?

I suppose that `--strip-debug` does not add any further optimization, so I'm fine with merging without it.

GabrielMajeri:
@tomaka It removes some metadata required for reflection, but I think it's OK, since the driver would throw it away anyway.

There are other options we could enable over time, described in this [white paper](https://www.lunarg.com/shader-compiler-technologies/white-paper-spirv-opt/).

rukai:
glsl-to-spirv is no more https://github.com/vulkano-rs/vulkano/pull/947

antiagainst:
Related to this, since we already include the whole copy of SPIRV-Tools in Shaderc, it should be natural to extend the current shaderc-rs API to also expose SPIRV-Tools ones, if there is a need.

PR #905, opened 2018-01-06
Enforce one ClearValue per Attachment

rukai:
Breaking change.

The vulkan spec says:
```
clearValueCount must be greater than the largest attachment index in renderPass that
specifies a loadOp (or stencilLoadOp, if the attachment has a depth/stencil format)
of VK_ATTACHMENT_LOAD_OP_CLEAR
```

The easiest way to enforce this is to ensure there is one clear value per attachment.

No need to update the changelog as it already mentions the addition of the (unreleased) clear value validation that I am modifying.


tomaka:
Thank you very much! That was a pretty nasty problem!

Issue #857, opened 2017-10-07
Bug after resizing window containing swap chain


porky11:
I created a program using windows and the swapchain.
The swapchain part is pretty similar to the triangle example.
(I may upload it in git if you want to see. currently it's just in pijul)
The problem occurs after rendering, when calling this on some result:
`.then_signal_fence_and_flush().unwrap();`
Then the window doesn't update anymore, because it just stops there.
When I wait some time, it generates an error message and a stack backtrace, I'll include:

```
thread 'main' panicked at 'unexpected error: DeviceLost', /home/porky11/.cargo/registry/src/github.com-1ecc6299db9ec823/vulkano-0.7.1/src/lib.rs:161:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DeviceLostError', /checkout/src/libcore/result.rs:906:4
stack backtrace:
   0:     0x55e535584b73 - std::sys::imp::backtrace::tracing::imp::unwind_backtrace::hcae9eaf44818c983
                               at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1:     0x55e53558127a - std::sys_common::backtrace::_print::h59558c687fc4cf51
                               at /checkout/src/libstd/sys_common/backtrace.rs:71
   2:     0x55e5355863a3 - std::panicking::default_hook::{{closure}}::h48460b9cd8e002ae
                               at /checkout/src/libstd/sys_common/backtrace.rs:60
                               at /checkout/src/libstd/panicking.rs:381
   3:     0x55e535586112 - std::panicking::default_hook::h9a7143460d24c4d8
                               at /checkout/src/libstd/panicking.rs:397
   4:     0x55e535586867 - std::panicking::rust_panic_with_hook::h76b4e13bdfc8824a
                               at /checkout/src/libstd/panicking.rs:577
   5:     0x55e535586734 - std::panicking::begin_panic::hb353b753134545f4
                               at /checkout/src/libstd/panicking.rs:538
   6:     0x55e535586639 - std::panicking::begin_panic_fmt::hd2175356d78b1606
                               at /checkout/src/libstd/panicking.rs:522
   7:     0x55e5355865ca - rust_begin_unwind
                               at /checkout/src/libstd/panicking.rs:498
   8:     0x55e5355b9ec0 - core::panicking::panic_fmt::he3a4d8a4eb6d8e51
[...]

PR #908, opened 2018-01-07
Fixed issue #857 / swapchain resize causes panic

Kurble:
This pull reqeust contains a fix for issue #857: the device can be lost while trying to resize the swapchain.
After enabling validation layers the cause for this issue seems to be that one time submit buffers can be submitted by vulkano up to 4 times!

This is because the PresentFuture submits the commandbuffer of the CommandBufferExecFuture before it all by itself, without notifying the CommandBufferExecFuture of this. Replacing the submit call by a flush() fixes the problem.
This is what happens if you resize the swapchain and get an OutOfDate error:

`FenceSignalFuture::flush()` => build submissions for underlying futures
`PresentFuture::build_submission()` => submits command buffer (first time), then fails to swap because `OutOfDate`
FenceSignalFuture::flush() now returns an error, but the future is still pending. What to do? we could drop it:
`FenceSignalFuture::drop()` => submission is built again, including command buffer submit (second time), but fails (technically still because of `OutOfDate`)=> VALIDATION ERROR
`PresentFuture::drop()` => submission is built again but fails (third time) => VALIDATION ERROR
`ExecCommandBufferFuture::drop()` => submission is built again, succeeds (fourth time)   => VALIDATION ERROR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Clear value validation for `AutoCommandBufferBuilder::begin_render_pass()`
 - Fix occasional truncation of glslang_validator when glsl-to-spirv is rebuilt
 - Fix linking against MoltenVK >= 0.19.0 
+- Fix panic on DeviceLost error after resizing swapchain on nvidia/amd hardware
 - Added `AutoCommandBufferBuilder::copy_image`
 - Added `VulkanObject::TYPE` to look up the `DebugReportObjectTypeEXT` of an object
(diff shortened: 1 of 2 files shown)


rukai:
Thankyou so much!
I tested this change on Windows + nvidia/amd and linux + nvidia/amd.
All handle resizing correctly.

tomaka:
Thanks! Nice catch! Could you add an entry in the Changelog as Well?

Kurble:
(don't know if commenting matters versus commits for notifications)
Fixed!

PR #909, opened 2018-01-07
Added documentation to vulkano_shader_derive

iirelu:
This isn't full documentation of everything that the crate generates, but it at least covers how to use it. I'm only working off my reading of the source code, so it's entirely possible I got things wrong, but crappy documentation is better than no documentation. I think.

Note that this has to add vulkano itself as a dev-dependency in order for the doctest to compile, but that's also going to help with writing tests if that ever happens.

An improvement over this would be to detail what the proc macro itself generates and the functions, structs, etc you get from it. The details on that are in vulkan-shaders, and are a bit harder to decipher.

Please let me know if anything needs fixing!


tomaka:
Thanks! It's a good start, and we will be able to expand that in the future.

PR #916, opened 2018-02-10
Publish 0.7.3

tomaka:


jasondavies:
I think you still need to publish to crates.io.  Thanks!

PR #919, opened 2018-02-14
Fixed resizing bug in examples (#303)

JohnDoneth:
Fixed the OutOfDate error in the examples by match unwrapping

By match unwrapping the future returned from `then_signal_fence_and_flush` and recreating the swapchain; The examples are now able to be resized without crashing on unwrapping the err value on my GTX 1050 and GTX 970.

This fixes issue #303 when dealing with unwrapping `Err(OutOfDate)`


tomaka:
Thanks!

PR #921, opened 2018-02-14
Added ImmutableBuffer Example

JohnDoneth:
I've created an example mirroring the teapot example that uses `ImmutableBuffer`'s as there is a lack of examples that use them. It's a good example to have at least one example showing their usage as they are more efficient (due to being in video memory) and should be used instead of `CpuAccessibleBuffer` for data that doesn't change often or at all.

This could override teapot.rs if desired for simplicity's sake.

The example could be improved by using a semaphore instead of blocking via fence. (See: Line 101)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Changed `ShaderInterfaceMismatchError` to be more verbose.
 - Allow depth/stencil images to be used with `AutoCommandBufferBuilder::copy_image_to_buffer()`
+- Allow `Surface` to own the window it references.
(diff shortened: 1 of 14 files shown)


newpavlov:
It could be worth to add a small example to the `ImmutableBuffer` documentations as well. Also a really good example will be a video texture.

JohnDoneth:
I didn't realize that `DeviceLocalBuffer`'s is what I had mistaken `ImmutableBuffer`'s for, therefore I am closing this pull request as it is not really entirely relevant.

PR #938, opened 2018-03-27
Replace tempdir with tempfile

newpavlov:
`tempdir` is [deprecated](https://github.com/rust-lang-nursery/tempdir/issues/45) in favor of `tempfile`. As a bonus vulkano will not depend on `winapi` crate for linux and macos builds.


tomaka:
Thanks!

PR #942, opened 2018-04-06
Rename vulkano-rs/vulkano-shaders/tests directory to data

casey:
The `.rs` files in a directory named `tests` are assumed by cargo to be integration tests. As part of refactoring `vulkano-shaders`, I needed to store a `.rs` file containing the pre-refactor generated source code, to make sure that it didn't change. This caused cargo to think it was a test and try to compile it. To avoid that I renamed the `tests` to `data`. I think this is also a good idea since people will likely expect a `tests` directory to contain test source files, and not static data needed by tests.


rukai:
I think this is fine as is. `tests/*.rs` files can happily live alongside related data files.

Issue #928, opened 2018-03-01
Vulkano shouldn't link to MoltenVK by default on macOS


nlordell:
OK, so this is going to be a long block of text, and some code (there is a TL;DR at the end). With the open source release of MoltenVK and the 1.0.69 release of the macOS LunarG Vulkan SDK I believe that the linking by default to the MoltenVK framework should no longer happen.

On Windows, Linux and Android, a shared object is loaded and the `vkGetInstanceProcAddr` symbol is loaded from it. They rely on `vulkan.{dll,so}` being available to load. However, through the `Loader` trait it allows crate users to load `vkGetInstanceProcAddr` however they see fit. That means that if I want to add `println!("cargo:rustc-link-lib=vulkan");` to my `build.rs` and provide `vkGetInstanceProcAddr` I can (I think... didn't try it to be honest).

As of the  LunarG Vulkan SDK 1.0.69 for macOS there is a `libvulkan.1.dylib` loader that acts more like its Windows, Linux and Android counterparts (more on that later). That means that if `libvulkan.1.dylib` is in an expected place, then the `DynamicLibraryLoader` implementation of `Loader` will work. If we choose to go down that path we are still linking to MoltenVK (and other frameworks) even if it is no longer needed.

But why not just use the MoltenVK framework? The LunarG Vulkan SDK loader has a bunch of cool features that you might want (eg. support for `VK_LAYER_PATH` and `VK_INSTANCE_LAYERS` environment variables). You can take a look at this gist: https://gist.github.com/nlordell/184d363cffe0d7895155ee5fee89aa9a
The important bit is the output of running with dynamic loading of `libvulkan.1.dylib` vs linking against `MoltenVK`:
```
$ export VK_LAYER_PATH="/path/to/layer/configs"
$ cargo run # with dynamic loading of libvulkan.1.dylib
   Compiling mvk v0.1.0 (file:///Users/nlordell/Developer/rust/mvk)
    Finished dev [unoptimized + debuginfo] target(s) in 1.51 secs
     Running `target/debug/mvk`
VK_LAYER_GOOGLE_unique_objects
VK_LAYER_GOOGLE_threading
VK_LAYER_LUNARG_standard_validation
VK_LAYER_LUNARG_core_validation
VK_LAYER_LUNARG_parameter_validation
VK_LAYER_LUNARG_object_tracker
$ cargo run --features "framework-linking" # with linking against MoltenVK
[...]

PR #948, opened 2018-04-18
Use libvulkan.dylib instead of MoltenVK by default on macOS

nlordell:
Note that this is a breaking change. I added some more docs to the README.md but would be willing to create an example that uses MoltenVK instead of the default libvulkan on macOS. This PR is for issue #928


tomaka:
After thinking a but about it, I think that this change makes the most sense.

Thanks!

PR #950, opened 2018-04-25
Expose SPIR-V binary in generated shader code and add unsafe function to load arbitrary new binary

casey:
To support safe shader hot reloading, we'll need to analyze the shader modules original SPIR-V binary and the new shader module SPIR-V binary to ensure that the new SPIR-V binary is compatible with the old interface and can be safely substituted.

As a first step, this change exposes the old SPIR-V module binary so that it can be analyzed for compatibility, as well as adding the ability to load a new binary. The function to load a new binary is marked as unsafe, since loading a new binary with a different interface may crash the program.

Edit: Added binary loading functionality.


AustinJ235:
I don't think this is compatible anymore since we no longer use glsl-to-spriv

PR #951, opened 2018-04-26
Allow subpass dependencies on VK_SUBPASS_EXTERNAL

branan:
This updates the asserts in `RenderPass` creation to allow `VK_SUBPASS_EXTERNAL` as a special value in addition to any value less than the total number of subpasses. This enables custom unsafe implementations of `RenderPassDesc` to define their external dependencies.


tomaka:
Thanks!

Issue #89, opened 2016-05-07
vk-sys should be autogenerated from vk.xml


cwabbott0:
Instead of hand-coding all the definitions, they should be parsed from the same XML definitions that are used to create vulkan.h. This should also add support for the recent extensions for free. I have some WIP code that parses it and generates something that at least seems like rust, I'll try and beat it into shape so it can autogenerate vk-sys/lib.rs.

tomaka:
There's also the [vk_generator](https://crates.io/crates/vk_generator) crate which was written for this purpose.

tomaka:
See also this issue: https://github.com/KhronosGroup/Vulkan-Docs/issues/210

kvark:
I'd like to bump this in priority, if possible, in relation to #525. Being able to run on Metal, D3D12, and even GL, would vastly expand the reach-ability of Vulkano and remove the blockers potential users may see in this library.

We can't replace the existing vk-sys types directly, but we could easily get away with changing the generation parameters if it was done automatically with `vk_generator`. Also see the issue on the other side - https://github.com/Osspial/vk-rs/issues/10

tomaka:
This is blocked by https://github.com/Osspial/vk-rs/issues/8 and https://github.com/Osspial/vk-rs/issues/11

kvark:
@tomaka all those vk-rs blockers are resolved now 🎉

Osspial:
I just published [vk_generator 0.3](https://crates.io/crates/vk_generator), which should resolve any issues blocking vulkano using it to autogenerate the functions.

Osspial:
I'm in the process of writing a PR to bring `vk-sys` over to using `vk_generator`, and there are a couple of sticking points:

* For bitflag types, the generator currently renames types ending with `FlagBits` to `Flags`, while the spec has both. This was intentional in an attempt to force consistency, but in retrospect it wasn't a good idea - the spec is weirdly inconsistent about which version it uses in types that have both variants, but it's probably best to keep both.
* `vk_generator` automatically wraps Vulkan handles in structs to add some simple type-checking, but vulkano's `VulkanHandle` trait is implemented on raw numeric types. I'm personally for manually implementing `VulkanHandle` on these types, but adding a flag to disable the wrapping shouldn't be hard if it's necessary.
* vk-sys has separate structs for holding different categories of function pointers (static functions, entrypoint functions, etc.). There's no good way to replicate this behavior in vk_generator, because
[...]

PR #952, opened 2018-04-26
Allow applications to access the loaded function pointers

branan:
Without this, an application has to create their own loader for the Vulkan function pointers if they need to drop down to raw Vulkan calls. It's not a lot of effort to load them, but it's nicer to just have them available on the Instance or Device rather than having to pass an extra parameter around for function pointer access.


branan:
This will probably interact with #89 in weird ways sometime in the future, but for now I'm just trying to toss up PRs for any patches I'm carrying in my private fork.

knappador:
I think we can merge this, and I support it since reducing the Vulkan API is pretty difficult without losing some power.  #89 shouldn't cause major issues because if the pointers no longer exist, they will get picked up by the compiler and replaced with the functions to get them.   @branan care to update and then ping Rukai for a merge?

PR #953, opened 2018-04-26
Update winit to 0.13.1

branan:
Winit 0.13.0 has some fixes for window management on X11 that avoid some messy behavior with Vulkan surfaces (see tomaka/winit#434).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Use dynamically loaded `libvulkan` like on other platforms instead of linking to MoltenVK on macOS
+- Updated winit to version 0.13.
 
 # Version 0.9.0 (2018-03-13)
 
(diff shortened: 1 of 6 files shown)


branan:
Updated this to point to Winit 0.13.1 to pull in the x11-dl dependency fix

francesca64:
...and now we're up to winit 0.15.0. Once again, no other changes needed.

DancingBard:
Are there really no more changes needed to update to winit 0.15.0? I tried locally and got `error[E0599]: no method named `build_vk_surface` found for type `winit::WindowBuilder` in the current scope` then... and I am wondering why atm. (Only change from winit ^0.11.0 right to ^0.15.0)

rukai:
You will also need to change the version in the cargo.toml for vulkano-win.

tomaka:
Thanks. This PR is no longer that relevant but it can't hurt to merge it.

Issue #923, opened 2018-02-21
OutOfHostMemory with Linux/Nvidia


AustinJ235:
So recently switched to using my nvidia card again. Running Void Linux with the 390 driver. Issue exists with the 384 driver also. Haven't had time to test on windows if this exists there yet, but using ``PresentMode::Fifo`` or ``PresentMode::Relaxed`` causes a panic on a unwrap of ``OutOfHostMemory``. ``PresentMode::Immediate`` works alright though. Using my application I can successfully match the error from ``aquire_next_image()`` and that returns ``OomError(OutOfHostMemory)``.

tomaka:
`OutOfHostMemory` is very weird! I have no idea when that is even supposed to happen.

AustinJ235:
Ok figured out how to monitor gpu ram usage without x. This issue causes to Xorg to lag out for some reason making it hard to read in the nvidia-settings.

I am writing this down as I observe...

The issue does not happen immediately. With no delay added in the loop it takes ~566 frames to spit out the error. Adding a 5ms delay to the rendering loop does increase the amount of frames it outputs before it returns and error, typically about 900-1200 frames instead. It also allows me to monitor gpu/cpu ram usage a little better as it happens to fast to get any valid gpu/cpu memory readings.

During runtime gpu memory remains the same while cpu memory is increasing roughly 15 MiB/s.

I am now printing out frame times. With 5ms delay in the loop I am getting frame times of ~10ms with ``Fifo`` and ~11ms with ``Relaxed``. Since my display is running at 60hz I'd expect frametimes of ~16.67ms.

I not very technical in terms of vulkan, so I may need some help with this info. But taking a guess, I think frames are being kept on the cpu because the graphics card isn't taking them fast enough.

I added a manual fps throttle to keep the fps at 60. It now runs as you'd expect. I think this confirms my guess. Could be wrong though.

Hope this helps :)

Kurble:
In the examples the cpu side never waits on the futures returned by `signal_fence_and_flush()` etc. So I guess this makes sense. The examples should perhaps contain some way that the cpu will wait for the pipeline to finish after 2 frames or whatever have been queued?

AustinJ235:
Call wait on the future after ``signal_fence_and_flush()`` does solve the issue.

AustinJ235:
Closing in favor of https://github.com/vulkano-rs/vulkano/issues/1247

Issue #627, opened 2017-07-08
Triangle example triggers OutOfHostMemory with gtx 1080


brandonson:
I'm running on an old box with an upgraded GPU. Some details:

Arch Linux
RAM: 12GB
CPU: i7-4770
GPU: GTX 1080

With 8-9GB of memory free, the triangle example runs rampant and triggers an OutOfHostMemory error within a few seconds (which seems fast, but I don't know enough to be sure).

Waiting for the fence at line 380 appears to fix the issue, presumably by letting things finish and be cleaned up at the top of the loop.

Not sure what the right fix (if any) is for this, but it certainly seems like a bug.

tomaka:
From my experience these out of memory errors are almost always a problem of some sort with the driver, but obviously I don't know if that's the case here.

brandonson:
I've got a backtrace now, if it helps:

#0  core::result::unwrap_failed<vulkano::OomError> (msg=..., error=vulkano::OomError::OutOfHostMemory) at /checkout/src/libcore/result.rs:860
#1  0x00005555555ce13e in core::result::Result<vulkano::sync::fence::Fence<alloc::arc::Arc<vulkano::device::Device>>, vulkano::OomError>::unwrap<vulkano::sync::fence::Fence<alloc::arc::Arc<vulkano::device::Device>>,vulkano::OomError> (self=...) at /checkout/src/libcore/result.rs:738
#2  0x00005555556842a9 in vulkano::sync::future::fence_signal::then_signal_fence<vulkano::swapchain::swapchain::PresentFuture<vulkano::command_buffer::traits::CommandBufferExecFuture<vulkano::sync::future::join::JoinFuture<alloc::boxed::Box<GpuFuture>, vulkano::swapchain::swapchain::SwapchainAcquireFuture>, vulkano::command_buffer::auto::AutoCommandBuffer<vulkano::command_buffer::pool::standard::StandardCommandPoolAlloc>>>> (future=..., behavior=...) at /home/brandon/.cargo/registry/src/github.com-1ecc6299db9ec823/vulkano-0.5.2/src/sync/future/fence_signal.rs:40
[...]

PR #955, opened 2018-05-04
Fix examples on nvidia linux

rukai:
The examples currently crash almost immediately on nvidia linux.
This is due to an OutOfHostMemory error as described in this issue: https://github.com/vulkano-rs/vulkano/issues/923, https://github.com/vulkano-rs/vulkano/issues/627
This PR changes the examples to wait on the previous frame before submitting the next frame.

I think this is the best solution but other solutions I thought of are:
*   call wait method immediately after submitting frame
    -   Only need the `Box<GpuFuture>` variable (dont need the `Option<Arc<vulkano::sync::FenceSignalFuture<_>>>`)
    -   Wastes time blocking
*   Add a wait method to GpuFuture trait that fails silently/panics/returns error enum if called on any future other than FenceSignalFuture
    -   Only need the `Box<GpuFuture>` variable
    -   Doesnt waste time blocking
    -   More potential runtime errors

### DIFF
--- a/examples/src/bin/image.rs
+++ b/examples/src/bin/image.rs
@@ -24,6 +24,7 @@ use vulkano_win::VkSurfaceBuild;
 use vulkano::sync::GpuFuture;
 
 use std::sync::Arc;
+use std::time::Duration;
 
 fn main() {
     // The start of this example is exactly the same as `triangle`. You should read the
@@ -143,10 +144,11 @@ fn main() {
 
     let mut recreate_swapchain = false;
 
-    let mut previous_frame_end = Box::new(tex_future) as Box<GpuFuture>;
+    let mut previous_frame_end_gpu = Box::new(tex_future) as Box<GpuFuture>;
+    let mut previous_frame_end_fence: Option<Arc<vulkano::sync::FenceSignalFuture<_>>> = None;
 
     loop {
-        previous_frame_end.cleanup_finished();
+        previous_frame_end_gpu.cleanup_finished();
         if recreate_swapchain {
 
             dimensions = surface.capabilities(physical)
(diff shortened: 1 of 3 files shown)


rukai:
As mentioned in https://github.com/vulkano-rs/vulkano/issues/627 this issue is caused by this unresolved bug in nvidia drivers (apparently windows as well as linux) https://vulkan.lunarg.com/app/issues/56ca3a477ef24d0001787448
Because its a driver bug, this PR should not be merged. However it should be kept open until the issue is resolved as it documents a work around.

jonathansty:
@rukai  just as a heads up that driver bug seems to be closed. The validation message definitely sounds like it's not a bug but just wrong usage in our case?
```
22:44:41 [ERROR] "Fence 0x12 is in use. The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkResetFences-pFences-01123)"
```

rukai:
oh, good to know, we'd better figure out whats going wrong then.

jonathansty:
I don't have enough knowledge on how vulkano handles the GPU futures. 

However I did give a try to using some raw functions and using fences instead of the `GpuFuture`s to wait. In my application [this commit](https://github.com/jonathansty/Rust-Experiments/commit/c4a8f007f1cdb6a4018601acf6ce722a4253aea6) fixed the issue. 

Although probably not the most optimal fix as I'm waiting before doing the submit and present and then also immediately waiting after.

knappador:
Motion to close; expiration date seems lapsed (the examples run on Nvidia hardware)

I do suspect we are missing a lot of behavior on synchronization given the validation, but mainly I'm just having difficulty getting Intel chips to run like Nvidia ones right now.

AustinJ235:
I know at least the image example still has the issue haven't tested the others. It is probably a good idea to resolve this pull request or open another one to allow the examples to run on nvidia cards.

AustinJ235:
closed in favor of https://github.com/vulkano-rs/vulkano/pull/1246

PR #958, opened 2018-05-24
Add overview of generated macro code to vulkano-shader-derive

Cldfire:
Right now using the results of the derive macro is not very straightforward and requires digging through code to understand. I've written up a small overview of what it generates to try and make it a bit easier.

This is what it looks like in rustdoc:
<details>
  <summary>Rustdoc screenshot</summary>

</details>


tomaka:
Thanks for this!

Issue #963, opened 2018-05-28
DescriptorDesc should contains the uniform name


realitix:
Hello,
First of all, thanks for this library.

When I use `vulkano-shaders`, I need to retrieve a set binding based on the uniform name in the shader.
Currently it's not possible, we can only retrieve name for input and output variables.
Would you be ok to add the name in the `DescriptorDesc` struct that we could fill in `vulkano-shaders` lib ?

Thanks,
Jean-Sébastien

realitix:
I created a pull request about it

realitix:
I close it after an answer on the pull request

PR #964, opened 2018-05-28
Add name in DescriptorDesc

realitix:
This commit add the uniform name in `DescriptoDesc`.
Using the `vulkano-shaders::reflect` function, it allows to increase
code introspection on shaders.

Fix the #963 issue

### DIFF
--- a/vulkano-shaders/src/descriptor_sets.rs
+++ b/vulkano-shaders/src/descriptor_sets.rs
@@ -23,6 +23,7 @@ pub fn write_descriptor_sets(doc: &parse::Spirv) -> String {
         desc_ty: String,
         array_count: u64,
         readonly: bool,
+        name: String
     }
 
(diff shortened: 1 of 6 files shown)


realitix:
I'm fixing it. I didn't see it on my build because my object file weren't cleaned.

realitix:
Any news on my pull request ?

tomaka:
Thanks for the PR.
`DescriptorDesc` is supposed to contain the information that we need to pass to Vulkan, so `name` doesn't really fit in there.

realitix:
Indeed @tomaka.
Finally I need to parse the spirv at runtime. It would be great if vulkano-shader could be used for runtime parsing and compile time parsing.
I pasted you module and adapted it to  my use case, it's an amazing work!

I close the pull request.

Issue #758, opened 2017-08-22
empty CpuBufferPool chunk (regression)


rukai:
I missed an issue with CpuBufferPoolChunk that occurs only when the first chunk is NOT empty, which I didnt test. (sorry)
This occurs on my windows install on the latest commit, and doesnt occur on the previous commit.

It panics with both:
```
thread 'main' panicked at 'assertion failed: chunk.num_gpu_accesses >= 1', /home/rubic/Projects/RustExamples/vulkano/vulkano/src/buffer/cpu_pool.rs:567:8
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', /checkout/src/libcore/result.rs:906:4
```
Branch with example demonstrating the issue:https://github.com/rukai/vulkano/commit/f355ea2d3d88c1fadba40ecae0ea7a7159c2b360
Full output log: [out.txt](https://github.com/vulkano-rs/vulkano/files/1242174/out.txt)

rukai:
@tomaka What was your reasoning for allowing empty buffers here? https://github.com/vulkano-rs/vulkano/pull/524

Issue #907, opened 2018-01-06
assertion failed: chunk.num_gpu_accesses >= 1 when drawing large number of meshes


franckv:
I'm drawing a scene with a moderatly large number of meshes (~500).
Each mesh has its own transformation matrix that i need to upload in a uniform.
To do this, I instanciate a new FixedSizedDescriptor for each mesh and attach a CpuBufferPool containing the matrix data as follows:
 
```
         let command_buffer = {
             let mut builder =
                 AutoCommandBufferBuilder::primary_one_time_submit(device.clone(), queue.family()).unwrap()
                 .begin_render_pass(framebuffers[idx].clone(), false, vec![[0., 0., 1., 1.].into()]).unwrap();
 
             let mut descriptor_sets_pool = FixedSizeDescriptorSetsPool::new(pipeline.clone(), 0);
 
             for mesh in &meshes {
                 let uniforms_data = vs::ty::Data {
                     transform: mesh.transform.into(),
                 };
 
                 let uniforms = uniforms_buffer.next(uniforms_data).unwrap();
 
                 let set = Arc::new(descriptor_sets_pool.next().add_buffer(uniforms).unwrap().build().unwrap());
 
                 builder = builder.draw_indexed(pipeline.clone(), dynamic_state.clone(),
                 mesh.vbuf.clone(), mesh.ibuf.clone(), set.clone(), ()).unwrap();
             }
 
             builder.end_render_pass().unwrap().build().unwrap()
         };

         let future = last_frame.join(acquire_future)
             .then_execute(queue.clone(), command_buffer).unwrap()
             .then_swapchain_present(queue.clone(), swapchain.clone(), idx)
             .then_signal_fence_and_flush().unwrap();
```

This code double panicks on my setup (Linux / Intel(R) HD Graphics 620 (Kaby Lake GT2)) if the number of meshes is greater than 400 (it works fine for lower values):
 
```
thread 'test::tests::run' panicked at 'assertion failed: chunk.num_gpu_accesses >= 1', /home/franck/.cargo/registry/src/github.com-1ecc6299db9ec823/vulkano-0.7.2/src/buffer/cpu_pool.rs:659:8
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread 'test::tests::run' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', src/libcore/result.rs:906:4
stack backtrace:
[...]

PR #970, opened 2018-06-07
Remove incorrect assertions

rukai:
Fixes: https://github.com/vulkano-rs/vulkano/issues/758 and https://github.com/vulkano-rs/vulkano/issues/907
I could be completely misunderstanding this but it looks like these assertions are incorrect?
Given the name of the function `increase_gpu_lock` and the lock state is a usize, it looks like this is a non exclusive lock, so this should be fine.

### DIFF
--- a/vulkano/src/buffer/cpu_access.rs
+++ b/vulkano/src/buffer/cpu_access.rs
@@ -375,7 +375,6 @@ unsafe impl<T: ?Sized, A> BufferAccess for CpuAccessibleBuffer<T, A>
             let read_lock = self.access.read().unwrap();
             if let CurrentGpuAccess::NonExclusive { ref num } = *read_lock {
                 let prev = num.fetch_add(1, Ordering::SeqCst);
-                debug_assert!(prev >= 1);
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


tomaka:
I think the assertions are correct.
See this comment: https://github.com/rukai/vulkano/blob/f10175e2127317adb2aa4871df0fb2605501e0c3/vulkano/src/buffer/cpu_access.rs#L90

If the number of accesses was 0, then we should have jumped to `Exclusive`. If we didn't, then there's a bug in vulkano.

tomaka:
But thanks for checking!

rukai:
Thanks for the feedback! I'll keep investigating.

rukai:
I think you might be getting Exclusive and NonExclusive mixed up? The comment says that "Number of non-exclusive GPU accesses. Can be 0." Also I dont see anything like this in cpu_pool.

rukai:
I mentioned this issue on irc and someone said that maybe increase_gpu_lock() is only intended to be called on already-locked CurrentGpuAccesses? Is that the case?

PR #975, opened 2018-06-13
[WIP] StorageImage initialization fix candidate

### LINKED ISSUE #974: Initialization procedure of StorageImage seems incorrect
My Vulkan validation layer told me that there's something wrong with image layouts in [the first example of the "Using images" section of the Vulkano guide](http://vulkano.rs/guide/image-export). It chokes when the command buffer is submitted to the device, even when it contains only a clear command, with the message:

    ERROR @ DS => Object: 0x7f8c5fce6c00 | Cannot submit cmd buffer using image (0x11) [sub-resource: aspectMask 0x1 array layer 0, mip level 0], with layout VK_IMAGE_LAYOUT_UNDEFINED when first use is VK_IMAGE_LAYOUT_GENERAL.

So I went to investigate a bit. Here is what I found:

- A StorageImage is, as far as I can tell, always created in the Undefined layout:
    * [StorageImage::new()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/image/storage.rs#L74) is a thin wrapper around [StorageImage::with_usage()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/image/storage.rs#L102)
    * StorageImage::with_usage() calls [UnsafeImage::new()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/image/sys.rs#L82) with its preinitialized_layout parameter set to false.
    * In this configuration, [UnsafeImage::new() creates images in the Undefined layout](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/image/sys.rs#L485).
    * No layout transition is subsequently performed by the UnsafeImage and StorageImage constructors.
- The [vkClearColorImage()](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#clears-outside) API requires images to be initially in the General or TransferDstOptimal layout, so clearly a layout transition will need to be performed at the beginning of the command buffer that calls this command.
- [AutoCommandBufferBuilder::clear_color_image()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/auto.rs#L747) is a thin wrapper around [AutoCommandBufferBuilder::clear_color_image_dimensions()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/auto.rs#L763).
- After some validation, AutoCommandBuffeBuilder::clear_color_image_dimensions() calls the inner [SyncCommandBufferBuilder::clear_color_image()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/commands.rs#L569), requesting a TransferDstOptimal image layout.
- SyncCommandBufferBuilder::clear_color_image() dispatches, in turn, to [SyncCommandBufferBuilder::append_command()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L426) and [SyncCommandBufferBuilder::prev_cmd_resource()](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L464), following the usual pattern of this builder. The first method stores the "raw" command object, the second method is of higher interest to us here as it handles resource-related pipeline barriers, including image layout transitions.
- SyncCommandBufferBuilder::prev_cmd_resource(), in turn, [does have a path for handling the case where images are initially in the wrong layout](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L622). We're getting close!
- On that code path, [it is assumed that img.initial_layout_requirement() tells the truth about the initial state of the image](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L652)...
- But that is wrong in our case, since [StorageImage always reports its initial layout to be General](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/image/storage.rs#L191), even when it is still in the Undefined state! Therefore, the initial pipeline barrier starts from the General state instead of starting from the Undefined state, which is why my validation layer is complaining.

At this point, my head feels a bit dizzy from all the layers of abstraction that I've just been through, but I think I have found where the inconsistency is. I can think of at least three ways of fixing this:

- The StorageImage constructor could silently initialize every image. But that is very invasive and unexpected (as it requires silently submitting command buffers to a device queue under the hood), so I do not expect this option to be considered.
- SyncCommandBufferBuilder::prev_cmd_resource() could special-case newly created images and create a different kind of layout transition barrier for them.
- StorageImage::initial_layout_requirement() could return something different when an image has just been created.

What should it be?

HadrienG2:
Here is a quick and dirty port of the ImmutableImage initialization mechanism to StorageImage, in an attempt to resolve #974 .

Please tell me if you like the general direction. If so, we can think about ways to clean this up further (e.g. sharing more code between StorageImageInitialization and ImmutableImageInitialization).

### DIFF
--- a/vulkano/src/image/storage.rs
+++ b/vulkano/src/image/storage.rs
@@ -9,6 +9,7 @@
 
 use smallvec::SmallVec;
 use std::sync::Arc;
+use std::sync::atomic::AtomicBool;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 
@@ -61,6 +62,9 @@ pub struct StorageImage<F, A = Arc<StdMemoryPool>>
     // Format.
     format: F,
 
+    // Image has been initialized
+    initialized: AtomicBool,
+
     // Queue families allowed to access this image.
     queue_families: SmallVec<[u32; 4]>,
 
@@ -68,8 +72,17 @@ pub struct StorageImage<F, A = Arc<StdMemoryPool>>
     gpu_lock: AtomicUsize,
 }
 
+// Must not implement Clone, as that would lead to multiple `used` values.
+pub struct StorageImageInitialization<F, A = Arc<StdMemoryPool>>
+    where A: MemoryPool
+{
+    image: Arc<StorageImage<F, A>>,
+    used: AtomicBool,
+}
+
 impl<F> StorageImage<F> {
     /// Creates a new image with the given dimensions and format.
+    #[deprecated(note = "use StorageImage::uninitialized instead")]
     #[inline]
     pub fn new<'a, I>(device: Arc<Device>, dimensions: Dimensions, format: F, queue_families: I)
                       -> Result<Arc<StorageImage<F>>, ImageCreationError>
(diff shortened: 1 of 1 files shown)


AustinJ235:
This is no longer needed as of https://github.com/vulkano-rs/vulkano/pull/1171

PR #978, opened 2018-06-21
Add pipeline barrier documentation to rust generated documentation

realitix:
Hello,
You have documented the pipeline barrier mechanism but it's not in the doc.


tomaka:
Hey,

This comment was indeed not intended to be in the docs. The struct in question handles the described mechanism automatically, and that block is intended to be read by the people who tweak it or want to read its code.

rukai:
As explained by tomaka, users of Vulkano don't need to understand barriers as they are abstracted over.

PR #980, opened 2018-06-25
Remove incorrect assertions

rukai:
I'm reopening this as the comment you linked to was for CurrentGpuAccess::Exclusive and my change is for the handling of CurrentGpuAccess::NonExclusive.
I mentioned this on the old PR but I dont know if you get notifications for closed PR's...
Link to the old one: https://github.com/vulkano-rs/vulkano/pull/970


rukai:
Thanks for explaining, that makes sense. I'll continue looking into it.

Issue #981, opened 2018-06-27
BufferAccess len() warnings/future errors


AustinJ235:
Running the latest nightly I got some warnings about ``BufferAccess``

```console
warning: the trait `buffer::traits::BufferAccess` cannot be made into an object
  --> /home/austin/workspace/engine/vulkano/vulkano/src/buffer/traits.rs:37:5
   |
37 | /     fn len(&self) -> usize
38 | |         where Self: TypedBufferAccess,
39 | |               Self::Content: Content
40 | |     {
41 | |         self.size() / <Self::Content as Content>::indiv_size()
42 | |     }
   | |_____^
   |
   = note: #[warn(where_clauses_object_safety)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #51443 <https://github.com/rust-lang/rust/issues/51443>
   = note: method `len` references the `Self` type in where clauses

warning: the trait `buffer::traits::BufferAccess` cannot be made into an object
  --> /home/austin/workspace/engine/vulkano/vulkano/src/buffer/traits.rs:37:5
   |
37 | /     fn len(&self) -> usize
38 | |         where Self: TypedBufferAccess,
39 | |               Self::Content: Content
40 | |     {
41 | |         self.size() / <Self::Content as Content>::indiv_size()
42 | |     }
   | |_____^
   |
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #51443 <https://github.com/rust-lang/rust/issues/51443>
   = note: method `len` references the `Self` type in where clauses
```

rukai:
This was fixed in: https://github.com/vulkano-rs/vulkano/pull/966
We will need to release vulkano 0.10.0 to fix the warning for crates.io users.
I'll try and submit a PR to upgrade to the latest winit before hand though. Give me a couple of days.

AustinJ235:
Oops just assumed on I was on the latest version, just cloned the my repo a few days ago. My vulkano submodule must be out of date.

rukai:
I would reopon this. Although it doesn't affect me or you, it's still a valid issue, as eventually vulkano will stop working on crates.io unless a new version is released.

PR #982, opened 2018-06-27
Upgrade deps + fix warnings in examples

rukai:
I upgraded all dependencies to their latest versions.
When I tested each example I noticed there were multiple warnings so I fixed those as well.

Somewhat unrelated:
We should really do a vulkano 0.10.0 release to prevent the new object safety warnings from occurring for people who use crates.io
e.g. https://github.com/vulkano-rs/vulkano/issues/981
I propose doing such a release after reviewing + merging the current waiting PR's (including this one ^.^)


tomaka:
> I propose doing such a release after reviewing + merging the current waiting PR's (including this one ^.^)

:+1:

Issue #984, opened 2018-07-02
Linux/Nvidia CpuBufferPool::chunk() runs out of memory


pypypypypypypypypyp:
buffer pool is declared like so: 
```rust
let vbuf_builder = CpuBufferPool::vertex_buffer(device.clone());
```
and then chunks are created from Vec<Vertex>s and stored in a hashmap: 
```rust
while limiter.frac_time_left().map(|t| t > 0.1).unwrap_or(false) {
    if let Ok((id, verts)) = new_vbufs.try_recv() {
        println!("adding vbuf with id: {:?} and {} total verts",id,verts.len());
        vbufs.insert(id, Arc::new(vbuf_builder.chunk(verts).expect("failed to create a vertex buffer")));
        use vulkano::buffer::BufferAccess;
        println!("size of the new vbuf is {} bytes",vbufs.get(&id).unwrap().size());
        println!("total number of vbufs is now {}",vbufs.len());
    } else {
        break;
    }
}
```
on windows this runs fine, on (arch) linux with a gtx 1050 Ti it panics when attempting to create the fourth buffer:
```
adding vbuf with id: Vec3 { x: 0, y: 0, z: 0 } and 12672 total verts
size of the new vbuf is 304128 bytes
total number of vbufs is now 1
adding vbuf with id: Vec3 { x: -1, y: 0, z: 0 } and 12672 total verts
size of the new vbuf is 304128 bytes
total number of vbufs is now 2
adding vbuf with id: Vec3 { x: -1, y: -1, z: 0 } and 12672 total verts
size of the new vbuf is 304128 bytes
total number of vbufs is now 3
adding vbuf with id: Vec3 { x: 0, y: -1, z: 0 } and 12672 total verts
thread '<unnamed>' panicked at 'failed to create a vertex buffer: OomError(OutOfDeviceMemory)',
```
-slowing down the loop and watching gpu memory usage shows the program never uses more than 30 MiB and total gpu memory usage never goes above 10%.
-removing the expect and printing the error causes the OomError to be returned every call after the first three.
-clearing the hashmap when an error occurs (and hence, freeing up the chunks) allows three more chunks to be allocated before OomError is returned again.
-doubling number of vertices in each buffer does not affect it, drastically increasing number of vertices still allows 3 chunks to be allocated at one time and no more.
-halving number of vertices allows 4, drastically reducing vertices in each chunk allows a lot more chunks to be allocated.
[...]

PR #986, opened 2018-07-03
Fix CpuBufferPool allocation strategy

newpavlov:
The previous strategy was quite sub-optimal and was failing hard with OOM when `chunk` was used for allocating big buffers, e.g. textures, due to the multiplication of `data.len()` with current capacity.

Fixes: #984

### DIFF
--- a/vulkano/src/buffer/cpu_pool.rs
+++ b/vulkano/src/buffer/cpu_pool.rs
@@ -315,12 +315,10 @@ impl<T, A> CpuBufferPool<T, A>
             Err(d) => d,
         };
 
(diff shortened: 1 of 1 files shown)


newpavlov:
It could be a good idea to provide `set_capacity(new_capacity: usize)` and `try_chunk` methods for more control over allocation process, the latter will return an error if there is not enough capacity in the underlying buffer.

AllenChong:
pull request

tomaka:
Oh thanks! Multiplying `data.len()` with the capacity is a reliquate from an old version of the buffer pool and definitely not intended to still be there.

PR #988, opened 2018-07-04
Add SingleInstanceBufferDefinition

andrewhickman:
This is something I needed when translating some code which used [`gfx::InstanceBuffer`](https://docs.rs/gfx/0.17.1/gfx/pso/buffer/type.InstanceBuffer.html). The code is pretty much copied from the other `VertexDefinition` implementations, I am not really familiar with the specific safety concerns involved.


vulkano/src/pipeline/vertex/instance.rs
```diff
@@ -0,0 +1,122 @@
...
+
+unsafe impl<'a, B, V> VertexSource<B> for SingleInstanceBufferDefinition<V>
+where
+    B: TypedBufferAccess<Content = [V]> + Send + Sync + 'static,
+    V: Vertex,
+{
+    #[inline]
+    fn decode(&self, source: B) -> (Vec<Box<BufferAccess + Send + Sync>>, usize, usize) {
+        let len = source.len();
+        (vec![Box::new(source) as Box<_>], len, 1)
```
tomaka: This should be `(vec![Box::new(source) as Box<_>], 1, len)`.

vulkano/src/pipeline/vertex/instance.rs
```diff
@@ -0,0 +1,122 @@
...
+{
+    #[inline]
+    fn decode(
+        &self,
+        mut source: Vec<Arc<BufferAccess + Send + Sync>>,
+    ) -> (Vec<Box<BufferAccess + Send + Sync>>, usize, usize) {
+        // FIXME: safety
+        assert_eq!(source.len(), 1);
+        let len = source[0].size() / mem::size_of::<V>();
+        (vec![Box::new(source.remove(0))], len, 1)
```
tomaka: This should be `(vec![Box::new(source.remove(0))], 1, len)`

tomaka:
Thanks for the PR!
Could you also please rename the newly-created `instance.rs` to `instance_buffer.rs` for example? The word `instance` is a bit too vague in the context of vertices and attributes.

PR #991, opened 2018-07-10
Don't force the surface to take ownership of the window

### LINKED ISSUE #920: vulkano-win shouldn't take ownership of my window
For reasons largely tangential to vulkano, I have a desire to hoist the winit::Window into the main function of my application. Unfortunately, vulkano-win likes to both create and assume ownership of the window, even though it seems to have no need of it after constructing the Surface.

Would you be opposed to moving to a model whereby the Surface constructor takes the window by reference, and doesn't store it at all?

andrewhickman:
This PR adds a new method `create_vk_surface` which allows creating a surface from anything outlived by a window. This is enforced by the `SafeBorrow` trait which is essentially the equivalent of `SafeDeref` for `Borrow`.

Closes #920


tomaka:
Thanks! Could you add an emtry to the CHANGELOG as well? Then I'll merge this.

PR #997, opened 2018-07-22
Docs: Fix swapchain image count computation.

jimblandy:
The previous expression gives unreasonable answers for some combinations of
`swapchain::Capabilities`' `min_image_count` and `max_image_count` values. For
example, for `3` and `None`, the expression evaluates to 2, which isn't a
permitted image count.

I looked for a terser expression (using `Option` methods, say), but the `match`
expression ended up being the most legible I came up with.


tomaka:
Thanks

PR #998, opened 2018-07-25
Fix the StructureType Ids for iOS and MacOS

JustinRyanH:
I was running the Lunar validation layer on my local macbook when I noticed errors in the console.

``` bash
VUID-VkMacOSSurfaceCreateInfoMVK-sType-sType(ERROR / SPEC): 
msgNum: 201502731 - vkCreateMacOSSurfaceMVK: 
parameter pCreateInfo->sType must be VK_STRUCTURE_TYPE_MACOS_SURFACE_CREATE_INFO_MVK. 

The spec valid usage text states 'sType must be VK_STRUCTURE_TYPE_MACOS_SURFACE_CREATE_INFO_MVK' 
(https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VUID-VkMacOSSurfaceCreateInfoMVK-sType-sType)
```

I check the local version, and found it was set to `1000000000 + (53 * 1000)`.

Specs said otherwise

https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkStructureType

https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VkStructureType


tomaka:
Oh it was probably changed when MoltenVK was integrated in the official stuff.
Thanks.

PR #1003, opened 2018-07-29
Decouple descriptor sets from pipeline

### LINKED ISSUE #1002: Decouple descriptor sets from pipeline layout
Right now, creating a `PersistentDescriptorSet` or `FixedSizeDescriptorSetsPool` requires specifying a pipeline layout and the set id within the pipeline/shader that the set binds to. However, there is nothing within Vulkan itself that mandates this; descriptor sets can be used pipeline-agnostically as long as the descriptor layout matches up with the pipeline when drawing. Requiring descriptor sets to be tied to a specific pipeline layout, with the set id already fixed, removes some of the flexibility that standard Vulkan does offer.

I therefore propose that these types no longer store a reference to a pipeline layout, but instead either create their own `UnsafeDescriptorSetLayout` object or accept one as a parameter. I'd be willing to try to implement this myself if desired.

Rua:
Resolves #1002.


Rua:
I've redone some of the code so that there is no more merge conflict. Apologies for the many commits, I'm still pretty new to how Git(Hub) works.

rukai:
Thanks for the rebase, your clearly still interested in getting this merged. I'll have to review it sometime soon.

knappador:
I started getting up to speed on this PR to get knowledgeable enough to review it.

One thought I had was that if the most likely source of a layout is the pipeline, and the pipeline is always guaranteed to have a layout, it might make sense to implement the `From` trait.  The example code is telling us that this type shortcut would be useful.

I took a quick pass and it wound up like:
```rust
impl<Mv, L, Rp> From<Arc<GraphicsPipeline<Mv, L, Rp>>> for Arc<PipelineLayoutDesc>
where
    L: PipelineLayoutAbstract,
    Rp: RenderPassAbstract,
    Mv: VertexSource<Vec<Arc<dyn BufferAccess + Send + Sync>>>,
{
    fn from(item: Arc<GraphicsPipeline<Mv, L, Rp>>) -> Arc<PipelineLayoutDesc> {
        item.layout().descriptor_set_layout(0).unwrap().clone()
    }
}
```

However, I'm still learning the type relationships in Vulkano, not to mention Rust, and I couldn't get ahead of the compiler.

Rua:
Thank you for having a look at the PR! The semantics of the `From` trait is that it's used specifically to convert one thing to another. For that reason, it normally accepts an owned value as its argument. Here, you are giving it an `Arc` which bypasses the ownership, which is fine to the compiler. However, it doesn't really agree with how the `From` trait is normally used, as you're not really converting a pipeline into a layout but rather getting a reference to the layout that's inside the pipeline. So I'm not sure if this is a good use case for it.

faulesocke:
@knappador You can't implement the `From` trait for external types (i.e. types that are outside of your current module) like `Arc`. That's probably why you couldn't get that working.

@Rua If you want to get rid of some of the commits just squash them together using `git rebase -i` (interactive rebase). However, I wouldn't recommend merging everything into just one single commit. For example you could have one commit for the implementation and one that fixes all the examples. I think 0214f36 and 8801733 have to go and you will get that with a rebase onto the current upstream master. Afterwards these commits should end up (nearly) empty so that you can drop them in a interactive rebase.

However since you are a git beginner it's probably easiest for you to just squash everything together into a single large commit and then rebasing onto the current upstream master. After that review your commit carefully using git show and git diff and then force push. This will also update this pull request.

Rua:
@rukai I updated the PR to make it merge and build again. Are you still interested in reviewing it?

AustinJ235:
Everything links good, but you appear to missing a changelog entry. Add that I think we are good to go with a merge.

simdimdim:
I'm getting some 

    Updating crates.io index
    Updating git repository `https://github.com/vulkano-rs/vulkano`
error: failed to load source for a dependency on `vk-sys`

Caused by:
  Unable to update https://github.com/vulkano-rs/vulkano

Caused by:
  failed to update submodule `glsl-to-spirv/glslang`

Caused by:
  no URL configured for submodule 'glsl-to-spirv/glslang'; class=Submodule (17)

Ideas?

PR #1005, opened 2018-07-30
Avoid gratuitous use of std::mem::replace.

jimblandy:
Either these calls to `replace` are unnecessary, or I'm going to learn something
I really need to know.

The only way difference I can see between `replace` and a simple assignment is
that `replace` returns ownership of the value to the caller, so the old value is
dropped after the new value has been put in place. But if Rust lets us assign to
or move from a variable, that means that no other alias can observe that
happening --- which I think means that the drop can't possibly care whether it
occurs before or after the move.

### DIFF
--- a/examples/src/bin/teapot.rs
+++ b/examples/src/bin/teapot.rs
@@ -155,8 +155,8 @@ fn main() {
                 Err(err) => panic!("{:?}", err)
             };
 
-            std::mem::replace(&mut swapchain, new_swapchain);
-            std::mem::replace(&mut images, new_images);
+            swapchain = new_swapchain;
+            images = new_images;
 
             let new_depth_buffer = vulkano::image::attachment::AttachmentImage::transient(device.clone(), dimensions, vulkano::format::D16Unorm).unwrap();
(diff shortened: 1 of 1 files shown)


rukai:
I don't use `mem::replace` in my vulkano code. https://github.com/rukai/PF_Sandbox/tree/master/pf_sandbox/src/vulkan
I think its fine to remove them.
Please remove these in your PR as well:

```
triangle.rs
357:            mem::replace(&mut swapchain, new_swapchain);
358:            mem::replace(&mut images, new_images);
373:            mem::replace(&mut framebuffers, new_framebuffers);

image.rs
164:            std::mem::replace(&mut swapchain, new_swapchain);
165:            std::mem::replace(&mut images, new_images);
178:            std::mem::replace(&mut framebuffers, new_framebuffers);

teapot.rs
158:            std::mem::replace(&mut swapchain, new_swapchain);
159:            std::mem::replace(&mut images, new_images);
162:            std::mem::replace(&mut depth_buffer, new_depth_buffer);
178:            std::mem::replace(&mut framebuffers, new_framebuffers);
```

Rua:
I noticed this too and never saw the point of it.

PR #1008, opened 2018-08-01
Avoid allocating a DynamicState every frame

andrewhickman:
This PR addresses the 'TODO' in the examples regarding creating a DynamicState (including a Vec) every frame. By taking `&mut DynamicState`, the `Vec` only needs to be recreated when the state needs to change.


andrewhickman:
After looking at the command buffer code a bit more I've changed it take a shared reference.

tomaka:
Thanks!

PR #1015, opened 2018-08-11
Export features and device extensions from the device module instead of the instance module

rukai:
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 
 # Version 0.10.0 (2018-08-10)
 
+- Export features and device extensions from the device module instead of the instance module
+    +   `instance::Features` -> `device::Features`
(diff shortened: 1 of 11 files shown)


rukai:
@tomaka Can I get your opinion on this?

DeviceExtensions was (accidentally?) exported in both device:: and instance:: but I couldn't tell if it was meant to be in device:: or instance::

I opted to resolve by:
*   `instance::Features` -> `device::Features`
*    exports at `instance::RawDeviceExtensions` and `instance::DeviceExtensions` are removed.
*    exports at `instance::RawInstanceExtensions` and `instance::InstanceExtensions` are kept.

I made this decision because from the users perspective, features and device extensions both serve the same purpose of configuring the device and so should be kept together.

PR #1023, opened 2018-08-23
Copy examples from example repo

rukai:
This PR copies the examples from the vulkano-examples repo.
It is the first step of a plan to resolve some issues with the current way examples are setup in vulkano.

As such i'll leave this PR open for a week so people have a chance to give feedback.
The plan is below:

## Current examples setup
*   Some examples are in the vulkano repo, some are in the vulkano-examples repo, some are in both. Initially they were meant to be all moved to vulkano-examples but then we would have no examples tracking the latest git, so thats not good.
*   The examples in the vulkano repo target latest git while vulkano-examples targets some crates.io release.
*   Vulkano-www also has some of its own examples.

## The issues
*   Initially all examples were meant to be moved to vulkano-examples but then we would have no examples tracking the latest git, so completing the move is not a great option.
*   There is a lot of duplicate work on examples between the repos because people dont realize the other repo exists.
*   Issues that should really be on the vulkano repo, end up on vulkano-examples. There are even duplicate examples across vulkano and vulkano-examples
*   The versions of the examples lag behind a few releases.
*   We also get a lot of issues from people confused that the examples don't work because of changes  in the examples since the latest release.

## Proposed solution
*   Copy all examples to the vulkano repo. CI will ensure these examples continue to compile on latest vulkano git.
*   Every time a new vulkano release is made, copy the examples to the vulkano-examples repo, targeting the new release. No more out of date examples.
*   Issues should be disabled on vulkano-examples. Direct people to file issues on the vulkano repo.
*   PRs can't be disabled but we should direct people to make PRs against the main vulkano repo.
*   Because the examples are built in the vulkano CI, responsiblity of updating examples after making breaking changes is put on the person making the breaking changes.
*   Examples used by vulkano-www should also be moved to vulkano and vulkano-examples, vulkano-www should then link to the file in vulkano-www
*   To direct people to the examples for the latest release a readme is added to the vulkano repo examples directory.

There is also the problem of vulkano-www getting out of date, at the moment I have added a line to the PR template instructing contributors to make a PR if their change will affect it, but we wont be able to merge such PRs until release, and then there will be merge conflicts. So maybe we should create a development branch for vulkano-www so that this works more smoothly.


rukai:
Ok here we go!

Issue #1019, opened 2018-08-17
Examples on mac panic with 'failed to create Vulkan instance: IncompatibleDriver'


CleanCut:
Should the examples in the master branch work on mac?  I've tried vulkan versions 1.1.82.0, 1.1.77.0, and 1.0.69.0, but I don't notice any difference in the error.

Am I missing something?

![image](https://user-images.githubusercontent.com/5838512/44241236-71f71b00-a177-11e8-9685-57c97c7a39cf.png)

rukai:
Have you installed the [vulkan sdk for mac](https://vulkan.lunarg.com/sdk/home) as per the readme?
Sadly I don't have a mac, so you might need to do some digging yourself.

catt-io:
Hi there, I get the same error with Vulkano 0.10.0 on mac. Used to work fine with version 0.9.0 when linking with the MoltenVK framework was default (before #948). I have version 1.1.82.0 of the LunarG Vulkan SDK installed.

rukai:
@nlordell any ideas?

catt-io:
Looks like I was missing a few environment variables, having `libvulkan.dylib` locatable on `DYLD_LIBRARY_PATH` was not enough, also needed `VK_ICD_FILENAMES ` and `VK_ICD_FILENAMES` setup as shown here: https://github.com/bwasty/vulkan-tutorial-rs/blob/master/mac-env.sh (more info at https://github.com/bwasty/vulkan-tutorial-rs#development-environment)

Works great now :)

rukai:
I would love for this sort of thing to be better documented, a PR would be great!

rukai:
Resolved by https://github.com/vulkano-rs/vulkano/pull/1026

PR #1026, opened 2018-08-27
Adding details about installing the Vulkan SDK for macOS

catt-io:
Attempting to make this process a bit clearer in the readme, tripped me up as mentioned in #1019


bwasty:
I can confirm, looks pretty much like my version: https://github.com/bwasty/vulkan-tutorial-rs/blob/master/mac-env.sh.

nicokoch:
Can confirm too :-)

PR #1027, opened 2018-08-27
Workaround moltenvk problems in examples.

### LINKED ISSUE #1022: Examples continue to run long after exit on Mac
The triangle example hangs for a long time (10-20 seconds) after one presses the close button on Mac (Mojave developer beta 7, using MoltenVK via vulkansdk-macos-1.1.82.1).

The Teapot example exposes something much more interesting - it isn't hanging per se, because the teapot continues to animate after the close event is processed. If you add a `println!` as the very last statement in `main`, you can see that the program exits the main loop entirely, but continues to render and animate for 10 seconds or so.

I'm not entirely sure where to start debugging this. It seems as thought the program is only actually displaying 1 frame each time it should display 2, and after the program ends it proceeds to play back all the missing frames that have been queued up. I can't really tell if that's a Vulkano issue, a MoltenVK issue, or an issue in the sample code?

nicokoch:
Fixes #1022

### DIFF
--- a/examples/src/bin/image.rs
+++ b/examples/src/bin/image.rs
@@ -222,6 +222,11 @@ fn main() {
 
         match future {
             Ok(future) => {
+                if cfg!(target_os = "macos") {
+                    // Workaround for moltenvk issue (hang on close)
+                    // FIXME Remove once motenvk is fixed
+                    future.wait(None).expect("waiting on fence failed");
+                }
                 previous_frame_end = Box::new(future) as Box<_>;
             }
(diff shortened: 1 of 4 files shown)


rukai:
Thanks for the PR.
I wont merge this because I don't want hacky work arounds, even in the examples.

But like this PR for a nvidia-linux problem, I'll keep the PR around as a reference for a workaround: https://github.com/vulkano-rs/vulkano/pull/955

AustinJ235:
closed in favor of https://github.com/vulkano-rs/vulkano/pull/1246

PR #1033, opened 2018-08-31
Fix instance_count when using draw_index with instance buffers

franckv:

Currently, AutoCommandBufferBuilder.draw_indexed() hard codes the value of instance_count to 1.
https://github.com/vulkano-rs/vulkano/blob/67ace6b1c5e7506dbf623741a7d940d4a77d2298/vulkano/src/command_buffer/auto.rs#L1076

This is different from the behavior of the draw() method which uses the size of the instance buffer used as a vertex input source.
This PR harmonizes both methods.


rukai:
Thanks!

Issue #298, opened 2016-10-05
vulkano-shaders generates wrong padding for vec3 array[2]


svenstaro:
So I have this:

```
struct MyStruct {
  vec3 this_breaks[2];
  uint something;
};
```

Which generates this excerpt:

```
    pub this_breaks: [[f32; 3]; 2] /* offset: 0 */,
    pub _dummy0: [u8; 8],
    pub something: u32 /* offset: 32 */,
```

Now the offset is obviously wrong. It should look like this:

```
    pub this_breaks: [[f32; 3 (and then we need a +1 padding here somehow)]; 2] /* offset: 0 */,
    pub something: u32 /* offset: 32 */,
```

This had led to some actual problems in my code. If use use a `vec4 array[2]` this works totally fine, by the way.

thiolliere:
I also have this issue with a mat3

singron:
It looks like the struct generator ignores ArrayStride and uses the default rust array layouts, which are wrong for vec3[x], mat3[x], and arrays of various non-square matrices as well.

I think it may also be ignoring MatrixStride as well. mat2, mat3, and mat4 all have a MatrixStride of 16 with glslangValidator, so it seems like this should mess with pretty much any matrix that doesn't have columns of length 4 even if they aren't in an array. I haven't confirmed that issue though so I might have just overlooked something.

Any ideas for fixing this? I don't think it's possible to tell the rust compiler to use an arbitrary array stride without wrapping each element of the array with a struct that has dummy padding. However, this makes the generated code difficult to use since there might be these wrapping structs all over the place depending on if rust and spirv agree on the array stride. I have a patch that just rejects these shader programs and requires the programmer to insert wrapping structs in the GLSL code. It's more work for the programmer, but then the GLSL code has the same structure as the rust code.

rukai:
Compile time error handling for this case implemented by https://github.com/vulkano-rs/vulkano/pull/1050

PR #1050, opened 2018-09-24
Reject spirv arrays that have incorrect stride in rust

singron:
SPIR-V allows the array stride and size of a type to differ, but rust defines them to be the same. Thus
certain types when represented in rust will have the wrong layout. E.g. an array of vec3 can have an array
stride of 16 in SPIR-V, but an array of [f32;3] in rust would have a stride of 12. Thus using one for the
other would cause corruption.

This suggests a workaround by using a wrapping struct or upgrading the size of the type to one where the size is the array stride.

I considered generating the wrapping struct for the user, but that seems very confusing for the user. We could generate wrapping structs for all vec and mat types in arrays, but that would be a large API change.

See #298.

  I did this because people affected by this bug will now get compile errors. This could be surprising, but their code was incorrect all along.
  I didn't see anything relevant to update.
  I didn't find anything relevant to update there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 - Fix instance_count when using draw_index with instance buffers
 - Added a `reinterpret` function to `BufferSlice`
 - Made `AttributeInfo` derive `Copy`, `Clone` and `Debug`
-
 - Use [google/shaderc](https://github.com/google/shaderc-rs) for shader compilation
+- Reject generation of rust types for SPIR-V arrays that would have incorrect array stride.
 
 # Version 0.10.0 (2018-08-10)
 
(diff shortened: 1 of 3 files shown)


rukai:
Thanks for the PR, sorry it took me a while to get to it.
The extra effort is appreciated: explanation of changes, changelog entry and tests
I'll merge it as is, but it has highlighted some potential follow up work.

It looks like decorations could also come from an `OpGroupDecorate`. However `OpGroupDecorate` is not used anywhere else in vulkano-shaders, so such a change would make sense to go in another PR and add it to all decoration code.
I think an abstraction like this would make sense:
```rust
impl Spirv {
    /// Returns the params held by the decoration for the specified id and type
    /// Searches raw OpDecorate and OpDecorationGroup
    fn get_decoration_params(self, id: u32, decoration: Decoration) -> Vec<u32> { }

    /// Returns the params held by the decoration for the member specified by id, member and type
    /// Searches raw OpMemberDecorate and OpMemberDecorationGroup
    fn get_member_decoration_params(self, id: u32, member: u32, decoration: Decoration) -> Vec<u32> { }
}
```

PR #1052, opened 2018-09-28
Add Clone/Debug to QueuesIter.

tarka:
This is primarily to allow debugging, e.g:

    println!("Queues: {:?}", queues.clone().count());
    let queue = queues.next().ok_or(VkError::NoAvailableQueue)?;

### DIFF
--- a/vulkano/src/device.rs
+++ b/vulkano/src/device.rs
@@ -556,6 +556,7 @@ unsafe impl<T> DeviceOwned for T
 }
 
 /// Iterator that returns the queues produced when creating a device.
+#[derive(Clone, Debug)]
 pub struct QueuesIter {
     next_queue: usize,
     device: Arc<Device>,


rukai:
Thanks for the PR, however:
I think adding `Clone` will break invariants. It would allow for multiple `Queue` structs referring to the same queue whereas previously we just had one `Arc<Queue>`.
You can can achieve what you want by doing this instead:
```rust
let queues_vec: Vec<_> = queues.collect();
println!("len: {}", queues_vec.len());
let get_iter_again_by_doing_this = queues_vec.iter();
```

Adding Debug is not correct because it's displaying implementation details. If there is a specific field you need access to, maybe we can talk about a proper way to expose it?
I noticed that `Queue` itself has Debug. I'm guessing that's just left there from early stages of development and should be removed. But maybe I'm wrong and exposing such things is useful for a library like vulkano.

tarka:
Thanks @rukai, I'm happy to defer to your opinion on this, I assumed it was just an omission.

rukai:
No worries

PR #1062, opened 2018-10-05
vulkano-shaders - proc macros 2.0

rukai:
* [x] Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR. https://github.com/vulkano-rs/vulkano-www/pull/41

Would like feedback on the new vulkano-shaders macro API.

This PR will need to wait for the stabilization of proc macros 2.0, which will happen in rust 1.30.
I'm pretty sure I'll need to remove the `#![feature(proc_macro_non_items)]` from the examples and just move the `vulkano_shader!` calls to the root of the module.

# Proposed syntax
Compact, consistent, everything is a field.
```rust
vulkano_shader!{
     mod_name: vertex_shader,
     ty: "vertex",
     src: "
#version 450

layout(location = 0) in vec3 position;

void main() {
   gl_Position = vec4(position, 1.0);
}"
}
```

# Alternative syntax

Tries to use the standard mod syntax to make it easier to understand what the mod setting does.
I included `Arguments` so its like a struct, otherwise its weird because a mod doesn't have fields.
I don't really like how this turned out at all, but I included it to hopefully motivate others to give some ideas.
```rust
vulkano_shader! {
    mod vs {
        Arguments {
            ty: "vertex",
            src: "
#version 450
layout(location = 0) in vec2 position;
void main() {
gl_Position = vec4(position, 0.0, 1.0);
}"
        }
    }
}
```


rukai:
With the release of rust 1.30 this can now be merged!

Ralith:
Is it absolutely necessary for the shader program to be quoted? Could rust tokens be used instead, sort of like [phaazon's quasiquoting macro] does?

rukai:
I would certainly like to try it at some point.

At the moment it would just be converted to a string again, so we can feed it to shaderc-rs.
So the only advantage would be having rust syntax highlighting on the glsl.
Are there any other advantages or things we could do, that you had in mind?

rukai:
Oh right, spans!
I might be able to make rustc give errors on the correct tokens, that would be awesome!

PR #1071, opened 2018-10-08
Add ranges for GPU locks

### LINKED ISSUE #1070: Writing to a range of a buffer while drawing another range causes an AccessError
Hi there,

I think I'm hitting a limitation with how the GPU lock is implemented in Vulkano. I'm trying to write to a buffer range (let's say 0..36), while another command buffer is drawing another range (say 36..72) of the same buffer. Since the ranges do not overlap, this should be safe, but executing the command buffer fails with an `AccessError`, telling me that the buffer is already in use.

From glancing at the code, it looks like `BufferSlice` locks the entire buffer in `try_gpu_lock` instead of only locking a range of the buffer. Indeed, if I comment out the function for debugging purposes, my code seems to work as expected.

Obviously, I can work around this by waiting for the write to complete before drawing the buffer, but that's not ideal.

My first thought was that this is a bug in Vulkano, but perhaps I'm supposed to do this some other way? Or maybe some of my assumptions are incorrect, and it's a bug in my code instead?

aleksijuvani:
* [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

Fixes #1070.

Would like to request feedback. There's some more work to be done before this should be merged, e.g. `try_gpu_lock` and related functions could be refactored into a `GpuAccess` struct which is then added as a field to all the different buffers.

Only the `DeviceLocalBuffer` currently supports the locking of subranges. Would it also be desirable to implement this for other types of buffers? For `ImmutableBuffer` I think locking the buffer is a no-op anyway, but what about `CpuAccessibleBuffer` and `CpuBufferPool`?

I think it would make sense to also support this for `CpuAccessibleBuffer`, but I'm not sure about `CpuBufferPool`. Considering how `CpuBufferPool` is generally used (a chunk of the buffer is filled with data, sent to the GPU, and then another chunk of the buffer is used the next frame), does it even make sense to support this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@
 - Use [google/shaderc](https://github.com/google/shaderc-rs) for shader compilation
 - Reject generation of rust types for SPIR-V arrays that would have incorrect array stride.
 - Removed the `Layout` prefix of the descriptions used for a render pass.
-- Added the ability to lock a range of `DeviceLocalBuffer` with `try_gpu_lock`
-- Changed the GPU to lock a range of `DeviceLocalBuffer` instead of the entire buffer when `BufferSlice`s are used
+- Added the ability to lock a range of a buffer with `try_gpu_lock`
+    +   Applicable to `DeviceLocalBuffer` and `CpuAccessibleBuffer`
+    +   `CpuBufferPool` still locks the entire buffer chunk
+- Changed the GPU to lock a range of the buffer instead of the entire buffer when `BufferSlice`s are used
 
 # Version 0.10.0 (2018-08-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Use [google/shaderc](https://github.com/google/shaderc-rs) for shader compilation
 - Reject generation of rust types for SPIR-V arrays that would have incorrect array stride.
 - Removed the `Layout` prefix of the descriptions used for a render pass.
+- Added the ability to lock a range of `DeviceLocalBuffer` with `try_gpu_lock`
+- Changed the GPU to lock a range of `DeviceLocalBuffer` instead of the entire buffer when `BufferSlice`s are used
 
 # Version 0.10.0 (2018-08-10)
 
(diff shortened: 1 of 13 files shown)


aleksijuvani:
This will also need further testing, but I've now also implemented GPU lock ranges for `CpuAccessibleBuffer`. I'm testing this out with a personal project, but it's still in early stages and it might take a while until I can properly test all these changes.

I'm a little worried about the amount of locks we'll potentially have to keep track of in a real application, and how it will affect performance. If this turns out to be a problem, there's an optimization we can do with interval trees. I'm still convinced that this is a valuable change, because right now Vulkano is too restrictive with its GPU lock, and it prevents valid use-cases.

rukai:
I tested https://github.com/rukai/PF_Sandbox with your changes and it works fine.
However this is hardly a comprehensive test, PF Sandbox rendering is very simple.

aleksijuvani:
Cool, thank you. I'm going to add a few unit tests for this and maybe a benchmark.

aleksijuvani:
So I wrote a benchmark, and locking and unlocking 100 locks takes around 33 µs on my system, which is acceptable, but as soon as we're up to 1000 locks it starts to take around 3.30 ms. It's not unusual for games to issue thousands of draw calls, so we should definitely try to optimize this.

I think we should be able to get this number down if we implemented the range overlap checks with an interval tree instead of naively looping through all the ranges to check if they overlap. I'll give it a try.

aleksijuvani:
Performance is now looking better, but we're still hitting 1.10 milliseconds once we reach 5000 locks in a command buffer. Similar to how Rust provides a flag to disable array bounds checks in release builds, maybe we should provide a flag to disable this particular safety check?

```
Benchmarking lock and unlock/100: Collecting 100 samples in estimated 5.0757 s

lock and unlock/100     time:   [18.538 us 18.700 us 18.869 us]
                        change: [-44.214% -43.778% -43.349%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Benchmarking lock and unlock/1000: Collecting 100 samples in estimated 5.5561 s
lock and unlock/1000    time:   [207.42 us 208.86 us 210.35 us]
                        change: [-93.639% -93.596% -93.553%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking lock and unlock/5000: Collecting 100 samples in estimated 5.5224 s

lock and unlock/5000    time:   [1.1030 ms 1.1097 ms 1.1166 ms]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
Benchmarking lock and unlock/10000: Collecting 100 samples in estimated 11.632 s

lock and unlock/10000   time:   [2.2417 ms 2.2565 ms 2.2716 ms]
```

PR #1076, opened 2018-10-15
Add logo image to HTML documentation

msmorgan:


rukai:
Looks nice! thanks.

msmorgan:
@rukai I only added this to the main package in this PR, but do you think it should be added to vulkano-shader-derive and vulkano-win as well?

rukai:
oh, good point. do it for:
*   vk-sys
*   vulkano-shaders
*   vulkano-win

vulkano-shader-derive will be removed in 2 weeks so no point in doing it there.

Issue #1078, opened 2018-10-16
Panic `Wrong value for vk::PresentModeKHR` on Android


LachlanHogan:
I am able to compile and run the teapot example in the `examples` directory using 
`cargo apk run --bin teapot`

The program panics when attempting to check the surface capabilities here: 
`let caps = surface.capabilities(physical).expect("failed to get surface capabilities");` 
with the panic error 
`thread 'main' panicked at 'Wrong value for vk::PresentModeKHR'`.

I have tried enabling the `khr_android_surface` instance extension with no extra luck

Do I perhaps need to create the surface from the native Android surface?

Any help would be appreciated, thanks!

rukai:
It looks like some present modes arent implemented yet, should be fairly easy to fix. Just need to handle all of the possible VkPresentModeKHR in capabilities.rs

LachlanHogan:
> It looks like some present modes arent implemented yet, should be fairly easy to fix. Just need to handle all of the possible VkPresentModeKHR in capabilities.rs

I thought `PresentMode::Fifo` was required to be supported by every platform? As it mentions in the LunarG docs 

>  VK_PRESENT_MODE_FIFO_KHR: ... This is the only value of presentMode that is required to be supported.

Source: https://vulkan.lunarg.com/doc/view/1.0.26.0/linux/vkspec.chunked/ch29s05.html

rukai:
It's not that `PresentMode::Fifo` is not supported.
The issue is that Vulkano is panicking if vulkan even mentions it supports a present mode that vulkano doesn't know about. https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/swapchain/capabilities.rs#L106

Can you please test this PR https://github.com/vulkano-rs/vulkano/pull/1079 ? Thanks.

LachlanHogan:
@rukai Tested the changes on your branch and the app now successfully runs! Thanks!

rukai:
Thanks!

PR #1079, opened 2018-10-17
Add missing present modes

rukai:

Fixes https://github.com/vulkano-rs/vulkano/issues/1078#issuecomment-430417152


vulkano/src/swapchain/capabilities.rs
```diff
@@ -103,7 +111,9 @@ pub fn supported_present_modes_from_list<I>(elem: I) -> SupportedPresentModes
             vk::PRESENT_MODE_MAILBOX_KHR => result.mailbox = true,
             vk::PRESENT_MODE_FIFO_KHR => result.fifo = true,
             vk::PRESENT_MODE_FIFO_RELAXED_KHR => result.relaxed = true,
-            _ => panic!("Wrong value for vk::PresentModeKHR"),
+            vk::PRESENT_MODE_SHARED_DEMAND_REFRESH_KHR => result.shared_demand = true,
+            vk::PRESENT_MODE_SHARED_CONTINUOUS_REFRESH_KHR => result.shared_continuous = true,
+            _ => panic!("Wrong value for vk::PresentModeKHR: {}", e),
```
aleksijuvani: Panicking here breaks existing applications if new present modes are added to Vulkan. Perhaps this should be changed into a `debug_assert!` instead?

```suggestion
            _ => debug_assert!(false, "Wrong value for vk::PresentModeKHR: {}", e),
```

rukai:
Good point, I've changed the PR to silently ignore unknown present modes.
Vulkan could easily add a new one in a minor patch.
We don't want existing applications to break.

PR #1080, opened 2018-10-17
Update .rustfmt.toml

aleksijuvani:
Some options have been removed, combined or renamed since this file was first written. I tried to find equivalents for these in the new options. See rust-lang-nursery/rustfmt#1974.


rukai:
Thanks for the PR.
I want to look into rustfmt settings to figure out if there are sensible enough options, such that we can enforce it via CI. I think the existing code was once formatted by rustfmt? and its imo really bad. There is a lot of that aligns the arguments of function calls to the first argument which looks awful when the first argument is already far to the right.

aleksijuvani:
I think there's value in enforcing formatting via CI, and I do it in my personal project, but some might find it too overbearing. I looked through the commits and `rustfmt` was ran once to format entire codebase, but it looks like it hasn't been used since.

If we enforce it via CI, we should also look over all the existing files for code that `rustfmt` doesn't handle well, and either disable formatting for those sections or find an option that handles it better.

Even if we don't end up enforcing it via CI, I think there's value in having the configuration file there. It's useful to be able to format any new files added to the project, which I ran into with my other pull request.

knappador:
Git hooks for commits.  Personally I have my editor set up to run the rustfmt, but everyone uses git.  I like getting whitespace out of commit history where possible.  Can lead to a lot of merge distraction.

@aleksijuvani  can you open up a branch with two commits:

1)  a fresh Rustfmt before the change to establish baseline
2)  Rustfmt after the change to see what's actually going to be different between the two configs

This is a faster method of reaching consensus on our rustfmt

Only after we change rustfmt should we tack on a major whitespace commit to get on a system.  They are terrible merge hazards.

mitchmindtree:
> I think the existing code was once formatted by rustfmt? and its imo really bad. 

I agree with this. I'm a bit of an advocate for just using the default `rustfmt`, whatever the result might be. There are always differing opinions about correct formatting - by just using the default we at least get a reasonable standard and we remove the problem of deciding on the "correct" styling out of vulkano itself.

Re enforcing via git, I don't mind so much either way, though it is nice to have consistent src :)

Issue #1073, opened 2018-10-11
Deferred example issue (linux, nvidia)


skemps:
Hi!

The example starts fine but after drawing first frame winit event triggers recreate swapchain which in the next frames fail with SwapchainCreationError::UnsupportedDimensions and than it loops forever since it never successfully recreates swapchain.
I am not actually resizing or moving window.

Any ideas?

aleksijuvani:
This example doesn't work properly for me on Windows either. I'm not sure what's supposed to happen, but it just seems to freeze indefinitely. I also have an Nvidia GPU.

rukai:
I ran the example on linux + nvidia using vulkano 0.10 and the latest git master.
It runs fine for me.

Can you print out the dimensions it is attempting to use and let me know?
Your window manager might be giving it invalid dimensions.
Which window manager are you using?
I have had an issue on windows before where it was attempting to use dimensions of (0, 0)

If this is the case, the solution here is probably to just remove the line
`winit::Event::WindowEvent { event: winit::WindowEvent::Resized (_), .. } => recreate_swapchain = true,`
Try that and let me know how it goes.

skemps:
Here it just seems that it runs fine because I can see the triangle. But actually it's not. For example CloseRequested doesn't trigger because of constant UnsupportedDimension loop.

Dimensions are [1195, 896]. This is probably 1024x768 + decoration.

If I remove the handling of the resized event than it works, but crashes if I actually resize the window (Result::unwrap()` on an `Err` value: OutOfDate on `let after_frame = after_future.unwrap()...`)

rukai:
Oh huh, the deferred example doesnt handle the swapchain being out of date like the other examples do.
I'll fix that.

rukai:
Can you give this PR a go? https://github.com/vulkano-rs/vulkano/pull/1074

rukai:
Thanks!

rukai:
@skemps what window manager are you using? X11 or wayland?
Looks like someone else has run into an issue where something similar to the old implemenation is required and I want to be able to reproduce this issue so i don't introduce a regression.

skemps:
I'm using X11.

rukai:
What window manager?

skemps:
I think it's kwin. I'm not sure. I'm not at my computer right now. It's the default debian stretch with kde I'm using.

rukai:
[...]

PR #1084, opened 2018-10-24
Fix triangle example.

dvc94ch:
Not sure if this is the right way to fix it but it works for me.

Some more information:
Happens with Fedora + Wayland.
current_extent always returns None.
WindowEvent::Resized happens on startup without resizing the window.

Before:
screenshot from 2018-10-24 15-14-59

After:
screenshot from 2018-10-24 15-15-45

### DIFF
--- a/examples/src/bin/triangle.rs
+++ b/examples/src/bin/triangle.rs
@@ -153,7 +153,12 @@ fn main() {
 
     // The dimensions of the surface.
     // This variable needs to be mutable since the viewport can change size.
-    let mut dimensions;
+    let mut dimensions = {
+        let window = surface.window();
+        let dim = window.get_inner_size().unwrap()
+            .to_physical(window.get_hidpi_factor());
+        [dim.width as u32, dim.height as u32]
+    };
 
     // Before we can draw on the surface, we have to create what is called a swapchain. Creating
     // a swapchain allocates the color buffers that will contain the image that will ultimately
@@ -163,8 +168,6 @@ fn main() {
         // pass values that are allowed by the capabilities.
         let caps = surface.capabilities(physical)
                          .expect("failed to get surface capabilities");
-
-        dimensions = caps.current_extent.unwrap_or([1024, 768]);
 
         // We choose the dimensions of the swapchain to match the current extent of the surface.
         // If `caps.current_extent` is `None`, this means that the window size will be determined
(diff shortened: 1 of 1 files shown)


rukai:
Thanks for the PR!

What window manager are you using?

This might conflict with my fix to https://github.com/vulkano-rs/vulkano/issues/1073 I'll need to test on their setup to make sure this doesn't break my fix.

Also this will need to be applied to every example in this repo as well as the examples in https://github.com/vulkano-rs/vulkano-www
BUT dont do that yet!
Wait until I've confirmed that this is the correct fix.

examples/src/bin/triangle.rs
```diff
@@ -348,9 +348,9 @@ void main() {
         // If the swapchain needs to be recreated, recreate it
         if recreate_swapchain {
             // Get the new dimensions for the viewport/framebuffers.
-            dimensions = surface.capabilities(physical)
+            /*dimensions = surface.capabilities(physical)
```
rukai: Please do not comment this out.

examples/src/bin/triangle.rs
```diff
@@ -482,6 +482,11 @@ void main() {
         events_loop.poll_events(|ev| {
             match ev {
                 winit::Event::WindowEvent { event: winit::WindowEvent::CloseRequested, .. } => done = true,
+                winit::Event::WindowEvent { event: winit::WindowEvent::Resized(dim), .. } => {
+                    let dim = dim.to_physical(surface.window().get_hidpi_factor());
```
rukai: The code you commented out above should already handle this, please remove this line.

rukai:
Oh wow I only just noticed you said "current_extent always returns None." my bad

examples/src/bin/triangle.rs
dvc94ch: I removed the commented out part and using surface.window().get_inner_size() to initialize dimensions instead of hard coding [1024, 768]. This way it should still work in case winit changes it's default or the user changes the default in the WindowBuilder.

rukai:
Currently your code still doesnt handle platforms where the swapchain must use a specific dimensions.

https://github.com/vulkano-rs/vulkano/blob/7c62a8cc5258d5095f811b13326f715161f17e61/examples/src/bin/triangle.rs#L398
For example we hit line 398, because the window was resized in between checking events and acquiring the next image. The swapchain will then be set to the wrong size because you have removed the code that sets the dimensions of the swapchain to `current_extent`

I'm proposing the following changes. I've tested it to work on x11 i3/kde. Can you verify that this branch fixes your issue? https://github.com/rukai/vulkano/tree/fix_wayland_resize
Windows+AMD is currently segfaulting but I need to sleep now...

I'm assuming that the physical dimensions of the window will always be the dimensions required by the swapchain. But that might not be the case ...

dvc94ch:
Yes that works

rukai:
I fixed the segfault on windows+amd by updating my amd drivers.

So either:
*   You take the changes in my branch and apply it to the triangle example + all other examples in your PR.
*   Or you can close your PR and I will apply it to the other examples and make my own PR.

dvc94ch:
Thank you for taking care of it

rukai:
Ok, no worries, thanks for the initial work!

PR #1089, opened 2018-10-26
Fix wayland resize issue

rukai:

Continuing on from https://github.com/vulkano-rs/vulkano/pull/1084

Will apply to the rest of the issues when I can confirm it doesn't regress https://github.com/vulkano-rs/vulkano/issues/1073


rukai:
Some drivers dont enforce a swapchain extent, so you need to check for
resize events from the os and then recreate the swapchain.
I took this opportunity to refactor the examples a bunch.
But I want to leave the rest to a follow up PR mostly making use of `using` consistent.

PR #1094, opened 2018-10-27
Descriptor set collection vec

jonathansty:

This PR implements `DescriptorSetsCollection` for `Vec<T>` so you can construct a list of descriptor sets more easily for use with `draw` and `draw_indexed`

```
let sets = vec![
    set0.clone(),
    set1.clone()
];

command_buffer = command_buffer.draw_indexed( pipeline.clone(), &dynamic_states, vec![obj.vertices().clone()],indices.clone(), sets, *matrix).unwrap();
```

Not sure if this was left out on purpose or not, but I found myself needing this to easily send descriptor sets around, I combine a uniform and per material set for textures etc.

### DIFF
--- a/vulkano/src/descriptor/descriptor_set/collection.rs
+++ b/vulkano/src/descriptor/descriptor_set/collection.rs
@@ -84,14 +84,18 @@ unsafe impl<T> DescriptorSetsCollection for Vec<T>
 
     #[inline]
     fn num_bindings_in_set(&self, set: usize) -> Option<usize> {
-        let set  = &self[set];
-        Some(set.num_bindings())
+        match self.get(set) {
+            Some(set) => Some(set.num_bindings()),
+            None => None,
+        }
     }
 
     #[inline]
     fn descriptor(&self, set: usize, binding: usize) -> Option<DescriptorDesc> {
-        let set = &self[set];
-        set.descriptor(binding)
+        match self.get(set) {
+            Some(set) => set.descriptor(binding),
+            None => None,
+        }
     }
 }
 
--- a/vulkano/src/descriptor/descriptor_set/collection.rs
+++ b/vulkano/src/descriptor/descriptor_set/collection.rs
@@ -84,12 +84,8 @@ unsafe impl<T> DescriptorSetsCollection for Vec<T>
 
     #[inline]
     fn num_bindings_in_set(&self, set: usize) -> Option<usize> {
-        match self.get(set) {
-            Some(set) => Some(set.num_bindings()),
-            None => None,
-        }
+        self.get(set).map(|x| x.num_bindings())
     }
-
     #[inline]
     fn descriptor(&self, set: usize, binding: usize) -> Option<DescriptorDesc> {
         match self.get(set) {
--- a/vulkano/src/descriptor/descriptor_set/collection.rs
+++ b/vulkano/src/descriptor/descriptor_set/collection.rs
@@ -88,10 +88,7 @@ unsafe impl<T> DescriptorSetsCollection for Vec<T>
     }
     #[inline]
     fn descriptor(&self, set: usize, binding: usize) -> Option<DescriptorDesc> {
-        match self.get(set) {
-            Some(set) => set.descriptor(binding),
-            None => None,
-        }
+        self.get(set).and_then(|x| x.descriptor(binding))
     }
 }
 
--- a/vulkano/src/descriptor/descriptor_set/collection.rs
+++ b/vulkano/src/descriptor/descriptor_set/collection.rs
@@ -70,6 +70,31 @@ unsafe impl<T> DescriptorSetsCollection for T
     }
 }
 
+unsafe impl<T> DescriptorSetsCollection for Vec<T>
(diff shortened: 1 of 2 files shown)


vulkano/src/descriptor/descriptor_set/collection.rs
```diff
@@ -70,6 +70,31 @@ unsafe impl<T> DescriptorSetsCollection for T
...
+
+    #[inline]
+    fn num_bindings_in_set(&self, set: usize) -> Option<usize> {
+        let set  = &self[set];
+        Some(set.num_bindings())
+    }
+
+    #[inline]
+    fn descriptor(&self, set: usize, binding: usize) -> Option<DescriptorDesc> {
+        let set = &self[set];
```
rukai: Same

rukai:
Is there a reason tuples are insufficient for your use? https://docs.rs/vulkano/0.10.0/vulkano/descriptor/descriptor_set/collection/trait.DescriptorSetsCollection.html

jonathansty:
@rukai  is there a way to pass in a `Vec` and can it convert to tuple? 

I wanted to construct a `Vec` of descriptor sets depending on the properties the object that I'm going to draw with has. With the tuples I would always have to at some point "hard code" the amount of descriptor sets. But I don't always know how many descriptor sets the "material" might add. 

Does that make sense?

example:
```
pub fn render_object(
    buffer : AutoCommandBufferBuilder,
    obj : &render_helpers::RenderObject,
    matrix : &Matrix4<f32>, 
    pipeline : &Arc<GraphicsPipelineAbstract + Send + Sync>, 
    dynamic_states : &vulkano::command_buffer::DynamicState,
    desc_collection : Vec<Arc<vulkano::descriptor::descriptor_set::DescriptorSet+ Send + Sync>>
    ) -> AutoCommandBufferBuilder {

    let mut command_buffer = buffer;
    command_buffer = command_buffer.draw_indexed(
                pipeline.clone(), &dynamic_states, vec![obj.vertices().clone()], (), desc_collection, *matrix)
            .unwrap();
    command_buffer
}
```

Calling this function would not be possible as there wasn't an implementation of `DescriptorSetCollection`. I wonder if it would be possible to implement this on a slice instead, although I'm not sure how.

vulkano/src/descriptor/descriptor_set/collection.rs
```diff
@@ -70,6 +70,31 @@ unsafe impl<T> DescriptorSetsCollection for T
...
+        let mut v = Vec::new();
+        for o in self {
+            v.push(Box::new(o) as Box<_>);
+        }
+        return v;
+    }
+
+    #[inline]
+    fn num_bindings_in_set(&self, set: usize) -> Option<usize> {
+        let set  = &self[set];
```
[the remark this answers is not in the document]
jonathansty: Sure! Thanks :+1:

vulkano/src/descriptor/descriptor_set/collection.rs
```diff
@@ -70,6 +70,35 @@ unsafe impl<T> DescriptorSetsCollection for T
...
+        let mut v = Vec::new();
+        for o in self {
+            v.push(Box::new(o) as Box<_>);
+        }
+        return v;
+    }
+
+    #[inline]
+    fn num_bindings_in_set(&self, set: usize) -> Option<usize> {
+        match self.get(set) {
```
rukai: You should be able to do self.get(set).map(|x| x.num_bindings())

vulkano/src/descriptor/descriptor_set/collection.rs
```diff
@@ -70,6 +70,35 @@ unsafe impl<T> DescriptorSetsCollection for T
...
+            Some(set) => Some(set.num_bindings()),
+            None => None,
+        }
+    }
+
+    #[inline]
+    fn descriptor(&self, set: usize, binding: usize) -> Option<DescriptorDesc> {
+        match self.get(set) {
```
rukai: You should be able to do self.get(set).flat_map(|x| x.descriptor(binding))

rukai:
That's reasonable.
If you cant get a slice working, I'm happy with adding an impl for vec.

vulkano/src/descriptor/descriptor_set/collection.rs
jonathansty: This one seems to not compile as `get`. 
```
error[E0599]: no method named `flat_map` found for type `std::option::Option<&T>` in the current scope
  --> /home/jonathan/projects/experiments/vulkano/vulkano/src/descriptor/descriptor_set/collection.rs:91:23
   |
91 |         self.get(set).flat_map(|x| x.descriptor(binding))
   |                       ^^^^^^^^
   |
   = note: the method `flat_map` exists but the following trait bounds were not satisfied:
           `&mut std::option::Option<&T> : std::iter::Iterator`
```

rukai: Oh, its actually called `and_then`, my bad, https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then

PR #1112, opened 2018-11-14
Expose `CopyImageError` and `DrawIndexedIndirectError`

mitchmindtree:
I noticed that these two were missing a pub export while reading through
the docs so I thought I'd add them!

* [ ] ~Updated documentation to reflect any user-facing changes - in this repository~


rukai:
Good catch, thanks for the PR!

PR #1116, opened 2018-11-18
Feature shader include

Limeth:
Adds support for `#include "..."` and `#include <...>` directives within source files.

* [ ] ~~Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.~~ Should not be needed, as this is an addition.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -9,7 +9,6 @@
 
 use std::io::Error as IoError;
 use std::path::Path;
-use std::fs;
 
 use syn::Ident;
 use proc_macro2::{Span, TokenStream};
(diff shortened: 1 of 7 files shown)


rukai:
Thanks for the PR!

Please include an example in the examples directory.
Use one of the compute examples as a base as they have less boilerplate.
In the one example, demonstrate both relative and standard includes in an external .glsl file and standard includes for embedded glsl.

This is important as the examples are not only helpful to users but also act as tests for the `shader!` macro.

Limeth:
Okay, I added a simple example and tried to polish the error messages. Unfortunately the way relative include directives are handled is, that if they fail, they fall back to inclusion from standard directories, which sometimes results in misleading error messages. I think that's something that could be improved upon with further PRs.

vulkano-shaders/src/codegen.rs
```diff
@@ -24,13 +25,130 @@ use entry_point;
...
+    };
+
+    let canonical_file_to_include = file_to_include.canonicalize()
+        .unwrap_or_else(|_| file_to_include);
+    let canonical_file_to_include_string = canonical_file_to_include.to_str()
+        .expect("Could not stringify the file to be included. Make sure the path consists of \
+                 valid unicode characters.")
+        .to_string();
+    let content = read_file_to_string(canonical_file_to_include.as_path())
+        .map_err(|_| format!("Could read the contents of file `{}` to be included in the shader \
```
rukai: Missed a word. should be "Could not read ..."

vulkano-shaders/src/codegen.rs
```diff
@@ -24,13 +25,130 @@ use entry_point;
...
+
+            found_requested_source_path.unwrap()
+        },
+    };
+
+    let canonical_file_to_include = file_to_include.canonicalize()
+        .unwrap_or_else(|_| file_to_include);
+    let canonical_file_to_include_string = canonical_file_to_include.to_str()
+        .expect("Could not stringify the file to be included. Make sure the path consists of \
+                 valid unicode characters.")
```
rukai: Looks like the expect here can be triggered on certain user setup/input. Can it be changed to a `map_err`?

vulkano-shaders/src/codegen.rs
```diff
@@ -24,13 +25,130 @@ use entry_point;
...
+                                    inclusion from standard directories (`#include <...>`), try \
+                                    using `#include \"...\"` instead. Requested path: {}",
+                                   requested_source_path_raw));
+            }
+
+            let mut found_requested_source_path = None;
+
+            for include_directory in include_directories {
+                let include_directory_path = Path::new(include_directory).canonicalize()
+                    .unwrap_or_else(|_| panic!("Invalid standard shader inclusion directory `{}`.",
```
rukai: Why do you panic if this `canonicalize` fails but not on the other calls to `canonicalize`?

Limeth: A panic generates a compilation error, and I thought it would be appropriate to generate one when the user specifies an include directory that doesn't exist within the project structure.

vulkano-shaders/src/codegen.rs
Limeth: Fixed, thanks. :ok_hand:

rukai:
I was being pedantic about returning Err vs panic, because the difference will be relevant when proper proc macro diagnostics is stabilized. However now that I think about it we will probably need to return the correct span as well. So may as well leave it now and fix it up when diagnostics is stable.

I'll merge the PR now, thanks for your contribution!

PR #1117, opened 2018-11-20
Remove seemingly unnecessary `undefined_layout` field from `swapchain::ImageEntry`?

### LINKED ISSUE #977: Can't execute command on swapchain image
Hello,

When I try to execute a command directly on a swapchain image, I got a layout error:

```
thread '<unnamed>' panicked at 'Can't execute clear command buffer: AccessError { error: ImageNotInitialized { requested: PresentSrc }, command_name: "vkCmdCopyImageToBuffer", command_param: "source", command_offset: 0 }', libcore/result.rs:945:5
```

With a command like this one:

```
let cb = AutoCommandBufferBuilder::primary_one_time_submit(device.clone(), queue.family(),).unwrap()
            .copy_image_to_buffer(
                swapchain_images[image_num].clone(),
                buffer.clone()
            )
            .unwrap()
            .build()
            .unwrap();
```

I got the same error with the `clear_color_image` command.
I searched in the API but I can't find a way to manually update the layout. Why the `AutoCommandBufferBuilder` doesn't expose the `vkCmdPipelineBarrier` function ?

Thanks,
JS

mitchmindtree:
Upon `Swapchain::new_inner`, `ImageEntry`s are created for each image with their `undefined_layout` field set to `true`.

Once set, this field is never altered within the vulkano codebase. The only time it is used following initialisation is when it is read during calls to `SwapchainAcquireFuture::check_image_access`. For reference, the case in which I am hitting this method is when attempting to execute
a command buffer that clears a swapchain image using `clear_color_image`.

If the `undefined_layout` field is `true` and the swapchain image layout is not `Undefined`, the `check_image_access` function returns an error. [See here](https://docs.rs/vulkano/0.11.1/src/vulkano/swapchain/swapchain.rs.html#828-835). This is a problem as the implementation of `ImageAccess::initial_layout_requirement` and `ImageAccess::final_layout_requirement` both return `ImageLayout::PresentSrc` which implies to me that by the time the command buffer is executed the swapchain image layout *must* be `PresentSrc` (the call to `add_image_memory_barrier` within
`SyncCommandBufferBuilder::build` [ensures this](https://docs.rs/vulkano/0.11.1/src/vulkano/command_buffer/synced/base.rs.html#720)) and in turn, this error condition will always be true.

In this PR I remove the unused `undefined_layout` field as well as the error condition that does not seem to make sense and as a result my downstream project seems to work fine. However, it is more than likely that there is some special case that @tomaka intended to cover here and
that there is a better, proper solution that involves checking that swapchain image layouts are actually initially `PresentSrc` (and not `Undefined`) and only then setting the `undefined_layout` field to `false`.

@tomaka could you share your thoughts on this briefly? I'm happy to dive in and do a proper fix if there is indeed some case you had in mind here that I'm overlooking.

Eventually, this should close #977.

* [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -192,8 +192,6 @@ pub struct Swapchain<W> {
 struct ImageEntry {
     // The actual image.
     image: UnsafeImage,
-    // If true, then the image is still in the undefined layout and must be transitioned.
-    undefined_layout: AtomicBool,
 }
 
 impl <W> Swapchain<W> {
@@ -443,7 +441,6 @@ impl <W> Swapchain<W> {
 
                 ImageEntry {
                     image: img,
-                    undefined_layout: AtomicBool::new(true),
                 }
             })
             .collect::<Vec<_>>();
@@ -823,15 +820,6 @@ unsafe impl<W> GpuFuture for SwapchainAcquireFuture<W> {
         let swapchain_image = self.swapchain.raw_image(self.image_id).unwrap();
         if swapchain_image.image.internal_object() != image.inner().image.internal_object() {
             return Err(AccessCheckError::Unknown);
-        }
-
-        if self.swapchain.images[self.image_id]
-            .undefined_layout
-            .load(Ordering::Relaxed) && layout != ImageLayout::Undefined
-        {
-            return Err(AccessCheckError::Denied(AccessError::ImageNotInitialized {
-                                                    requested: layout,
-                                                }));
         }
 
         if layout != ImageLayout::Undefined && layout != ImageLayout::PresentSrc {


tomaka:
Well, swapchain images are initially undefined and need to be transitioned to the `PresentSrc` layout. The thing that's not implemented is the transitioning.

mitchmindtree:
@tomaka thanks for the quick reply!

If swapchain images are initially undefined, should the implementation of `ImageAccess::initial_layout_requirement` for `SwapchainImage` return `Undefined` rather than `PresentSrc` as it currently does? This way the call to `add_image_memory_barrier` on either a command buffer builder method or on the build method itself would insert the transition for us, right? 

Making this change also seems to fix the behaviour downstream, though it does not address the issue that the `undefined_layout` field is never changed after initialisation. Can/should this be set to `false` when a barrier is inserted that transitions the image layout to a variant that is not `Undefined`?

tomaka:
To be honest the reason why this is only partially implemented is that the whole thing is extremely complicated. Right now there are trade-offs between optimizations (don't have unnecessary barriers), user-friendliness, and correctness.

mitchmindtree:
For anyone else running into this, we're currently doing some investigation into a correct solution at nannou-org/nannou#271.

mitchmindtree:
Closing in favour of #1171 which is almost ready

PR #1125, opened 2018-11-30
Add a `union` method to the extensions types.

mitchmindtree:
Currently the extensions types implement `intersection` and `difference` methods but provide no simple way to produce a `union`. This PR addresses this by adding a `union` method.

* ~Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.~

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -2,6 +2,7 @@
 
 - Add support for `#include "..."` and `#include <...>` directives within source
   files.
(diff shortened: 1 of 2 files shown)


rukai:
Thanks for the PR! Can you rebase to fix the changelog conflict?

rukai:
The other methods have `#[inline]` I assume we also want that here?

Otherwise lgtm.

PR #1133, opened 2018-12-10
Implement `VertexMember` for cgmath and nalgebra vectors

aleksijuvani:

This change will allow the vector types from [rustgd/cgmath](https://github.com/rustgd/cgmath) to be used inside vertices in Vulkano. To avoid adding additional dependencies, this feature is optional and only enabled if the user enables the "cgmath" feature flag for Vulkano.


rukai:
I'll need to investigate this further, but here is my initial thoughts.

Despite being optional it still adds a maintenance overhead.
Also vulkano users may see this feature as meaning cgmath is preferred over other libraries.
But I think nalgebra-glm has replaced cgmath as cgmath is no longer well maintained.
I'm looking to swap my personal projects over to nalgebra-glm as cgmath pulls in outdated dependencies.

If I do merge something like this I'll want to also handle nalgebra-glm in the same way.

Is mint a possible solution to this problem?

aleksijuvani:
Yes, support for nalgebra (or nalgebra-glm) should be added as well. This PR is just initially providing cgmath support because that is what I'm personally using at the moment. It's not my intention to prefer any library over the other. Despite cgmath not being actively maintained, a lot of people are still using it and I think it's valuable to support. The maintenance burden of implementing `VertexMember` is negligible. It's literally a couple of lines.

I'm not familiar with mint. Is there any overhead in the conversions? If we implemented `VertexMember` for mint types instead, would the user be required to explicitly use mint types in their vertices, instead of cgmath or nalgebra types? If so, I would rather just `VertexMember` be implemented for nalgebra types directly.

aleksijuvani:
I've gone ahead and implemented this for nalgebra vectors also. Looks like they have the same data layout as cgmath (at least the lower dimensions do), so it was relatively straightforward. I didn't explore mint any further.

jmi2k:
This seems interesting, maybe it should be reconsidered. Having as much type information for as long as possible it's always nice, and it even gives more context to what kind of data it's being passed to the shaders.

PR #1028, opened 2018-08-27
doc: suggest brew on macOS

gibfahn:
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users
- [x] Updated documentation to reflect any user-facing changes

Having to hard-code the version is a bit of a pain, raised an issue in brew to discuss (https://github.com/Homebrew/homebrew-cask/issues/51330), but I'm not sure what the best solution is.

Issue #927, opened 2018-03-01
Could MoltenVK be automatically installed by Cargo?


mcclure:
Hello, I am new to both Rust and Vulkan. I tried out the Triangle example and it worked on Windows. Then I tried it on Mac and it failed because MoltenVK was not installed. Well of course not, I didn't install it.

But: Cargo was able to install all the *other* libraries it needed. I understand that there are some C based projects already in Cargo, for example libgit2 https://crates.io/crates/libgit2-sys/0.4.2 Would it be possible to make a Cargo package for MoltenVK, so that the requirement to externally install MoltenVK to use Vulkano could be waived, and we could get the versioning advantages of cargo for MoltenVK?

(I see there are a couple PRs right now related to this week's open source release of MoltenVK, but they seem to just be about making Vulkano work with an externally installed MoltenVK.)

mitchmindtree:
@freesig and I are also very interested in this! I think @freesig is going to have a go at automating  this process in a standalone crate first, then we'll probably do a PR here and get some feedback.

One thing that I'm guessing will be necessary will be for this to be optional. @rukai perhaps having an opt-in `moltenvk_installer` feature for this would be enough?

@freesig maybe before getting started it would be worth listing all the individual steps here that the `moltenvk_installer` crate would perform at build time so we can get some feedback?

rukai:
Yeah a feature sounds good, in the future we might also allow for gfx-rs as a backend via a similar feature.

freesig:
Oh just saw this. I already built it, it's pretty much just a curl with some bash commands then you set the environment variables.
I'm having some issues with curl though. I'll post it here once it's working properly for some feedback

freesig:
Just thought I'd post this here just incase anyone would like to use it. I will try and make a PR that integrates it into the build.rs aswell
https://crates.io/crates/moltenvk_deps

PR #1145, opened 2018-12-27
Mac install

freesig:

This PR automatically downloads and installs the latest Vulkan SDK from lunarg.
It will only do this if `VULKAN_SDK` is not set.
The sdk will be installed at `$HOME/.vulkan_sdk`
 and the following environment variables will be set in the users `.profile` or `.bash_profile`
```sh
export VULKAN_SDK=$HOME/.vulkan_sdk/macOS
export PATH=$VULKAN_SDK/bin:$PATH
export DYLD_LIBRARY_PATH=$VULKAN_SDK/lib:$DYLD_LIBRARY_PATH
export VK_ICD_FILENAMES=$VULKAN_SDK/etc/vulkan/icd.d/MoltenVK_icd.json
export VK_LAYER_PATH=$VULKAN_SDK/etc/vulkan/explicit_layer.d
```
The only thing that isn't automated is the user will get a error about failing to link the vulkan libs. This is because the changes to their profile are not automatically refreshed.
They will need to either `source ~/.profile` or `source ~/.bash_profile` or reopen their terminal after the first time they build vulkano.

If anyone knows how to automate this that would be super useful.
So far I have tried to pass the env var settings along from the build script to the lib using `cargo:rustc_env` but that only works if you are directly using Vulkano and not if it's a dependency.

One thing we could do is assume that if the env vars are not set when Vulkano is run then they must just be set to the defaults and but the user has not sourced their profile. Then just `std::env::set_var` to set them temporarily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ The easiest way to get vulkano up and running on macOS is to install the
 Vulkano will find it and dynamically link with `libvulkan.dylib`:
 
 Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
+The sdk will be installed at `$HOME/.vulkan_sdk`.
 If you would prefer a manual installation complete the following instructions before
 building vulkano.
 
--- a/README.md
+++ b/README.md
@@ -107,8 +107,10 @@ Vulkano will find it and dynamically link with `libvulkan.dylib`:
 Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
 The sdk will be installed at `$HOME/.vulkan_sdk`.
 If you get an error the first time you run vulkano then
-you will need to source your `source ~/.profile` or `source ~/.bash_profile`
+you will need to source your `source ~/.bash_profile`
 or simply reopen your terminal.
+If you are not using bash as your shell you will need to manually add
+the following environment variables.
 If you would prefer a manual installation complete the following instructions before
 building vulkano.
 
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ Vulkano will find it and dynamically link with `libvulkan.dylib`:
 
 Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
 The sdk will be installed at `$HOME/.vulkan_sdk`.
+If you get an error the first time you run vulkano then
+you will need to source your `source ~/.profile` or `source ~/.bash_profile` 
+or simply reopen your terminal.
 If you would prefer a manual installation complete the following instructions before
 building vulkano.
 
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Vulkano will find it and dynamically link with `libvulkan.dylib`:
 Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
 The sdk will be installed at `$HOME/.vulkan_sdk`.
 If you get an error the first time you run vulkano then
-you will need to source your `source ~/.profile` or `source ~/.bash_profile` 
+you will need to source your `source ~/.profile` or `source ~/.bash_profile`
 or simply reopen your terminal.
 If you would prefer a manual installation complete the following instructions before
 building vulkano.
--- a/README.md
+++ b/README.md
@@ -96,23 +96,29 @@ sudo apt-get install build-essential git python cmake libvulkan-dev vulkan-utils
 
 ### macOS and iOS Specific Setup
 
+TLDR; In the best case, vulkano should work on macOS without having to take any further action, read on if you need more information or something goes wrong.
+
 Vulkan is not natively supported by macOS and iOS. However, there exists [MoltenVK](https://github.com/KhronosGroup/MoltenVK)
 a Vulkan implementation on top of Apple's Metal API. This allows vulkano to build and run on macOS
 and iOS platforms.
 
-The easiest way to get vulkano up and running on macOS is to install the
-[Vulkan SDK for macOS](https://vulkan.lunarg.com/sdk/home). To install the SDK so that
-Vulkano will find it and dynamically link with `libvulkan.dylib`:
+The Vulkan SDK for macOS also comes with the iOS framework.
+
+Note that as of writing, MoltenVK has some bugs that show up in the examples.
+Some minor modifications may be required as workarounds: see https://github.com/vulkano-rs/vulkano/pull/1027.
+The examples also do not work properly on macOS 10.11 and lower without workarounds due to MoltenVK's Metal backend not getting
+the required features until macOS 10.12. See https://github.com/vulkano-rs/vulkano/issues/1075 for workarounds.
+
+#### macOS and iOS Automatic Setup
 
 Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
-The sdk will be installed at `$HOME/.vulkan_sdk`.
-If you get an error the first time you run vulkano then
-you will need to source your `source ~/.bash_profile`
-or simply reopen your terminal.
-If you are not using bash as your shell you will need to manually add
-the following environment variables.
-If you would prefer a manual installation complete the following instructions before
-building vulkano.
+The SDK will be installed at `$HOME/.vulkan_sdk`.
+If you are not using bash as your shell you will need to manually add the environment variables listed in step 2 of "macOS and iOS Manual Setup".
+
+#### macOS and iOS Manual Setup
+
+Follow these steps to manually install the [Vulkan SDK for macOS](https://vulkan.lunarg.com/sdk/home) so that
+Vulkano will find it and dynamically link with `libvulkan.dylib`.
 
 1. Download the latest macOS release and unpack it somewhere, for the next step
 we'll assume that's `~/vulkan_sdk`.
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ The easiest way to get vulkano up and running on macOS is to install the
 [Vulkan SDK for macOS](https://vulkan.lunarg.com/sdk/home). To install the SDK so that
 Vulkano will find it and dynamically link with `libvulkan.dylib`:
 
+Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
+If you would prefer a manual installation complete the following instructions before
+building vulkano.
+
 1. Download the latest macOS release and unpack it somewhere, for the next step
 we'll assume that's `~/vulkan_sdk`.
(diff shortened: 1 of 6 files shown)


mitchmindtree:
cc @mcclure this is an attempt at what you were after in #927 - any chance you're still interested and would like to review/test this branch? No worries if you're no longer interested, that issue is pretty old :)

mitchmindtree:
Here's the [moltenvk_deps src](https://github.com/freesig/moltenvk_deps/blob/master/src/lib.rs#L117) in case anyone would like to see what it's doing in detail.

mitchmindtree:
Hey I just tried running this on my dad's MBP!

It almost worked fine, however the env vars did not load automatically when I closed and re-opened the terminal. However, if I manually did `source ~/.profile` it worked fine afterwards. Perhaps `.profile` is not `source`d automatically when bash starts up? Maybe we're better off putting these env vars in `.bash_profile`?

freesig:
Yeh I think you're right I might remove the .profile stuff and just have a .bash_profile. It means we only support bash but that is the default on macos since 10.1 and I'll put a note in the readme for other shells.

freesig:
Ok I've made those changes, you may need to run cargo update before testing it.

README.md
```diff
@@ -104,6 +104,16 @@ The easiest way to get vulkano up and running on macOS is to install the
...
+Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
+The sdk will be installed at `$HOME/.vulkan_sdk`.
+If you get an error the first time you run vulkano then
+you will need to source your `source ~/.bash_profile`
+or simply reopen your terminal.
+If you are not using bash as your shell you will need to manually add
+the following environment variables.
+If you would prefer a manual installation complete the following instructions before
+building vulkano.
+
```
rukai: I think the readme should be restructured a bit more:
```
### macOS and iOS Specific Setup

TLDR; In the best case, vulkano should work on macOS without having to take any further action, read on if you need more information or something goes wrong.

Vulkan is not natively supported by macOS and iOS. However, there exists [MoltenVK](https://github.com/KhronosGroup/MoltenVK)
a Vulkan implementation on top of Apple's Metal API. This allows vulkano to build and run on macOS
and iOS platforms.

The Vulkan SDK for macOS also comes with the iOS framework.

Note that as of writing, MoltenVK has some bugs that show up in the examples.
Some minor modifications may be required as workarounds: see https://github.com/vulkano-rs/vulkano/pull/1027.
The examples also do not work properly on macOS 10.11 and lower without workarounds due to MoltenVK's Metal backend not getting
the required features until macOS 10.12. See https://github.com/vulkano-rs/vulkano/issues/1075 for workarounds.

#### macOS and iOS Automatic Setup

Vulkano will attempt to automatically download and install the latest Vulkan SDK for macOS.
The SDK will be installed at `$HOME/.vulkan_sdk`.
If you get an error the first time you run vulkano then you will need to run `source ~/.bash_profile`" or simply reopen your terminal.
If you are not using bash as your shell you will need to manually add the environment variables listed in step 2 of "macOS and iOS Manual Setup".

#### macOS and iOS Manual Setup

Follow these steps to manually install the [Vulkan SDK for macOS](https://vulkan.lunarg.com/sdk/home) so that
Vulkano will find it and dynamically link with `libvulkan.dylib`.

1. Download the latest macOS release and unpack it somewhere, for the next step
we'll assume that's `~/vulkan_sdk`.
2. Modify your environment to contain the SDK bin directory in PATH and the SDK lib directory in
DYLD_LIBRARY_PATH. We also need to set VK_ICD_FILENAMES and VK_LAYER_PATH. When using the Bash
shell, which is the default for macOS, it's easiest to do this by appending the following to the
`~/.bash_profile` file and then restarting the terminal.

```sh
export VULKAN_SDK=$HOME/vulkan_sdk/macOS
export PATH=$VULKAN_SDK/bin:$PATH
export DYLD_LIBRARY_PATH=$VULKAN_SDK/lib:$DYLD_LIBRARY_PATH
export VK_ICD_FILENAMES=$VULKAN_SDK/etc/vulkan/icd.d/MoltenVK_icd.json
export VK_LAYER_PATH=$VULKAN_SDK/etc/vulkan/explicit_layer.d
```a

It is also possible to link with the MoltenVK framework (as vulkano did in previous versions) by adding the
appropriate cargo output lines to your build script and implementing your own
`vulkano::instance::loader::Loader` that calls the MoltenVK `vkGetInstanceProcAddr` implementation.
```

Another concern is that previously, both the developer and the user need to install the SDK. Now the developer has automated install but the user does not. Is there a way to make shipping the Vulkan SDK with your application easy. Basically a runtime version of this? Maybe we should just recommend that the application call `moltenvk_deps` at runtime? Probably want a dialog box/loading window too so it doesn't look like the program isn't opening?
This is very much something for a separate PR, but its something to think about because this PR makes it easy for the developer but potentially confuses them when a shipped app doesn't work on a users machine.

Other than that, I am happy with this PR.
Note that I havent tested this PR because I dont have a mac.
I will just point people to file an issue to your repo when something goes wrong :)
Very little maintenance for vulkano.
Thankyou for your work! 🎉

rukai:
There is also this PR which wanted to make homebrew the recommend way to install vulkan sdk https://github.com/vulkano-rs/vulkano/pull/1028
My understanding is this would automatically handle updates. Is that something this PR can do?

mitchmindtree:
@rukai I think recommending homebrew is a little controversial on macos as [users are often split between macports and homebrew](https://stackoverflow.com/questions/21374366/what-is-the-difference-usage-of-homebrew-macports-or-other-package-installation). It's also worth noting neither are installed by default, whereas this PR just uses default mac tooling which should already exist if they have rust setup.

I guess it wouldn't hurt to mention it as an option in the case that the user already has homebrew installed?

rukai:
I think adding a mention of macports and homebrew would just further complicate an already large readme without much gain.

What we need is a way to update the vulkan sdk installed by vulkano.
Currently `moltenvk_deps` will install the vulkan sdk if `VULKAN_SDK` is not set.
However we should also install if `VULKAN_SDK` is set to the value that `moltenvk_deps` sets it to and the path points to a directory that does not exist.
This way the user can update by just deleting `~/.vulkan_sdk`.
This will also need to be documented in the readme.

freesig:
I would hesitate to use homebrew to install the sdk as the user would need to call `homebrew update` to get the new sdk anyway and it just adds another dependency.
I like the idea of moving the install to runtime. I think it should be possible to check the version is the latest with moltenvk_deps but I'm not sure all users would want to automatically update. So maybe it's better to just print a notice saying it's out of date with an option to update.

I'll make these changes and see what I can come up with.

freesig:
I'm just thinking through running the install at runtime. 
Is it possible to build for macos from another os due to all the vulkan dependencies?
If so we might need to check for the dependencies at compile time and runtime.

freesig:
I'm stuck on automatic updates because I cannot find anyway to reliably get the latest version of the vulkan_sdk without downloading it.

Additionally I can't rely on the user deleting `~/.vulkan_sdk` because they may have chosen to install it manually in a different location.
I cannot get them to unset `VULKAN_SDK` because that is the case when they haven't source'd bash anyway.

We could however get them to set `UPDATE_VULKAN_SDK=1` temporarily. This would then trigger an update. So the command would be used like:
```
UPDATE_VULKAN_SDK=1 cargo run
```
Is this an appropriate solution?

rukai:
So earlier I said "However we should also install if VULKAN_SDK is set to the value that moltenvk_deps sets it to and the path points to a directory that does not exist."
Because I figured that we dont want vulkano touching anything if the user has already, at least partially, setup their system. (just to be on the safe side)
However we could definitely change this to just "However we should also install if VULKAN_SDK points to a directory that does not exist."

Then we don't need the `UPDATE_VULKAN_SDK=1` which is imo less intuitive.
Although being able to use as `UPDATE_VULKAN_SDK=1 cargo run` isnt too bad. Its just annoying to have too lookup what the magic env var is when you need to update. Its easier to just remember to delete the vulkan directory.

freesig:
Yeh good call, that's a lot simpler.
I've made it so it will update if the `VULKAN_SDK` is set to the default directory and that directory is empty as you said. (I agree it's probably not good to touch anything if the user has manually set it)
So all you need to do now is remove the `~/.vulkan_sdk` directory to update the sdk to latest.

mitchmindtree:
Hmm now it seems that although linux is building fine, macos is error-ing [on a timeout in the tests](https://travis-ci.org/vulkano-rs/vulkano/jobs/484728917#L424). @freesig are you able to run the tests locally on macos or do they timeout for you too? I wonder if something can be done about the culprit test.

freesig:
Yeh this is a little weird. If the test was triggering a moltenvk_deps to ask for a sdk install then there would be text in the log. It can't try and download without the users permission. I will run the test locally and report back.

freesig:
I get different failures on my machine
```
    memory::device_memory::tests::oom_single
    pipeline::compute_pipeline::tests::spec_constants
    sync::event::tests::event_create
    sync::event::tests::event_pool
    sync::event::tests::event_reset
    sync::event::tests::event_set
```
I'll keep digging. I wonder which version of the sdk travis is using.

freesig:
I just remembered that cargo test captures output. I think it might be detecting that the sdk isn't installed and asking the user if it would like to install it.
I'm not sure the best way to get around this. I could time out, just not install it if the user doesn't respond in a certain time, although that's not great

mitchmindtree:
I would personally be fine without any command-line user prompt whatsoever and just going ahead and installing if we can't find an existing installation. I think it would be better to have a function/method/parameter of some sort which allows a developer to opt out of auto-installation in their code in case they want to run their own GUI installer or something. I'd imagine most downstream games would want to go ahead and install vulkan or provide their own GUI installer rather than have a terminal open up for the user to answer?

freesig:
Yeh I see what you mean but also its ~125mb file and the user might not be in the best position to start a download (bad net connection etc.) and would rather continue the install themselves later or perhaps even manually. Not everyone would be comfortable with an automated install and without a prompt there's not really any good way to warn them.

Perhaps I could make a function that in the vulkano api for the developer that's using vulkano to manually create a check for the dependencies before the instance is loaded (and us that in their gui). (although I'm not sure how easy this would be as the instance get's the library path with `lazy_static`

freesig:
I managed to get this working on the current vulkano master so I will close this PR now. But I'm happy to make another PR if you would like instructions for how to use moltenvk_deps in your readme.

Issue #1137, opened 2018-12-13
Relative shader includes not working / shader-include example fails to compile on Windows


aleksijuvani:
*   Version of vulkano: 9a08414054fadfc534abfb63102da04b6446f877
*   Version of rustc: 1.31.0 (abe02cefd 2018-12-04)
*   OS: Windows 10 Version 1803
*   GPU (the selected PhysicalDevice): NVIDIA GeForce GTX 960
*   GPU Driver: 397.31

# Issue

```
    Checking examples v0.1.0 (C:\Users\flaradr_ljos\Development\vulkano\examples)
error: proc macro panicked
  --> examples\src\bin\shader-include\main.rs:37:13
   |
37 | /             vulkano_shaders::shader!{
38 | |                 ty: "compute",
39 | |                 // We declare what directories to search for when using the `#include <...>`
40 | |                 // syntax. Specified directories have descending priorities based on their order.
...  |
60 | | }"
61 | |             }
   | |_____________^
   |
   = help: message: called `Result::unwrap()` on an `Err` value: "4 compilation error(s): shader.glsl: error: \\\\?\\C:\\Users\\flaradr_ljos\\Development\\vulkano\\examples\\src\\bin\\shader-include\\standard-shaders\\common.glsl:6: \'#include\' : Invalid inclusion path `\\\\?\\C:\\Users\\flaradr_ljos\\Development\\vulkano\\examples\\src\\bin\\shader-include\\standard-shaders\\../relative-shaders/relative-include.glsl`. for header name: ../relative-shaders/relative-include.glsl\nshader.glsl: error: \\\\?\\C:\\Users\\flaradr_ljos\\Development\\vulkano\\examples\\src\\bin\\shader-include\\standard-shaders\\common.glsl:9: \'multiply_by_2\' : no matching overloaded function found\nshader.glsl: error: \\\\?\\C:\\Users\\flaradr_ljos\\Development\\vulkano\\examples\\src\\bin\\shader-include\\standard-shaders\\common.glsl:9: \'multiply_by_3\' : no matching overloaded function found\nshader.glsl: error: \\\\?\\C:\\Users\\flaradr_ljos\\Development\\vulkano\\examples\\src\\bin\\shader-include\\standard-shaders\\common.glsl:9: \'return\' : type does not match, or is not convertible to, the function\'s return type\n"

error: aborting due to previous error

error: Could not compile `examples`.
```

aleksijuvani:
[...]

PR #1148, opened 2019-01-05
Run CI on windows

rukai:

This will prevent issues like https://github.com/vulkano-rs/vulkano/issues/1137 from occurring in the future.

This will be an iterative PR as I will likely need to add C++ deps for building shaderc-rs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: xenial
 os:
   - linux
   - osx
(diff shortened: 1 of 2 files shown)


rukai:
Oh wow, looks like everything "just works".
However I do need to disable the failing example reported in https://github.com/vulkano-rs/vulkano/issues/1137.

PR #1152, opened 2019-01-13
Mutable descriptor sets

zedrian:


CHANGELOG_VULKANO.md
```diff
@@ -5,6 +5,8 @@
 - Add a `union` method for the extensions types.
 - Make `BufferUsage` derive `PartialEq` and `Eq`
 - Fix a compiler error in struct code generated by the `shader!` macro.
+- Add `inner_mut` method for `FixedSizeDescriptorSet` and `PersistentDescriptorSet` providing mutable access to inner `UnsafeDescriptorSet`.
+- Add `empty()` method for `FixedSizeDescriptorSet` and `next_empty()` method for `PersistentDescriptorSet` that construct descriptor sets without any bound descriptors.
```
mitchmindtree: Out of curiosity, what is the use case for these `empty()` methods?

zedrian: The idea is to divide two different parts of logic: the first one controls object construction (such as descriptor set), and the second one controls which resources will be bound to the descriptor set (which exactly buffers, images and so on).
I work with computational shaders and for my needs it's much more convenient to modify existing descriptor sets instead of creating new ones.

faulesocke:
Isn't this what descriptor set pools are there for?

zedrian:
> Isn't this what descriptor set pools are there for?

AFAIK, the "right" way to construct new descriptor set using a descriptor set pool is to pass all resources that you want being bound to future `DescriptorSet` via chained `.next().add_*()....build()` call. 

But what if some slots should hold the same resources as in previous render call, while others should be re-bond? Something like this:
```rust
renderer.bind_image(0, atlas);
renderer.bind_buffer(1, camera_buffer);
renderer.bind_buffer(2, renderables_buffer);

render();

if should_modify_gamma {
    renderer.bind_image(0, modified_gamma_atlas);
}

render();
```
It seems that your resource wrapper (that holds descriptor set pool inside) should have some cache of resources (references?) that should be bond to new `DescriptorSet` - and update such references using calls like in example above.

But **Vulkan** itself already have such caching mechanism - `DescriptorWrite` objects (`DescriptorCopy` are also cool, but not implemented in Vulkano yet).

So if you have a mutable access to `UnsafeDescriptorSet` (like in this PR), you can update existing `DescriptorSet` by re-binding the only resources that differ from previous state.

faulesocke:
Ah, now I see where this is going.

zedrian:
@mitchmindtree, @rukai - is this motivation still insufficient?

> > Isn't this what descriptor set pools are there for?
> 
> AFAIK, the "right" way to construct new descriptor set using a descriptor set pool is to pass all resources that you want being bound to future `DescriptorSet` via chained `.next().add_*()....build()` call.
> 
> But what if some slots should hold the same resources as in previous render call, while others should be re-bond? Something like this:
> 
> ```rust
> renderer.bind_image(0, atlas);
> renderer.bind_buffer(1, camera_buffer);
> renderer.bind_buffer(2, renderables_buffer);
> 
> render();
> 
> if should_modify_gamma {
>     renderer.bind_image(0, modified_gamma_atlas);
> }
> 
> render();
> ```
> 
> It seems that your resource wrapper (that holds descriptor set pool inside) should have some cache of resources (references?) that should be bond to new `DescriptorSet` - and update such references using calls like in example above.
> 
> But **Vulkan** itself already have such caching mechanism - `DescriptorWrite` objects (`DescriptorCopy` are also cool, but not implemented in Vulkano yet).
> 
> So if you have a mutable access to `UnsafeDescriptorSet` (like in this PR), you can update existing `DescriptorSet` by re-binding the only resources that differ from previous state.

AustinJ235:
Still trying to wrap my head around the advantage of this versus the existing method of just using set pools and creating a new set every frame. Is there some great performance advantage to this?

zedrian:
I'm using this method in my pet project with computational shaders for more than two years (at first in C++, then migrated to Rust) and after all this time I even don't remember why I actually chose such approach. Now I tried to find some arguments for that in official Vulkan specification, but it seems that creating new descriptor set is more efficient than modifying the existing one...
I'll come back to this PR after some performance checks.

zedrian:
Okay, even without special performance checks, let's track the workflow.

Existing Vulkano way to get a descriptor set with required resources:
1. Ask descriptor set pool for available descriptor set (create new descriptor set or reuse some that was used in previous frame and then returned back to pool)
2. Call `UnsafeDescriptorSet::write()` with list of `DescriptorWrite` structures collected by descriptor set builder.

Way that is proposed by this PR:
Immediately go to step 2 from previous flow - but in quite unsafe way, because it becomes a user responsibility to provide correct list of `DescriptorWrite` structures.

Is this new way more efficient? Yes, a bit (because user explicitly provides which descriptor set object to use). Is this new way safe? It isn't, because it allows user to call `UnsafeDescriptorSet::write()` method that is `unsafe`. The question is: is it okay?

othelarian:
@zedrian you're workflow is a great solution for optimization, but one of the most important keyword associated with rust is "safety", so including an unsafe workflow seems paradoxal. There is possibility to define options in Cargo.toml, maybe it's the right way to let people choose, between classic safe and optimized unsafe workflow?

zedrian:
@othelarian you are right, defining such functionality in a separate future is really a good solution, thank you!
I'll update the PR later.

AustinJ235:
I am not so keen on the idea of separation. Surely exposing unsafe functions I am not opposed to. The main issue with this code is that it is sorta redundant until it is proven that is a faster means to the solution. Currently ``CpuBufferPool`` handles rapidly changing descriptor sets. While partially modifying it does seem like it would be faster, real world performance gains haven't been tested or shown. If there are some tests or users saying that this means is a faster alternative then ``CpuBufferPool``'s then I'll be happy to merge, otherwise it is just extra code in vulkano that needs to be maintained.

zedrian:
@AustinJ235 it seems that I'm a single user of such functionality (possibly due to the specifics of my Vulkan usage). If so, it really doesn't make sense adding extra code that is unsafe, used by only one person, but requires maintaining. You are right.
For now, I'm closing this PR. If I find a solution for mutable descriptor sets that will be safe and useful for more people, I'll come back with another PR.

PR #1160, opened 2019-01-29
Add support for VK_KHR_16bit_storage and VK_KHR_storage_buffer_storage_class

pablode:
This pull request introduces six new SPIR-V capabilities and one new SPIR-V storage class according to [SPV_KHR_16bit_storage](https://www.khronos.org/registry/spir-v/extensions/KHR/SPV_KHR_16bit_storage.html) and [SPV_KHR_storage_buffer_storage_class](https://www.khronos.org/registry/spir-v/extensions/KHR/SPV_KHR_storage_buffer_storage_class.html).

Additionally, it allows them to require the corresponding Vulkan device extensions [VK_KHR_16bit_storage](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_KHR_16bit_storage) and [VK_KHR_storage_buffer_storage_class](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_KHR_storage_buffer_storage_class). New definitions have been added to vk-sys.

The guide does not need to be updated.


rukai:
I haven't reviewed anything yet, but I tried testing your changes and get the following warnings:
```warning: unreachable pattern
   --> vulkano-shaders/src/enums.rs:27:29
    |
27  |                               $value => Ok($name::$elem),
    |                               ^^^^^^
...
39  | / enumeration! {
40  | |     typedef enum SourceLanguage_ {
41  | |         SourceLanguageUnknown = 0,
42  | |         SourceLanguageESSL = 1,
...   |
556 | |     } Capability;
557 | | }
    | |_- in this macro invocation
    |
    = note: #[warn(unreachable_patterns)] on by default

warning: unreachable pattern
   --> vulkano-shaders/src/enums.rs:27:29
    |
27  |                               $value => Ok($name::$elem),
    |                               ^^^^^^
...
39  | / enumeration! {
40  | |     typedef enum SourceLanguage_ {
41  | |         SourceLanguageUnknown = 0,
42  | |         SourceLanguageESSL = 1,
...   |
556 | |     } Capability;
557 | | }
    | |_- in this macro invocation
```

vulkano-shaders/src/enums.rs
```diff
@@ -546,5 +547,11 @@ enumeration! {
         CapabilityStorageImageReadWithoutFormat = 55,
         CapabilityStorageImageWriteWithoutFormat = 56,
         CapabilityMultiViewport = 57,
+        CapabilityStorageUniformBufferBlock16 = 4433,
+        CapabilityStorageBuffer16BitAccess = 4433,
+        CapabilityStorageUniform16 = 4434,
```
knappador: Related to the compiler warnings around the shader macro, I noticed these duplicate values and that at least one other project has chosen to deal with them by treating them as aliases and removing half:
https://github.com/google/clspv/blob/master/cmake/spirv_c_strings.cmakescript#L39-L41

For our API it may be sufficient to support only one (or use associated constants)?   The Khronos hpp includes the duplication but C++ can't detect this anyway.

This SO post shows the associated constants route, which may not allow match arms to work.  Evidently rust dropped the unique requirement since we're only getting warnings? 
https://stackoverflow.com/questions/38025330/multiple-enum-variants-with-same-value

pablode:
I've removed the duplicates which resolves the warning.

We are trying to the deduce the capability from a number, which is ambiguous in this case. Having two enum entries with the same value does not make this possible.

Since the capabilities `StorageBuffer16BitAccess` and `UniformAndStorageBuffer16BitAccess` have the same values as `StorageUniformBufferBlock16` and `StorageUniform16` but result from extension interaction, they have been removed in favor of the more basic definitions.

knappador:
I'm surprised to see `4434` get deleted, but it sounds like it was mapping to the same value in Vulkano, causing one of the warnings?  Since Rust is saner about enums, I'd also be okay with adding a _semantic_ duplicate like `StorageBuffer16BitAccess2` or something if it gets us broader platform support.  Other than that, Changelog update and then you should be good to go.  I don't have merge permission or I would merge at that point.  **Overtly lobbying for organization membership**

PR #1161, opened 2019-01-31
Make buffer::sys::UnsafeBuffer::new panic if the usage parameter is empty

CephalonRho:
This PR adds an assertion in buffer::sys::UnsafeBuffer::new to catch invalid usage of the Vulkan API.
It's a small addition, but it could still cause new errors so I have added a line to the changelog.
Changing the guide shouldn't be necessary.


rukai:
Thankyou!
This is an important safety fix.

PR #1164, opened 2019-02-05
Make `BufferUsage` derive `PartialEq` and `Eq`

Limeth:

A small quality of life improvement.


rukai:
When would you want to check for equality of BufferUsage?

Limeth:
In my case, I enable the specific usage fields depending on how a buffer is used within glTF. The `BufferUsage` struct is used like a builder and then it's checked whether a buffer with such usage should even be used. See https://github.com/metaview-org/ammolite/blob/master/src/model/import.rs#L262-L264

rukai:
Thanks!

PR #1166, opened 2019-02-07
Fix compiler error in code generated for GLSL struct

mitchmindtree:
I was running into a compiler error generated by the `shader!` macro
while writing a compute shader for a downstream commercial project.
Using `cargo-expand` I was able to find the generated code for the
struct and noticed that the current implementation would always generate
an invalid type. *`{}` is never a valid length for a fixed size array of
bytes.*

It seemed to me that `{}` was used within the `quote!` macro
accidentally rather than `#diff` as this particular field is used to
specify a padding amount for matching the memory layout of the struct.
By changing the fixed size array len to use `#diff` I was able to
complete compilation successfully and my code runs as expected.

@tomaka can you confirm that this was likely a bug and that this is the
correct fix? I want to check that this was not an intentional compile
error pending future work or a better error message.

* [ ] ~Updated documentation to reflect any user-facing changes - in this repository~


rukai:
Thanks for the PR!
Your change is correct.
If you check the history of this file, I messed up when converting the shader! macro to use syn instead of raw strings.
https://github.com/vulkano-rs/vulkano/commit/d779829cbd46909df137daa31fa11f6dd56f6947#diff-138e56c5781e74b3af091b197fbff71a

PR #1172, opened 2019-02-21
Propagate newlines correctly on error messages

### LINKED ISSUE #345: Format SPIR-V Cross-Compilation Errors
The errors given back from the SPIR compiler contain `\n` sequences as opposed to actual newlines. It would be a minor improvement to have them replaced with newlines to make reading the errors easier. 

Current example error:
```
/tmp/glslang-compile.3l8vmVJ0cS8t/0.vert\nERROR: #version: ES shaders for Vulkan SPIR-V require version 310 or higher\nWarning, version 310 is not yet complete; most version-specific features are present, but some are missing.\nERROR: 1 compilation errors.  No code generated.\n\n\n\nLinked vertex stage:\n\nERROR: Linking vertex stage: Missing entry point: Each stage requires one entry point\n\nSPIR-V is not generated for failed compile or link\n
```

Preferred example error:
```
/tmp/glslang-compile.3l8vmVJ0cS8t/0.vert
ERROR: #version: ES shaders for Vulkan SPIR-V require version 310 or higher
Warning, version 310 is not yet complete; most version-specific features are present, but some are missing.
ERROR: 1 compilation errors.  No code generated.



Linked vertex stage:

ERROR: Linking vertex stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link
```

jsarasin:
This resolves #345 for me. It's not ideal as I think the fix can happen upstream in shaderc, however this provides a more mature way to deal with error output. If it's appropriate or there is interest I can add color/bold terminal codes to make the error output more clear.

I'm a bit confused exactly what is happening here. I believe the issue is in shaderc. This is also bit of a hacky fix to correct the output of shaderc so there is a newline after the initial `message: 22 compilation error(s):` 

The newline characters are outputting on my Linux terminal, but they're not actually causing a newline to happen? I tried it every which way and this is the simplest solution which works for me.

When I have more time I will try and fix this upstream in shaderc.


knappador:
Reading up on:
https://github.com/vulkano-rs/vulkano/issues/345

Binearo says that they cannot reproduce the issue.  Is this still alive or should we close the PR as fixed in upstream?

knappador:
I've reproduced this since several times.  Especially when there are 3-4 error messages, it's hard to see through.  Newlines are coming out escaped I believe.

AustinJ235:
Is this still a thing and does this patch still work?

jsarasin:
haven't tested in awhile. I'll check in a couple days

PR #1175, opened 2019-03-07
Add function execute_commands_from_vec to accept multiple command buffers from a vector.

WarpspeedSCP:


knappador:
Changes look good.

Is there context for supporting N command buffers vs single command buffers?

On the side, I'm asking on Rust IRC if there's a general way to implement coercion form singular to lists for situations where it makes semantic sense to have both signatures.

WarpspeedSCP:
I have a use case where I want to batch renders with multiple secondary command buffers, so this addresses that. The vulkan API itself does support batching like this, the signature of `vkCmdExecuteCommands` looks like

```c++
void vkCmdExecuteCommands(
    VkCommandBuffer                             commandBuffer,
    uint32_t                                    commandBufferCount,
    const VkCommandBuffer*                      pCommandBuffers);
``` 

I ran into some weird errors when I tried to make multiple calls to the old version of `AutoCommandBufferBuilder::execute_commands`; maybe because it was making multiple calls to `vkCmdExecuteCommands`. I don't exactly understand the problem, but it seems like this might be a good idea any way to conform better to the underlying API.

WarpspeedSCP:
Currently it is a bad idea to make multiple calls to `AutoCommandBufferBuilder::execute_commands`, because doing so will try to re-add command buffers added in previous calls to it.

WarpspeedSCP:
wait, that's a bug.

vulkano/src/command_buffer/auto.rs
```diff
@@ -1244,13 +1244,15 @@ impl<P> AutoCommandBufferBuilder<P> {
     /// **This function is unsafe for now because safety checks and synchronization are not
     /// implemented.**
     // TODO: implement correctly
-    pub unsafe fn execute_commands<C>(mut self, command_buffer: C)
+    pub unsafe fn execute_commands<C>(mut self, command_buffers: Vec<C>)
                                       -> Result<Self, ExecuteCommandsError>
         where C: CommandBuffer + Send + Sync + 'static
     {
         {
             let mut builder = self.inner.execute_commands();
```
knappador: @WarpspeedSCP   suspect side-effects behind this call

knappador:
@rukai  I'm lobbying for organization membership and merge authorization from @tomaka  so that I can push buttons instead of just roleplaying project manager :man_cartwheeling: 

@WarpspeedSCP since I can't actually merge, I'll pretend to request changes

[ ]  Root cause the accumulation of buffers in the builder before changing API
[ ]  In line with your argument that Vulkan is designed to work with multiple buffers, an API shortcut for vector vs scalar secondary buffers is warranted.  The most likely use case of secondary buffers is to create more than one.  Recommend `execute_buffers` or similar to avoid unnecessary API breaking change.
[ ]  Calling rustfmt was a bit heavy handed over one piece of whitespace.  Recommend rebasing that out to avoid merge hazards with other branches.  (in general I prefer picking a semver and doing whitespace changes after adopting a standard such as in https://github.com/vulkano-rs/vulkano/pull/1080)

vulkano/src/command_buffer/auto.rs
WarpspeedSCP: We could create a temporary buffer to hold the newly added command buffers and run the API call with the temporary buffer instead of the internal one, then append temp buf to internal buf... Does that sound right?

WarpspeedSCP:
Ah I didn't notice that rustfmt changed the formatting of the entire document. I'll revert that and fix my formatting.

vulkano/src/command_buffer/auto.rs
knappador: > vulkano also allows you to _use_ alternative command pool implementations _and use them_

We are on the right track! :train2: :boom: 

> create a temporary buffer to hold the newly added command buffers and run the API call with 
the temporary buffer instead of the internal one, then append temp buf to internal buf

I can imagine this working, but I am afraid that whatever is breaking may still become unpredictable in successive calls to `execute_commands`.

## Documenting my trip into the call

* Starting in the `AutoCommandBufferBuilder` that [returns](https://github.com/vulkano-rs/vulkano/blob/671c1451ea43f146498a18ca8891a48b1ddb0014/vulkano/src/command_buffer/auto.rs#L1248) `Self` after presumably adding the secondary buffer somewhere
*  Creates a builder of `Sync..ExecuteCommands` via call to `inner`, a `SyncCommandBufferBuilder<StandardCommandPoolBuilder>`.  The returned `Sync..ExecuteCommands` struct references the `SyncCommandBufferBuilder`, and no mutation has been performed so far.
*  Invalidation of the `state_cacher` appears not to touch any commands and is assumed independent for now
*  Our `Sync..ExecuteCommands` struct `add` and `submit` are reported by @WarpspeedSCP to cause duplication of commands whenever `add` and `submit` are both called on each secondary buffer but not if only `add` is called per buffer followed by a final `submit`.  It's unknown presently if subsequent calls to `submit` will again exhibit duplication.

### Add
We [add the commands](https://github.com/vulkano-rs/vulkano/blob/71d90e562a63f2a9aee40abaa3b59949e909015d/vulkano/src/command_buffer/synced/commands.rs#L2155-L2157) to an unsafe `inner` and to `command_buffers`.   Both of these were just created and are empty.

### [Submit](https://github.com/vulkano-rs/vulkano/blob/71d90e562a63f2a9aee40abaa3b59949e909015d/vulkano/src/command_buffer/synced/commands.rs#L2187-L2190)
`SyncCommandBufferBuilder` `append_command`'s a struct that contains our unsafe command and our safe `command_buffers`

Since the `..ExecuteCommands` are all created with [fresh buffers](https://github.com/vulkano-rs/vulkano/blob/71d90e562a63f2a9aee40abaa3b59949e909015d/vulkano/src/command_buffer/synced/commands.rs#L1365), and `append_commands` doesn't itself explain duplication, I think we can guess that a later phase contains the bug and any combination of the higher up API that results in appending multiple [`SyncCommandBufferBuilderExecuteCommand::Cmd`](https://github.com/vulkano-rs/vulkano/blob/71d90e562a63f2a9aee40abaa3b59949e909015d/vulkano/src/command_buffer/synced/commands.rs#L2162) is going to duplicate some commands.

`append_command` itself [looks benign](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L434)

### Build

At length, unsent commands have [`send`](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L699) and [`into_final_command`](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/base.rs#L747) called.  In our case, `SyncCommandBufferBuilderExecuteCommand` will call `UnsafeCommandBufferBuilder::execute_commands` [here](https://github.com/vulkano-rs/vulkano/blob/71d90e562a63f2a9aee40abaa3b59949e909015d/vulkano/src/command_buffer/synced/commands.rs#L2173)

### Send

The [`UnsafeCommandBufferBuilder::execute_commands`](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/sys.rs#L1176) is straightforward.

[`UnsafeCommandBufferBuilderExecuteCommands`](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/sys.rs#L1584-L1611) doesn't appear to accumulate commands.

>  because when I add a command buffer via execute_commands, it adds the command buffer to a vector and then calls vkCmdExecuteCommands once.

I intended to familiarize myself enough to review the effectiveness of changes.  However, @WarpspeedSCP, I haven't so fare seen the mechanism of accumulating commands.  Specifically the call to `SyncCommandBufferBuilder::execute_commands` returns a fresh builder that itself has a fresh `UnsafeCommandBufferBuilderExecuteCommands` which again has a fresh `SmallVec` and I've traced these to the [call](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/sys.rs#L1183) to `vkCmdExecuteCommands`. Since your statement sounds like you've seen where the commands are accumulating, perhaps you already know where the duplication is occurring?

WarpspeedSCP: I think I've been jumping to conclusions regarding this. When I encountered the problem, I looked into it a little and based on what I saw of the add method, changed the signature of `execute_commands`. 

I've been looking back on this, and I've even tried using the old version of execute_commands again, and it seems that the behaviour I got was caused by some manner of synchronisation error on my part, not vulkano's. I've rerun my code with the old version and now it works, so it seems I've been making mountains out of molehills.

WarpspeedSCP: I dug into the command buffer code, and what I've found is that there is definitely no duplication going on. all previously recorded commands are dropped between calls to `execute_commands`

knappador: @WarpspeedSCP   no worries.  I was going to have to learn the source to start working with it productively.

On the one hand, adding to the API for `execute_buffers` might still make sense if there's an obvious reason that multiple secondary buffers is the norm for each `vkCmdExecuteBuffers` call.  I'm curious if the docs would say we're missing any functionality by not having multiple buffers per `vkCmdExecuteBuffers` call.  I think that trivia basically decides if we still want to pursue things on the API side.

knappador:
Recording some trivia around secondary command buffers:

https://vulkan.lunarg.com/doc/view/1.0.37.0/linux/vkspec.chunked/ch05s06.html
> Once vkCmdExecuteCommands has been called, any prior executions of the secondary command buffers specified by pCommandBuffers in any other primary command buffer become invalidated, unless those secondary command buffers were recorded with VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT.

I'll check on the docs later to be sure we have this mentioned since it might work but would cause a very horrible user experience.  We probably shouldn't allow it because it's so silly.

Asked about multiple vs single secondary command buffers per call to `vkCmdExecuteCommands`.  Let's see what  internet comes up with and then I'll try to recommend some path forward.

vulkano/src/command_buffer/auto.rs
WarpspeedSCP: It's probably less efficient to make a call to vkCmdExecuteCommands every time we want to add a command buffer, and the spec doesn't say anything about how to batch secondary command buffers. I'm inclined to assume they want us to record multiple secondary commands at a time.

knappador: @WarpspeedSCP I know nobody is going to say, "You should never send multiple buffers to that signature that accepts an array of buffers" and I buy the efficiency argument.  My followup about behavior will apply equally to the current API or the prospective multiple-buffer signature. 

Recommend implementing `execute_buffers` and leaving alone poor souls who used `execute_commands`.   The modification you made is correct if only it was a new function.  Re-use of inner command struct smart.  Just update docs & changelog for the new API call and I will pretend to merge the PR until @rukai or @tomaka can make my dreams come true.

WarpspeedSCP: I'll get to that then.

knappador:
:+1:  nominating for merge.  @rukai

Vulkan subreddit TL;DR is no opinions opposed to multiple submissions in our API.

https://www.reddit.com/r/vulkan/comments/b0w0iy/secondary_command_buffers_vkcmdexecutecommands/

:up:  and to the :arrow_right: 
:o: :outbox_tray:

vulkano/src/command_buffer/auto.rs
```diff
@@ -1259,6 +1259,28 @@ impl<P> AutoCommandBufferBuilder<P> {
         Ok(self)
     }
 
+    /// Adds a command that executes all the commands in a vector.
+    ///
+    /// **This function is unsafe for now because safety checks and synchronization are not
+    /// implemented.**
```
mitchmindtree: Would it be possible to mention which checks need to be added? Or even better, add them before merging? I'm just thinking that, with our limited maintainership, it would be nice if we could try and slowly reduce vulkano's TODO rather than add to it :)

knappador: Wish I knew the history.  I didn't notice anything relatively less safe than other commands unless the dependency checks on primary buffers has no idea about secondary buffer dependencies.  Could be some auto-barrier stuff missing.  Personally I'm not too concerned about it since it's a runtime check rather than some nice type / borrow check based solution.

WarpspeedSCP: @mitchmindtree  I had asked about it on the gitter a while ago, but I didn't really get a response, so I am not sure of what checks would be necessary to make this safer.

knappador: I think the TODO is actually mildly out of date, mildly because I don't have a tremendous amount of faith that the barrier behavior is a comprehensive solution or path to a solution yet.

The blame on that TODO:
https://github.com/vulkano-rs/vulkano/commit/fbcdd77aaa24ff0632a0091d653b941d728ded9d

The blame on the command  buffer part that added barriers for secondaries is:
https://github.com/vulkano-rs/vulkano/commit/a4589e7a16b462f288d58d54fc7b650ff5fadf14#diff-4eeb919345afc97f16ffc03e0fc5843eR707

I development docs way more than I need user docs :slot_machine:

PR #1179, opened 2019-03-20
Fix shader include for virtual workspaces

TheEdward162:
Fixed proc macro `shader!` to correctly handle cases where crate root != proc macro root.

In `include_callback` function in `codegen.rs` variable `contained_within_path_raw` is used to calculate paths for relative includes. This variable is a RELATIVE path passed from the `shader` proc macro and it is relative to the crate root (as passed by the user from the `path` attribute of the macro). However, in cases where virtual workspaces are used, crate root is not the same as workspace root. This is a problem because proc macros resolve relative paths against workspace root.

I've added a fix to the `shader` macro to correctly prefix the user-passed crate-relative path into workspace-relative path for correct resolution in `include_callback`.

Also I tested this on the shader-include example (uncommented, on linux) and it didn't break anything.

Additionally, while researching this I've noticed that shader include directories (for includes with `#include <path>`) also require paths relative to the workspace root, not crate root. I don't want to make any breaking decisions but I think we should consider prefixing those as well so they are always crate-root relative, same as shader paths.

* [ ] ~~Added an entry to `CHANGELOG_VULKANO.md` or `CHANGELOG_VK_SYS.md` if knowledge of this change could be valuable to users~~
* [ ] ~~Updated documentation to reflect any user-facing changes - in this repository~~
* [ ] ~~Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.~~

### DIFF
--- a/vulkano-shaders/src/lib.rs
+++ b/vulkano-shaders/src/lib.rs
@@ -290,40 +290,40 @@ pub fn shader(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 
     let (path, source_code) = match input.source_kind {
         SourceKind::Src(source) => (None, source),
-		SourceKind::Path(path) => {
-			let crate_root_string = env::var("CARGO_MANIFEST_DIR").unwrap_or(".".to_string());
-			let crate_root = Path::new(&crate_root_string);
-			let source = {
-				let full_path = crate_root.join(&path);
+        SourceKind::Path(path) => {
+            let crate_root_string = env::var("CARGO_MANIFEST_DIR").unwrap_or(".".to_string());
+            let crate_root = Path::new(&crate_root_string);
+            let source = {
+                let full_path = crate_root.join(&path);
 
-				if full_path.is_file() {
-					read_file_to_string(&full_path)
-						.expect(&format!("Error reading source from {:?}", path))
-				} else {
-					panic!("File {:?} was not found; note that the path must be relative to your Cargo.toml", path);
-				}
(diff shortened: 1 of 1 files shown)


rukai:
Thanks for the PR!

The rest of this file uses spaces, please change your tabs to spaces.

I dont seem to be able to reproduce the problem.
The `shader-include` example works fine when copied to a virtual workspace, however it should fail because its shaders are using relative paths?

As for:
```
Additionally, while researching this I've noticed that shader include directories (for includes with #include <path>) also require paths relative to the workspace root, not crate root.
I don't want to make any breaking decisions but I think we should consider prefixing those as well so they are always crate-root relative, same as shader paths.
```
Good catch, please make a separate PR for this.

TheEdward162:
Hi,

I have converted tabs to spaces.

As for the relative include, the way the example is currently set up, it is impossible to properly test relative includes without them falling back to standard ones (and working). I have created a [branch in my fork](https://github.com/TheEdward162/vulkano/tree/shader-include-examples/virtual-examples) to demostrate the problem.

TheEdward162:
I have created a new draft PR https://github.com/vulkano-rs/vulkano/pull/1185 that deals with `includes` folders and prefixing. It is a draft because it builds on commits from this one (I hope I am doing this whole PR thing right).

I am not sure if I should close this PR or if this should get merged before the other one gets merged.

Issue #1174, opened 2019-03-07
supports_transfers always returns true


Julusian:
# Template

*   Version of vulkano: 0.11.1
*   OS: ubuntu18.04
*   GPU (the selected PhysicalDevice): Quadro P4000
*   GPU Driver: 390.116.0.0
*   Upload of a reasonably minimal complete `main.rs`: [main.zip](https://github.com/vulkano-rs/vulkano/files/2942852/main.zip)

# Issue

The supports_transfers check is ignoring the flags and always returning true.
If https://www.reddit.com/r/vulkan/comments/4x0lbu/nvidias_lone_transfer_queue/ is accurate (I am struggling to find anything else arguing either way) then being able to identify the DMA transfer queue will be crucial to decent performance when lots of transfers are needed.

Introduced by https://github.com/vulkano-rs/vulkano/pull/536

My simple test program is producing:
```
#0: g:true c:true, t:true
#1: g:false c:false, t:true
#2: g:false c:true, t:true
```
for the GPU: https://vulkan.gpuinfo.org/displayreport.php?id=5364#queuefamilies

knappador:
Recommend closing or doing a PR on the docs to reference the original source of the assumption.

https://vulkan-tutorial.com/Vertex_buffers/Staging_buffer#page_Transfer_queue
> The good news is that any queue family with VK_QUEUE_GRAPHICS_BIT or VK_QUEUE_COMPUTE_BIT capabilities already implicitly support VK_QUEUE_TRANSFER_BIT operations. The implementation is not required to explicitly list it in queueFlags in those cases.

Vulkano docs related to the assumption that graphics or compute implies transfer:
https://docs.rs/vulkano/0.11.1/vulkano/instance/struct.QueueFamily.html#method.supports_transfers
> Note: Queues that support graphics or compute operations also always support transfer operations. As of writing this, this function will always return true. The purpose of this function is to be future-proofed in case queues that don't support transfer operations are ever added to Vulkan.

Related source
https://github.com/vulkano-rs/vulkano/blob/d8fbef1a20465b67d53b8f329f0828eadb18fb2f/vulkano/src/instance/instance.rs#L1081-L1085

Blame doesn't reference the docs that back up the assumption.
https://github.com/vulkano-rs/vulkano/commit/8574583409cddc201cfb995d4f48f995060482d4

Reproduced on GeForce GTX 1060:
https://vulkan.gpuinfo.org/displayreport.php?id=5227#queuefamilies

[...]

PR #1187, opened 2019-03-28
supports_transfer tell the truth

knappador:
The future is here.  Some graphics cards are now only setting the transfer bit on queues with some special DMA controller relationship that can perform higher speed transfers.  Let's tell the truth.

https://github.com/vulkano-rs/vulkano/issues/1174

### DIFF
--- a/vulkano/src/instance/instance.rs
+++ b/vulkano/src/instance/instance.rs
@@ -1073,15 +1073,13 @@ impl<'a> QueueFamily<'a> {
     }
 
     /// Returns true if queues of this family can execute transfer operations.
-    ///
-    /// > **Note**: Queues that support graphics or compute operations also always support transfer
-    /// > operations. As of writing this, this function will always return true. The purpose of
-    /// > this function is to be future-proofed in case queues that don't support transfer
-    /// > operations are ever added to Vulkan.
+    /// > **Note**: While all queues that can perform graphics or compute operations can implicitly perform
+    /// > transfer operations, graphics & compute queues only optionally indicate support for tranfers.
+    /// > Many discrete cards will have one queue family that exclusively sets the VK_QUEUE_TRANSFER_BIT
+    /// > to indicate a special relationship with the DMA module and more efficient transfers.
     #[inline]
-    pub fn supports_transfers(&self) -> bool {
-        (self.flags() & vk::QUEUE_TRANSFER_BIT) != 0 || self.supports_graphics() ||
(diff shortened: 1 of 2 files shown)


knappador:
@rukai  @tomaka  Organization membership.　There is nobody who can press buttons right now.

rukai:
I'll look into it this weekend

vulkano/src/instance/instance.rs
```diff
@@ -1075,13 +1075,10 @@ impl<'a> QueueFamily<'a> {
     /// Returns true if queues of this family can execute transfer operations.
     ///
     /// > **Note**: Queues that support graphics or compute operations also always support transfer
-    /// > operations. As of writing this, this function will always return true. The purpose of
-    /// > this function is to be future-proofed in case queues that don't support transfer
-    /// > operations are ever added to Vulkan.
+    /// > operations.
```
rukai: Change this line to:
 ` /// > operations, regardless of the value this method returns.`

vulkano/src/instance/instance.rs
```diff
@@ -1075,13 +1075,10 @@ impl<'a> QueueFamily<'a> {
...
     #[inline]
     pub fn supports_transfers(&self) -> bool {
```
rukai: The name of the function should be changed to `explicitly_supports_transfers`.
This makes it clear from the name alone that its not like the other `supports_*` methods. i.e. Even when false it may still support transfers.

knappador:
[x] explicitly_supports_transfers
[x] updated CHANGELOG
[x] deprecated `supports_transfers` instead of removing it outright.  Anyone affected by the behavior change is using it incorrectly most likely to begin with.
[x] commit squashed
[x] tests run
[x] notes are shorter version of the actual Vulkan docs:

```
All commands that are allowed on a queue that supports transfer operations are also allowed on a queue that supports either graphics or compute operations. Thus, if the capabilities of a queue family include VK_QUEUE_GRAPHICS_BIT or VK_QUEUE_COMPUTE_BIT, then reporting the VK_QUEUE_TRANSFER_BIT capability separately for that queue family is optional.
```

rukai:
The old function should just be removed not deprecated, because:
*   More maintenance in the future
*   Changing the functionality and deprecating the function is really weird. It was doing exactly what it was documented to do.

knappador:
This is ready to go AFAIK

rukai:
Whoops, I missed this was ready.

PR #1192, opened 2019-04-12
Ensure Device::active_queue_families contain no duplicates

NiklasJonsson:

Each entry in Device::active_queue_families should be unique (as I've understood it). See for example the creation of ImmutableBuffer in Immutable::raw_impl, which will set the sharing mode to Concurrent if we have more than one (used) queue family on the device.

### DIFF
--- a/vulkano/src/device/mod.rs
+++ b/vulkano/src/device/mod.rs
@@ -300,6 +300,13 @@ impl Device {
                                                   *const _
                                           });
 
+        let mut active_queue_families: SmallVec<[u32; 8]> = SmallVec::new();
+        for (queue_family, _) in output_queues.iter() {
+            if let None = active_queue_families.iter().find(|&&qf| qf == *queue_family) {
+                active_queue_families.push(*queue_family);
+            }
+        }
+
         let device =
             Arc::new(Device {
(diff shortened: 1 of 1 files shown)


knappador:
Seems like a bug =)  

> which will set the sharing mode to Concurrent if we have more than one (used) queue family on the device.

This choice seems like a hack to delay implementing the buffer ownership?

There's a lot of ugly stuff in this file.  

The change is more correct than what was there, so I would be happy to merge it if someday I get organization membership @rukai @tomaka

NiklasJonsson:
So is anything gonna happen here? Is there something else I should do?

rukai:
Hey sorry about the wait, I've been putting my time into wgpu instead.

Good catch.
This change fixes an invalid usage of VkBufferCreateInfo
```
If sharingMode is VK_SHARING_MODE_CONCURRENT, each element of pQueueFamilyIndices must be unique and must be less than pQueueFamilyPropertyCount returned by either vkGetPhysicalDeviceQueueFamilyProperties or vkGetPhysicalDeviceQueueFamilyProperties2 for the physicalDevice that was used to create device
```

Issue #1183, opened 2019-03-24
ShaderC Compile & CI


knappador:
Our CI server currently failed one of the builds due to shaderc compile stalling.   I believe we need the option to use pre-compiled shaderC for the shaderc Rust package and am chasing this over to the ShaderC repo.

Gathering up some info here.

https://travis-ci.org/vulkano-rs/vulkano/jobs/506588484
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

Packages exist:
https://packages.gentoo.org/packages/media-libs/shaderc
https://www.archlinux.org/packages/extra/x86_64/shaderc/

This might be a lie on our CI setup:
[`# The shaderc dependencies are already installed on the base images`](https://github.com/vulkano-rs/vulkano/blob/5233221ad908bfffbd9379da025e8cb62c9a84ea/.travis.yml#L1)

knappador:
https://github.com/google/shaderc-rs/issues/44

Going to try to kill this at the Rust package level.

knappador:
https://github.com/google/shaderc-rs/pull/46

Headline change is using a static build path instead of the dynamic ones that cargo generates for `out` directory.

I don't know an Ubuntu package to make this process easy.  Who even controls the keys to the CI servers?  @rukai @tomaka

Rua:
This seems to have been addressed by ShaderC?

PR #1197, opened 2019-04-22
Shaderc version bump

knappador:

```
01:59 < knappador> Can I pass a file path into a dependency via features?
02:00 < knappador> Trying to figure out how to give downstream consumers a way to configure 
                 the location of a pre-build library that's slow to compile
02:00 < viny> not really
```
Sad.

I took a look at the CI files but I think it's opaque from the perspective of Vulkano maintainers.

To enable this on a CI build:
```
cargo build SHADERC_LIB_DIR=/path/to/the/glslang/SPIRV/and/libshaderc/unless/monolithically/built/
```

I'm also waiting for [a valid pc file](https://github.com/google/shaderc/pull/581) to start hitting maintained packages to use pkg-config to pick up the system libraries more reliably.  Ubuntu has no ppa, so basically we're limited to setting `SHADERC_LIB_DIR` or building from source in CI on Ubuntu.

The API from 0.3.X was not bumped to 0.5 for any reason other than punching crates.io, so there are no anticipated breaking changes (and I will support on shaderc-sys related issues)

#1183  related, not fixed due to CI

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,5 +1,5 @@
 # Unreleased (breaking)
-
+- Update shaderc to 0.5.  New shaderc has improved pre-built options for libshaderc that significantly reduce package build time and are appropriate for use in CI
 - `QueueFamily::explicitly_supports_tranfers` only returns true if `vk::QUEUE_TRANSFER_BIT` is set instead of also always returning true.  Removed `supports_transfers`.
 - Update to winit 0.19
 - Add support for `#include "..."` and `#include <...>` directives within source
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Vulkano uses [shaderc-rs](https://github.com/google/shaderc-rs) for shader compi
 
 Unless you provide libshaderc, in order to build libshaderc with the shaderc-sys crate, the following tools must be installed and available on `PATH`:
 - [CMake](https://cmake.org/)
+- [Ninja](https://ninja-build.org/) Is optional except when building with MSVC.  It may speed up build time for libshaderc.
 - [Python](https://www.python.org/) (works with both Python 2.x and 3.x, on windows the executable must be named `python.exe`)
 
 These requirements can be either installed with your favourite package manager or with installers
--- a/README.md
+++ b/README.md
@@ -53,8 +53,9 @@ To get started you are encouraged to use the following resources:
 
 ## Setup
 
-Vulkano uses [shaderc-rs](https://github.com/google/shaderc-rs) for shader compilation. In order to
-build the shaderc-rs crate the following tools must be installed and available on `PATH`:
+Vulkano uses [shaderc-rs](https://github.com/google/shaderc-rs) for shader compilation.  Refer to shaderc-rs documentation to provide a pre-built libshaderc for faster build times.
+
+Unless you provide libshaderc, in order to build libshaderc with the shaderc-sys crate, the following tools must be installed and available on `PATH`:
 - [CMake](https://cmake.org/)
 - [Python](https://www.python.org/) (works with both Python 2.x and 3.x, on windows the executable must be named `python.exe`)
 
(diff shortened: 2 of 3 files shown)


knappador:
Checking out the CI builds, I noticed that we don't cache the target directory.  In the rare circumstance that a clean is needed to evade a cargo bug, we can punch this manually.

https://github.com/e-nguyen/e-nguyen/blob/master/.circleci/config.yml#L43

These 20min CI builds are quite up in the pain threshold for collaboration.

I would offer to rip up the existing CI infrastructure if I had a Mac solution ready.  Caching the target and providing libshaderc_combined.a via these changes should pretty much reap all the same potential gains as moving to an E-Nguyen style CI.

knappador:
Dear reviewer, if your soul is and endless spring of misanthropic spite, it appears there is Mac support on Circle.  I am likely capable of getting it running without a local mac to work on, but I am unwilling to replace the CI infrastructure without great cause and mutual will to change.

https://circleci.com/docs/2.0/hello-world-macos/

If our current CI is basically inaccessible, I'm in favor of the move because these build times have latency so huge that figuring out how you broke Mac etc is definitely a turnoff to any contributors.

mitchmindtree:
Nice @knappador, good to hear this at least improves compile times a bit if anything!

Silly question perhaps, but just wanted to check - have you tested all the vulkano-examples still compile and behave the same after applying this?

> 01:59 < knappador> Can I pass a file path into a dependency via features?
> 02:00 < knappador> Trying to figure out how to give downstream consumers a way to configure 
>                  the location of a pre-build library that's slow to compile
> 02:00 < viny> not really
>
> Sad.

Totally agreed, this is one of the biggest issues we run into with our [nannou](https://github.com/nannou-org/nannou) workshops as well - spending half of a one-hour workshop compiling huge deps on many machines with varying levels of performance is not fun! Having the option to specify a source for pre-built dependencies in the Cargo manifest (where it falls back to src if the desired target triplet build doesn't exist or the md5 doesn't match) would be a *huge* win. I wonder if there's an issue for this somewhere at the cargo repo...

knappador:
> have you tested all the vulkano-examples

Well, I do get a segmentation fault on the `debug` example :boom: on this branch and on master :boom: :boom: 

They all compile and all run the way I recall them running before.


@mitchmindtree  https://github.com/google/shaderc-rs/pull/46
Tests at the very bottom demonstrate valid ways to call the build.  The environment variables pass through.  Just use absolute paths.

knappador:
@mitchmindtree  This might help:
https://github.com/google/shaderc/issues/580

While there is no Ubuntu package for libshaderc yet, shaderc itself just added a valid pkg-config .pc file, and whenever that gets added, I'll add pkg-config support to shaderc-sys and we can stop global warming.

rukai:
Works on Arch Linux with and without the shaderc package installed.
Still need to test on windows, it sounds like it only supports ninja for msvc now??

knappador:
Ninja is now _required_ on msvc but is still optional on other platforms.  It's only used on from-source builds AFAIK and doesn't affect usage.  

https://github.com/google/shaderc-rs/blob/master/shaderc-sys/build/build.rs#L224-L230

I added a line to docs about Ninja being required on MSVC.

knappador:
shaderc-rs planning to drop Python2
https://github.com/google/shaderc-rs/commit/c5bb3296815aa0433be7eed24aafc866ed43b83d

_will_ someday affect https://github.com/vulkano-rs/vulkano/pull/1197/files#diff-04c6e90faac2675aa89e2176d2eec7d8R61

Which is not changed in this PR.

rukai:
Please add the `mingw-w64-x86_64-ninja` package to the command in step 4 of the windows-msvc Specific Setup

knappador:
Should be good to go.  

Do the python3 packages work?

rukai:
No they dont, as the executable is named python3 instead of python.
All good to merge what we have now though.

PR #1203, opened 2019-05-24
Publish 0.12.0 of `vulkano`, `vulkano-win` and `vulkano-shaders`.

mitchmindtree:
Also updates the CHANGELOG to include the image layout transition
improvements.

cc @rukai it might finally be that time :)

-----

* ~Updated documentation to reflect any user-facing changes - in this repository~
* ~Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR~.


m4b:
I’m ok with merging this because I want the winit update but just FYI pretty sure the image layout transitions broke me, or at least regressed. I can post more details later if I get time, but since it’s only a visual regression/breakage and I haven’t identified yet if I’m doing something weird (don’t think so!) we should just merge; just wanted to add this before I forgot :)

mitchmindtree:
It seems the macOS stable travis build timed out, but the other builds seem to be fine.

@m4b ah interesting - if you decide to create an issue about it please cc @freesig and I as we'd be curious to investigate! One thing that comes to mind that might be worth trying - have you removed all manual cases of specifying `initial_layout` and `final_layout` in your `renderpass!` declarations? In theory, vulkano should be able to work these out automatically now, whereas in the past you kind of had to add them sometimes as a hack to work around layout transitions that couldn't be inferred automatically.

rukai:
I'm happy to do a release now, just checking for low hanging fruit that should be thrown in.

rukai:
You should do a PR to the guide to update the version number used.

Issue #1207, opened 2019-06-21
Shader compilation fails with: "No Rust equivalent for a floating-point of width 32"


dthul:
*   Version of vulkano: 0.12
*   OS: Windows 10
*   GPU (the selected PhysicalDevice): GTX 1080
*   GPU Driver: 418.96

# Issue

I updated to the latest Rust nightly (1.37.0-nightly (1d9981f04 2019-06-20)) and now shader compilation doesn't work anymore:

```rust
vulkano_shaders::shader! {
    ty: "vertex",
    path: "src/render/diffuse/vertex.glsl"
}
```

fails with:

```
error: proc macro panicked
  --> viewer\src\render\checkerboard\mod.rs:10:5
   |
10 | /     vulkano_shaders::shader! {
11 | |         ty: "vertex",
12 | |         path: "src/render/diffuse/vertex.glsl"
13 | |     }
   | |_____^
   |
   = help: message: No Rust equivalent for a floating-point of width 32
```

It seems like this issue was also reported in [nannou-org/nannou_timeline#1](https://github.com/nannou-org/nannou_timeline/issues/1) and [nannou-org/nannou#338](https://github.com/nannou-org/nannou/issues/338).

milkey-mouse:
I have the same issue. Oddly on [my project](https://github.com/milkey-mouse/planets) `cargo build` works fine, but `cargo build --release` does not, even after running `cargo clean`.

I am using rustc `1.37.0-dev (8d6f4b96d 2019-06-21)` & vulkano `0.12.0` on Linux.

rukai:
A quick look at the code shows that this is either some evil UB somewhere in vulkano or rustc is miscompiling things.

https://github.com/vulkano-rs/vulkano/blob/master/vulkano-shaders/src/structs.rs#L294 This branch can never be taken when width = 32, and yet it is taken...

norru:
Just been bitten by this. Only happens in `build --release`. I see @rukai has worked this out already, when is the fix scheduled to be release on `crates.io`? :)

```
cargo 1.37.0-nightly (4c1fa54d1 2019-06-24)
rustc 1.37.0-nightly (17e62f77f 2019-07-01)
vulkano-shaders v0.11.1
```

PR #1209, opened 2019-06-25
rustc is now making use of the fact that derefencing a null pointer is UB. Remove all null dereferences

rukai:

Uh oh..
rustc is now making use of the fact that derefencing a null pointer is UB.
So we need to remove all null dereferences, of which there are surprisingly large amount of. In order to get vulkano working AT ALL on nightly and presumably soon beta + stable.

@tomaka You have a pretty good idea what is and isnt UB. Why did you write these null dereferences, is there something you needed from them that I am missing?

Fixes https://github.com/vulkano-rs/vulkano/issues/1207

### DIFF
--- a/examples/src/bin/deferred/frame/ambient_lighting_system.rs
+++ b/examples/src/bin/deferred/frame/ambient_lighting_system.rs
@@ -136,7 +136,7 @@ impl AmbientLightingSystem {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Default, Debug, Clone)]
 struct Vertex {
     position: [f32; 2]
 }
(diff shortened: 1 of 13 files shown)


rukai:
Ok, this PR now runs on the latest nightly.

Unfortunately I could not find a way to do so without a breaking change.
The vertex struct provided by the user now needs to `#[derive(Default)]`
If anyone can provide a fix that isn't breaking I will gladly take it!

I'll wait a few days to get feedback from people. @mitchmindtree  @freesig 
Then I'll go ahead and release this as a breaking change so people can transition before it breaks on stable rust.

tomaka:
It was a way to be sure that the offset is actually correct. In other words, a replacement for the `offsetof!` macro that Rust is lacking.

rukai:
Is the variable called size poorly named and should really be another type of offset?

tomaka:
I guess `size_of` *should* work, but is not necessarily 100% future-proof. In practice it should be fine, I guess.

Issue #960, opened 2018-05-25
`triangle` example crashes after rendering first frame with `PoisonError`


dannyfritz:
> Windows 10
> Using device: Intel(R) HD Graphics 620 (type: IntegratedGpu)   
> rustc 1.26.0 (a77568041 2018-05-07)

When I run the triangle example (or any other `GraphicsPipeline` example for that matter), it crashes after rendering the first frame with a [`PoisonError`](https://doc.rust-lang.org/std/sync/struct.PoisonError.html).

So for teapot, I see the teapot, then it crashes. For image, I see the logo, then it crashes.

They all crash at [this line](https://github.com/vulkano-rs/vulkano/blob/c112085/examples/src/bin/triangle.rs#L334). Removing the `cleanup_finished()` line allows me to resize the window and operate normally, but after about 10 seconds crashes with a stack overflow.

For compute, it seems to run fine without erroring and exits cleanly. 

<details>
  <summary>Stacktrace:</summary>

```
C:\Users\danny\Dev\vulkano\examples (master)
λ cargo run --bin triangle
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `C:\Users\danny\Dev\vulkano\target\debug\triangle.exe`
Using device: Intel(R) HD Graphics 620 (type: IntegratedGpu)
thread 'main' panicked at 'assertion failed: {
    let dur = Some(Duration::new(0, 0));
    self.fence.as_ref().map(|f| f.wait(dur).is_ok()).unwrap_or(true)
}', vulkano\src\swapchain\swapchain.rs:867:13
stack backtrace:
   0:     0x7ff7b18d873a - std::sys::windows::backtrace::unwind_backtrace
                               at C:\projects\rust\src\libstd\sys\windows\backtrace\mod.rs:65
   1:     0x7ff7b18e231f - std::sys_common::backtrace::_print
                               at C:\projects\rust\src\libstd\sys_common\backtrace.rs:71
   2:     0x7ff7b18e231f - std::sys_common::backtrace::print
                               at C:\projects\rust\src\libstd\sys_common\backtrace.rs:59
   3:     0x7ff7b18d0f8e - std::panicking::default_hook::{{closure}}
                               at C:\projects\rust\src\libstd\panicking.rs:207
   4:     0x7ff7b18d0c3a - std::panicking::default_hook
                               at C:\projects\rust\src\libstd\panicking.rs:223
[...]

PR #1225, opened 2019-07-28
Remove assert that was based on bad assumption

nickwilcox:
* ~[ ] Updated documentation to reflect any user-facing changes - in this repository~
* ~[ ] Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR~

This pull request addresses issue #960 

The assumption that the fence must already be signalled is incorrect. There is nothing in the Vulkan spec regarding `VkAcquireNextImageInfoKHR` that says what order the fence and semaphore are signalled in, or that they are signalled at exactly the same time from the API users perspective.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-*No changes as of yet.*
+- Removed faulty debug_assert in `SwapchainAcquireFuture::drop`.
 
 # Version 0.13.0 (2019-07-02)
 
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks a lot for digging into this @nickwilcox and apologies for the delay. I haven't checked the spec on this myself due to lack of time, but going off your conversation with @mickare I think this makes sense. This debug assertion seems to be more of a sanity check than a requirement for vulkano's safety assumptions, otherwise I'd expect to have seen a non-debug assertion here. Going to merge!

PR #1226, opened 2019-07-30
Update shaderc to 0.6 in order to significantly reduce compile-times …

Abendstolz:
…on (Arch/Void)Linux again by linking towards shared libraries

@mitchmindtree fyi

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-*No changes as of yet.*
+- Update shaderc to 0.6. This again allows to use locally installed libraries which reduces the build-time significantly on Arch/Voidlinux (see https://github.com/google/shaderc-rs/issues/58)
 
 # Version 0.13.0 (2019-07-02)
 
--- a/vulkano-shaders/Cargo.toml
+++ b/vulkano-shaders/Cargo.toml
@@ -21,4 +21,7 @@ quote = "0.6"
 proc-macro2 = "0.4"
 
 [dev-dependencies]
-vulkano = { version = "0.13", path = "../vulkano" }
+vulkano = { version = "0.13", path = "../vulkano"} 
+
+[features]
+shaderc-build-from-source = ["shaderc/build-from-source"]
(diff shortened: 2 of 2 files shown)


mitchmindtree:
@Abendstolz nice! I think using pre-installed libraries was already active in 0.5, but at least this gets it working again on Arch and Void :+1:

Could you also add a `shaderc-build-from-source` feature to `vulkano-shaders` which enables the `shaderc/build-from-source` feature? I just realised this feature is currently inaccessible via `vulkano-shaders` and might save us temporarily if shaderc's linking to pre-built libraries breaks again in the future.

Abendstolz:
@mitchmindtree sure thing. Done in https://github.com/vulkano-rs/vulkano/pull/1226/commits/5db99c837330f3a0656e9c6827a1fde24421465b

mitchmindtree:
Thanks a lot @Abendstolz - it looks like travis is just complaining about some trailing whitespace now, but after that's fixed I'd be happy to merge!

@rukai would you mind if we published 0.14 after this as another emergency patch to make it possible for Arch/Void users to build vulkano-shaders? Maybe we can also include #1227 and #1225, though I'm not well versed enough to understand if #1225 is correct/safe.

rukai:
Sure, feel free to publish 0.14 if need be, the website would additionally need to be updated.

I cannot review PRs as i am already using my free time to work on gfx Hal/wgpu stack. Sorry. It might be a simple change but it's hard to tell without investing a lot of time in the first place.

AustinJ235:
@rukai I'd like to see this go through and a version published, my projects are still on vulkano 11.1 due to the linking errors with shaderc. That issue is solved by this pull request.

m4b:
can this be emergency pushed? basically all vulkano projects are broken now on archlinux et. al

rukai:
You can workaround it by adding the build-from-source feature to shaderc

 Can @abendstolz fix the CI errors? Then I can proceed with an emergency release

Abendstolz:
@rukai or @mitchmindtree go for it ;)

PR #1227, opened 2019-08-01
Add binary_shader! macro and make shaderc optional

myfreeweb:
Since the vulkano-shaders codegen parses SPIR-V and does not depend on the GLSL source, it's easy to allow SPIR-V binary shaders (which can come from other, non-GLSL compilers) to be included directly.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-*No changes as of yet.*
+- Add the `binary_shader!` proc macro that allows inclusion of arbitrary SPIR-V binary shaders.
+- Make `shaderc` an optional but default feature of `vulkano-shaders`, so that it's possible to not have the dependency if using `binary_shader` exclusively.
 
 # Version 0.13.0 (2019-07-02)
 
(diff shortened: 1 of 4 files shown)


mitchmindtree:
Thanks @myfreeweb, I think this could be very useful for folk compiling their SPIR-V using methods other than `shaderc`! Speaking of, would you mind sharing your current approach / use case? I'm mostly just curious as I'm interested in alternative solutions to `shaderc`.

The only thing that strikes me about this PR is that it uses a lot of feature flags - is it possible to consolidate the shaderc code into its own module so that we only have to feature gate the module?

Also, perhaps we should add a line to `.travis.yml` to test building with the `shaderc` feature disabled as well? Otherwise it might be easy to accidentally break the `binary_shader!` support in the future.

myfreeweb:
> use case?

I was trying to run [clspv](https://github.com/google/clspv) output. It turned out to not be very friendly to the codegen here :) and even with everything corrected by hand I'm getting a crash without anything detected by standard validation layers…

> consolidate the shaderc code into its own module

Sure

lewisclement:
This is a feature I'd like to use. Is there something currently blocking the merge?

AustinJ235:
Is this still worth pursing, as in it is functional for other options available, or should this be closed?

AustinJ235:
Closing due to inactivity and likely needing some rework. Others are open, to open another pull request if they would like to implement this.

PR #1228, opened 2019-08-10
Fix `ImageDimensions::mipmap_dimensions` and `max_mipmaps` where the original size is not a power of two.

IcyDefiance:
* [x] ~Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.~ (No reference to this function in the guide.)

---

Mipmap sizes are not rounded up to the nearest power of two ([source](https://www.khronos.org/registry/vulkan/specs/1.1/html/chap11.html#resources-image-miplevel-sizing)), so `.next_power_of_two()` shouldn't be there. It happens to produce the correct result when the preceding calculation produces a power of two, but only because it doesn't do anything.

In addition, the `((width - 1) >> level) + 1` trick is cool, but it will always round up instead of down, which is also incorrect. It happens to produce the correct result when the original size is an even number, but only because it doesn't need to round after dividing by 2.

Scrapping both and replacing them with a call to `cmp::max()` is correct for all image sizes.

Here's one of the validation errors I have when trying to blit to the first mipmap of an image with a size of 192x256 px:

> Object: 0x29505b92840 (Type = 6) | vkCmdBlitImage(): region [0] dstOffset[].x values (0, 128) exceed dstSubresource width extent (96). The Vulkan spec states: dstOffset[0].x and dstOffset[1].x must both be greater than or equal to 0 and less than or equal to the destination image subresource width (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageBlit-dstOffset-00248)


IcyDefiance:
It seems max `max_mipmaps()` is also off by one under the same conditions. I'll add a fix for that in a second.

IcyDefiance:
For bonus points, my partner figured out that because `.leading_zeroes()` only cares about the highest bit, we can just do a bitwise OR on each dimension instead of using `cmp::max()`. It should be slightly faster now.

AustinJ235:
Wonder if this was causing any weird issues with my things. Your changes seem to check out to me. There is now a conflict however with this merge.

IcyDefiance:
Okay, the conflict is handled.

AustinJ235:
Well interesting enough mac is broken.

AustinJ235:
So looking at the build log it appears we are using metal 13.1, which is from dec of 2018. Perhaps it is time we update this. I don't have a mac to do testing and am not sure how how the large the changes required to do the upgrade are.

So does vulkano even work for people on mac os atm?

PR #1229, opened 2019-08-11
support for compressed formats in check_copy_buffer_image

nickwilcox:
* [x ] Added an entry to `CHANGELOG_VULKANO.md` or `CHANGELOG_VK_SYS.md` if knowledge of this change could be valuable to users
* [x ] Updated documentation to reflect any user-facing changes - in this repository

This is my take on supporting compressed formats in check_copy_buffer_image.
* Texture Formats now have a 2d block size. For non block formats this is simply (1,1)
* The size is now given as the size of a block (so for 1x1 blocks this is still the element size)
* The destination size check rounds up the image dimensions to the nearest block before calculating buffer size.

All the block dimensions and sizes taken from https://www.khronos.org/registry/DataFormat/specs/1.1/dataformat.1.1.html


mitchmindtree:
Thanks for the PR! After a quick look this looks good to me. Tests all seem to complete successfully locally too (at least, no more fail than they did before). Going to merge for inclusion in #1230.

PR #1235, opened 2019-09-20
Discrete fix

MacTuitui:
It looks like compute shaders exhibit the same behavior again (that is, they do not compute, but nothing complains). Could you merge upstream so I can test?


MacTuitui:
Sorry, did not mean to ask a PR on this repo...

PR #1240, opened 2019-10-13
Replaced all mem::uninitialized() by MaybeUninit

nifker:
No changes for the user.
I will try to find a way to initialize the remaining mem::uninitialized() with MaybeUninit too.


nifker:
I know got all mem::uninitalized() occurrences removed!

AustinJ235:
Looks good.

PR #1244, opened 2019-10-20
Update Metal

AustinJ235:


AustinJ235:
So turns out the pull request before where the winit updated I forgot this repo has auto release. So at this time vulkano-win 0.15 is out when it shouldn't be. Will probably have to yank 0.15 and publish 0.16 when we are ready.

@tomaka would you mind yanking vulkano-win 0.15?

AustinJ235:
CI is having issues now....

PR #1248, opened 2019-10-22
update shader include resolution to fix issues with windows

nickwilcox:
* [x ] Added an entry to `CHANGELOG_VULKANO.md` or `CHANGELOG_VK_SYS.md` if knowledge of this change could be valuable to users
* [ x] Updated documentation to reflect any user-facing changes - in this repository
* [ ] ~~Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.~~

This change in the behaviour of shader includes in order to make it consistent against the command line `glslc` from shaderc, and make it work on Windows.

* Remove the use of canonicalize - this is the source of windows issues, and results in incorrect behaviour when symlinks are used.
* Include folders are relative to `Cargo.toml` for consistency with the paths to shaders
* Don't check for includes going outside their path - this is allowed by `glslc` so should be allowed here


AustinJ235:
Looking over the code it seems all good.

nickwilcox:
Thanks.

PR #1275, opened 2019-11-14
fix descriptor set codegen for SSBO after spirv update

nickwilcox:

The upgrade to spirv 1.3 lead to bugs in the `shader!` around buffers in descriptor sets.

The `BufferBlock` decoration is deprecated (see https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_decoration_a_decoration) and the new way is to use the storage class on the variable.

Tested in the basic-compute-shader example, which I also modified to require the `khr_storage_buffer_storage_class` extension when initialising.


AustinJ235:
Don't seem to be able to restart builds at the moment. You might have to push another dummy commit to trigger it.

nickwilcox:
CI still seems to be failing at the very start on one of the jobs

PR #1277, opened 2019-11-20
Sendable DebugCallback

jbit:

On the `master` branch the `DebugCallback` struct can not be sent between threads due to the `user_callback` field being non `Send`able. 

It's useful to be able to instantiate Vulkano on one thread and then send the instances to a worker thread for use, but right now this would require the worker thread to instantiate it's own `DebugCallback`, which could cause two to be registered at the same time which results in double logging. 

This change simply adds the `Send` trait bound to the `user_callback` field to fix this. The `DebugCallback::new` function already required the `Send` trait bound, so I do not believe this can break any existing code. I also don't believe this introduces any unsoundness since the field is only there to keep the callback object alive. 

Including error to aid searchability:
```
error[E0277]: `(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)` cannot be sent between threads safely
   --> vulkano/src/instance/debug.rs:356:9
    |
356 |         thread::spawn(move || {
    |         ^^^^^^^^^^^^^ `(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)` cannot be sent between threads safely
    |
    = help: the trait `std::marker::Send` is not implemented for `(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)>`
    = note: required because it appears within the type `std::boxed::Box<(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)>`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<std::boxed::Box<(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)>>`
    = note: required because it appears within the type `std::boxed::Box<std::boxed::Box<(dyn for<'r, 's> std::ops::Fn(&'r instance::debug::Message<'s>) + 'static)>>`
    = note: required because it appears within the type `instance::debug::DebugCallback`
    = note: required because it appears within the type `std::result::Result<instance::debug::DebugCallback, instance::debug::DebugCallbackCreationError>`
    = note: required because it appears within the type `[closure@vulkano/src/instance/debug.rs:356:23: 358:10 callback:std::result::Result<instance::debug::DebugCallback, instance::debug::DebugCallbackCreationError>]`
```


AustinJ235:
Reading through vulkan documentation I see nothing where the callback should be pinned to a single thread; therefore, I don't see an issue accepting this.

PR #1279, opened 2019-11-21
Prevent building a GraphicsPipeline with a NULL handle.

jbit:

I've found that the Adreno driver on Android returns a NULL handle from `vkCreateGraphicsPipelines` if invalid shaders are passed to a pipeline (for example, shaders that exceed the limits of GPU). The current behavior of Vulkano is to ignore this and create a Vulkano `GraphicsPipeline` instance with a NULL handle, this causes lots of weird and hard to debug issues later on in the application including crashes inside the graphics driver.

This change simply adds a panic if this case is detected, so the rest of the program cannot proceed with an invalid `GraphicsPipeline`.


AustinJ235:
Although it doesn't hurt to have this check I feel like things such as limits should be checked before this stage perhaps when doing the ``Shader::load()`` function we should check that the generated code is compatible with the device passed.

PR #1287, opened 2019-12-26
Make Runtime-Shader Example work for the Version 0.16

0xReself:
I was testing the example and saw that the Swapchain Constructor dont need a Colorspace as an Parameter only an old Swapchain so i fixed that


0xReself:
I checked the docs for the older versions but even there the swapchain wants an old swapchain and not a ColorCode. I hope im not mistaking anything

AustinJ235:
The current git does. Thanks for checking though

PR #1289, opened 2020-01-08
I added a from_cstrings method on RawInstanceExtensions to...

SamHSmith:
...ease integration with windowing api's. My current solution to integrate glfw in my project has been to create my own RawInstanceExtensions struct and transmuting it to a vulkano RawInstanceExtensions. This solution is very ugly so I added the helper function and some documentation in the code. 

The function is unsafe to make it clear to the user that they need to be careful while using it. I know Vulkano is about creating a safe api but this should be unsafe since misuse could cause odd behaviour.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Added RawInstanceExtensions::from_cstrings(Vec<CString>) - which allows for easier integration with windowing api's like glfw
 - Added Swapchain::surface() - which returns the saved surface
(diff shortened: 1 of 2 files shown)


SamHSmith:
My change is not the cause of the checks failing so, I will conduct and investigation of why.

AustinJ235:
This seems redundant. ``RawInstanceExtensions`` current has the function ``pub fn new<I>(extensions: I) -> Self where
    I: IntoIterator<Item = CString>, ``

Assuming you want to pass ``Vec<CString>`` why not do ``RawInstanceExtensions::new(my_vec_of_cstrings.into_iter())``?

SamHSmith:
Very good point. There is no point for this feature and I will close the pull request.

Issue #1265, opened 2019-10-28
Pipeline barriers are not inserted in the right places due to stale hashmap key


Xenopathic:
If the same resource is used in two commands, Vulkano will correctly put a pipeline barrier between the commands. If it is used in a third command however, Vulkano will not put a second pipeline barrier in. This can be easily demonstrated by dispatching the same compute shader and descriptor set multiple times.

The problem lies in the `resources` hashmap, used by `SyncCommandBufferBuilder::prev_cmd_resource()`. Vulkano only decides to insert a pipeline buffer (and flush the command queue) if the collision command ID is greater or equal to the first unflushed command ID, i.e. the colliding command has not been flushed. However, the colliding command is inferred from the hashmap, which references the *first* command that used a given resource, so after the first flush for a resource, any future uses of the command are incorrectly assumed to have already been flushed, so pipeline barriers are not inserted.

One fix is to update the command ID inside the hashmap key to the latest command ID (which, by construction, also conflicts with the resource), but this doesn't seem super clean. I think this code needs a rethink and a refactor to move the command ID into the value, rather than the key.

AustinJ235:
I have confirmed that a long standing bug in my Basalt framework is caused by this: https://github.com/AustinJ235/basalt/issues/6

I have already had a look at the code to see if I can understand what is going on. I agree with the author that some rethinking needs to be done with this code.

In order for me to fix this issue I will need to understand vulkan a little better to create a proper fix. If anyone has more experience in this area pull requests are welcome.

AustinJ235:
I have been going through the code some more learning the process of how this stuff works. I think just updating the command id is out of question. 

From what I understand right now when a command adds a resource that already exists it creates a key, but the key isn't very unique so it grabs whatever entry had that resource first which is going to contain the command id of whichever command created the key in the first place. So what is happening here is that when a second command uses the resource it doesn't update the command id at all so when a third one is added it doesn't know to insert a barrier because as far as it is concerned there are no conflicting commands in the pending commands.

So in reality command_id should really be a list of all the commands that this resource has been used in. If it were a value it would be a matter of just changing the types around, but since it is a key it kinda makes matters a bit more difficult.

We could still use a hashmap but instead of a key based on a struct of what really should be a value we could use the conflict_key() function of the resources to be used a key. This would still allow for a more optimized route over just a list of these structs.

[...]

PR #1290, opened 2020-01-11
Fix issue with pipeline barriers not being inserted

AustinJ235:

Fixes: https://github.com/vulkano-rs/vulkano/issues/1265

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,16 @@ language: rust
 jobs:
   fast_finish: true
   include:
-    - os: windows
-      env: PYTHONPATH=C:\Python38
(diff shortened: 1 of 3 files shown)


AustinJ235:
When converting to a CBKey I drop the list of command_ids and only use the first one. I am not sure if this is important that I retain these or not. Currently this would be the same as the current behavior.

https://github.com/vulkano-rs/vulkano/pull/1290/files#diff-4eeb919345afc97f16ffc03e0fc5843eR255

I was wondering if any other members @rukai @freesig or even the man himself @tomaka have a comment on this patch.

Issue #1294, opened 2020-01-18
Validation layers say Vulkano is submitting a single-use command buffer more than once


tolziplohu:
I'm trying to upload data to an image that I'm also using for rendering. Ideally I shouldn't have to worry about sync since I'm only using one queue, but unless I signal a fence and wait for it after rendering and before uploading, I get an `AccessError: AlreadyInUse`. If I don't also wait for the GPU after uploading, the application seems to work as intended but the validation layer I have active complains:

```
VUID-vkQueueSubmit-pCommandBuffers-00071(ERROR / SPEC): msgNum: 0 - VkCommandBuffer 0x55df70cb2f80[] is already in use and is not marked for simultaneous use. The Vulkan spec states: If any element of the pCommandBuffers member of any element of pSubmits was not recorded with the VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT, it must not be in the pending state (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkQueueSubmit-pCommandBuffers-00071)
    Objects: 1
        [0] 0, type: 6, name: NULL
UNASSIGNED-CoreValidation-DrawState-CommandBufferSingleSubmitViolation(ERROR / SPEC): msgNum: 0 - VkCommandBuffer 0x55df70cb2f80[] was begun w/ VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT set, but has been submitted 0x2times.
    Objects: 1
        [0] 0, type: 6, name: NULL
```

Here's some pieces of my code, I'll try to make a minimum working example when I have time.

```rust
        // Draw
        let command_buffer =
            AutoCommandBufferBuilder::primary_one_time_submit(win.device(), win.queue.family())
                .unwrap()
                .begin_render_pass(frame.framebuffer, false, clear_values.clone())
                .unwrap()
                .draw(
                    pipeline.clone(),
                    &win.dynamic_state,
                    BufferlessVertices {
                        vertices: 4,
                        instances: 1,
                    },
                    desc.clone(),
                    pc,
                )
                .unwrap()
                .end_render_pass()
                .unwrap()
                .build()
                .unwrap();
        let f = future
            .join(frame.acquire)
            .then_execute(win.queue.clone(), command_buffer)
            .unwrap()
            .then_swapchain_present(win.queue.clone(), win.swapchain.clone(), frame.image_num)
            .then_signal_fence_and_flush();

[...]

PR #1313, opened 2020-01-28
Stop Join Future from doing the same submit multiple times.

KillianDS:

The implementation of `build_submission `in JoinFuture called queue submits of joined futures directly. Since the `build_submission` contract allows it to be called mutiple times that could lead to submits being duplicated. Amongst others this leads to vulkan validation layer errors as seen in #1294, which should be fixed with this.

This fix replaces the direct submit calls by a flush of the appropriate joined futures. This is similar to the approach taken in `PresentFuture` and as far as I can tell is in line with the contracts of `build_submission` and `flush`. I ran all tests and examples to do some basic validation and saw no issues.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -14,6 +14,7 @@
 - Added Swapchain::surface() - which returns the saved surface
 - Propogate new lines correctly in shader compile errors.
 - `Queue` and `QueueFamily` now implement `PartialEq` and `Eq`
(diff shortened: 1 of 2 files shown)


AustinJ235:
Good catch I verify when I get home.

AustinJ235:
The explanation in the comment seems valid and the comment from who last touched says:

> // TODO: review this function

so there was most likely some uncertainty there. Again thanks!

PR #1314, opened 2020-01-29
Add Eq and Hash implementations to types used in draw calls

### LINKED ISSUE #1309: Implement Eq and Hash for buffers and images
I would like to have these traits implemented on all buffer and image types in Vulkano. It's often useful to group render batches by texture or buffer. To allow for this, they should be allowed to be used as the key in hashmaps, or inserted into hashsets. These traits should be implemented not just on the concrete buffer and image types, but also ideally on `dyn BufferAccess` and `dyn ImageAccess`, so that these trait objects can also be used in hashable contexts.

I can give implementing this a try, but I'm not sure what would be a suitable source of an equality test/hash. `BufferAccess` and `ImageAccess` both have methods to return an inner unsafe object, which implements `VulkanObject` and thus can give you a raw Vulkan object handle. This seems like it would be suitable, but I'd like to have verification by someone who knows Vulkan better than I do.

Rua:
Fixes #1309

* [ ] (N/A) Updated documentation to reflect any user-facing changes - in this repository
* [ ] (N/A) Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -17,6 +17,7 @@
 - `Swapchain::acquire_next_image()`` now returns ``(image_id, suboptimal, aquire_future)``
     + *suboptimal indicates that the swapchain is usable, but should be recreated*
 - Fixed Join Future implementation to not submit joined command buffers twice.
+- Added `PartialEq`, `Eq` and `Hash` implementations for `Device`, `dyn BufferAccess`, `dyn ImageAccess`, `dyn ImageViewAccess` and various buffer and image types.
 
 # Version 0.16.0 (2019-11-01)
 
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -17,7 +17,12 @@
 - `Swapchain::acquire_next_image()`` now returns ``(image_id, suboptimal, aquire_future)``
     + *suboptimal indicates that the swapchain is usable, but should be recreated*
 - Fixed Join Future implementation to not submit joined command buffers twice.
-- Added `PartialEq`, `Eq` and `Hash` implementations for `Device`, `dyn BufferAccess`, `dyn ImageAccess`, `dyn ImageViewAccess`, `GraphicsPipeline` and various buffer and image types.
+- The traits `GraphicsPipelineAbstract` and `DescriptorSet` now require `DeviceOwned`.
+- Added `PartialEq`, `Eq` and `Hash` implementations to all types involved in a draw call, including:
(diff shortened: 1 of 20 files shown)


Rua:
I wanted to include graphics pipelines in the changes, but the `GraphicsPipelineAbstract` trait doesn't actually allow one to access the device, which is necessary for comparing two objects. So for now only the base `GraphicsPipeline` type implements the traits.

AustinJ235:
Haven't look at any of your changes so far, but I think with buffers there is a possibility that the underlying buf would be the same but the offsets different and length be different

AustinJ235:
As for the abstract pipeline do the non abstract versions have access to the device if so then it would be as simple as adding another function to the trait ao we could access the device from the trait object

Rua:
The buffers use `BufferInner` as the basis for comparison, which includes the offset. It doesn't include the size, though, which is a good point. I can add a `size` member to `BufferInner` to remedy that.

Regarding `GraphicsPipelineAbstract`, I think the trait could simply be made to extend `DeviceOwned`. I don't know if that would break anything though?

Rua:
I've gone ahead and implemented these extra changes. Please let me know if there is anything problematic.

AustinJ235:
Thinking if checking the device is enough. I have seen just checking the device in other areas of the code, but if there were say multiple instances utilizing the same physical device would these implementations hold up? The only changes that would need to be made is for the `PartialEq` for `Device`. The questions to ask would be can an application have multiple instances? Is the `vk::Device` unique to that `Instance`? Otherwise, I think we are good to go here.

Rua:
I don't think they can share a device, but device handles are just handles, and it's perfectly possible for each instance to hand out the same handle. I've added a check for Instance now too.

Rua:
While writing code that makes use of this, I noticed that `DescriptorSetsCollection` requires the `Send` and `Sync` traits, but `Arc<dyn ImageViewAccess + Send + Sync>` can't be put in a hashmap. Apparently the addition of the extra traits means they no longer implement `Eq` and `Hash`. Are there any solutions to this?

AustinJ235:
well they aren't included in the trait object so `Arc<dyn ImageViewAccess + Send + Sync + PartialEq + Eq + Hash>`, that can get quite lengthy so you might want to typeded it.

Rua:
That's not allowed, only auto-traits are allowed as additional constraints on trait objects. So the only practical solution I see is to change the existing implementations of `PartialEq`, `Eq` and `Hash` to be for `dyn ImageViewAccess + Send + Sync`.

PR #1326, opened 2020-02-15
Add VertexMember implementations for cgmath types

jmi2k:

Still very WIP, it fulfills the same needs of #1133. I've implemented it only for cgmath because it's what I'm using currently, but I'd have no problem extending this to, for example, nalgebra types.

The former PR replaces vertex members of the examples with types from cgmath. I think they should be self-contained, and should run without more external dependencies, so I kept them as-is.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -5,6 +5,8 @@
 - **Breaking** Update dependency `winit` 0.21 -> 0.22
 - Update dependency `half` 1.4 -> 1.5
 - Update dependency `smallvec` 0.6 -> 1.2
+- Add an optional feature, `cgmath-vertex` (which requires cgmath 0.17), which allows common cgmath types to be used inside vertex data definitions
+    + Types supported: `Basis`, `Deg`, `Point`, `Quaternion`, `Rad`, `Vector`, `Matrix`
 
 # Version 0.17.0 (2020-02-09)
 
--- a/vulkano/Cargo.toml
+++ b/vulkano/Cargo.toml
@@ -14,9 +14,13 @@ build = "build.rs"
 
 [dependencies]
 crossbeam = "0.7"
+cgmath = { version = "0.17", optional = true }
 fnv = "1.0"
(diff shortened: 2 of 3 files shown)


louisgjohnson:
this would be great for https://github.com/rustsim/nalgebra as well

jmi2k:
I have no prior experience with nalgebra, but I'll give it a try next week. I was waiting for #1327 to be merged, but it depends on a Rust issue to be solved first. I'll go on with this PR, and hold the other back until there is a reasonable way of implementing it without resorting to undefined behavior.

louisgjohnson:
@jmi2k that's fine, I think getting cgmath would be a huge win for now anyway

jmi2k:
It seems that `nalgebra` types can have various storage types, and their layout is undefined: there is no restriction on their contents (they can be structs with metadata) or their shape (they can be non-contiguous, or not ordered).

I believe they have a subset of storage types which can guarantee being only a wrapper of primitive types, and the implementation could be parametized, but I don't have any prior experience with nalgebra and their type-level machinery is a bit convoluted, so I'll leave it out of this PR. @louisgjohnson open an issue once this is merged, `nalgebra` looks very complete and packed with useful features, definitely worth supporting.

AustinJ235:
I am not personally sure how well maintained cgmath is. I have switched my projects over to nalgebra personally. cgmath last release was in January of 2019.

AustinJ235:
I think I am not going to go through with this pull request due to cgmath being a bit out of date. I don't know if you want to open another for nalgebra or use this one, up to you. We should also remove cgmath use from the examples and replace it with the more favorable nalgebra.

AustinJ235:
Although support for other libraries is nice to have, sadly cgmath seems to have gone inactive and this would bring in soon to be dead code to the project that will end up being removed.

I am open to this if you decide to implement nalgebra types, but since this pull request changes seem all related to cgmath I am closing this one. You are welcome to implement this for nalegbra and that'll likey go through. nalgebra seems to be the way forward.

PR #1327, opened 2020-02-15
Add vulkano-derive with custom derive for Vertex

jmi2k:

WIP. Currently it does the same as `impl_vertex!` but as a derive macro. There is a lot of room left to improve the code, and even use features only available to procedural macros.

### DIFF
--- a/vulkano-derive/Cargo.toml
+++ b/vulkano-derive/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "vulkano-derive"
+version = "0.17.0"
+edition = "2018"
+authors = ["Pierre Krieger <pierre.krieger1708@gmail.com>", "The vulkano contributors"]
+repository = "https://github.com/vulkano-rs/vulkano"
+description = "Custom derive for vulkano traits"
+license = "MIT/Apache-2.0"
+documentation = "https://docs.rs/vulkano"
+homepage = "https://vulkano.rs"
+keywords = ["vulkan", "bindings", "graphics", "gpu", "rendering"]
+categories = ["rendering::graphics-api"]
+
+[lib]
+proc-macro = true
+
+[dependencies]
+quote = "1.0"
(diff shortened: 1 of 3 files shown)


jmi2k:
Right now an implementation of `Default` is required for the vertex struct, in order to make the macro work. It cannot be removed AFAIK because we use a dummy instance of it to calculate offsets for each of the members, which is impossible to do right now without an existing value of the struct (there are attempts out there using raw pointers and uninitialized data, but right now they rely on undefined behavior more or less).

I'd like to remove that restriction in the future, but right now we'll have to live with it.

vulkano-derive/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "vulkano-derive"
+version = "0.17.0"
```
AustinJ235, line 3: Vulkano 0.18 is the current release.

AustinJ235:
Aside from the mismatched version there are some issues with adding another crate. Version publishing is automatically handled by travis ci, so that'll need to be updated via the travis config to publish that also.

Eliah-Lakhin:
Perhaps, the goal if this PR is already at least partially covered by my other PR https://github.com/vulkano-rs/vulkano/pull/1460 that now allows custom derives in existing macro(from vulkano-shaders crate).

Also to mention, that now we(maintainers) have capabilities to maintain crates.io publishings directly(thanks to Tomaka, the owner of the project), so the issue with the side crates is probably not that big deal now, however I would suggest to avoid of adding additional crates to the project ecosystem whenever it possible. It might complicate maintainability.

There were no updates in the Pull Request nor comments about the development progress from the author( @jmi2k ) quite a while. It seems like the development was abandoned. As such I would like to close the PR for now. If the development continues please reopen the Pull Request again. We are glad to accept any infrastructure improvements that could benefit Vulkano user community experience.

Issue #914, opened 2018-01-28
dynamic work group size in compute shaders?


arianvp:
You now need to define  the local work group size twice:

once in the shader:
```
layout(local_size_x = 16) in;
```

And once in the kernel invocation:
```
        builder.dispatch([dimensions[0] / 16, dimensions[1] / 16, 1],
                      self.pipeline.clone(),
                      self.next_set(input),
                      ())

```

Apparently glsl (and vulkan) has the concept of a `local_size_x_id` where you then pass the size information at runtime thorugh `VkSpecializationMapEntry` . Such that you don't have to hardcode the size in the shader. However, I can't find support for this in vulkano

See here for an example:
https://www.reddit.com/r/vulkan/comments/5eqgly/has_anybody_got_a_working_example_of_work_group/

arianvp:
Is this the accompanying behaviour in vulkano? https://docs.rs/vulkano/0.7/vulkano/pipeline/shader/trait.EntryPointAbstract.html#associatedtype.SpecializationConstants

tomaka:
Whenever you compile a shader with the vulkano-derive crate, a strict that implements the trait you linked will be generated. You will then have to pass an instance of it when building the pipeline.

I haven't tested if that works with compute shaders invocation size, but I don't see why it shouldn't.

arianvp:
Ahh cool. What is the name of this struct?

tomaka:
It's named `SpecializationConstants`. I thought there was an example of this somewhere, but I can't find it unfortunately.

arianvp:
I'll see if I can turn my experiments in some examples for in the repo.

PR #1338, opened 2020-04-03
Anonymous specialization constants support

### LINKED ISSUE #1320: vulkano-shaders panics on local_size_x_id
*   Version of vulkano: 18189e07
*   OS: Arch Linux
*   GPU (the selected PhysicalDevice): AMD RX 580
*   GPU Driver: amdgpu
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue:
```rust
mod cs {
    vulkano_shaders::shader! {
        ty: "compute",
        src: "
#version 450

layout (local_size_x_id = 0) in;

void main() {
}"
    }
}

fn main() {
}
```

# Issue

I wanted to specify the workgroup size via specialization constants, but vulkano-shaders panics when I try to specify `local_size_*_id`:

```
error: proc macro panicked
  --> src/main.rs:2:5
   |
2  | /     vulkano_shaders::shader! {
3  | |         ty: "compute",
4  | |         src: "
5  | | #version 450
...  |
10 | | }"
11 | |     }
   | |_____^
   |
   = help: message: called `Option::unwrap()` on a `None` value
```

Both glslangValidator and glslc compile this shader fine.

Eliah-Lakhin:
Fixes Vulkano Shaders bug when compute shader local group layout values bound to specialization constants.

Prior to this fix it was not possible to redefine local group size of compute shader in run time. Attempts to do so lead to macro compilation errors.

Now it is possible to define the layout in form of
```glsl
layout(local_size_x_id = 12, local_size_y_id = 13, local_size_z = 1) in;
```
and then set the values as
```rust
SpecializationConstants {constant_12: 8, constant_13: 4, ...}
```
during Compute Pipeline creation.

This patch fixes #1320 and #914

Tested with `AMD RADV VEGA M`, Ubuntu 18.04.4, Vulkan 1.2.131.


AustinJ235:
Is this a standard way for the names. Seems weird to me that you'd set the variable names to the values instead of a key.

Eliah-Lakhin:
@AustinJ235 Well, glsl requires integer literals in local size layout definition. Even for the `local_size_<x>_id` fields. Unfortunately there is no way to specify constant name as far as I know. As such shaderc doesn't provide constant name as well and vulkano simply skipping such constants when building `SpecializationConstants` struct because a field name cannot start with a number in Rust. My suggestion is to consider `constant_<x>` as Vulkano interface convention. It doesn't violate Vulkan/GLSL standard anyway.

AustinJ235:
Thanks I just wanted some clarification. Seems reasonable enough. I think you may need to push a dummy commit to re-trigger travis ci. It was offline for a bit and acting funny, so It isn't aware of this pull request.

Issue #1318, opened 2020-02-03
ImageUsage supported by surface but not by PhysicalDevice with some format combinations


ghost:
*   Version of vulkano: git master
*   OS: Linux (Pop!_OS 19.10)
*   GPU (the selected PhysicalDevice): Intel(R) UHD Graphics 630 (Coffeelake 3x8 GT2)
*   GPU Driver: i915
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: https://gist.github.com/heartles/a79e732928735211b10d688ad7ec5e50

The uploaded file is a slightly modified `examples/src/triangle.rs`. The only changes are that validation layers are enabled, and an extra line of code that is commented out at line 142. If this line is uncommented, the issue disappears.

# Issue

When I attempt to run the triangle example with validation layers on using integrated graphics, I see the following error appear in my logs:

```VUID-VkImageViewCreateInfo-usage-02275(ERROR / SPEC): msgNum: 0 - vkCreateImageView(): pCreateInfo->format VK_FORMAT_B8G8R8A8_SRGB with tiling VK_IMAGE_TILING_OPTIMAL does not support usage that includes VK_IMAGE_USAGE_STORAGE_BIT. The Vulkan spec states: If usage contains VK_IMAGE_USAGE_STORAGE_BIT, then the image view's format features must contain VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-usage-02275)```

This error appears right as I create a swapchain. I can see from API logs that `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT` is not set in the output of `vkGetPhysicalDeviceImageFormatProperties`

After looking at the Vulkan documentation and my API logs, I've found out that the supported ImageUsage flags for any given Format can be different than the supported usages as reported by `vkGetPhysicalDeviceSurfaceCapabilitiesKHR`. The supported usages are defined by the intersection of the surface capabilities and the physical device properties as reported by [vkGetPhysicalDeviceImageFormatProperties.](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#_surface_queries)

[...]

PR #1339, opened 2020-04-04
new function to check image/device compatibility

jeffw387:

This is in reference to issue #1318. 

I added a new function to Device to check if the given image and related configuration will be compatible with the Physical Device. It is called during swapchain creation and if required, returns an error stating the image configuration is not compatible with the Physical Device.

Let me know if you feel like anything needs work!


jeffw387:
I did test it on my machine (Intel Skylake laptop) and it returns the expected error if I enable the storage usage for the swapchain images. Then it creates the swapchain without complaint if I remove the storage usage.

PR #1350, opened 2020-04-28
Ray tracing (NV and KHR extensions)

bwrsandman:

I'm proposing these additions to vulkano. I'm setting as draft because I would like some help/feedback. I'm fairly new to Rust and there may be some tricks I don't know about.

There are five main parts to this addition:

### Extension
There are two extensions which allow ray tracing: `khr_ray_tracing` and `nv_ray_tracing`.
The NV extension is likely only available with select NVidia cards. I have a 1070 which supports this on Linux so it is not specifically limited to the RTX line of cards.

The KHR extension is newer and should have wider adoption in the future. As it is now, for NVidia, it requires a beta driver to test. I have not finished implementing this.
The two extensions differ slightly but I wrapped builders to have a unique API. This is done with a `bool` but maybe it could be better with a Generics.

### Ray Tracing Pipeline
This is different from both Compute and Graphics pipeline.
There can be multiple stages of the Ray Tracing Pipelines of the same type, for example, there can be two miss shaders.

Additionally, there is a concept of shader groups. Ray Generation, Miss and Execution Shaders are alone in their vulkan groups. Actual groups include optional Closest Hit, Any Hit and Intersection Shaders. There can be more than one of these groups.

If a group has an Intersection Shader, the geometry is procedural, otherwise, the geometry should be triangles. This automatically handled in this PR.

This has been challenging to me as I don't know of a way to make variable length generics or to wrap the builder stages in traits. It would be nice to have help here.

In the sample, I create procedural geometry and I would like to add a triangle group but this requires a solution to the variable length of generics.

### Acceleration Structures

To ray trace, the pipeline must have access to the full scene in the trace command. Vulkan offers acceleration structures (a bounding volume hiearchy of axis-aligned bounding boxes) to quickly traverse the geometry (bounding boxes for procedural geometry or triangles).

For one scene representation, Vulkan offers multiple bottom level acceleration structure for different meshes and only one top level acceleration structure to group them.

A static scene only needs to be built once but a dynamic scene requires a rebuild with the `build_acceleration_structure` command at each change.

In this PR, `AccelerationStructureBuilder` automatically creates the top level acceleration structure and all lower level structure created with `add_aabbs` for procedural geometry and `add_triangles` for triangle meshes.

The structure requires memory for storage and a scratch buffer for rebuilding.

Adding triangles is not yet implemented.

### Shader Binding Tables

The shader binding table is a pointer list the shaders use to call each other.
In this PR, the table has to be built and uploaded in a buffer manually, but it might be possible to do this automatically in the pipeline. Different `trace` calls can use different tables and offsets.

### nv-ray-tracing example
nv-tray-tracing example


Eliah-Lakhin:
@bwrsandman Any updates regarding the progress on this feature?

It's hard to find volunteers to help with development unfortunately, but the feature would be definitely very beneficial for Vulkano infrastructure.

bwrsandman:
I haven't progressed as I require a bit of help / review on the the mentioned points.

MarijnS95:
> * Regarding the difference between NV and KHR variants, I suggest doing only KHR first, and adding NV in a separate PR later. While I don't think Vulkano has anything against vendor-specific functionality, it is not as important as KHR and EXT.

The KHR extension has wide enough adoption that it's wasteful to support NV, unless you wish to use Vulkano on one specific ancient Nvidia driver. The semantics changed a bit in some key areas, too.

PR #1355, opened 2020-05-05
Extending descriptor set builders to support sampled image descriptor arrays

leftadjoint:
Vulkan/GLSL supports variable-length descriptor arrays (i.e. `sampler2D textures[]`), however it isn't currently possible to decide the length of the array at runtime using standard vulkano constructs.

This is distinct from an 'array texture', i.e. a GLSL `sampler2DArray`. These textures must all be the same dimensions, whereas the images in a descriptor like `sampler2D textures[]` can vary in dimensions and most other attributes.

Note: While image samplers can be arrayed like this, I'm not certain exactly which other descriptors are supported as I haven't checked the spec.

The primary barrier to trying to work around this is the typed builder pattern (i.e. [`PersistentDescriptorSetBuilder`](https://docs.rs/vulkano/0.18.0/vulkano/descriptor/descriptor_set/struct.PersistentDescriptorSetBuilder.html)). Each additional descriptor added to the builder changes the type, so there is no way to add multiple descriptors in a loop or iterator.

To address this, this change adds the following struct:
  - `PersistentDescriptorSetImgArray`

This serves the same function as an underlying type for the builder to manage an array of images, analagous to `PersistentDescriptorSetImg`.

This change also adds the following method to [`PersistentDescriptorSetBuilder`](https://docs.rs/vulkano/0.18.0/vulkano/descriptor/descriptor_set/struct.PersistentDescriptorSetBuilder.html)) and [`FixedDescriptorSetBuilder`](https://docs.rs/vulkano/0.18.0/vulkano/descriptor/descriptor_set/struct.PersistentDescriptorSetBuilderArray.html), along extending [`FixedSizeDescriptorSetBuilder`](https://docs.rs/vulkano/0.18.0/vulkano/descriptor/descriptor_set/struct.FixedSizeDescriptorSetBuilder.html) and it's array counterpart.
  - `add_sampled_image_array(mut self, image_views: Vec<T>, sampler: Arc<Sampler>)`

Which adds the array to the builders and performs the appropriate book-keeping.

Usage of `add_sampled_image_array` with a dynamically-sized array in a shader (i.e. `uniform sampler2D textures[]`) can be unsafe, as vulkano-shaders has no way of determining the correct number of elements -- at the moment it appears to produce an array length of 1 in it's descriptor output, which should generate an error if an array of length greater than 1 is passed.

This means using `add_sampled_image_array` with a dynamically sized array requires constructing an [`UnsafeDescriptorSetLayout`](https://docs.rs/vulkano/0.18.0/vulkano/descriptor/descriptor_set/struct.UnsafeDescriptorSetLayout.html) separately from the length of the array being passed at runtime. This should be a sufficient warning to the user that they must determine the length of the array correctly.

If used with a statically-sized array  (i.e. uniform sampler2D textures[N]) in the shader it should be safe; providing an incorrect length to the descriptor set will produce an error.

This change also provides an additional example using based on the image example, `image-array`.

Utilizing dynamically (but not statically) sized arrays in a shader requires the GL_EXT_nonuniform_qualifier GLSL extension, which currently breaks the vulkano-shaders parser. Until this is fixed, it's necessary to load the SPIRV separately.

A related topic for consideration is that eventually 'array textures' (`sampler2DArray`) may be added to vulkano, and how a builder function for that should be named to distinguish it from `add_sampled_image_array`.


AustinJ235:
I feel like we should include changes to vulkano-shaders at the same time as adding this. It might give users the idea that this is more supported then it is. Overall this seems like a welcome change.

Eliah-Lakhin:
@leftadjoint Any updates on the feature? Do you need a hand? I probably will need this feature in my project too, so will be glad with coding if needed.

leftadjoint:
> @leftadjoint Any updates on the feature? Do you need a hand? I probably will need this feature in my project too, so will be glad with coding if needed.

@Eliah-Lakhin I think the simple issue at this point is to update vulkan-shaders to work with `GL_EXT_nonuniform_qualifier`, but that requires better SPIRV and vulkano-shaders internals knowledge than I have at the moment. I'm hoping to get a little more time to dig into that part in about a month.

Rua:
@leftadjoint do you know which part of `GL_EXT_nonuniform_qualifier` is currently causing trouble for Vulkano-shaders? I'd like to have a look at it.

Rua:
Reading about this some more, I see several distinct concerns going on here:

1. The ability to have an array of samplers in a shader (`sampler2D foo[42]`), and the ability to provide bindings for such arrayed descriptors from the Vulkano side. This requires no features or extensions.
2. The ability to index such an array with a dynamically uniform value in the shader instead of only a constant expression. This needs the `shaderSampledImageArrayDynamicIndexing` feature enabled in Vulkan.
3. The ability to declare the array without a compile-time size (`sampler2D foo[]`). This requires the `GL_EXT_nonuniform_qualifier` extension in the shader.
4. The ability to do further fancy things in the shader, enabled by the `VK_EXT_descriptor_indexing` Vulkan extension, or core 1.2.

These items seem to be roughly on a scale of simple/commonly supported to difficult/less support. I would suggest trying to get point 1 working first in a PR, and then progressively implementing each point in a separate PR. It's better to have something than nothing at all IMO.

faulesocke:
@Rua I've just tried 1 and 2 and they already work in the current master.

Rua:
Huh, but then what's the `PersistentDescriptorSetImgArray` struct for, that this PR adds?

faulesocke:
Maybe for 3? I was just able to create this descriptor set:
```rust
        let desc_set = self
            .dir_ds_pool
            .next()
            .add_sampled_image(self.diffuse_input.clone(), self.sampler.clone())
            .unwrap()
            .add_sampled_image(self.normals_input.clone(), self.sampler.clone())
            .unwrap()
            .add_sampled_image(self.position_input.clone(), self.sampler.clone())
            .unwrap()
            .enter_array()
            .unwrap()
            .add_sampled_image(shadow_views[0].clone(), self.sampler.clone())
            .unwrap()
            .add_sampled_image(shadow_views[1].clone(), self.sampler.clone())
            .unwrap()
            .add_sampled_image(shadow_views[2].clone(), self.sampler.clone())
            .unwrap()
            .add_sampled_image(shadow_views[3].clone(), self.sampler.clone())
            .unwrap()
            .leave_array()
            .unwrap()
            .add_buffer(light.uniform_buffer().clone())
            .unwrap()
            .add_buffer(camera_ub)
            .unwrap()
            .build()
            .unwrap();
```

For this shader:
```glsl
layout(binding = 0) uniform sampler2DArray gbuffer_diffuse;
layout(binding = 1) uniform sampler2DArray gbuffer_normals;
layout(binding = 2) uniform sampler2DArray gbuffer_position;

layout(binding = 3) uniform sampler2D shadow_maps[NUM_SHADOW_CASCADES];

layout(binding = 4) uniform Light {
    DirectionalLight light;
};

// TODO simplify: layout(...) uniform Camera cam;
layout(binding = 5) uniform Cam {
    Camera cam;
};
```

Rua:
Ahh I see now what's going on. In your code, you aren't able to add images in a loop, because each call on the builder changes its type. You have to add a fixed number of images at compile time, which isn't very convenient. This PR lets you provide a single `Vec` with all the array elements at once, which works better.

faulesocke:
I was aware of this limitation. Glad to see it fixed with this.

AustinJ235:
👀

Rua:
It would be great if the internals of descriptor set builders could be modified so that they no longer have any type parameters for the values they contain. Unfortunately, because the functions for adding resources take any type that implements a trait, the only solution I see for this is to put all these resources in a `Vec<Box<Trait>>` (or alternatively `SmallVec`), which never sits right with me.

I've been thinking about the possibility of designing the Vulkano API regarding resources like buffers and images a bit differently. Instead of taking any `T where T: Trait` as many functions do now, they would take either `Arc<T> where T: Trait` or even `Arc<dyn Trait + Send + Sync>`. This means that everything that implements traits like `BufferAccess`, `ImageAccess` or `DescriptorSet` must always be wrapped in an `Arc`, which removes the need for a `Box`. The descriptor set builder would then contain e.g. `Vec<Arc<dyn BufferAccess + Send + Sync>>` instead. This is of course a rather major design change in Vulkano so I'd appreciate feedback on this.

Most resource types in Vulkano already wrap themselves in an `Arc` when you create them, e.g. `ImmutableBuffer`. But there are a few exceptions like `BufferSlice`, `CpuBufferPoolChunk` or indeed `PersistentDescriptorSet`, which would need to be modified to produce an `Arc` as well.

AustinJ235:
I honestly wouldn't mind having everything wrapped in an `Arc`. If you look at of any of my stuff that is how I like to do things anyways. With most of vulkano already having most things in an `Arc` and doing things with traits, I don't think this would be awful. The above mentioned types probably commonly get wrapped into an `Arc` anyways by the user I'd assume.

I think this would be a very useful feature for anyone working with multiple texture/material sources.

I did find this, https://gist.github.com/DethRaid/0171f3cfcce51950ee4ef96c64f59617 which may be helpful for anyone implementing this. Specifically this pull request focuses on sampled image arrays, but this should be possible with any descriptor as long as it is the last one in a set.

Notably this seems to be fairly widely supported now with devices that support vulkan 1.2 (about 96.6% according to https://vulkan.gpuinfo.org/listfeaturescore12.php)

AustinJ235:
Closed in favor of https://github.com/vulkano-rs/vulkano/pull/1674

PR #1357, opened 2020-05-09
Dynamic Compute Shader local size

Eliah-Lakhin:
This PR is the second part of my work in making Compute Shader's local size to be computed in run time.

**Motivation**

When working with heavily Compute Shader computation tasks it is very important to set shader's local size properly in order to obtain maximal performance that particular GPU can provide. Failure to do so may lead to significant performance penalty.

Ideally local size cube volume must match the exact number of threads of the GPU. This number can vary between particular video card manufacturers. For example most Intel and nVideo cards provide 32 threads whereas AMD provides 64 threads.

My previous PR #1338 enables an option to specify shader's local size through specialization constants, but Vulkano still doesn't provide generic way to compute local size in run time based on device capabilities.

**Solution**

The number of available threads can be fetched from `VK_KHR_get_physical_device_properties2` Instance extension's substructure called `VkPhysicalDeviceSubgroupProperties`: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceSubgroupProperties.html

In the current implementation Vulkano already supports fetching of some of the properties of the physical device from the `STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2_KHR` structure(to collect metadata about Queue Families capabilities) whenever the appropriate Instance extension is enabled: https://github.com/vulkano-rs/vulkano/blob/f211a3af2ace62f7abf99ddbaf947eb1907dc084/vulkano/src/instance/instance.rs#L293 , but it ignores almost all possible substructures of this structure. In my PR I implemented fetching additional data from the mentioned substructure to get `subgroupSize` value and exposing this value as an `Option<u32>` of the Physical Device for the end-user(in safe way).

In addition I implemented `PhysicalDeviceExtendedProperties` high-level structure that is currently providing only this single property, but we can extend this future to fetch more options from `VK_KHR_get_physical_device_properties2` structure in the future depending on the needs.

In the end the end-user can access `subgroupSize` value this safe way:

```rust
let instance = Instance::new(None, &InstanceExtensions {
  khr_get_physical_device_properties2: true,
  ..InstanceExtensions::none()
}, None).unwrap();
let physical: PhysicalDevice = ...
let subgroup_size: Option<u32> = physical.extended_properties().subgroup_size();
```

Rustdoc for `PhysicalDeviceExtendedProperties` struct and it's function provided in this PR too.

Also in this PR I prepared an example(based on vulkano-www compute shader example) of this new feature that illustrates how to use both `subgroup_size()` function and features from PR #1338 to compute Compute Shader's local size in runtime.

**Test**

Tested with `AMD RADV VEGA M`, `NVIDIA GeForce GTX 1060 3GB` and `Intel(R) HD Graphics 630 (Kaby Lake GT2)` on Windows and Linux.


CHANGELOG_VULKANO.md
```diff
@@ -1,5 +1,11 @@
 # Unreleased
 
+- Provides new API to fetch additional information of the Physical Device from the
+  `PhysicalDeviceProperties2KHR` structure whenever it possible. In the current
+  implementation only `subgroupSize` property is fetching. This interface can be
+  extended in the future to obtain more metadata depending on community needs.
+- `dynamic-local-size` compute shader example added showing how to utilize
+  `subgroupSize` to compute and set shader's local size in run time.
```
AustinJ235, line 8: Extra white space at of of line here will prevent travis ci from returning successful.

Eliah-Lakhin: @AustinJ235 Fixed.

PR #301, opened 2016-10-05
Support for MoltenVK

tomaka:
cc #300

Remains to do:
- [ ] Expose the `ActivateMoltenVKLicenseMVK` function somewhere in vulkano.
- [x] Tweak `winit` to allow exporting the `NSView` or `UIView` of the window. For MacOS this is done by adding an additional function here: https://github.com/tomaka/winit/blob/41e1a768e575667a0dfff6bc9c8f56c30d662122/src/os/macos.rs#L9
- [x] Update `vulkano-win` to create surfaces on MacOS and iOS.
- ~~Check that it's a correct thing to do to dlopen the MoltenVK library.~~

After steps 2 and 3, it should "just work" if you put the `MoltenVK` file next to the executable. Step 1 is optional for now because the free trial doesn't require activating a license.

I'm not sure that I'll finish this PR soon, so feel free to submit PRs to the `moltenvk` branch if you want to work on it. It should be fairly easy to do steps 2 and 3 (feel free to ask questions).

I'm also unsure about whether we'll want to merge this branch until MoltenVK is stable enough. I don't really want to have obsolete code in `vk-sys` because they decided to modify their extensions.

PR #1358, opened 2020-05-11
Fix an `abort()` in the tests.

DiamondLovesYou:
Not sure how widespread this issue is, I'm on Mesa 20.0 currently. I've re-enabled a heap size check for RADV only because MoltenVK apparently has an issue with the reported heap sizes.

### DIFF
--- a/vulkano/src/instance/instance.rs
+++ b/vulkano/src/instance/instance.rs
@@ -867,14 +867,12 @@ impl<'a> PhysicalDevice<'a> {
 
     /// Returns the human-readable name of the device.
     #[inline]
-    pub fn name(&self) -> String {
-        // FIXME: for some reason this panics if you use a `&str`
+    pub fn name(&self) -> &str {
         unsafe {
-            let val = self.infos().properties.deviceName;
+            let val = &self.infos().properties.deviceName;
             let val = CStr::from_ptr(val.as_ptr());
             val.to_str()
                 .expect("physical device name contained non-UTF8 characters")
-                .to_owned()
         }
     }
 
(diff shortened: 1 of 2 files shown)


AustinJ235:
Should this be enabled for all drivers except for MoltenVK? This original change to disable this check was made in this pull request: https://github.com/vulkano-rs/vulkano/pull/301 part of enabling support for molkenvk. The issue first arised here: https://github.com/vulkano-rs/vulkano/pull/301#issuecomment-303151762. It seems that molkenvk then returned 0 for the current heap size instead of anything meaningful.

Maybe we could remove the specific radv check in favor of something that just checks to make sure the heap size is anything but zero before comparing. Also since this issue was from in 2017, when they had intentions of implementing this in molkenvk, this may have been done since. 

> MoltenVK does not currently use iOS Metal Heaps. They will be introduced in an upcoming release.

Source: https://moltengl.com/forums/topic/vkmemoryheap-size/

I would still be a bit hesitant on re-enabling in full due to my lack of possession of macos hardware, but I have reason to believe this is supported on MacOS.

DiamondLovesYou:
> Should this be enabled for all drivers except for MoltenVK? This original change to disable this check was made in this pull request: #301 part of enabling support for molkenvk. The issue first arised here: #301 (comment). It seems that molkenvk then returned 0 for the current heap size instead of anything meaningful.

> Maybe we could remove the specific radv check in favor of something that just checks to make sure the heap size is anything but zero before comparing. Also since this issue was from in 2017, when they had intentions of implementing this in molkenvk, this may have been done since.

Sounds fine to me.

PR #1359, opened 2020-05-11
Add support for the Physical Storage Buffer Access feature.

DiamondLovesYou:
This feature was recently promoted from an extension to the core spec in Vulkan 1.2. It is already supported on many recent drivers: NVidia, Mesa RADV, and AMDVLK.

Other than supporting enabling the feature and the basic buffer address query, I made no attempt to create safe interfaces for its use. Such will come later, after I've experimented with its use in Geobacter.


DiamondLovesYou:
@AustinJ235 Rebased.

PR #1361, opened 2020-05-11
Added more detailed explenation to docs

SamHSmith:
Added a more detailed explanation about the image copy to buffer process.


AustinJ235:
```
$ git diff-tree --check $(git hash-object -t tree /dev/null) HEAD
vulkano/src/command_buffer/auto.rs:993: trailing whitespace.
+    /* 
vulkano/src/command_buffer/auto.rs:995: trailing whitespace.
+    
The command "git diff-tree --check $(git hash-object -t tree /dev/null) HEAD" exited with 2.
```

PR #1367, opened 2020-05-17
Ran rustfmt and added a check to .travis.yml

VictorKoenders:
I was going to apply clippy suggestions throughout the project when I noticed that my editor automatically ran `cargo fmt` on all files I tried saving.

Assuming that most people use `cargo fmt`, I think it'll be good to keep the entire project properly formatted.

This pull request runs `cargo fmt` over the entire repository, as well as adding a check to `.travis.yml` to see if rustfmt has any suggestions.

There are several options to configure rustfmt, the easiest would be to add a `.rustfmt.yml` file in the root of the project. All options can be found [on rustfmt's github page](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md)

This pull request will probably break other open pull requests, but will not introduce any API or documentation changes. I figured it was not worth adding this to any of the changelogs.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,4 @@
 * [ ] Added an entry to `CHANGELOG_VULKANO.md` or `CHANGELOG_VK_SYS.md` if knowledge of this change could be valuable to users
 * [ ] Updated documentation to reflect any user-facing changes - in this repository
 * [ ] Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.
+* [ ] Ran `cargo fmt` on the changes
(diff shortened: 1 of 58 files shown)


AustinJ235:
Isn't it possible that rustfmt ran on nightly versus stable, and various versions on the host machine versus what is on travis ci may have differing suggests / formatting? I think that it is just best to run this time to time on the entire code base or users can opt to run it when the create pull requests. I don't think it should be a requirement.

VictorKoenders:
Rustfmt should :tm: always produce the same formatted code between versions and platforms, as long as the config in `.rustfmt.toml` stays the same

AustinJ235:
I think about it some. I am still a bit hesitant to add more things to travis ci. It is already annoying as is that the trailing whitespace cause the build to fail, and when they take upwards of 30 minutes that is the cache system doesn't time out.

I think I would be more open to a change like this once we have a more competent ci system in place other than travis ci. One thing that is keeping this project on travis ci atm is that it is the sole means of publishing versions, as tomaka is the only person with publish rights. He is a bit hard to get a hold of to say the least.

VictorKoenders:
Removed the travis check,  pulled in the latest master and ran rustfmt again

PR #1374, opened 2020-06-01
publish 0.19.0 & vk-sys 0.5.2

AustinJ235:


AustinJ235:
As https://github.com/vulkano-rs/vulkano/pull/1376 has been merged, no longer is this a minor version, so major it is. I wasn't too sure about the minor bump anyways with the additional image compatibility checks likely to cause some issues among users upgrading, so at least that'll be an implied update instead of one that comes along breaking their project unexpectedly.

Would still like to see the spec we use upgraded to 1.2, but some drivers are still a bit behind. I think it is safe to stay on 1.1 for the time being.

PR #1379, opened 2020-06-04
Add properties method to Format

Rua:

This adds a `properties` method to `Format`, which can be used to query the features a given device supports for images of this format.


AustinJ235:
Would it be better to call it FormatFeatures? I think that would make it less confusing for those coming from vulkan or are familiar with vulkan struct names.

Rua:
The Vulkan name of the struct is `vkFormatProperties`, and its three members are `VkFormatFeatureFlags`. So I tried to keep that naming and went with `FormatProperties` and `FormatFeatures`, omitting "flags" because that seems more like an implementation detail.

AustinJ235:
ah I was just glancing on it mobile. All is good then.

PR #1380, opened 2020-06-05
Return Send + Sync trait object from FramebufferAbstract

Rua:

This adds a `Send + Sync` bound to the trait object that gets returned from `FramebufferAbstract::attached_image_view` and `AttachmentsList::as_image_view_access`. This presumably means that framebuffers can now only be created from images that are `Send + Sync` themselves, but this requirement is met by all the Vulkano image types. So unless someone has made a custom image type that fails this requirement, it shouldn't break anyone's code.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Added a `properties` method to `Format`.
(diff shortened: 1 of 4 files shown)


Rua:
Well, after thinking about it more, I realise that this doesn't do what I wanted it to do. I wanted to be able to use the returned image elsewhere, but that would require returning an `Arc`, not a mere reference. So I'm closing this now, sorry for the trouble.

PR #1407, opened 2020-07-30
Copy buffer dimensions

Tetrel:

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -4,6 +4,7 @@
 - Added function to create surface from RawWindowHandle
 - Added a `properties` method to `Format`.
 - Added additional device feature flags for enabling SPIR-V related capabilities.
+- Added method `copy_buffer_dimensions` that allows copying parts of buffers containing arrays.
 
 # Version 0.19.0 (2020-06-01)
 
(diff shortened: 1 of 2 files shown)


Rua:
Could the function be named `copy_buffer_with_dimensions` instead? It's longer, but your name makes me think that it copies the dimensions of the buffer or something.

AustinJ235:
@Rua suggestion seems to make the method name confusing. Other than that this seems like a sound idea. This is already possible with buffer slices, but this way would be less combersum.

Tetrel:
Thank you for your response and feedback.
Yes, sure, changing the name is trivial. I chose the name in analogy to `copy_buffer_to_image_dimensions`, but since the buffers contain slices and the idea is to allow copying subslices I would rather suggest the name `copy_buffer_range` or `copy_buffer_with_range`.

Eliah-Lakhin (approved):

Eliah-Lakhin:
@Tetrel I merged the PR. Thank very much for your contribution! And I'm sorry for delay with review.

Personally I don't see any issues with proposed naming, but we can change it later if needed. We are still on the version behind `v1.0`, so changes in public API names would be fine I think.

Issue #1430, opened 2020-11-02
Make AutoCommandBufferBuilder::execute_commands safe to use


Rua:
This is marked as unsafe in the documentation, but it is a pretty important function to have. Can it be made safe to use, or at least have a good documentation on what a user needs to do in order to produce sound code? Specifically, if the command buffer provided as the argument is another `AutoCommandBuffer`.

AustinJ235:
I think it may be unsafe to the nature that fences are not automatically inserted like they are on the rest of the methods on `AutoCommandBufferBuilder`. If were were to change how fences are done, which is fairly complicated in its current state, it may be possible to check this. But in its current state, I think this isn't possible.

Rua:
For future implementors, these checks as detailed in [the documentation of `vkCmdExecuteCommands`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdExecuteCommands.html) still need to be implemented as of #1431:

* [ ] If any element of `pCommandBuffers` was not recorded with the `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` flag, and it was recorded into any other primary command buffer which is currently in the executable or recording state, that primary command buffer becomes invalid.
* [ ] Each element of `pCommandBuffers` **must** be in the pending or executable state
* [ ] If any element of `pCommandBuffers` was not recorded with the `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` flag, it **must** not be in the pending state
* [ ] If any element of `pCommandBuffers` was not recorded with the `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` flag, it **must** not have already been recorded to `commandBuffer`
* [ ] If any element of `pCommandBuffers` was not recorded with the `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` flag, it **must** not appear more than once in `pCommandBuffers`
* [ ] Each element of `pCommandBuffers` **must** have been allocated from a `VkCommandPool` that was created for the same queue family as the `VkCommandPool` from which `commandBuffer` was allocated

[...]

PR #1431, opened 2020-11-04
Some safety checks for AutoCommandBufferBuilder::execute_commands

Rua:

This is an attempt to improve the safety of the `execute_commands` method, by adding checks. I've implemented most of the checks listed in [the documentation of `vkCmdExecuteCommands`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdExecuteCommands.html), but not all of them.

Synchronisation was not implemented for secondary command buffers yet, as this is a bit lower-level stuff that I don't feel as comfortable working on. I would guess that `SyncCommandBufferBuilderExecuteCommands::submit` should add/expose the resources of the secondary command buffer to the top-level one, much like it already does with descriptor sets. However, it's unclear to me who is responsible for calling `lock_submit` on the secondary command buffers, the inner `Sync` command buffer or the outer `Auto`.

The documentation also mentions a few requirements on the execution state of the secondary buffer at the moment it's recorded to the primary one. Essentially, the requirements are the same as if one would execute the buffer, except that resources don't need to be locked, only the command buffer itself, and the lock can be released once `vkCmdExecuteCommands` returns. So the `CommandBuffer` trait would need some new methods to perform a lock/unlock without propagating it to the inner `Sync` buffer. However, once the command buffer has been handed over to the inner builder, it is no longer possible for the outer `Auto` to unlock it, unless we require that users wrap it in an `Arc` so that it can be cloned. So should this also become the responsibility of `SyncCommandBufferBuilder`?

Finally, none of the checks related to queries have been implemented, because `AutoCommandBufferBuilder` itself does not expose them. Whoever implements it in the future will have to add the appropriate checks to `execute_commands`, too.

In any case, as it stands now, the PR can be merged. It is usable in its current state, just not completely rid of unsafety unfortunately.


AustinJ235:
Well didn't see this when I made the comment on https://github.com/vulkano-rs/vulkano/issues/1430. Good work.

PR #1439, opened 2020-11-17
Add debug marker commands to AutoCommandBufferBuilder

dobrakmato:

---------------

I saw that there are no methods for recording the debug markers in `AutoCommandBufferBuilder` but the methods are implemented for `UnsafeCommandBufferBuilder`. This PR adds support for these methods:
- [x] `AutoCommandBufferBuilder::debug_marker_begin`
- [x] `AutoCommandBufferBuilder::debug_marker_end`
- [x] `AutoCommandBufferBuilder::debug_marker_insert`

Todo:
- [x] validate / document required extension
- [x] document safety for all `unsafe` fns
 
I am unsure about these things:
- is the `'static` lifetime bound on `&CStr` necessary? I looked up the function in [Vulkan Specification][1], but it says nothing about when the pointer should be valid.
- ~is `debug_assert!` the right choice or should we rather make the method return `Result`?~

@Eliah-Lakhin ?

[1]:https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap44.html#VkDebugUtilsLabelEXT

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -10,6 +10,7 @@
 - Added a `properties` method to `Format`.
 - Added additional device feature flags for enabling SPIR-V related capabilities.
 - Added method `copy_buffer_dimensions` that allows copying parts of buffers containing arrays.
+- Added `debug_marker_begin`, `debug_marker_end` and `debug_marker_insert` to `AutoCommandBufferBuilder`.
 - Fixed surface creation function on Windows(PR #1410).
 - Travis CI Linux Nightly job temporary disabled until #1423 resolved.
 
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -70,6 +70,7 @@ use pipeline::ComputePipelineAbstract;
 use pipeline::GraphicsPipelineAbstract;
 use query::QueryPipelineStatisticFlags;
 use sampler::Filter;
+use std::ffi::CStr;
 use sync::AccessCheckError;
 use sync::AccessFlagBits;
 use sync::GpuFuture;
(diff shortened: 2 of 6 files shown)


Eliah-Lakhin:
@dobrakmato Thank you for your contribution!

As of the second question, Vulkano in almost all cases returning `Result`. So, you need to introduce or enrich existing appropriate error object and return it in wrapped `Result` instead of `debug_assert`.

As of the first question, yes it seem to be tricky. Perhaps, exposing static `CStr` reference in the public interface would be the only option at this moment which is not preferable from API design point of view, and probably was a reason why this API wasn't exposed before. And even if the pointer's lifetime would be more clear in Vulkan's specs, it's still unclear where to store the CString ownership in Vulkano's structs. I'm not sure if Vulkano keeps `Command` ownership long enough too. But if you need this feature let's do it this way for now with `'static` lifetime. At least it would guarantee safety.

dobrakmato:
Hey, thank you for fast response!

I changed asserts to results + added some validation for arguments. I also thought that implementing of the `debug_marker_insert` fn should be probably part of this PR, so I will add that. 

Then I will squash, `cargo fmt` and add information to changelog. I will ping you then for review.

dobrakmato:
It should be ready for review @Eliah-Lakhin

CHANGELOG_VULKANO.md
```diff
@@ -10,6 +10,7 @@
 - Added a `properties` method to `Format`.
 - Added additional device feature flags for enabling SPIR-V related capabilities.
 - Added method `copy_buffer_dimensions` that allows copying parts of buffers containing arrays.
+- Added `debug_marker_begin`, `debug_marker_end` amd `debug_marker_insert` to `AutoCommandBufferBuilder`.
```
Eliah-Lakhin, line 13: @dobrakmato Fix a typo, please. The implementation looks good, I will merge it.

dobrakmato: oops I missed that one. fixed now.

Eliah-Lakhin:
@dobrakmato PR merged. Thank you! :)

PR #1443, opened 2020-11-27
Finish pipeline cache

### LINKED ISSUE #1442: Finish the implementation of the PipelineCache
I stumbled over [`vulkano::pipeline::cache::PipelineCach`](https://docs.rs/vulkano/0.19.0/vulkano/pipeline/cache/struct.PipelineCache.html) and noticed that it isn't fully implemented yet.

I dug through the code base and I think only a few parts are missing. A field in the *PipelineBuilder and a method to add a Cache to the builder and the call to the vk-sys function needs to be changed from 0 to the [`cache`](https://docs.rs/vulkano/0.19.0/src/vulkano/pipeline/cache.rs.html#42) field of the `PipelineCache`. The heavy work should be done by the Vulkan implementation, right?

If I got this right and you are interested I could try to implement this.

Gosuo:
Added the method build_with_cache to set an Option that holds an Arc
containing a PipelineCache. This could probably be optimized to just
hold a vk::PipelineCache(u64) and start with a vk::NULL_HANDLE and
set the value to the internal_object of the PipelineCache. Using the Arc
is probably safer for now, but it is an idea.

I'm not sure if anything needs to be changed in the guide. An example wouldn't be too bad, but I wanted to get feedback before I invest the time.

I haven't added the changes to the ComputePipeline as those would be breaking changes. There is no builder for the ComputePipeline afaik and thus you would have to pass the cache to the functions `new`, `with_pipeline_layout`, and `with_unchecked_pipeline_layout`. I'm not sure if that is desirable.

Fixes #1442.


Eliah-Lakhin:
@Gosuo Thank you for the quick solution!

This feature would definitely be useful. And as someone who are actively using `ComputePipeline` of Vulkano I wouldn't mind about implementation for Compute Pipeline too even with API breaking changes. In my opinion this breaking is not too bad, and the benefits we could gain from the cache worth it!

Please let me know when the PR is ready for final review. I will gladly accept it.

Gosuo:
@Eliah-Lakhin I would guess the implementation is done for now.

I would like to write some tests for the PipelineCache, but I'm not sure how to actually test this, as there is no real response from the PipelineCache of the implementation. I could test if it is possible to save a cache to a binary file and load a new one from there, but I don't know how to handle file creation and clean up after the tests.

Also, someone should check, if there are performance gains. I will see if I can test this, but it would take me some time.

vulkano/src/pipeline/cache.rs
```diff
@@ -14,15 +14,15 @@
...
+//! Note: This is currently only possible with the GraphicsPipelineBuilder
 //! The Vulkan implementation will then look in the cache for an existing entry, or add one if it
 //! doesn't exist.
 //!
 //! Once that is done, you can extract the data from the cache and store it. See the documentation
 //! of [`get_data`](struct.PipelineCache.html#method.get_data) for example of how to store the data
 //! on the disk, and [`with_data`](struct.PipelineCache.html#method.with_data) for how to reload it.
-//!
 
+// TODO: implement the cache functionality for ComputePipeline
```
Eliah-Lakhin, line 25: @Gosuo Is it outdated? :)

Gosuo: Yea I just saw that, can I change that without running the tests again?

Eliah-Lakhin:
> I would like to write some tests for the PipelineCache, but I'm not sure how to actually test this, as there is no real response from the PipelineCache of the implementation

@Gosuo You can write a simple test that will just create a pipeline, then reinstantiate it over with the cache object from the previous instantiation. No need to save to disk I think.

As of performance, I will try it on compute shaders in my personal project. The compute shader there is quite large, and it currently takes around 20-25 seconds to load, so I will try to use save-to-disk strategy. Hopefully performance impact will be clear.

Eliah-Lakhin:
@Gosuo Just tested your branch with my compute shaders. Not quite accurate benchmarking, but consequent loading speed seem to be raised from 3-8 seconds to 14-20 milliseconds constantly when the cache is loading from disk. Which is truly great!

vulkano/src/pipeline/cache.rs
Eliah-Lakhin: @Gosuo Github actions tests will trigger anyway, but no need to retest it locally. Thanks!

Gosuo:
@Eliah-Lakhin Cool, that is really nice to hear.

I think I need help with building a simple GraphicsPipeline for the tests as I'm new to Vulkan and vulkano. How do I build a simple Pipeline without creating too many unnecessary structs? I tried to build a simple RenderPass, I think it can be shortened though, and add a Subpass from that to the GraphicsPipelineBuilder, but I think the builder needs more information than that.

Eliah-Lakhin:
@Gosuo Not sure if creation of Graphics Pipeline without instantiating some additional stuff(like renderpass) is doable. Perhaps, to make things simpler, you can create a minimal example of cache usage here instead of unit tests: https://github.com/vulkano-rs/vulkano/tree/master/examples/src/bin In examples you can even use save/load-to-file strategy for demonstration purposes. For the end-users it would be even better.

Gosuo:
@Eliah-Lakhin Ok, let's see if the checks pass.

I added an example and I added some tests. If those checks pass and you are happy with the code this can be merged.

As the Vulkan implementation does not necessarily check the binary data that is passed to it, it might be in the scope of vulkano to rename the old methods of the `PipelineCache` to `unchecked_*` and implement new ones, that will add extra information to the header of the file to increase the safety of the loading process. This [blog](https://zeux.io/2019/07/17/serializing-pipeline-cache/) has some interesting approaches.

Eliah-Lakhin:
@Gosuo Thank you so much for your great work! The implementation and tests look very good and clean. PR is merged.

Eliah-Lakhin:
@Gosuo Regarding the safety topic. Yes, it could be unsafe, so the thing that we added it to the public interface is controversial, but I as far as I understand it could be considered as safe as far as the end user using it on the same device.

In the future we probably should move implementation to the deeper internals marking it unsafe, and exposing the safe wrapper that would somehow parse metadata(which might be difficult to do as this is vendor-specific) or add extra headers to the binary which is controversial too due to extra(but probably small) performance impact, and due to the fact the the user might want to load actual caching data prebuilt somewhere outside of Vulkano.

But anyway I would live it for the future improvements for now.

PR #1444, opened 2020-11-29
Check usage bits on image when creating image view

### LINKED ISSUE #1392: Validation error on image usage when creating swapchain
I create a swapchain with the following parameters:
```Rust
Swapchain::new(
	device,
	surface,
	4,
	B8G8R8A8Unorm,
	[800, 600],
	1,
	ImageUsage {
		transfer_destination: true,
		..ImageUsage::none()
	},
	SharingMode::Exclusive,
	SurfaceTransform::Identity,
	CompositeAlpha::Opaque,
	PresentMode::Mailbox,
	FullscreenExclusive::Default,
	true,
	ColorSpace::SrgbNonLinear,
)
```
The Vulkan implementation then gives me these validation errors:
```
UNASSIGNED-CoreValidation-MemTrack-InvalidUsageFlag: Invalid usage flag for VkImage 0x40000000004[] used by vkCreateImageView(). In this case, VkImage should have VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT|DEPTH_STENCIL_ATTACHMENT|INPUT_ATTACHMENT|SHADING_RATE_IMAGE|FRAGMENT_DENSITY_MAP]_BIT set during creation.
UNASSIGNED-CoreValidation-MemTrack-InvalidUsageFlag: Invalid usage flag for VkImage 0x50000000005[] used by vkCreateImageView(). In this case, VkImage should have VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT|DEPTH_STENCIL_ATTACHMENT|INPUT_ATTACHMENT|SHADING_RATE_IMAGE|FRAGMENT_DENSITY_MAP]_BIT set during creation.
UNASSIGNED-CoreValidation-MemTrack-InvalidUsageFlag: Invalid usage flag for VkImage 0x60000000006[] used by vkCreateImageView(). In this case, VkImage should have VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT|DEPTH_STENCIL_ATTACHMENT|INPUT_ATTACHMENT|SHADING_RATE_IMAGE|FRAGMENT_DENSITY_MAP]_BIT set during creation.
UNASSIGNED-CoreValidation-MemTrack-InvalidUsageFlag: Invalid usage flag for VkImage 0x70000000007[] used by vkCreateImageView(). In this case, VkImage should have VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT|DEPTH_STENCIL_ATTACHMENT|INPUT_ATTACHMENT|SHADING_RATE_IMAGE|FRAGMENT_DENSITY_MAP]_BIT set during creation.
UNASSIGNED-CoreValidation-MemTrack-InvalidUsageFlag: Invalid usage flag for VkImage 0x80000000008[] used by vkCreateImageView(). In this case, VkImage should have VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT|DEPTH_STENCIL_ATTACHMENT|INPUT_ATTACHMENT|SHADING_RATE_IMAGE|FRAGMENT_DENSITY_MAP]_BIT set during creation.
```
`Capabilities::supported_usage_flags` is:
```Rust
ImageUsage {
    transfer_source: true,
    transfer_destination: true,
    sampled: true,
    storage: true,
    color_attachment: true,
    depth_stencil_attachment: false,
    transient_attachment: false,
    input_attachment: false,
}
```

If I add the `color_attachment` flag to the creation, these errors go away. But it's not clear to me which part of the Vulkan spec says the flags cannot be only `transfer_destination`.

Rua:

Fixes #1392. It's only a quick fix using an assert statement, but since `UnsafeImageView` already has a bunch of checks with asserts, it felt appropriate.

I believe that, in the future, it may be good to split off image views as a type separate from images. In https://github.com/vulkano-rs/vulkano/issues/1392#issuecomment-653622619 I mentioned that, in my particular use case, I have no need for an image view on `Swapchain`, and therefore there shouldn't be a need to add one of the usages that views require. This could be solved by having `SwapchainImage` contain only a `vkImage`, with the user then free to create a `vkImageView` for it if they need to.

### DIFF
--- a/vulkano/src/image/sys.rs
+++ b/vulkano/src/image/sys.rs
@@ -356,12 +356,10 @@ impl UnsafeImage {
                 array_layers,
             } => {
                 if width == 0 || array_layers == 0 {
-                    return Err(ImageCreationError::UnsupportedDimensions {
-                        dimensions: dimensions,
-                    });
+                    return Err(ImageCreationError::UnsupportedDimensions { dimensions });
                 }
                 let extent = vk::Extent3D {
-                    width: width,
+                    width,
                     height: 1,
                     depth: 1,
                 };
@@ -374,18 +372,14 @@ impl UnsafeImage {
                 cubemap_compatible,
             } => {
                 if width == 0 || height == 0 || array_layers == 0 {
-                    return Err(ImageCreationError::UnsupportedDimensions {
-                        dimensions: dimensions,
-                    });
+                    return Err(ImageCreationError::UnsupportedDimensions { dimensions });
                 }
                 if cubemap_compatible && width != height {
-                    return Err(ImageCreationError::UnsupportedDimensions {
-                        dimensions: dimensions,
-                    });
+                    return Err(ImageCreationError::UnsupportedDimensions { dimensions });
                 }
                 let extent = vk::Extent3D {
-                    width: width,
-                    height: height,
+                    width,
+                    height,
                     depth: 1,
                 };
                 let flags = if cubemap_compatible {
@@ -401,14 +395,12 @@ impl UnsafeImage {
                 depth,
             } => {
                 if width == 0 || height == 0 || depth == 0 {
-                    return Err(ImageCreationError::UnsupportedDimensions {
-                        dimensions: dimensions,
-                    });
+                    return Err(ImageCreationError::UnsupportedDimensions { dimensions });
                 }
                 let extent = vk::Extent3D {
-                    width: width,
-                    height: height,
-                    depth: depth,
+                    width,
+                    height,
+                    depth,
                 };
                 (vk::IMAGE_TYPE_3D, extent, 1, 0)
             }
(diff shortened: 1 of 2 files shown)


Eliah-Lakhin:
@Rua Can you add Changelog entry, please?

Eliah-Lakhin:
@Rua Merged. Thank you! :+1:

vulkano/src/image/sys.rs
```diff
@@ -1034,6 +1016,19 @@ impl UnsafeImageView {
         assert!(mipmap_levels.end <= image.mipmaps);
         assert!(array_layers.end > array_layers.start);
         assert!(array_layers.end <= image.dimensions.array_layers());
+        assert!(
```
Eliah-Lakhin, line 1019: @Rua I discovered that this is probably too restrictive. There are cases when the user may decide to create a temporary image to read from or to write to in command buffers only. I mean without access in render passes or in the shader. For example when loading texture from file to be later mip-map blit into another image.

In other words I think that we should relax this restriction to accept Transfer usages only.

Can you fix it, please?

Rua: The restriction was added because of the Vulkan specification, which says that [when creating an image view](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkImageViewCreateInfo.html):

> `image` **must** have been created with a `usage` value containing at least one of the usages defined in the valid image usage list for image views

Where the [usage list then lists](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#valid-imageview-imageusage):
> When creating a `VkImageView` one of the following VkImageUsageFlagBits **must** be set:
> - `VK_IMAGE_USAGE_SAMPLED_BIT`
> - `VK_IMAGE_USAGE_STORAGE_BIT`
> - `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT`
> - `VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT`
> - `VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT`
> - `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT`
> - `VK_IMAGE_USAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR`
> - `VK_IMAGE_USAGE_FRAGMENT_DENSITY_MAP_BIT_EXT`

Allowing anything else would violate the spec, so we can't do that. I think the real problem is what I mentioned in the first post, that every image type in Vulkano automatically creates an image view. This then brings with it the restrictions Vulkan imposes on views. The proper solution IMO would be make Vulkano images no longer contain a view, but instead create a new safe type for views, and make creating a view a separate optional step.

Issue #1447, opened 2020-11-30
UB when reading ImmutableImage with General layout


faulesocke:
I get random behaviour (so likely this is UB) when reading a `ImmutableImage` with layout `General` using `AutoCommandBufferBuilder::blit_image`. The minimum example below reproduces this error:

```rust
use vulkano::{
    buffer::{BufferUsage, CpuAccessibleBuffer},
    command_buffer::{AutoCommandBufferBuilder, CommandBuffer},
    device::{Device, DeviceExtensions},
    format::Format,
    image::{Dimensions, ImageLayout, ImageUsage, ImmutableImage},
    instance::{Instance, InstanceExtensions, PhysicalDevice},
    sampler::Filter,
    sync::GpuFuture,
};

fn main() {
    let instance = Instance::new(None, &InstanceExtensions::none(), None).unwrap();
    let physical = PhysicalDevice::enumerate(&instance).next().unwrap();
    println!(
        "Using device: {} (type: {:?})",
        physical.name(),
        physical.ty()
    );

    let queue_family = physical
        .queue_families()
        .find(|&q| q.supports_graphics())
        .unwrap();

    let (device, mut queues) = Device::new(
        physical,
        physical.supported_features(),
        &DeviceExtensions::none(),
        [(queue_family, 0.5)].iter().cloned(),
    )
    .unwrap();
    let queue = queues.next().unwrap();

    let dimensions = Dimensions::Dim2d {
        width: 128,
        height: 128,
    };
    let format = Format::R8G8B8A8Srgb;
    let usage = ImageUsage {
        transfer_source: true,
        transfer_destination: true,
        ..ImageUsage::none()
    };
    let tex1 = {
        let data = vec![0; 128 * 128 * 4];

        // NOTE THIS
        let layout = ImageLayout::General;

        let (tex, init) = ImmutableImage::uninitialized(
            device.clone(),
            dimensions,
            format,
            1,
            usage,
            layout,
            device.active_queue_families(),
        )
        .unwrap();

        let buffer = CpuAccessibleBuffer::from_iter(
            device.clone(),
            BufferUsage::transfer_source(),
            false,
            data.into_iter(),
        )
        .unwrap();

        let mut cbb =
            AutoCommandBufferBuilder::primary_one_time_submit(device.clone(), queue_family)
                .unwrap();
        cbb.copy_buffer_to_image(buffer, init).unwrap();

        let cb = cbb.build().unwrap();
        cb.execute(queue.clone())
[...]

Issue #1441, opened 2020-11-23
vkCmdBlitImage on itself. Resource conflicts issue


qnope:
# Template

I am beginning in Rust, but I have some experiences with Vulkan in Cpp.
I am trying to implement the mipmap generation into ImmutableImage. (As I see it is in Todo list)

I was inspired from https://github.com/vulkano-rs/vulkano/issues/989, siince it seems to work (even if there was some layout transition issues that was apparently solved now)

So what I tried is that :

```rust
pub fn from_buffer<B, P>(
        source: B,
        dimensions: Dimensions,
        mipmaps: MipmapsCount,
        format: F,
        queue: Arc<Queue>,
    ) -> Result<
        (
            Arc<Self>,
            CommandBufferExecFuture<NowFuture, AutoCommandBuffer>,
        ),
        ImageCreationError,
    >
    where
        B: BufferAccess + TypedBufferAccess<Content = [P]> + 'static + Clone + Send + Sync,
        P: Send + Sync + Clone + 'static,
        F: FormatDesc + AcceptsPixels<P> + 'static + Send + Sync,
        Format: AcceptsPixels<P>,
    {
        let usage = ImageUsage {
            transfer_destination: true,
            transfer_source: true,
            sampled: true,
            ..ImageUsage::none()
        };
        let layout = ImageLayout::ShaderReadOnlyOptimal;

        let (buffer, init) = ImmutableImage::uninitialized(
            source.device().clone(),
            dimensions,
            format,
            mipmaps,
            usage,
            layout,
            source.device().active_queue_families(),
        )?;

        let mut cbb = AutoCommandBufferBuilder::new(source.device().clone(), queue.family())?;
        cbb.copy_buffer_to_image_dimensions(
            source,
            init,
            [0, 0, 0],
            dimensions.width_height_depth(),
            0,
            dimensions.array_layers_with_cube(),
            0,
        )
        .unwrap();

        let img_dim = buffer.dimensions.to_image_dimensions();
        for level in 1..buffer.mipmap_levels() {
            let [xs, ys, ds] = img_dim.mipmap_dimensions(level - 1).unwrap().width_height_depth();
            let [xd, yd, dd] = img_dim.mipmap_dimensions(level).unwrap().width_height_depth();

            cbb.blit_image(
                buffer.clone(), //source
                [0, 0, 0], //source_top_left
                [xs as i32, ys as i32, ds as i32], //source_bottom_right
                0, //source_base_array_layer
[...]

PR #1449, opened 2020-12-02
Unlock images correctly on Err

Gosuo:
If the CommandBuffer encountered an error in the lock_submit function,
it would unlock all the previously locked resources and return the
error. It would unlock all the images with a None value which would
panic in an assertion in the unlock method of the images. Now it checks
for the final_layout and unlocks the image this way.
This change should help with the panics mentioned in #1441 and #1447.
With this fix, it should return the error that is occurring instead of
panicking.

~~* [ ] Updated documentation to reflect any user-facing changes - in this repository~~
~~* [ ] Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.~~

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -16,6 +16,7 @@
 - Travis CI Linux Nightly job temporary disabled until #1423 resolved.
 - Renamed feature from `shader_f3264` to `shader_float64`.
 - Added method `build_with_cache` to the `GraphicsPipelineBuilder` that enables pipeline caching.
+- Fixing an assertion panic in the SyncCommandBuffer. If the buffer encountered an error while locking the necessary resources, it would unlock all previously locked resources. Images were unlocked incorrectly and an assert in the image unlock function would panic.
 
 # Version 0.19.0 (2020-06-01)
 
--- a/vulkano/src/command_buffer/synced/base.rs
+++ b/vulkano/src/command_buffer/synced/base.rs
@@ -1195,7 +1195,7 @@ impl<P> SyncCommandBuffer<P> {
 
         // If we are going to return an error, we have to unlock all the resources we locked above.
         if let Err(_) = ret_value {
-            for key in self.resources.keys().take(locked_resources) {
+            for (key, val) in self.resources.iter().take(locked_resources) {
                 let (command_ids, resource_ty, resource_index) = match *key {
                     CbKey::Command {
                         ref command_ids,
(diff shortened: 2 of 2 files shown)


Gosuo:
@Eliah-Lakhin Well, I'm open for a review here actually. I don't know if this is correct, but the images are unlocked the same way in the `unlock` method of the buffer.

The panics of the asserts in the mentioned issues disappear with this fix though.

Eliah-Lakhin:
@Gosuo Can you add an entry to the changelog file, please?

Gosuo:
@Eliah-Lakhin Oh, I thought the change log is only for user-facing changes, but I can add an entry as soon I have time.

Eliah-Lakhin:
@Gosuo Well, this is an actual(and important!) bug fix affecting the end-user experience, so I would log it anyway :)

Gosuo:
@Eliah-Lakhin Added the information to the changelog.

Eliah-Lakhin:
@Gosuo PR merged. Thank you very much for this fix! Now it will unblock a number of higher level features progress.

faulesocke:
I guess this is not correct. When a command buffer execution fails (which should never happen with vulkano), it is unspecified what the contents and layout of the images it touched are.

However in the unlock methods parameter you specify what the layout is after command buffer execution and I guess the `None` value is exactly for this purpose: Telling the image that the layout is unknown because of an execution error.

The solution here is obvious: Fix the actual error that causes the command buffer execution to fail.

Gosuo:
@faulesocke But the actual underlying Vulkan buffer wasn't executed at that point. It just locked the needed resources from the Rust point of view. That way the image wasn't touched at that point, or am I wrong?
Sounds logical though that the current unlocks with the final layout are not correct. Maybe unlock it with the initial layout?

faulesocke:
Well but when nothing was executed the image is certainly not in the desired final layout, is it?

Gosuo:
What do you suggest? Revert the change?
Your example in #1447 fails at the insertion introduced in #1444, so I'm not sure how to test this further.

faulesocke:
Not sure what to do. This needs further investigation. I might have time for it during the weekend.

Eliah-Lakhin:
@faulesocke, @Gosuo Please keep updating me on the issue progress. These features are completely above of my current understanding of Vulkano internals, so I will leave it to your hands.

PR #1450, opened 2020-12-05
CpuAccessibleBuffer: switch to parking_lot RwLock

rroohhh:
This allows ReadLock and WriteLock to be Send and Sync.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking** `CpuAccessibleBuffer` now uses `RwLock` from `parking_lot`.
 - **Breaking** The `Kind` and `SubpassContents` types have been moved to the root of the `command_buffer` module.
 - **Breaking** On `AutoCommandBufferBuilder`, the methods `begin_render_pass` and `next_subpass` now take `SubpassContents` instead of a boolean value.
 - **Breaking** The `CommandBuffer` trait now has an additional required method, `kind`.
--- a/vulkano/Cargo.toml
+++ b/vulkano/Cargo.toml
@@ -20,3 +20,4 @@ smallvec = "1.4"
 lazy_static = "1.4"
 vk-sys = { version = "0.5.2", path = "../vk-sys" }
 half = "1.6"
+parking_lot = { version = "0.11.1", features = ["send_guard"] }
--- a/vulkano/src/buffer/cpu_access.rs
+++ b/vulkano/src/buffer/cpu_access.rs
@@ -30,9 +30,6 @@ use std::ptr;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
-use std::sync::RwLock;
-use std::sync::RwLockReadGuard;
-use std::sync::RwLockWriteGuard;
 
 use buffer::sys::BufferCreationError;
 use buffer::sys::SparseLevel;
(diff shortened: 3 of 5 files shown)


rroohhh:
@Eliah-Lakhin do you mind taking a look at this?

Eliah-Lakhin:
@rroohhh Thank you for your contribution!

Why do you think having `Send+Sync` for CPU Accessible buffer would benefit Vulkano users? Can you describe use cases, please.

There are a number of questions we probably should cover with this migration first:

1) We are adding new dependency to the Vulkano crate which complicates build process.
2) Having `Send+Sync` is good per se, but CPU Accessible Buffer is probably not the only restricted place. Maybe we need a more generalized solution.
3) I'm not sure if `Send+Sync` wouldn't brake the Vulkan resource locking logic. The underlying Vulkan resource might lost even if the Guard is still valid, so it would lead to unsafe behavior. I'm not sure about this part, but just to mention that just changing the Lock Guard type may be not enough.

rroohhh:
@Eliah-Lakhin 

> Why do you think having `Send+Sync` for CPU Accessible buffer would benefit Vulkano users? Can you describe use cases, please.

Ok, so we are using this in [axiom-recorder](https://github.com/apertus-open-source-cinema/axiom-recorder/tree/stream_processing) to do multi-threaded zerocopy processing. Currently this is used to process a image in one thread in a vulkan compute shader, which writes its output to a CPU Accessible buffer and then ships off that buffer to other threads for further processing (like writing the result to a file, or stuffing it into ffmpeg to do h264 encoding).

To avoid the need to lock the CPU Accessible buffer we are creating the ReadLock on the thread that does the vulkan processing and use [owning-ref](https://github.com/Kimundi/owning-ref-rs) to combine it with the CPU Accessible buffer itself (see [this](https://github.com/apertus-open-source-cinema/axiom-recorder/blob/51dcbb64fee72636babc41e78debbae26f11df84/src/gpu/gpu_util.rs#L80-L86)). We are creating this in the thread that does the GPU Processing to avoid having the threads that will receive the buffer need to known about  CPU Accessible buffers and having to lock them first. This way it is completely transparent to them and looks just like any other `T : Deref<Target = [u8]>`.

> 1. We are adding new dependency to the Vulkano crate which complicates build process.

`parking_lot` is a widely used crate (about 40000 weekly download from `crates.io`) so this should not cause big problems.
> 2. Having `Send+Sync` is good per se, but CPU Accessible Buffer is probably not the only restricted place. Maybe we need a more generalized solution.

CPU Accessible Buffer is already `Send+Sync`, only the lock guards are not, I am not sure there are more restricted places.
> 3. I'm not sure if `Send+Sync` wouldn't brake the Vulkan resource locking logic. The underlying Vulkan resource might lost even if the Guard is still valid, so it would lead to unsafe behavior. I'm not sure about this part, but just to mention that just changing the Lock Guard type may be not enough.

I am not sure this should introduce any new problems, before CPU Accessible buffer was already `Send+Sync` so it could already be send to other threads. The returned guard still has the same lifetime as before (the lifetime of the underlying CPU Accessible buffer) so you cannot use the guard after the CPU Accessible buffer gets dropped. The only actual way to get the guard itself on another thread is using something like `owning-ref` or a scoped thread api like [`crossbeam::scope`](https://docs.rs/crossbeam/0.8.0/crossbeam/fn.scope.html).

Eliah-Lakhin:
@rroohhh I'm happy more projects migrating to Vulkano, and I don't see any more blockers for this change too. Thank you for detail explanation. PR is merged.

PR #1451, opened 2020-12-06
[Vulkano-1441] Adding the possibility to generate mipmaps at image creation

qnope:
I am a rust beginner, so, if there is anything to change, tell me.

To do so, I created a SubImage that implements the ImageAccess trait
This SubImage will decorate an ImageAccess image and will represent
one or more mip level / array layer level

Originally, I wanted to modify both `ImmutableImage` and `InitializationImmutableImage` in order to have the possibility to manage the mip levels / array layer levels directly on them, but I don't know vulkano enough to be able to do such a thing...

Maybe, it could be nice to have the `SubImage` type in another file, since it could be used by storage image also.

Hope it can help :)

P.S: This PR address the issue https://github.com/vulkano-rs/vulkano/issues/1441

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking** The `ImmutableImage::from_iter` and `ImmutableImage::from_buffer` can build Mipmaps
 - **Breaking** `CpuAccessibleBuffer` now uses `RwLock` from `parking_lot`.
 - **Breaking** The `Kind` and `SubpassContents` types have been moved to the root of the `command_buffer` module.
 - **Breaking** On `AutoCommandBufferBuilder`, the methods `begin_render_pass` and `next_subpass` now take `SubpassContents` instead of a boolean value.
--- a/vulkano/src/image/immutable.rs
+++ b/vulkano/src/image/immutable.rs
@@ -47,6 +47,7 @@ use memory::pool::MemoryPoolAlloc;
 use memory::pool::PotentialDedicatedAllocation;
 use memory::pool::StdMemoryPoolAlloc;
 use memory::DedicatedAlloc;
+use sampler::Filter;
 use sync::AccessError;
 use sync::NowFuture;
 use sync::Sharing;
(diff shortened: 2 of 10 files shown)


CHANGELOG_VULKANO.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking** The `ImmutableImage::from_iter` and `ImmutableImage::from_buffer` can build Mipmaps
```
Eliah-Lakhin, line 3: Please sync with the latest `master`.

vulkano/src/image/immutable.rs
```diff
@@ -65,10 +66,97 @@ pub struct ImmutableImage<F, A = PotentialDedicatedAllocation<StdMemoryPoolAlloc
     layout: ImageLayout,
 }
 
+pub struct SubImage<Img> {
```
Eliah-Lakhin, line 69: A new public API object probably needs some rustdoc description.

vulkano/src/image/immutable.rs
```diff
@@ -270,21 +358,31 @@ impl<F> ImmutableImage<F> {
...
             dimensions,
             format,
-            MipmapsCount::One,
+            mipmaps,
             usage,
             layout,
             source.device().active_queue_families(),
         )?;
 
+        let init = SubImage::new(
```
Eliah-Lakhin, line 377: What would happen if the end-user use `uninitialized`'s initializer and the `SubImage` objects both?

vulkano/src/image/immutable.rs
```diff
@@ -270,21 +358,31 @@ impl<F> ImmutableImage<F> {
     {
         let usage = ImageUsage {
             transfer_destination: true,
+            transfer_source: true,
```
Eliah-Lakhin, line 361: Internally I would separate implementation into two cases: when the user specifies more than One mipmaps(current implementation) and specifies just a single mipmap(previous implementation) to avoid extra steps for a single-mipmap case which might be a common use case.

vulkano/src/image/immutable.rs
```diff
@@ -296,14 +394,27 @@ impl<F> ImmutableImage<F> {
...
+        );
+
         let cb = cbb.build().unwrap();
 
         let future = match cb.execute(queue) {
             Ok(f) => f,
-            Err(_) => unreachable!(),
+            Err(e) => {
+                eprintln!("{:?}", e);
+                unreachable!();
```
Eliah-Lakhin, line 411: These two lines could be combined together `unreachable!("{:?}", e)`

Eliah-Lakhin (changes requested):
I left a few comments.

Also, mind you add a few tests: for multi-mipmap cases and for single-mipmap cases both to cover the old and new implementations?

Thank you.

qnope:
I will do the modification in the following days :)
Thanks !

vulkano/src/image/immutable.rs
qnope: I don't know why I deleted it, it is fixed

vulkano/src/image/immutable.rs
qnope: I hope what I did is correct about the doc

qnope:
I hope I have fixed all the comments you made.
Regarding the test, you mean adding an example that uses mipmaps?

Eliah-Lakhin:
@qnope Thank you very much for the fixes! The code looks good, I would merge it.

How do you think do we need unit tests before I merge? Maybe without the actual mipmap content checking, but just checking if image creation process with and without mipmaps passes without failures. That would help for future maintainability to verify if we have regressions.

qnope:
I am going to try to write some little unit tests before the weekend :).
I'll let you know when it will be mergeable :)

qnope:
I have written very very tiny tests... (not really proud ahaha) But I hope it'll be enough :)
Thanks

Eliah-Lakhin:
@qnope PR merged. Thank you very much for your help to the project and the community! :)

Rua:
Looking back at this code now, @qnope I wonder if you could explain something to me? You added two new methods to the `ImageAccess` trait, namely `current_miplevels_access` and `current_layer_levels_access`. These are supposed to return the range of mipmaps and array layers that an `ImageAccess` represents, relative to the "full" image. However, the `ImageInner` type, which is also returned from the `ImageAccess` trait, seems to contain these same values, except that your new type `SubImage` doesn't correctly set them. Was this an oversight on your part? Please let me know what the intent was here, as I'm working on a PR that changes this.

qnope:
Hello @Rua 

When I did the fix, I was (and actually I am still) a novice both in rust and in vulkano rs.
I did not have knowledge that the  `ImageInner` type existed. So it is highly likely that there exists a much better way for adding mipmaps and I encourage you to change it to make things better :).

When your PR is ready, please quote me to see your changes :)

Rua:
Thank you for clarifying!

PR #1455, opened 2020-12-17
Add support for including precompiled shaders in vulkano-shaders

Arc-blroth:

This PR adds an option `bytes` to the vulkano shader macro for directly loading spri-v bytecode rather than compiling from source. This is useful in circumstances where developers may want to use a different build system for shader compilation (In my case I'm using https://github.com/EmbarkStudios/rust-gpu to compile the shader).

Feel free to bikeshed/improve line 360 in vulkano-shaders/src/lib.rs - while it works on my system, I'm not sure how endianess might affect converting the u8 slice to a u32 slice, or if there's a better way to perform that conversion than pointer magic 👀


Eliah-Lakhin:
@Arc-blroth The change is not breaking, and having an ergonomic way to load precompiled shaders was something people around asking from time to time, so it would be definitely helpful to have this option. Thank you! PR is merged.

Rua:
I think there is a mistake in the documentation for `bytes`: "Cannot be used in conjunction with the `src` or `bytes` field." That should probably say `src` or `path`?

PR #1458, opened 2020-12-19
Fix vulkano-shaders bytes documentation

Arc-blroth:
> I think there is a mistake in the documentation for `bytes`: "Cannot be used in conjunction with the `src` or `bytes` field." That should probably say `src` or `path`?

_Originally posted by @Rua in https://github.com/vulkano-rs/vulkano/issues/1455#issuecomment-748470182_

Whoops! Thanks for catching that. This PR should fix the documentation for `bytes` now.


Eliah-Lakhin:
@Arc-blroth merged. Thanks for the fix! And thanks to @Rua for catching the issue! I missed it too :)

Issue #1454, opened 2020-12-17
`AutoCommandBufferBuilder::begin_render_pass` doesn't clear all cubemap layers


Tenebryo:
*   Version of vulkano: `0.19.0` from crates.io
*   OS: Windows 10 x64
*   GPU (the selected PhysicalDevice): GeForce RTX 2070 SUPER
*   GPU Driver: 460.79
*   See bottom of issue for a minimal example

# Issue

`AutoCommandBufferBuilder::begin_render_pass` only clears the first layer of cubemap attachments. After some digging through the source, it appears this is because `Dimensions::array_layers` (called from `FramebufferBuilder::add`) returns 1 instead of 6 for `Dimensions::Cubemap{size}`. I noticed that if you convert the dimensions to a `ImageDimensions` it does return 6.

With my limited familiarity with the codebase, I see two possible solutions:
* ~~change the behavior of `Dimensions::array_layers` so that it returns 6 for cubemaps [here](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/image/mod.rs#L304)~~ I didn't see the `Dimensions::array_layers_with_cube` function, so this is probably not the solution
* convert the `Dimensions` to `ImageDimensions` in `FramebufferBuilder::add` (via `Dimensions::to_image_dimensions`) [here](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/framebuffer/framebuffer.rs#L181)

# Minimal Bug Reproduction
```rust
use vulkano::buffer::{BufferUsage, CpuAccessibleBuffer};
use vulkano::image::{Dimensions, StorageImage};
use vulkano::instance::InstanceExtensions;
use vulkano::command_buffer::AutoCommandBufferBuilder;
use vulkano::device::{Device, DeviceExtensions};
use vulkano::framebuffer::Framebuffer;
use vulkano::image::ImageUsage;
use vulkano::instance::{Instance, PhysicalDevice};
use vulkano::format::Format;
use vulkano::sync;
use vulkano::sync::GpuFuture;

use std::sync::Arc;

fn main() {

    let instance = Instance::new(None, &InstanceExtensions::none(), None).unwrap();

    let physical = PhysicalDevice::enumerate(&instance).next().unwrap();

    println!(
        "Using device: {} (type: {:?})",
        physical.name(),
        physical.ty()
    );

    let queue_family = physical
        .queue_families()
        .find(|&q| {
            q.supports_graphics()
        })
        .unwrap();

[...]

PR #1459, opened 2020-12-19
Fix Cubemap and CubemapArray framebuffer attachments

Tenebryo:
A fix for issue #1454 

Graphics pipeline did not clear or render to all the layers in a `Cubemap` or `CubemapArray` because framebuffers incorrectly read the number of array layers when adding an attachment. The proposed fix is to convert the `Dimensions` that is currently used to determine the attachment size to `ImageDimensions` which correctly calculates the number of array layers.


Eliah-Lakhin:
@Tenebryo Thank you for the fix! Could you fix the merge conflict with the change log file, please? And then I will merge it.

Eliah-Lakhin (approved):

Tenebryo:
I fixed the merge conflict

Eliah-Lakhin:
@Tenebryo Merged. Thanks!

PR #1480, opened 2021-01-31
[Breaking] Add support for dynamic buffers

Arc-blroth:

This PR (finishes) adding support for:
- [dynamic uniform buffers and dynamic storage buffers][1],
- specifying [dynamic offsets][2] in draw and dispatch calls, so that the dynamic buffers are actually useful,
- updating buffers with unsized data to make aligned allocations easier,
- and public usage of `PipelineLayoutDescTweaks` for tweaking compute pipeline layouts

In total, that's 4 breaking changes! Since this PR adds an extra argument to the draw and dispatch methods in `AutoCommandBufferBuilder`, the website will need to be updated along with this PR. All of the existing examples and tests in this repository have been updated, and I've also added a [new example][3] to demostrate using dynamic uniform buffers with vulkano.

Currently, this PR only adds basic support for dynamic buffers - it checks that dynamic offsets are properly aligned, but it doesn't validate the buffer data itself or handle aligned allocation. In the future (maybe after the Rust Allocator API is stablized), we might want to add an API that handles unsafe aligned allocation for dynamic buffers.

[1]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#descriptorsets-uniformbufferdynamic
[2]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdBindDescriptorSets.html#VUID-vkCmdBindDescriptorSets-pDynamicOffsets-01971
[3]: https://github.com/Arc-blroth/vulkano/blob/staging/dynamic_buffers/examples/src/bin/dynamic-buffers.rs

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,10 +1,14 @@
 # Unreleased
 
+- **Breaking** On `AutoCommandBufferBuilder`, methods that bind a descriptor set now take a `dynamic_offsets` parameter
+- **Breaking** On `AutoCommandBufferBuilder` and `SyncCommandBufferBuilder`, the `update_buffer` method now takes `data` by reference
+- **Breaking** Made `PipelineLayoutDescTweaks` public, for use with compute pipelines
 - Added support for `ImageAspect` and YV12/NV12 formats,  for use with the UnsafeImage API.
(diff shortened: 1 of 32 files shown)


Eliah-Lakhin:
@Arc-blroth Pull Request reviewed and merged. Thank you very much for your contribution! :+1:

Rua:
A bit late, but instead of an empty `vec![]` in the examples you could also have used `std::iter::empty()`.

PR #1481, opened 2021-02-05
Fix descriptor pool allocation for layouts that contain arrays

Arc-blroth:

This PR fixes an error where when allocating memory for an `UnsafeDescriptorSetPool`, vulkano did not consider that arrays could have more than one element. This led to a `OutOfPoolMemory` error being reported from Vulkan, which then led to, in the wise words of Rust:
> ```thread 'main' panicked at 'internal error: entered unreachable code'```


Eliah-Lakhin:
@Arc-blroth After merging of #1480 there are merging conflicts. Could you fix them, please?

Arc-blroth:
Fixed 👍

Eliah-Lakhin:
@Arc-blroth Merged. Thank you.

PR #1483, opened 2021-02-05
Update various dependencies

Rua:

This updates various dependencies to their latest versions. A few changes have also been made to the code, to comply with breaking changes or deprecations. I've taken the liberty of removing the patch semvers on all the dependencies, so that they will be more flexible. If that's not desired I can undo it.


Eliah-Lakhin:
@Rua Merged. Thank you for your help with regular maintainability of dependency system.

 >  I've taken the liberty of removing the patch semvers on all the dependencies

I see that most changes are about examples, so this is not critical part anyway.

I'm a bit worried about the shaderc upgrade, but at least it is passing our current tests. I will do a deeper testing of the shader system later.

PR #1486, opened 2021-02-06
Replace filter map 

### LINKED ISSUE #1485: Error when getting surface capabilities
*   Version of vulkano: 0.20
*   OS: macOS 11.1
*   GPU (the selected PhysicalDevice): M1 SOC GPU
*   GPU Driver: Vulkan SDK (moltenVK)

```rust
let instance = {
let extensions = vulkano_win::required_extensions();
    Instance::new(None, &extensions, None).expect("failed to create Vulkan instance")
};
let physical = PhysicalDevice::enumerate(&instance).next().expect("no device available");

let el = EventLoop::new();
let surface = WindowBuilder::new()
    .build_vk_surface(&el, instance.clone())
    .unwrap();

let queue_family = physical.queue_families()
    .find(|&q| q.supports_graphics() && surface.is_supported(q).unwrap_or(false))
    .expect("couldn't find a graphical queue family");

let (device, mut queues) = {
    let device_ext = vulkano::device::DeviceExtensions {
        khr_swapchain: true,
        .. vulkano::device::DeviceExtensions::none()
    };
    Device::new(physical, physical.supported_features(), &device_ext,
        [(queue_family, 0.5)].iter().cloned()).expect("failed to create device")
};
let queue = queues.next().unwrap();
let caps = surface.capabilities(physical)
    .expect("failed to get surface capabilities");
```

The above code panics at runtime with message: 
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', .cargo/registry/src/github.com-1ecc6299db9ec823/vulkano-0.20.0/src/swapchain/surface.rs:639:63
```
https://github.com/vulkano-rs/vulkano/blob/4a98caacbff0fdf8c8438fb18a9b468e7d8ef201/vulkano/src/swapchain/surface.rs#L637-L642

I suspect that the Vulkan driver is reporting some kind of feature or format that isn't included in `vk-sys`, which is why the unwrap call fails.

EthanYidong:

Fixes #1485 

I replaced map and unwrap with a filter_map to prevent panics when not all formats reported by Vulkan are supported in `vk-sys`. I don't think this is a good permanent solution. A better solution would be to let `vk-sys` build against the Vulkan sdk on your system, but that's probably best left for another time, and I don't see any downsides to using this as a temporary fix.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -9,6 +9,7 @@
 - Fixed race condition in `StandardCommandPool` when allocating buffers.
 - Fixed potential stack overflow error in loading large shaders by storing the bytecode as static.
 - Fixed descriptor set layouts with arrays containing more than one element triggering unreachable code.
+- Fixed panic on some machines when calling `Surface::capabilities`
 - Added basic support and safety checks for dynamic uniform/storage buffers
 - Updated dependencies:
   - `crossbeam` 0.7 -> 0.8
--- a/examples/src/bin/push-constants.rs
+++ b/examples/src/bin/push-constants.rs
@@ -104,7 +104,13 @@ fn main() {
     let mut builder =
         AutoCommandBufferBuilder::primary_one_time_submit(device.clone(), queue.family()).unwrap();
     builder
-        .dispatch([1024, 1, 1], pipeline.clone(), set.clone(), push_constants, vec![])
+        .dispatch(
+            [1024, 1, 1],
+            pipeline.clone(),
+            set.clone(),
+            push_constants,
+            vec![],
+        )
         .unwrap();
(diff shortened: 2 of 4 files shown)


Eliah-Lakhin:
@EthanYidong Thank you for the catch.

Your solution is filtering out formats that might actually be supported by Vulkan without any indication for the API user of skipped values. That could confuse the user. But I think it's fine as a temporary workaround.

Can you put this specific of behavior in the `capabilities` function docs, and also put a `TODO` comment in it's implementation so we don't miss to implement it properly in the future, please?

EthanYidong:
@Eliah-Lakhin Done!

Eliah-Lakhin:
@EthanYidong Pull Request merged. Thank you for the fix!

PR #1487, opened 2021-02-10
segfaulting: DebugUtilsMessengerCallbackDataEXT::pMessageIdName may be null in debug callback

DaAitch:
Running on MacOS the `cargo run --bin debug` example is segfaulting (after fixing a layer problem: layer `VK_LAYER_LUNARG_standard_validation` doesn't exist on MoltenVK but it has to be `VK_LAYER_KHRONOS_validation`, but that's another point not in this scope).

- Source: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkDebugUtilsMessengerCallbackDataEXT.html#VUID-VkDebugUtilsMessengerCallbackDataEXT-pMessageIdName-parameter
- `(*callback_data).pMessageIdName` is a null pointer for some debug callbacks (at least for two on my machine while testing)

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking** `Message::layer_prefix` turned to Option to prevent segfaults when Vulkan message didn't provide `pMessageIdName` value
 - **Breaking** On `AutoCommandBufferBuilder`, methods that bind a descriptor set now take a `dynamic_offsets` parameter
 - **Breaking** On `AutoCommandBufferBuilder` and `SyncCommandBufferBuilder`, the `update_buffer` method now takes `data` by reference
 - **Breaking** Made `PipelineLayoutDescTweaks` public, for use with compute pipelines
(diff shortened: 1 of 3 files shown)


Eliah-Lakhin:
@DaAitch Thank you for the fix. Can you add an entry to the changelog, please?

DaAitch:
@Eliah-Lakhin Sure.

Do you have an idea for this string https://github.com/vulkano-rs/vulkano/pull/1487/files#diff-365d1440144c1c4c8329f5e8f963d7b182434c8d6bd74c25b67d80d303535b9bR105 I think "Unknown Layer" is not correct anymore. It's called `layer_prefix` but maybe sth. like`"Unknown message id name"` is better? I don't know.

Eliah-Lakhin:
@DaAitch I think the more proper way of fixing it would be turning `Message::layer_prefix` into `Option`.

Also, in such case it will be a breaking change(which is fine), just we need to mention it in release logs. Something like: "**Breaking** `Message::layer_prefix` turned to Option to prevent segfaults when Vulkan message didn't provide `pMessageIdName` value"

DaAitch:
@Eliah-Lakhin please check.

Eliah-Lakhin:
@DaAitch Merged. Thank you!

Issue #1482, opened 2021-02-05
Error: vkCreateDevice: VK_KHR_portability_subset must be enabled when calling Device::new


drhaynes:
# Template

*   Version of vulkano: 0.20
*   OS: macOS (using MoltenVK)
*   GPU (the selected PhysicalDevice): GTX 680
*   GPU Driver: 
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: TODO

# Issue
When trying to create a logical device on macOS, it's giving the following error:

`Validation layer: "Validation Error: [ VUID-VkDeviceCreateInfo-pProperties-04451 ] Object 0: handle = 0x7f8115c2d8f0, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x3a3b6ca0 | vkCreateDevice: VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x7f8115c2d8f0[] supports it The Vulkan spec states: If the [VK_KHR_portability_subset] extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensions must include \"VK_KHR_portability_subset\". (https://vulkan.lunarg.com/doc/view/1.2.162.1/mac/1.2-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451)"`

`VK_KHR_portability_subset` does not appear to be present in the repo or documentation. Is there 1. A way to detect this is required, and 2. A way to enable it manually?

Eliah-Lakhin:
@drhaynes This is default behavior when the user is trying to use some feature from not-included extension.

> A way to detect this is required

It would shade more light on the issue if you provide a code snippet. It's sometimes tricky and implementation-wise.

> A way to enable it manually

Device extensions(and Instance extensions too) are specifying during initialization: https://github.com/vulkano-rs/vulkano/blob/master/examples/src/bin/triangle.rs#L131 I don't see if we support this particular extension in Vulkano (there is unfortunately only a small subset of Vulkan extensions that we currently support), but if you need this one you can implement it and contribute to the project. It's usually not too difficult, the list of extensions reside here: https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/device/extensions.rs#L106

DaAitch:
hey,

[...]

PR #1489, opened 2021-02-10
fix vkDeviceCreate enabled extensions have to contain "VK_KHR_portability_subset" if supported by device

DaAitch:
- Ref: #1482
- https://vulkan.lunarg.com/doc/view/1.2.162.1/mac/1.2-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451

### DIFF
--- a/examples/src/bin/debug.rs
+++ b/examples/src/bin/debug.rs
@@ -118,7 +118,7 @@ fn main() {
     let (_, mut queues) = Device::new(
         physical,
         physical.supported_features(),
-        &DeviceExtensions::none(),
+        &DeviceExtensions::required_extensions(physical),
         vec![(queue_family, 0.5)],
     )
     .expect("failed to create device");
--- a/vulkano/src/device/extensions.rs
+++ b/vulkano/src/device/extensions.rs
@@ -21,7 +21,7 @@ use vk;
 use VulkanObject;
 
 macro_rules! device_extensions {
-    ($sname:ident, $rawname:ident, $($ext:ident => $s:expr,)*) => (
+    ($sname:ident, $rawname:ident, [$($ext_req_if_supported:ident,)*], $($ext:ident => $s:expr,)*) => (
         extensions! {
             $sname, $rawname,
             $( $ext => $s,)*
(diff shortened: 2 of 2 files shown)


Eliah-Lakhin:
@DaAitch Thank you for the fix.

I like the idea of automatically selecting extensions required by Vulkan Docs. The problem here is that the user may call `DeviceExtensions::supported_by_device` explicitly too, so in the end this function will be called twice. Current Vulkano design is that it lefts the decision on enabled extensions completely to API user side. Maybe it wasn't the right approach in Vulkano from the beginning, but this patch in my opinion would make a contradiction with existing practice.

But I would like to hear community opinions too.

@drhaynes, @Rua, @EthanYidong, @Arc-blroth What do you think, guys?

Rua:
I prefer explicitness and user control, which is what Vulkan is about in the first place. So I'm generally in favour of requiring the user to do something and giving an error if they don't. However, in this case, how does Vulkano decide that this particular extension is required? Can it determine it without having to call `DeviceExtensions::supported_by_device`, for example through an error from the underlying Vulkan? Or does Vulkan silently ignore the mistake and keep going, like it's prone to do?

DaAitch:
@Eliah-Lakhin @Rua I agree.
Explicitness is way more better than magic things happening. I'll prepare a new PR this evening. I would keep `khr_portability_subset` added to the extensions, but nothing more, right?

@Rua Vulkan prints an error but somehow keeps going. I don't know which impact ignoring this error has, but at least the examples work.

Eliah-Lakhin:
@DaAitch No need to open a new PR, you can just add more commits to existing branch. They will propagate to the PR automatically.

Also, please add a Changelog entry.

Rua:
> @Rua Vulkan prints an error but somehow keeps going. I don't know which impact ignoring this error has, but at least the examples work.

I'm guessing the error is being caught by the validation layers, but that's not something Vulkano can rely on, and it doesn't cause `vkCreateDevice` to fail. So I guess there is no other way to detect the mistake, than to query the available extensions and see if this one is listed, like you currently do. However, perhaps the result of this function can be cached? I recall looking at its code a while ago and noticing that it does that, so maybe calling it a second time is less expensive than it seems.

Eliah-Lakhin:
@Rua Good point!

Yes, I think it would be a good idea from API design view to introduce `DeviceExtensions::required_extensions(phys)` that will be responsible for detection of what is required. Similar to vulkano-win. Calling `DeviceExtensions::supported_by_device` under the hood is probably fine, yes. It would also left us an opportunity to implement caching in the future, but for now I think it's fine to leave caching aside of this patch scope.

DaAitch:
@Eliah-Lakhin @Rua what do you think about the last commit?

Eliah-Lakhin:
@DaAitch Pull Request merged. Thank you for your help and fixes! :)

Rua:
Should all the other examples also be updated to call `required_extensions` when creating a device? This is something that every user should be doing from now on after all, lest they write code incompatible with Macs, or with future required extensions on other platforms.

PR #1490, opened 2021-02-10
Add VK_KHR_portability_subset device extension

drhaynes:
Adds support for the VK_KHR_portability_subset device extension.

From the [spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_portability_subset.html)
>This extension allows a non-conformant Vulkan implementation to be built on top of another non-Vulkan graphics API, and identifies differences between that implementation and a fully-conformant native Vulkan implementation.

This extension is required when using validation layers on e.g. MoltenVK platforms, such as macOS.


drhaynes:
Closed as dupe of https://github.com/vulkano-rs/vulkano/pull/1489

PR #1491, opened 2021-02-10
Fix reuse command buffers test

Databean:
The per-thread weak reference to StandardCommandPoolPerThread in StandardCommandPool can expire forcing the recreation of the pool and its internal objects. This is a problem for the reuse_command_buffers test which asserts that a destroyed command pool must be reused if possible.

This change fixes the test by introducing an extra command buffer that is held through the rest of the test, introducing an extra strong reference to the command pool that keeps the weak reference alive for the duration of the test.

This test fails consistently on my machine with a NVidia 1070 Max-Q otherwise.


Eliah-Lakhin:
@Databean Thank you for the catch! PR is merged.

PR #1497, opened 2021-02-20
Fix generated descriptor set layouts for shader modules with multiple entrypoints

Arc-blroth:

SPIR-V modules can [contain multiple entrypoints][1], each with seperate interfaces and descriptor set layouts. However, vulkano-shaders currently only generates 1 descriptor set layout struct and uses it for all entrypoints in a module (which makes sense, since until #1455 modules were essentially limited to a single entrypoint anyway).
This PR moves the logic that writes descriptor set layout information to execute once per entrypoint, rather than once per the entire shader module.

[1]: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_module


Eliah-Lakhin:
@Arc-blroth Thank you for your help! And I'm sorry for later review.

I tried it locally, and it seems that the change is breaking pipeline creation for compute shaders. For example a `basic-compute-shader` example is finishing with segfault on Pipeline creation(or maybe on shader loading, not quite sure).

Also, this change is breaking in general since it renames generated `Layout` name to `MainLayout` when the user uses the macro in ordinary way(with glsl code for shaderc). It needs to be indicated in changelogs, and also in the docs. Also, it would be nice to cover the failing case somehow with unit tests(if possible).

The feature would be very useful for us, but we need to fix appearing issues first to merge it to upstream.

Arc-blroth:
Huh, I didn't realize this broke compute shaders, thanks for the heads up on that! I'll try getting around to working on this PR later today/tommorow.

<br>

> Also, it would be nice to cover the failing case somehow with unit tests(if possible)

@Eliah-Lakhin Could you please clarify what exactly is the failing case here? I'm not sure if you meant renaming `Layout` &rarr; `MainLayout` or something else.

Arc-blroth:
It turns out that the problem of figuring out if a variable belongs to an entrypoint is harder than I originally thought.
When compiling for SPIR-V 1.0 to 1.3, shaders do not include any variables other than `Input`s and `Output`s ([specification link][1]) in the `OpEntryPoint` interface:
<pre>
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 32
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
<b>               OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID</b>
               ; --- snip ---
               OpDecorate %data DescriptorSet 0
               OpDecorate %data Binding 0
               ; --- snip --
       %data = OpVariable %_ptr_StorageBuffer_Data StorageBuffer
</pre>
But when compiling for SPIRV 1.4 and up, shaders will include variables of all storage classes as part of the interface:
<pre>
; SPIR-V
; Version: 1.4
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 32
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
<b>               OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID %data</b>
               ; --- snip ---
               OpDecorate %data DescriptorSet 0
               OpDecorate %data Binding 0
               ; --- snip --
       %data = OpVariable %_ptr_StorageBuffer_Data StorageBuffer
</pre>
Because of this, the only way to accurately determine which variables are actually used in a shader, and therefore need to be bound to a descriptor set, is to \*insert sad programmer noises\* manually inspect the instruction tree of each entrypoint. 

[1]: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpEntryPoint

Arc-blroth:
I've updated the `write_descriptor_sets` logic so that it now actually writes descriptor sets correctly - the compute shader now runs! (at least on my computer). I've also updated the changelog with the 3 breaking changes:

1. Prefixed `shader!` generated descriptor set `Layout` structs with the name of the entrypoint the layout belongs to. For shaders generated from GLSL source, this means `Layout` has been renamed to `MainLayout`.
2. `shader!` will no longer generate descriptor information for variables that are declared but not used in a shader.
3. `shader!` now accepts structs in shader interfaces decorated with `BufferBlock` rather than `Block`

Change 2 might come as a surprise to some downstream consumers, but I don't see it as too big of an issue, since the usecase is  small.

Eliah-Lakhin:
@Arc-blroth It took for me some time to adopt new changes from your feature branch(because of N2) in my project that has quite large and complicated compute shader with a number of various bindings and specializations. I used it as a form of a test suite in addition to test you implemented in your PR.

But in the end I can confirm that it works perfectly and as expected, and passing all tests. Code looks good too. PR is merged. Thanks again for your great work and for the fixes! :+1:

Issue #1492, opened 2021-02-10
Make image views a separate type from images


Rua:
Mentioned in https://github.com/vulkano-rs/vulkano/issues/1392#issuecomment-653622619, https://github.com/vulkano-rs/vulkano/pull/1444#issue-529094324, https://github.com/vulkano-rs/vulkano/pull/1444#discussion_r572160885 and https://github.com/vulkano-rs/vulkano/pull/1449#issuecomment-740042773. I propose making a separate `ImageView` type, that is created from an image (presumably with an `Arc`, as one image can have many views). Images would no longer contain image views inside them, but instead the user would have to create a view, if their use case requires it.

This would lift some restrictions on particular use cases. Vulkano currently creates views for every image, whether a view is needed or not, and views bring limitations with them. In particular, certain image usage flags are required when creating a view, but not every image needs such usage in practice (e.g. swapchain images that are only transferred onto).

The new type would be a safe wrapper around the current `UnsafeImageView`, and most likely the safety checks currently in the latter should be in this new type, and implemented in a better way than with `assert`s. This follows the Vulkano practice that the low-level "unsafe" types do not check for correctness, but leave this to higher-level wrappers. `ImageViewAccess` would naturally no longer be implemented on the images, but rather on this new type, although it may no longer be necessary if there is only one single view type.

Eliah-Lakhin:
I like this idea :+1:

PR #1499, opened 2021-02-26
Major rearrange of image and view types

### LINKED ISSUE #1494: Dimensions versus ImageDimensions, what's their function? Can they be merged?
Vulkano has had two types for representing the dimensions of an image, `Dimensions` and `ImageDimensions`. Originally there was just one, but the split between these types was added already in 2016 in #260.

Neither of the types is documented, so it's not clear what their function is, nor which is to be used for which purpose. The result is a lot of duplication and an IMO very confusing situation. For example, `ImageAccess::dimensions()` and `UnsafeImage::dimensions()` both give you an `ImageDimensions`, while `ImageViewAccess::dimensions()` and `ImmutableImage::dimensions()` return `Dimensions`. `Dimensions` has a function to convert to `ImageDimensions`, but there is no function to perform the reverse operation. Either that simply never got added, or the conversion is lossy and `Dimensions` contains more information.

I would find it a lot clearer if there was just one type, or if the purpose of each one were documented. Does anyone who has worked with Vulkano longer have any idea what's going on? If there is a clear distinction, can they be renamed to make that clearer? If nobody knows, I'm going to try changing all instances of `ImageDimensions` to `Dimensions` in a future PR.

Rua:

This fixes #1494 and #1492. I'm doing it as a draft because it's such a big change, but I'll keep the changelog inside the PR updated so it's clear what has been changed.


Rua:
This is ready for review. The change is quite huge in the end, requiring major changes to users' code. I made sure all the tests pass, but I can't guarantee that something didn't slip through anyway!

Eliah-Lakhin:
@Rua Thank you very much for your work!

I like the idea of explicit distinction between Images and Views in the API, and also the unification between "Dimensions*" objects too. Also, I can confirm that this branch works fine in my personal project, so I didn't discover any regressions too.

A few things I would like to cover before we merge it to upstream:

1. In `ImageDimensions::Dim2` there is an `array_layers` parameter. Does your implementation covers zero value safely?
2. Previously Images had a View implementations on top of them(that I didn't like too), now the user has to crate views explicitly to describe them(that I like). Do I understand correctly, there is no additional performance penalty of this explicit strategy? Probably not, but just want to double check.
3. `ImageCreateFlags` are now explicitly exposed in Image creation interface. Maybe it would be better to somehow cast them from the usage flags and/or dimensions under the safe API hood? Is it possible if the user misused them? I'm not very familiar with this aspect of Vulkan API(but probably many other users too:).
4. As far as I understood Views also allow component swizzling. It is currently is not covered by the current implementation. Would it be possible to cover this feature in the future improvements?

Please respond to my questions when you have time and also merge `master` to your branch, there are conflicts with changelog. And then we will proceed with merging.

Thanks again!

Rua:
1. I haven't really changed anything in how images handle the possibilty of 0 `array_layers`. This check is still done in the constructor of `UnsafeImage`.
2. There may be a slight extra overhead because the compatibility with the image and arguments have to be checked. This is because `ImageView` is a "safe" type.
3. Create flags aren't the same as usage flags though, they're separate things in Vulkan so I think we should keep them separate too. I have added some checks to avoid misuse, but it's very crude and not ideal. I don't know how to implement proper support for all the flags at the moment.
4. I haven't messed with component swizzling stuff, the methods and such are still there. They just need to be implemented by a future editor.

Eliah-Lakhin:
@Rua ok, Pull Request is merged :+1:

Regarding Create Flags. I think we need to deal with this thing in relatively foreseeing future, as the current implementation is actually unfinished. But let's move it aside from this PR context.

For swizzling. It would be nice to have it too, but probably this is lesser important for us at this moment.

PR #1504, opened 2021-03-05
UnsafeCommandBuffer and SyncCommandBuffer no longer contain a command pool reference

Rua:

This was brought on in an attempt to make #1430 a bit easier to implement, as the removal of the associated type makes it feasible to keep `dyn CommandBuffer`s around. But it also makes more sense in the grander scheme of Vulkano design, where "Unsafe" types aren't expected to keep resources alive. Now it's instead the `AutoCommandBuffer` that is responsible for that.


Rua:
I have thought this over and think it's not quite ready yet. I think they should hold a reference after all, but differently.

Rua:
Is this PR ok to be merged too, or do I need to change something?

Eliah-Lakhin:
@Rua The code looks fine at a first glance, but I would like to spend more time to understand the topic, and especially the question you raised in #1430. And I would like to proceed with review tomorrow. Does it block your current development?

Rua:
A bit, yes. I have another PR waiting that converts the code to the 2018 edition and fixes some clippy issues. But because that touches the entire codebase I'd rather wait until everything has been merged so there's less conflicts. I don't mind if you take your time though, I appreciate a thorough review.

As for that question, I think I already answered it for myself, and I think I know how to tackle this problem. My plan is to keep a record of all the images and buffers added during the build process, with their accesses and such, inside the finished `SyncCommandBuffer`. Then, when it needs to be added as a secondary command buffer to a primary one, this record is "replayed" and the builder will do its usual work adding pipeline barriers.

Eliah-Lakhin:
@Rua Understood. ok, I will review and try to promote it today, just a bit later. I'm reviewing PRs all the day today, just want to get a short break :)

Rua:
Hah, very understandable. You did great work today!

Eliah-Lakhin:
@Rua I read original Issue #1430 and the implementation more carefully, and I have a couple of thoughts about this topic.

First, regarding this Pull Request. While it is probably not too critical to clear the enforcement of keeping the pool in unsafe API, this move will make related objects even more unsafe. There is no formal definition of the levels of "unsafety", but in my opinion the lesser rules to check unsafe code introduces the less logical errors users will miss. This is not only important for the crate users, but also for the crate maintainers/contributors, especially in the future when and if the current active team of contributors and maintainers completely changes. So, basically this change could complicate future maintainability. Which is a risk for the project.

However, this risk is acceptable if it worth the benefits we would gain in the end. And that is about my second question. Honestly speaking, I don't see why do we need `AutoCommandBufferBuilder::execute_commands` method at all. In it's unsafe form it was probably left in the API for some rare edge cases, because the normal safe synchronization was intended to be managed by GpuFutures, and it is capable to cover almost all use cases if the user keeps command buffers relatively short. If the users need to put something in between the execution sequence they could organize this execution through a series of small command buffers put together in a single GpuFuture, and they could even reuse and remix these command buffers many times.

Another controversial thing is that creation of the Command buffer has relatively small performance cost which is also a benefit of Vulkan/Vulkano, that I think we should persist. And I'm not sure if it would be feasible by introducing safe `execute_commands`. Also as I understood from the Issue #1430 making it safe would require a lot of work of duplicating the implementations already covered by GpuFuture. As such it could complicate maintainability too.

To sum up, I would like to better understand how safe `execute_commands` would benefit Vulkano users, which cases would it cover that Vulkano is currently is not able to cover well, and how would it fit in the current Vulkano code design. The costs of this change might appear to be too high for us, and maybe we need rethink the approach.

Rua:
The `execute_commands` command is tied to the distinction between primary and secondary command buffers, and is essential to multithreaded rendering. The distinction is:
- A primary command buffer can be submitted to a queue, must contain whole render passes (each start must be matched by an end), and can have secondary buffers embedded inside it using `execute_commands`.
- A secondary command buffer _can't_ be submitted to a queue, it _must_ be executed by embedding it in a primary command buffer, and for rendering they must fit entirely inside a render pass (can't start/end passes in them).

A command buffer builder is neither `Send` nor `Sync`, so you can't pass it between threads during the recording process. Therefore, if you have a lot of command buffer recording to do and want to split the work up into multiple threads, then your only option is to record a separate _secondary_ command buffer on each thread, then build them, and then send the finished secondary command buffers (which are `Send + Sync`!) to a master thread where they are embedded in a primary command buffer via `execute_command`, then submitted. You can't do this with primary command buffers, because they must contain whole render passes.

The ability to multithread your recording is one of the big power features of Vulkan, so lacking that ability in Vulkano is a big downside that has bothered me for a while now. That's why I have been trying to make it safe. And I definitely do think it's possible and desirable, and `execute_commands` should definitely not be removed!

Regarding the unsafety of this particular PR, the worst that can happen is a Vulkan memory leak, which is not good but not UB either. What I could do is have `UnsafeCommandPoolAlloc` contain a reference not to any `CommandPool`, but specifically to `UnsafeCommandPool`.

However, I must note that the other "pool" object type, `UnsafeDescriptorSet`, doesn't implement `Drop` currently either and will leak if it goes out of scope instead of being returned to its pool. In this case, it would actually be UB to implement `Drop` naively, because the `reset` method on the pool will immediately free any descriptor sets it owns, and if `Drop` frees it afterwards that's a double-free. But that's a matter for another discussion, I just wanted to point out that Vulkano already has a "leaky" type. The `reset` of command pools doesn't free command buffers, so for them `Drop` is safe.

Eliah-Lakhin:
@Rua ok, that all makes sense. Thank you for detailed explanation of the use case. I see that the checks are failing but the errors seem to be related to the recent issues in `master` that already fixed there. I'm going to merge your Pull Request.

Eliah-Lakhin:
@Rua PR is merged. Thank you! Let's proceed with other steps towards the plan.

PR #1510, opened 2021-03-18
More memory changes with some thoughts on error handling

gurchetansingh:
* [ x] Ran `cargo fmt` on the changes

Hey, I know the tests don't pass yet, but wondering for some feedback on the VulkanoResult.

CC: @hakolao @Rua @Eliah-Lakhin @maximbaz

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -1,6 +1,8 @@
 # Unreleased
+- **Breaking ** DeviceMemoryBuilder::new() takes in `memory_index` rather than `MemoryType`.
+- Opaque fd and dma-buf import support on `Linux`.
+- `DeviceMemoryMapping` to separate device memory and mappings.
 - Added external memory support for `DeviceLocalBuffer` for `Linux`
-
 - Fixed `shader!` generated descriptor set layouts for shader modules with multiple entrypoints.
   - **Breaking** Prefixed `shader!` generated descriptor set `Layout` structs with the name of the entrypoint the layout belongs to. For shaders generated from GLSL source, this means `Layout` has been renamed to `MainLayout`.
   - **Breaking** `shader!` will no longer generate descriptor information for variables that are declared but not used in a shader.
--- a/vulkano/src/memory/device_memory.rs
+++ b/vulkano/src/memory/device_memory.rs
@@ -9,6 +9,7 @@
 
 use std::error;
 use std::fmt;
+use std::marker::PhantomData;
 use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::ops::DerefMut;
@@ -16,11 +17,12 @@ use std::ops::Range;
 use std::os::raw::c_void;
 use std::ptr;
 use std::sync::Arc;
+use std::sync::Mutex;
 
 #[cfg(target_os = "linux")]
 use std::fs::File;
 #[cfg(target_os = "linux")]
-use std::os::unix::io::FromRawFd;
+use std::os::unix::io::{FromRawFd, IntoRawFd};
 
 use check_errors;
 use device::Device;
@@ -33,6 +35,29 @@ use vk;
 use Error;
 use OomError;
 use VulkanObject;
+
+pub struct BaseOutStructure {
+    pub s_type: i32,
+    pub p_next: *mut BaseOutStructure,
+}
+
+pub(crate) unsafe fn ptr_chain_iter<T>(ptr: &mut T) -> impl Iterator<Item = *mut BaseOutStructure> {
+    let ptr: *mut BaseOutStructure = ptr as *mut T as _;
+    (0..).scan(ptr, |p_ptr, _| {
+        if p_ptr.is_null() {
+            return None;
+        }
+        let n_ptr = (**p_ptr).p_next as *mut BaseOutStructure;
+        let old = *p_ptr;
+        *p_ptr = n_ptr;
+        Some(old)
+    })
+}
+
+pub unsafe trait ExtendsMemoryAllocateInfo {}
+unsafe impl ExtendsMemoryAllocateInfo for vk::MemoryDedicatedAllocateInfoKHR {}
+unsafe impl ExtendsMemoryAllocateInfo for vk::ExportMemoryAllocateInfo {}
+unsafe impl ExtendsMemoryAllocateInfo for vk::ImportMemoryFdInfoKHR {}
 
 /// Represents memory that has been allocated.
 ///
@@ -55,6 +80,7 @@ pub struct DeviceMemory {
     size: usize,
     memory_type_index: u32,
     handle_types: ExternalMemoryHandleType,
+    mapped: Mutex<bool>,
 }
 
 /// Represents a builder for the device memory object.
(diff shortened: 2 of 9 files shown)


Rua:
I don't think using `VulkanoResult` like this is a very good idea. Having an explicit error type means the user knows in advance what errors a function is going to return, and they can exhaustively match on it. Now every function can return every error, so it's much less clear and the user has to cover every possible variant. What was the motivation for making this change?

Aside from that, in the Ash backend conversion I've been working on, the `check_errors` function and the `Error` and `Success` enums disappear. `ash::prelude::VkResult` converts directly to each individual error type a function returns. For example:
```Rust
impl From<ash::vk::Result> for InstanceCreationError {
    #[inline]
    fn from(err: ash::vk::Result) -> Self {
        use ash::vk::Result as R;
        match err {
            R::ERROR_OUT_OF_HOST_MEMORY | R::ERROR_OUT_OF_DEVICE_MEMORY => {
                OomError::from(err).into()
            }
            R::ERROR_INITIALIZATION_FAILED => Self::InitializationFailed,
            R::ERROR_LAYER_NOT_PRESENT => Self::LayerNotPresent,
            R::ERROR_EXTENSION_NOT_PRESENT => Self::ExtensionNotPresent,
            R::ERROR_INCOMPATIBLE_DRIVER => Self::IncompatibleDriver,
            _ => panic!("unexpected error: {:?}", err),
        }
    }
}```
This method could be used for Vulkano's current state as well, converting from `vk::Result`. You simply place a `?` after the called function. You just need to pay attention when the function could return a value that is neither `SUCCESS` nor an `ERROR`, e.g. `vkAcquireNextImageKHR`.

**Comment by gurchetansingh** [comment]
Thank you rua for your feedback.  I have dropped the Vulkano result from this series and look forward to your Ash change, as I view it fundamental to the project going forward!!  

I guess my thinking is I just don't see the overall structure for Vulkano error handling yet.  Add details in the commit message and the commit.  I think if we want to be a validation layer we should just return the VUID violation and in theory the user can match on that.  I think it can be simplified and improved, but of course I'm just raising the issue and would welcome other solutions.  For now, it's more a "quality of life" thing and not a priority, so dropped.

**Comment by Eliah-Lakhin** [comment]
@gurchetansingh Thank you for your contribution. Please let me know when it's ready for final review.

Regarding VulkanoResult/Error topic. I would keep error system API as it is, without changes.

**Comment by gurchetansingh** [comment]
@Eliah-Lakhin ready for final review now.

**Comment by gurchetansingh** [comment]
One thing to note is we no longer use `memory_type: MemoryType` in the builder.  The reason is I think we should autogenerate most of these builders as well [after the Ash change], so working with raw types and then validating later was chosen.

**Comment by Eliah-Lakhin** [comment]
@gurchetansingh Sorry for delaying. I usually don't have enough time to manage Pull Requests properly during the week, and I left this job for my days off. So I will take a closer look on Sat-Sun.

For now, can you explain:

1. What is the purpose of this change?
2. Why do we need to remove `MemoryType` and how is it connected to the PR's context? If this are irrelevant topics to the main PR's task context, maybe consider moving to a different Pull Request?
3. If there are breaking changes, these changes need to be highlighted in the changelogs.

**Comment by gurchetansingh** [comment]
@Eliah-Lakhin thank you for taking time to contribute this project!

1)  For my use case, I need to:

* Import external memory from one process to another.
* A wrapper around `DeviceMemoryMapping`, so I can pass the raw pointer to the hypervisor.
* Also a cross-beam simplification, which makes Vulkano integration easier.
 
Take a look at [crrev.com/c/2786286](crrev.com/c/2786286) for the use case [I've modified it a bit, but the Vulkano side remains unchanged]

2) It's just a minor inconsequential detail -- anyone using `DeviceMemoryBuilder` already can just use memory_type.id().  The bigger issue is we're mixing Vulkano objects (`MemoryType`) and the builders -- which will prevent auto-generated builders around Ash in the future.  But you are right, that's too far ahead and opaque to a reviewer such as yourself 🙂.

Since getting things merged can take a while, I do tend to put multiple discrete changes in one MR.  In future, I definitely agree smaller, easier to review MRs are better ... I hope the community gets there!

3) Done, added a changelog entry for the minor `DeviceMemoryBuilder` breaking change!

File: CHANGELOG_VULKANO.md [hunk 1]
```diff
@@ -1,4 +1,7 @@
 # Unreleased
+- **Breaking ** DeviceMemoryBuilder::new() takes in `memory_index` rather than `MemoryType`.
```
Eliah-Lakhin, line 2: Extra whitespace

vulkano/src/memory/device_memory.rs
```diff
@@ -87,28 +112,21 @@ impl<'a> DeviceMemoryBuilder<'a> {
...
     /// - Panics if `size` is 0.
     /// - Panics if `memory_type` doesn't belong to the same physical device as `device`.
-    pub fn new(device: Arc<Device>, memory_type: MemoryType, size: usize) -> DeviceMemoryBuilder {
-        assert!(size > 0);
-        assert_eq!(
-            device.physical_device().internal_object(),
-            memory_type.physical_device().internal_object()
-        );
-
+    pub fn new(device: Arc<Device>, memory_index: u32, size: usize) -> DeviceMemoryBuilder<'a> {
```
Eliah-Lakhin, line 115: Is this method still safe? What would happen if the user passes incompatible value of `memory_index`?

vulkano/src/memory/device_memory.rs
```diff
@@ -87,28 +112,21 @@ impl<'a> DeviceMemoryBuilder<'a> {
     ///
     /// - Panics if `size` is 0.
     /// - Panics if `memory_type` doesn't belong to the same physical device as `device`.
```
Eliah-Lakhin, line 114: `memory_type` argument was replaced, so this comment needs to be corrected accordingly.

vulkano/src/memory/device_memory.rs
```diff
@@ -120,38 +138,28 @@ impl<'a> DeviceMemoryBuilder<'a> {
...
-                .as_ref()
-                .map(|i| i as *const vk::MemoryDedicatedAllocateInfoKHR)
-                .unwrap_or(ptr::null()) as *const _;
-
-            if let Some(ref mut export_info) = self.export_info {
-                export_info.pNext = ptr;
-            } else {
-                self.allocate.pNext = ptr;
-            }
+        if self.dedicated_info.is_some() {
```
Eliah-Lakhin, line 141: Why assertion replaced with explicit panic?

vulkano/src/memory/device_memory.rs
```diff
@@ -160,67 +168,155 @@ impl<'a> DeviceMemoryBuilder<'a> {
...
-        assert!(self.import_info.is_none());
-
-        // Only extensions tested with Vulkano so far.
-        assert!(self.device.loaded_extensions().khr_external_memory);
-        assert!(self.device.loaded_extensions().khr_external_memory_fd);
-
-        let handle_bits = handle_types.to_bits();
-        if handle_bits & vk::EXTERNAL_MEMORY_HANDLE_TYPE_DMA_BUF_BIT_EXT != 0 {
-            assert!(self.device.loaded_extensions().ext_external_memory_dmabuf);
+        if self.export_info.is_some() {
```
Eliah-Lakhin, line 175: Why assertion replaced with explicit panic?

Eliah-Lakhin (changes requested):
@gurchetansingh Thank you for detailed explanation. I would like to proceed with merging of this PR, but I left a few questions/comments to address first. Mostly cosmetic.

vulkano/src/memory/device_memory.rs
gurchetansingh: Fixed.

vulkano/src/memory/device_memory.rs
gurchetansingh: Fixed.

vulkano/src/memory/device_memory.rs
gurchetansingh: Fixed.

vulkano/src/memory/device_memory.rs
gurchetansingh: Yes, it is.  Validation is now done in `DeviceMemoryBuilder::build()`, since on the fly validation while adding to the builder I feel does not scale or lend itself to autogeneration (say we a ImageBuilder or BufferBuilder in the future).

gurchetansingh:
@Eliah-Lakhin thank you for the review!

Eliah-Lakhin:
@gurchetansingh Pull Request merged. Thank you for your contribution and for the fixes!

PR #1511, opened 2021-03-18
Make execute_commands safe to use

### LINKED ISSUE #1430: Make AutoCommandBufferBuilder::execute_commands safe to use

Rua:

This fixes #1430.

`SyncCommandBufferBuilder::execute_commands` should now avoid or disallow any GPU data races. The current implementation is a bit "overeager" and may insert more pipeline barriers than are really needed. This is because the secondary command buffer may have its own pipeline barriers, which could "shield" its resources from the primary and therefore make an additional barrier in the primary unnecessary. This is quite difficult subject matter however, and I don't fully understand it, so I'd rather leave it with a functional but potentially inefficient implementation.

As noted in https://github.com/vulkano-rs/vulkano/issues/1430#issuecomment-800284644, I had to remove a `Mutex` that seemed to do nothing, in order to make this work. All the tests still complete just fine, but I can't say for sure issues with this won't crop up later. A lot of testing on real codebases is the only way to know for sure.

The truly safe version, `AutoCommandBufferBuilder`, hasn't been completed yet. Only the third of the three bullet points at https://github.com/vulkano-rs/vulkano/issues/1430#issuecomment-725311354 is now implemented, the first two still need doing. I intend to continue working on this, but I've submitted this draft to see how it goes.

A potential improvement, aside from efficiency, is the handling of multiple secondary command buffers in a single `execute_commands` call. These secondary command buffers cannot be synchronised against each other because they are part of the same command, so if there are resource conflicts between the two secondaries, trying to submit the command will produce a conflict error. This can be worked around by submitting them in separate `execute_commands` calls. It would be nice if the builder could handle these cases automatically instead of erroring out, by splitting a single `execute_commands` into several and inserting barriers between. But that would need more thinking and probably some redesigning, and it's an ergonomics issue rather than a correctness issue.


Rua:
The more I work on this, the more I'm thinking that primary and secondary command buffers should really be separate types entirely. The methods on the `CommandBuffer` trait (both current ones, and ones I need to add for secondaries) are almost entirely disjoint, each method with the exception of `inner` works either on primary or secondary command buffers, but not both. Almost all functions that want a command buffer as an argument want either a primary or a secondary, again not really many that are indifferent to the command buffer type. The only functions with significant overlap between the types are the builders.

So what I'm thinking is that `AutoCommandBuffer` and the `CommandBuffer` trait are split into two types, for primary and secondary respectively. `AutoCommandBufferBuilder` could probably stay a single type, but it would get an extra type parameter that specifies whether it builds into a `PrimaryAutoCommandBuffer` or a `SecondaryAutoCommandBuffer`. Some methods on `AutoCommandBufferBuilder`, for example `begin_render_pass` or `execute_commands`, could be implemented only for one of the two types.

Rua:
Ok, I think I have done what I need to do to make `execute_commands` safe, so I've now removed the `unsafe` marker from the function. There are still two points I want to look at improving though:
- Splitting primary and secondary, as noted above.
- Making `SyncCommandBufferBuilder` more robust in the case of errors. Right now, if an error occurs while building, it leaves the builder in an inconsistent state, which has potential safety implications if the user continues to use the same buffer. I'd like to make it more "atomic" so that its internal state is only changed once all error checks have been made.

Rua:
I think this is now ready for review. I haven't solved the problem of errors in `SyncCommandBufferBuilder`, but I have added a poisoning mechanism so that you get a panic if you use a poisoned builder. That's better than silently continuing to use a broken builder.

I've decided to leave the primary/secondary split to a future PR. I'd like to know what people think about the idea first. The plan is to split the `CommandBuffer` trait and the `AutoCommandBuffer` type in two, while keeping `AutoCommandBufferBuilder` as one type but parameterised by the type of `AutoCommandBuffer` that you're constructing. Commands that can only be called on one type are only implemented on one specialisation, so that runtime checks are not needed, they become compile errors.

Eliah-Lakhin:
@Rua Thank you. I will take closer look on the nearest days off. Too busy during the week. But I'm glad we are getting closer to the final goal!

Rua:
Closer? Doesn't this PR reach it? :p

Eliah-Lakhin:
@Rua All tests passing and, I didn't detect any regressions in my personal project too from your feature branch. Pull Request merged. Thank you for your great work!

Issue #1264, opened 2019-10-28
Shader codegen always treats descriptors as readonly


RobinMcCorkell:
Vulkano 0.14

I have a GLSL file with some readonly descriptors and some not, however the code generated by vulkano-shaders marks all of the descriptors as readonly. This has knock-on effects like the pipeline barrier code not working, since it looks for buffers that are not readonly in order to generate barriers in the right places.

It seems like vulkano-shaders always considers descriptors to structs (and by rule induction, buffers too) as readonly, which doesn't seem correct: https://github.com/vulkano-rs/vulkano/blob/master/vulkano-shaders/src/descriptor_sets.rs#L203

Ziqul:
Any recepie for workaround?

Rua:
I just ran into this bug myself. Vulkano-shader always sets the `readonly` attribute on the descriptors it generates to `true`, whether the shader writes to it or not. This has knock-on effects for memory safety in `SyncCommandBufferBuilder`: since the builder thinks the buffer isn't being written, it doesn't insert a pipeline barrier between the write and any subsequent reads, leading to a data race on the GPU.

PR #1513, opened 2021-03-24
Fix descriptor set TypeStruct readonly detection

### LINKED ISSUE #1297: Implement barrier insertions in between dispatch calls
Originally brought up under https://github.com/vulkano-rs/vulkano/issues/1265

> 
> 
> I think I'm hitting a similar issue where I have two dispatch calls to two different compute shaders in one command buffer, where the first one writes to a buffer and the second one reads from it, and there's no pipeline barrier call between the two according to RenderDoc. The code looks like this:
> 
> ```rust
> let layout = self.pipeline.descriptor_set_layout(0).unwrap();
> let set = PersistentDescriptorSet::start(layout.clone())
>     .add_buffer(prev.clone())
>     .unwrap()
>     .add_buffer(curr.clone())
>     .unwrap()
>     .add_buffer(next.clone())
>     .unwrap()
>     .build()
>     .unwrap();
> 
> let vertex_layout = self.vertex_pipeline.descriptor_set_layout(0).unwrap();
> let vertex_set = PersistentDescriptorSet::start(vertex_layout.clone())
>     .add_buffer(next.clone())
>     .unwrap()
>     .add_buffer(self.vertex_buffer.clone())
>     .unwrap()
>     .build()
>     .unwrap();
> 
> let command_buffer =
>     AutoCommandBufferBuilder::new(self.device.clone(), self.queue.family())
>         .unwrap()
>         .dispatch(
>             [SIZE, SIZE, 1],
>             self.pipeline.clone(),
>             set,
>             self.drop.take().unwrap_or((-1., -1.)),
>         )
>         .unwrap()
>         // vulkano doesn't insert a barrier here for some reason?
>         .dispatch(
>             [SIZE, SIZE, 1],
>             self.vertex_pipeline.clone(),
>             vertex_set,
>             (),
>         )
>         .unwrap()
>         .build()
>         .unwrap();
> let future = command_buffer.execute(self.queue.clone()).unwrap();
> future
>     .then_signal_fence_and_flush()
>     .unwrap()
>     .wait(None)
>     .unwrap();
> ```

Currently from looking around the code pertaining to dispatch calls, there doesn't seem to be any resource checks between dispatch calls, therefore no barrier insertions when the user may expect them.

danielzgtg:
* [x] (`CHANGELOG_VULKANO.md` only) Added an entry to `CHANGELOG_VULKANO.md` or `CHANGELOG_VK_SYS.md` if knowledge of this change could be valuable to users
* [x] (examples) Updated documentation to reflect any user-facing changes - in this repository
* [x] (N/A) Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.

This fixes #1297.

Before this PR, my Intel GPU was working but my AMD GPU outputted random data. After this PR, both my GPUs work in my application. I have confirmed that before this, `readonly: true` was used everywhere as shown by `cargo expand`, and after this, `readonly: false` is properly emitted when the shader doesn't use `readonly` in front of `buffer`

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -26,6 +26,9 @@
 - Implemented synchronization for `SyncCommandBufferBuilder::execute_commands`.
 - `AutoCommandBufferBuilder::execute_commands` is now fully safe to use.
 - `SyncCommandBufferBuilder` now becomes poisoned when it returns an error, to prevent using the builder in an inconsistent state.
+- Fixed missing barriers in dispatch calls
+  - **Breaking** `shader!` no longer marks descriptor sets as readonly as a fallback when it doesn't know
+    - **Breaking** The keyword `readonly` might need to be added in front of the `buffer` keyword in GLSL files to get them working again
 
 # Version 0.21.0 (2021-03-05)
 
--- a/examples/src/bin/immutable-buffer-initialization.rs
+++ b/examples/src/bin/immutable-buffer-initialization.rs
@@ -57,11 +57,11 @@ fn main() {
 
 layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;
 
-layout(set = 0, binding = 0) buffer Data {
+layout(set = 0, binding = 0) restrict buffer Data {
     uint data[];
 } data;
 
-layout(set = 0, binding = 1) buffer ImmutableData {
+layout(set = 0, binding = 1) readonly restrict buffer ImmutableData {
     uint data;
 } immutable_data;
 
--- a/vulkano-shaders/src/descriptor_sets.rs
+++ b/vulkano-shaders/src/descriptor_sets.rs
@@ -378,7 +378,7 @@ fn descriptor_infos(
         .iter()
         .filter_map(|i| {
             match i {
-                &Instruction::TypeStruct { result_id, .. } if result_id == pointed_ty => {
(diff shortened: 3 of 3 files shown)


Rua:
Funny that you made this, I ran into #1264 just yesterday.

Rua:
I just realised that the same issue exists with storage images as well. `TypeImage` subtypes `DescriptorDescTy::TexelBuffer` and `DescriptorDescTy::Image` with `sampled: false` can be both read from and written to in a shader, when the `readonly` attribute is not present. Do you think you could fix this as well in your PR? Might as well get it over with.

danielzgtg:
Fixed in a33b144c. I've ran all the automated tests successfully, but I don't have a use case of mine to manually test

Rua:
Shouldn't the `readonly` attribute also be read for images and texel buffers?

danielzgtg:
As I said, I'm not familiar with images and texel buffers. I just reverted the commit, and will leave the texture stuff to someone more experienced

Eliah-Lakhin:
@danielzgtg Thank you very much for the fix! Can you please resolve merging conflicts in the changelog file, and I will continue with merging.

danielzgtg:
Rebased

Eliah-Lakhin:
@danielzgtg Merged. Thanks!

PR #1516, opened 2021-03-25
Add dispatch_indirect to AutoCommandBufferBuilder

### LINKED ISSUE #1501: There are no draw_indirect command in AutoCommandBufferBuilder
There are no [vkCmdDispatchIndirect](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdDispatchIndirect.html) command in [AutoCommandBufferBuilder](https://docs.rs/vulkano/0.20.0/vulkano/command_buffer/struct.AutoCommandBufferBuilder.html).

But it is exists in [SyncCommandBufferBuilder](https://docs.rs/vulkano/0.20.0/vulkano/command_buffer/synced/struct.SyncCommandBufferBuilder.html): https://docs.rs/vulkano/0.20.0/vulkano/command_buffer/synced/struct.SyncCommandBufferBuilder.html#method.dispatch_indirect

Rua:

Fixes #1501. I have no code to test this on, maybe @p0lunin can give it a try first?


p0lunin:
ok i'll try check it tomorrow.

p0lunin:
for now i can't attach an image to the descriptor set due to error:
```
error[E0277]: the trait bound `AttachmentImage: SafeDeref` is not satisfied
   --> rencan-core\src\app.rs:294:24
    |
294 |             .add_image(output_image.clone())
    |                        ^^^^^^^^^^^^^^^^^^^^ the trait `SafeDeref` is not implemented for `AttachmentImage`
    |
    = note: required because of the requirements on the impl of `ImageViewAbstract` for `AttachmentImage`
    = note: 1 redundant requirements hidden
    = note: required because of the requirements on the impl of `ImageViewAbstract` for `Arc<AttachmentImage>`
```

Rua:
Hmm that's strange, I didn't change anything about how descriptor sets work in this PR. There was recently a change to images/views though. `AttachmentImage` no longer is an image view, you have to wrap it in an `ImageView` manually, and pass the image view to `.add_image`.

Eliah-Lakhin:
@Rua, @p0lunin That division implicitly affecting DescriptorSets as they related to template signatures from Images.

In short, all of this should work if the user pass ImageView instead of Image to the DescriptorSet.

p0lunin:
now it fails after several seconds of working with
```
thread 'main' panicked at 'assertion failed: inner.offset < buffer.size()'
```
in https://github.com/Rua/vulkano/blob/dispatch-indirect/vulkano/src/command_buffer/sys.rs#L1040
i use `CpuBufferPool` for indirect buffer

Rua:
I think I found the problem. It was a long-standing bug that seems to have not been noticed until now: `inner.offset` is the offset of the `CpuBufferPool`'s chunk within a larger buffer object, but `buffer.size()` is the size of the chunk itself. It's comparing to the wrong thing. I've now changed the latter to `inner.buffer.size()`, which is the size of the buffer object.

p0lunin:
I tested it and it works.
Only one thing: vulkano does not provide memory barriers for indirect buffer automatically, so you need to write somewhere in documentation that user need to synchronize access it yourself.

Rua:
The code to specify the memory access for the buffer is present: https://github.com/vulkano-rs/vulkano/blob/e81d9a02f7cd5ad563f4cfd20bb1e7675178d0ac/vulkano/src/command_buffer/synced/commands.rs#L1374 It looks correct at first glance.

Where are you noticing that a barrier isn't being inserted, when it should be? Do you have example code?

p0lunin:
You show code for the `draw_indirect`. There are code for `dispatch_indirect`: https://github.com/Rua/vulkano/blob/dispatch-indirect/vulkano/src/command_buffer/synced/commands.rs#L1372-L1392. I think there must be something like
```rust
PipelineStages {
  compute_shader: true,
  ..PipelineStages::none()
},
```
for `stages` field, because `dispatch_indirect` allowed only for compute shaders.

Rua:
According to the table at https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccessFlagBits.html, `VK_ACCESS_INDIRECT_COMMAND_READ_BIT` is only allowed with the pipeline stages `VK_PIPELINE_STAGE_DRAW_INDIRECT_BIT` and `VK_PIPELINE_STAGE_ACCELERATION_STRUCTURE_BUILD_BIT_KHR`. So changing the pipeline stages to `compute_shader` would be incorrect.

Again, do you have some code I can use to reproduce the problem?

p0lunin:
My code is big and a bit confusing. [There is code](https://github.com/p0lunin/rencan/tree/rework_shaders_v2), [there is dispatch call](https://github.com/p0lunin/rencan/blob/rework_shaders_v2/rencan-render/src/commands/lightning.rs#L72).
I use Nvidia Nsight Graphics to debug it, and it shows that there are not barriers with `VK_ACCESS_INDIRECT_COMMAND_READ_BIT`.
API calls:
image
Barrier 1:
image

Barrier 2:
image

Rua:
Thank you for your code, this allows me to look at it in much more detail!

If I am reading your code correctly, there are two locations where the indirect buffer is being written to:
1. Here, where you put what I presume is a dummy placeholder value: https://github.com/p0lunin/rencan/blob/rework_shaders_v2/rencan-core/src/app.rs#L169
2. Here, where you pass it to a compute shader to fill it with usable data: https://github.com/p0lunin/rencan/blob/rework_shaders_v2/rencan-render/src/commands/ray_trace.rs#L112

I ran `rencan-render` through `cargo expand` to expand all the macros. For the `divide_cs` shader used in case 2, I found that vulkano-shaders is marking the buffer as read-only, even though the shader is writing to it. This is part of what's generated:
```Rust
fn descriptor(&self, set: usize, binding: usize) -> Option<DescriptorDesc> {
    match (set, binding) {
        (0usize, 0usize) => Some(DescriptorDesc {
            ty: DescriptorDescTy::Buffer(DescriptorBufferDesc {
                dynamic: None,
                storage: true,
            }),
            array_count: 1u32,
            stages: self.0.clone(),
            readonly: true,
        }),
        _ => None,
    }
}
```
That seems to be related to #1513, which was supposed to fix this but apparently it didn't completely. That will need closer examination. In any case, the bug doesn't seem to be in this PR. It's not adding a barrier because the `readonly: true` being generated by vulkano-shaders is telling `SyncCommandBufferBuilder` that it doesn't need a barrier.

Eliah-Lakhin:
@Rua, @p0lunin Thank you very much for digging down the root cause of the issue. Could it be that the `Rua:dispatch-indirect` branch didn't pull the latest master changes from the #1513 merge?

Rua:
No that's included already. It's in the changelog (although I moved the message in this PR so that all breaking changes are listed first).

Eliah-Lakhin:
@Rua ok, I think we should proceed with merging as there are no breaking changes, and the issue reported by @p0lunin seem to be(at a first glance) irrelevant to this PR changes. PR is merged. Thank you!

Rua:
@Eliah-Lakhin @p0lunin I tested the rencan code again with the 0.22 release, and now it correctly marks the descriptor as not readonly. It turns out it was using the base vulkano library from my PR, but not the vulkano-shaders library, which was still on the 0.21 release and therefore lacked the readonly fix.

PR #1522, opened 2021-03-29
Split `AutoCommandBuffer` into primary and secondary

Rua:

The `AutoCommandBuffer` is split into `PrimaryAutoCommandBuffer` and `SecondaryAutoCommandBuffer`, and the trait `CommandBuffer` is also split into `PrimaryCommandBuffer` and `SecondaryCommandBuffer`. `AutoCommandBufferBuilder` is still one type, but it has a type parameter for which type it will build.

I've also renamed `Kind` and changed how it works a little, so that it now holds a `CommandBufferInheritance` for secondary command buffers. The `SecondaryCommandBuffer::inheritance` trait method returns this value.

Pros:
- It turns out that there are actually no functions in Vulkano that can accept either type of command buffer. Functions can now specifically ask for one type or the other, giving the wrong type is a compile-time error. This removes a runtime check.
- Likewise, the trait methods of `CommandBuffer` only make sense for primary _or_ secondary command buffers, the only one that works for both is `inner`.
- Some methods on `AutoCommandBufferBuilder` are now only implemented for `AutoCommandBufferBuilder<PrimaryAutoCommandBuffer>`, so this removes a runtime check.

Cons:
- It's now a bit harder to write a function that works for both primary and secondary command buffer builders. You'd have to make it a generic function. Time will tell if this proves to be an obstacle?

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -3,6 +3,9 @@
     Please add new changes at the bottom, preceded by a hyphen -.
     Breaking changes should be listed first, before other changes, and should be preceded by - **Breaking**.
 -->
+- **Breaking** `AutoCommandBuffer` and the `CommandBuffer` trait have been split in two, one for primary and the other for secondary command buffers. `AutoCommandBufferBuilder` remains one type, but has a type parameter for the level of command buffer it will be create, and some of its methods are only implemented for builders that create `PrimaryAutoCommandBuffer`.
+- **Breaking** `Kind` has been renamed to `CommandBufferLevel`, and for secondary command buffers it now contains a single `CommandBufferInheritance` value.
+- **Breaking** `CommandBufferInheritance::occlusion_query` and `UnsafeCommandBufferBuilder::begin_query` now take `QueryControlFlags` instead of a boolean.
 
 # Version 0.22.0 (2021-03-31)
 
--- a/examples/src/bin/deferred/frame/ambient_lighting_system.rs
+++ b/examples/src/bin/deferred/frame/ambient_lighting_system.rs
@@ -9,9 +9,9 @@
 
 use vulkano::buffer::BufferUsage;
 use vulkano::buffer::CpuAccessibleBuffer;
-use vulkano::command_buffer::AutoCommandBuffer;
 use vulkano::command_buffer::AutoCommandBufferBuilder;
 use vulkano::command_buffer::DynamicState;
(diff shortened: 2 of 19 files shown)


Rua:
I don't really understand why this branch is including a whole lot of commits that have already been merged. But I think it's ok? Git is still a bit magic to me.

Eliah-Lakhin:
@Rua This is not a strong requirement to accept this PR, but I detected false-positive issues with the IntelliJ Idea Rust plugin. This is a minimal example:

```rust
use std::sync::Arc;
use vulkano::command_buffer::{AutoCommandBufferBuilder, PrimaryAutoCommandBuffer};
use vulkano::device::{Device, Queue};

fn foo(device: Arc<Device>, queue: Arc<Queue>) -> PrimaryAutoCommandBuffer {
    AutoCommandBufferBuilder::primary(device, queue.family())
        .unwrap()
        .build()
        .unwrap()
}
```

This code compiles perfectly fine with Rust 1.51, there are no compile-time errors, but Idea plugin(I tried nightly and stable versions both) cannot resolve `build` method correctly:

Screenshot from 2021-04-04 09-00-17

Although, this error normally should be addressed to the plugin developers, I think this is an indication of some mess in the Vulkano internal reexports. And since Idea plugin is widely used by the users, I want to ask you to debug your code a little further and try to fix this issue. Even if this is a blame of Idea plugin, it would be good if we align our code such that it would be more ergonomic for Vulkano users.

Rua:
The `build` method is implemented in this block
```Rust
impl<P> AutoCommandBufferBuilder<PrimaryAutoCommandBuffer<P::Alloc>, P>
where
    P: CommandPoolBuilderAlloc,
```
I'm not really sure what the plugin is getting confused about, because at least in Rust this makes perfect sense. As you say, the compiler figures it out just fine. But Idea appears to not recognise that the builder matches this `impl` block, somehow. I don't know what to do about that other than poke the developers.

Eliah-Lakhin:
@Rua ok, it probably will be difficult to fix, and after talking to some people I figured out that the Idea plugin is not that widely used these days, so we probably shouldn't relay on it that much. Please resolve merging conflicts, and I will proceed with merging. Thank you!

Rua:
Done.

Eliah-Lakhin:
@Rua Merged :+1:

 > Looks like someone already made an issue referencing this too, nice!

Yes, I reported it to developers in a chat, and they opened the issue :) Hope it will be resolved eventually, but I believe IDE plugin shouldn't be a blocker to push our project forward.

Issue #1349, opened 2020-04-27
Shader macro not recompiling on change


WSeegers:
```
[dependencies]
vulkano = "0.18"
vulkano-shaders = "0.18"
```

# Issue
When compiling shaders using the path.

```
vulkano_shaders::shader! {
    ty: "compute",
    path: "somepath/someshder.glsl"
}
```
The compilation seems to be cached and does not detect changes to the file. This results in shaders not being updated in subsequent runs.

The same is does not happen when using `src: "some glsl"`.

AustinJ235:
I recall some previous issues about this. But if you use the `include!` macro just to have the compiler watch it that seems to be a workaround.

SamHSmith:
@WSeegers Surely this is a rust issue and not a vulkano one? As in it is up to the rust compiler and cargo to recompile the file at which point you shader gets recompiled.

AustinJ235:
I think there is something that we can hint to the compiler but it might be only available in the build.rs

⁣Get BlueMail for Android ​

SamHSmith:
I'm not so sure. If the compilation has been done and none of the rust source code has changed then a cargo run will completely side-step the compilation stage and run the executable it compiled last time. Fixing this would involve loading and compiling the shaders at runtime. Which may be a feature that someone wants to implement, dynamic runtime spirv compilation. If it isn't already implemented upstream.

est31:
cc https://github.com/rust-lang/rust/issues/73921 and #890

Rua:
I believe this is now fixed?

Issue #890, opened 2017-12-06
Emit throwaway `include_bytes!()` in shader-derive so changes to shader files trigger recompilation


abonander:
In [this Reddit discussion](https://www.reddit.com/r/rust/comments/7hffz6/hey_rustaceans_got_an_easy_question_ask_here/dqttf8m/?context=3), the asker was unsure of how to get Cargo to recognize shader files as source files to trigger recompilation. I suggested either emitting `rerun-if-changed` from a build script or using `include_bytes!("<shader file>")` in their Rust code in order for Cargo to detect changes to shader files when deciding whether or not to recompile.

It occurred to me that shader-derive *could* be doing the latter automatically, but I haven't found any discussion on why it doesn't currently. The additional data would inflate debug builds, but the unused rvalues *should* be optimized out in release builds.

flaghacker:
A potential solution should also consider included files, making it a bit harder then just emitting `rerun-if-changed` in the build script.

Rua:
This seems to be fixed by #1523.

PR #1523, opened 2021-03-30
Use include_bytes! in shader! for automatic recompilation

### LINKED ISSUE #1231: vulkano_shaders::shader! "path:" parameter doesn't incorporate changes to the referenced file by default
*   Version of vulkano: 0.14
*   rustc 1.39.0-nightly
*   OS: Windows 10
*   GPU (the selected PhysicalDevice): GTX 1080
*   GPU Driver: GeForce GRD 431.60
*   See [this demo repo](https://github.com/colekurk/vulkano_shader_issue) for an easy example, though any file using `shader!` with the `path` argument should work.

# Issue

When using `vulkano_shaders::shader!` macro with the `path` variant, the file indicated by `path` is not watched by `cargo`. Changes to the shader file must then be coordinated with a change in the rust file the shader macro runs in (or with a clean build) for the changes to be reflected by the executable. 

It appears to be possible to add an arbitrary file dependency to cargo in *some* cases as indicated by [this code](https://github.com/frehberg/test-generator/blob/master/build-deps/src/lib.rs) which appears to be intended for use in a cargo build script. I'm unsure if a similar technique could work for a macro, or if there's a different API the macro could use that would result in more consistent behavior.

If it isn't currently possible to make this work reliably the docs should at least include a warning of some sort, I think.

This obviously doesn't affect the `src` variant of `shader!`, which by nature changes the file when the shader code changes.

mvilim:
This PR makes use of `std::include_bytes` to automatically recompile path-loaded shaders when they change, as suggested in #890. The resulting variable is left unused, so the compiler is free to optimize it away.

I confirmed that the shader bytes are only included in the binary if `debug = true` (with debug off and `opt-level = 0`, the bytes are not present in the final binary). Given the small size of shader files, I assume including their source by default is reasonable. If we would rather not include it by default, I can add a flag to the shader! macro to control the functionality.

I believe this should also close #1231 and #1349.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -191,6 +191,7 @@ pub(super) fn reflect(
     name: &str,
     spirv: &[u32],
     types_meta: TypesMeta,
+    full_path: Option<&str>,
     dump: bool,
 ) -> Result<TokenStream, Error> {
     let struct_name = Ident::new(&name, Span::call_site());
@@ -255,6 +256,12 @@ pub(super) fn reflect(
         }
     }
 
+    let include_bytes = full_path.map(|s| quote! {
+        // using include_bytes here ensures that changing the shader will force recompilation.
+        // The bytes themselves can be optimized out by the compiler as they are unused.
+        let _bytes = ::std::include_bytes!( #s );
+    }).unwrap_or(TokenStream::new());
+
     let structs = structs::write_structs(&doc, &types_meta);
     let specialization_constants = spec_consts::write_specialization_constants(&doc, &types_meta);
     let uses = &types_meta.uses;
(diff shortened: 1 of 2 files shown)


Eliah-Lakhin:
@mvilim Merged. Thank you for the improvement!

Regarding #1231 and #1349. Do I understand correctly, it affects glsl source changes detection as well?

Eliah-Lakhin:
@mvilim From my observations it didn't fully resolve the issue. This is an example of the macro use:

```
vulkano_shaders::shader! {
    ty: "compute",
    include: [ "./src/shader" ],
    path: "./src/shader/main.glsl",
    types_meta: {
        use serde::{Deserialize, Serialize};

        #[derive(Clone, Copy, PartialEq, Debug, Default, Serialize, Deserialize)]

        impl Eq
    }
}
```

When I change a file indirectly included in the entry-point(`./src/shader/main.glsl`) through the shaderc's `#include <..>` directive, it didn't detect changes. However, if I change the entry-point file itself it does detect the changes.

Perhaps, the more proper solution would be reading such directives and include their sources as well.

Would you mind working on this additional improvement?

mvilim:
I didn't even notice the include directive. Thanks for catching that issue. I have opened #1529 with a fix. Thankfully `shaderc` exposes a mechanism for obtaining the set of files used in the compilation.

PR #1525, opened 2021-03-31
Release 0.22.0

Eliah-Lakhin:
Versions bump and changelogs update for new release.


Rua:
I think for future editors it might be good to include a HTML comment in the changelog, to tell them how to format it. Something like:
`<!-- Please add new changes at the bottom. Breaking changes should be listed first, before other changes. -->`

PR #1529, opened 2021-03-31
Consider included files in automatic shader recompilation

mvilim:
This PR is the follow-up to #1523. The detection of shader changes now also applies to shader files provided via the `include` directive (detected via the `set_include_callback` of `shaderc`).

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -538,7 +538,17 @@ mod tests {
     use super::*;
     use std::path::PathBuf;
 
-    fn prepend_root(root_path: &Path, paths: &[&str]) -> Vec<String> {
+    #[cfg(not(target_os = "windows"))]
+    pub fn path_separator() -> &'static str {
+        "/"
+    }
+
+    #[cfg(target_os = "windows")]
+    pub fn path_separator() -> &'static str {
+        "\\"
+    }
+
+    fn convert_paths(root_path: &Path, paths: &[String]) -> Vec<String> {
         paths
             .iter()
             .map(|p| path_to_str(root_path.join(p).as_path()).to_owned())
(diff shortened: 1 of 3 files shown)


Eliah-Lakhin:
@mvilim Thank you very much for the fix. Some tests are failing probably due to a directory separator(slash/backslash) stringification on Windows: https://github.com/vulkano-rs/vulkano/pull/1529/checks?check_run_id=2241689107

mvilim:
I see that the tests are failing on windows due to path formatting. I will put together a fix.

mvilim:
Oops, I commented before I saw that you had noticed the same thing. Working on a fix...

vulkano-shaders/src/codegen.rs
```diff
@@ -691,22 +701,26 @@ mod tests {
...
         assert_eq!(
             includes_with_relative,
-            prepend_root(
+            convert_paths(
                 &root_path,
                 &[
-                    "tests/include_dir_a/target_a.glsl",
-                    "tests/include_dir_a/../include_dir_b/target_b.glsl"
+                    vec!["tests", "include_dir_a", "target_a.glsl"].join(path_separator()),
+                    vec!["tests", "include_dir_a", "../include_dir_b/target_b.glsl"]
```
Rua, line 714: Should the slashes here be backslashes too on Windows?

mvilim: I considered that myself, but the code before my changes already generated slashes on Windows (inside the path resolution code due to the use of slashes in the literal shader strings/files, e.g. `../include_dir_b/target_b.glsl` in the tests. These strings can lead to mixed path styles, e.g. `shader-include` yields the include "vulkano\examples\src/bin/shader-include\standard-shaders/common.glsl". I tested on Windows 10 and these mixed style paths work for me, in that both the shaders compile (unsurprisingly, as the CI build for Windows passes) and `include_bytes!` properly detects file changes.

My change just adds assertions on the value of those strings while the values of the strings remain unchanged.

I am a bit hesitant to modify the paths without a clear reason since such things tend to be very finicky. If this is something that should be cleaned up (though it's a bit unclear to me how, either via templatizing the shaders or post-processing them) then I think it should probably be a separate PR.

Eliah-Lakhin:
@mvilim Works as expected. Merged. Thank you for the improvement!

mvilim:
Great! Thank you both for taking the time to review it.

PR #1531, opened 2021-04-01
Render pass restructuring

Rua:

This is a major rearranging of render passes and framebuffers. The changelog gives more details about what exactly has been changed. The main goal was to eliminate the type parameter of `RenderPass`, by using a single type for all render pass descriptions instead of a custom type for each one. As a result, a lot of traits became superfluous and could be eliminated, making the API simpler to understand without sacrificing any flexibility.

I also renamed the `framebuffer` module to `render_pass`, because in my mind the render pass is the main higher-level object and framebuffers are just a detail of its interface. It didn't make sense to me to consider render passes a subtopic of framebuffers, the other way around fits better.

I looked at eliminating the one remaining type parameter of `Framebuffer`, which would allow removing the `FramebufferAbstract` trait as well and further simplify some other types. However, this is complicated by the existence of `AttachmentsList`, so I left it for now. This PR works just fine as it is.


Eliah-Lakhin:
@Rua Can you describe please, what problems this rearrangement is trying to resolve, and how would it benefit Vulkano users?

Rua:
My main aim was to reduce the reliance on traits, when single types will do just fine. I realised that the `RenderPassDesc` trait could easily be implemented as a standard `struct`, since the interface of the trait basically reimplemented slices. That then allowed the elimination of a whole bunch of type parameters and other traits, which became superfluous.

Regarding benefits, I think that having less type parameters and traits, without having less functionality, makes the API a bit easier to understand. Users can now store `Arc<RenderPass>` instead of having to figure out `Arc<RenderPassAbstract + Send + Sync>`. Users who want to write their render pass description by hand can now also do so by calling `RenderPassDesc::new` with three `Vec`s, instead of having to implement a whole new type and implement traits on it. I feel this brings the Vulkano API a little closer to Vulkan's own API, which I think is a benefit too. Finally, because graphics pipelines and framebuffers now have methods to return a reference to `Arc<RenderPass>` directly, it's much easier to get the render pass from these objects and that gives the user new options they didn't before.

Finally, regarding the rename of the module, I explained that in the original post. It just makes more sense to me. I can revert that, if you want.

Eliah-Lakhin:
@Rua I didn't detect any issues on my end, but my project doesn't use Renderpass that much extensively(it is compute pipeline-based mostly), so unfortunately I can't give precise evaluation on whether the new API flows well or not. At first glance this refactoring makes sense to me too, and I would proceed with merging.

Can you fix merging conflicts, please?

Eliah-Lakhin (approved):

Eliah-Lakhin:
@Rua Merged. Thank you!

PR #1534, opened 2021-04-02
Fix and refine the generation of `readonly` descriptor attribute

### LINKED ISSUE #1533: Vulkano-shaders incorrectly marks uniform buffers as not readonly
Presumably due to #1513, uniform buffers now get marked as not readonly in descriptor sets. This causes a mismatch error when creating a buffer to bind to the shader. Uniform buffers are never writable, so this shouldn't be happening.

This is the uniform declaration in the shader, for reference:
```glsl
layout(set = 0, binding = 0) uniform Matrices {
	mat4 proj;
	mat4 view;
	mat4 billboard;
	float extra_light;
};
```

Rua:

This fixes #1533 and also improves upon #1513, by properly marking storage image descriptors as readonly only if they are marked such in the shader.

I would recommend releasing a version with this fix (and any other non-breaking ones that are waiting) as 0.22.1, perhaps even yanking 0.22.0. If my experiences are an indication, #1533 breaks any code that uses uniform buffers, which is probably almost all of them.

EDIT: After some more digging, it turns out the bug isn't quite as severe as I thought. It's happening because my code defines its own descriptor instead of using the one from the pipeline. My descriptor of course sets `readonly`, so that causes the mismatch. So it probably doesn't break everyone's code, but if you define your own descriptor for a uniform buffer it likely will.

### DIFF
--- a/vulkano-shaders/src/descriptor_sets.rs
+++ b/vulkano-shaders/src/descriptor_sets.rs
@@ -11,7 +11,7 @@ use std::cmp;
 
 use proc_macro2::{Ident, TokenStream};
 
-use crate::enums::{Decoration, Dim, ImageFormat, StorageClass};
+use crate::enums::{AccessQualifier, Decoration, Dim, ImageFormat, StorageClass};
 use crate::parse::{Instruction, Spirv};
 use crate::{spirv_search, TypesMeta};
 use std::collections::HashSet;
@@ -389,10 +389,14 @@ fn descriptor_infos(
                         decoration_block ^ decoration_buffer_block,
                         "Structs in shader interface are expected to be decorated with one of Block or BufferBlock"
                     );
-                    let is_ssbo = pointer_storage == StorageClass::StorageClassStorageBuffer;
+
+                    // false -> VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER
+                    // true -> VK_DESCRIPTOR_TYPE_STORAGE_BUFFER
+                    let storage = pointer_storage == StorageClass::StorageClassStorageBuffer;
 
                     // Determine whether there's a NonWritable decoration.
-                    let readonly = {
+                    // Uniforms are never writable.
+                    let readonly = !storage || {
                         let mut readonly = vec![false; member_types.len()];
                         for i in doc.instructions.iter() {
                             if let Instruction::MemberDecorate { target_id, member, decoration, .. } = i {
@@ -408,7 +412,7 @@ fn descriptor_infos(
                     let desc = quote! {
                         DescriptorDescTy::Buffer(DescriptorBufferDesc {
                             dynamic: None,
-                            storage: #is_ssbo,
+                            storage: #storage,
                         })
                     };
 
@@ -421,11 +425,17 @@ fn descriptor_infos(
                     ms,
                     sampled,
                     ref format,
+                    ref access,
                     ..
                 } if result_id == pointed_ty => {
                     let sampled = sampled.expect(
                         "Vulkan requires that variables of type OpTypeImage \
                                               have a Sampled operand of 1 or 2",
+                    );
+
+                    let readonly = matches!(
+                        access,
+                        Some(AccessQualifier::AccessQualifierReadOnly)
                     );
 
                     let arrayed = match arrayed {
@@ -435,7 +445,7 @@ fn descriptor_infos(
 
                     match dim {
                         Dim::DimSubpassData => {
-                            // We are an input attachment.
+                            // VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT
                             assert!(
                                 !force_combined_image_sampled,
                                 "An OpTypeSampledImage can't point to \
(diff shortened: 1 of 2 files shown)


vulkano-shaders/src/descriptor_sets.rs
```diff
@@ -459,25 +469,32 @@ fn descriptor_infos(
...
                             let desc = quote! {
                                 DescriptorDescTy::TexelBuffer {
-                                    storage: #not_sampled,
+                                    storage: #storage,
                                     format: None, // TODO: specify format if known
                                 }
                             };
 
-                            Some((desc, true, 1))
+                            Some((desc, !storage || readonly, 1)) // Uniforms are never writable.
```
Eliah-Lakhin, line 485: @Rua Does it mean that bindings such as e.g. `uniform image2D foo` are treated as readonly? I don't think this is always true. The shader can write to such uniforms. For example, in compute shader. And this is a normal practice.

Rua: This line is correct at least from the Vulkan side. The Vulkan standard doesn't allow the shader to write to descriptors of the types `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER`, `VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER`, `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`, `VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE` (and others). Only "storage" type descriptors permit writing in the shader.

The question is how GLSL gets translated to SPIR-V and how Vulkano-shaders then assigns Vulkan descriptor types to what it finds. At least the second half of the question I haven't changed, I assumed Vulkano-shaders' interpretation of the shader was correct, and based the `readonly` conclusion on that. As for how GLSL becomes SPIR-V, I can't tell. You'd have to write some and see what it puts out.

Rua:
I'm experimenting a bit with `shaderc` now to see what kind of output it generates. It appears that I got the code for the readonlyness of images wrong after all, but not in the way you suspected. I ask not to merge this until I resolved it.

Rua:
Ok, this time it actually works, I've tested it properly.

Eliah-Lakhin:
@Rua This is an example that doesn't work for me(but it worked previously).

Error when calling `CommandBufferBuilder::dispatch` for compute-shader pipeline: `Conflict { command1_name: "vkCmdBindDescriptorSets", command1_param: "Image bound to descriptor 1 of set 1", command1_offset: 2, command2_name: "vkCmdBindDescriptorSets", command2_param: "Image bound to descriptor 2 of set 1", command2_offset: 2 }`

These are descriptors it is referring to:

```
layout(set = 1, binding = 0, r8) uniform image2D STORAGE_MASK;
layout(set = 1, binding = 1, r32f) uniform image2D STORAGE_SINGLE_A;
```

Both images I'm attaching through the descriptor set builder are of the form `Arc<ImageView<Arc<StorageImage<Format>>>>`. They have correct underlying formats, it worked previously, and there are no errors during the descriptor set building.

Rua:
This error is coming from `SyncCommandBufferBuilder`, and it's saying that the two image accesses conflict with each other. That should only happen if the same image is bound to both descriptors, is that the case here? If not, then that's a bug elsewhere in the code.

This PR is doing its work correctly here: these storage images are being marked as writable because the shader doesn't explicitly mark them with the `readonly` decoration. It worked previously because Vulkano-shaders treated them as `readonly` regardless of what the shader declared. But that would be a data race if the shader wrote to the images, because `SyncCommandBufferBuilder` would think there wasn't a conflict. The fact that conflicts are appearing shows that Vulkan-shaders is now correctly treating these descriptors as writable.

Eliah-Lakhin:
@Rua You were right. I double checked images ids and realized they had the same ids. Fixing this issue in my code, resolved reported problem.

However, it raises another question with your implementation and maybe #1513 implementation too. The current implementation relaying on the fact that if the user didn't explicitly specify `readonly` or `writeonly` qualifiers, Vulkano should treat attachment as writetable, which might not always be true.

Example(glsl):
```glsl

layout(constant_id = 0) const uint MY_SPECIALIZATION_CONSTANT = 0;

layout(set = 0, binding = 0, rgba32f) uniform image2D SOME_IMAGE;
// May be more binding here

ivec2 INDEX = ivec2(gl_GlobalInvocationID.xy);

void main() {

if (MY_SPECIALIZATION_CONSTANT == 0) {
  imageStore(SOME_IMAGE, INDEX, vec4(1.0));
} else {
  vec4 data = imageLoad(SOME_IMAGE, INDEX);

 // do something with data, but don't write to SOME_IMAGE
}

}

```

Then the user creates two independent pipelines from this shader: first one with specialization constant set to `0`, the second one set to let say `1`.

The driver will eliminate dead code of one of the branches in the condition above based on the specialization constant value during Pipeline creation. And as such corresponding binding(`SOME_IMAGE`) appears writeable or readable accordingly in corresponding Pipeline.

In Vulkano we can establish a rule that without explicit qualifier it would always be treated as `readonly writeonly` both, but I think it contradicts glsl memory model, or at least restricting it. Default option(without qualifiers) doesn't establish such restriction if I understand glsl memory model correctly

Rua:
Unfortunately GLSL follows a C-like model where everything is readable and writable unless qualified. Dealing with the situation like your example would mean that Vulkano-shaders has to read through the entire shader to see if anything actually writes to the variable or not. I don't know if that's even feasible. What Vulkano does now with this PR is the safe option: if you don't know (no decoration) then assume it's potentially writable. That does restrict some use cases for the moment, but I'm not sure how far Vulkano can go to make this "perfect".

Eliah-Lakhin:
@Rua I think for now this solution is the best we can do with reasonable efforts. Another alternative is a deep traversing of spir-v assembly and semi-interpretation which is too hard. Thank you for the fix, and for your help with discovering of the bug in my code too! :)

PR is merged :+1:

PR #1538, opened 2021-04-05
Add component mapping to ImageView, create ImageViewBuilder

Rua:

This adds support for non-identity component mapping/swizzling to image views. Because the situation with the constructors of `ImageView` was getting a bit messy, I decided to create a builder for it, so the user can choose to override only the values they want. For the common case, the default constructor `ImageView::new` is kept, which internally calls the builder and then immediately builds it.

I also added some checks to `PersistentDescriptorSetBuilder`, to ensure that image views are identity swizzled in cases where the standard requires it.


Eliah-Lakhin:
@Rua Works good. PR is merged.

With component swizzling, split of Images and Views makes much more sense for the end user. Thank you for the improvement!

PR #1542, opened 2021-04-07
Partial safe support for queries

Rua:

This gets queries to a working state in Vulkano. There was already a `query` module, but it wasn't really usable for very much. Every query command in core Vulkan is now present on both `SyncCommandBufferBuilder` and `AutoCommandBufferBuilder`, and also some other things elsewhere. I also documented everything query-related, which should help too.

Most safety checks have also been implemented, but not all. The commands that rely on the available/unavailable state do not yet synchronise this, so it's possible for the user to use the same query in multiple command buffers concurrently. These functions are still marked `unsafe`, and a safety condition is included in the documentation.

Hopefully, the remaining safety conditions can be implemented in a future PR. I may do it myself at some point, but I need to figure out how it should work first.


Eliah-Lakhin:
@Rua Would you mind add an example showing how to use queries, please? Also, as usual, we have changelog conflicts.

Rua:
I've never actually used queries before myself, so I'm not really sure what would be a good real-world example. I suppose I could make an example that draws two triangles, one behind the other, and demonstrates that the occlusion query succeeds on the first and fails on the second? I would need some more time to work that out, and would prefer to add that in a separate PR.

Eliah-Lakhin:
@Rua ok, let's implement an example in separate PR, but adding an example is strongly recommended, otherwise it might be unclear on how to use this feature from Vulkano API point of view. Even a trivial example should work.

fyi, we have merging conflicts after https://github.com/vulkano-rs/vulkano/pull/1531 merging.

Eliah-Lakhin:
@Rua PR merged. Thank you!

PR #1550, opened 2021-04-10
Remove format structs and some related traits

### LINKED ISSUE #1518: Remove the format specification structs
At the moment, the `format` module in Vulkano has two different ways to specify formats: either the `Format` enum, or a long list of individual structs. Both of these do the exact same thing. I feel it's kinda silly to have both of these existing side by side, and would like to remove one of them, specifically the individual structs. The `Format` enum would become the only way to specify formats.

It would make downstream code in Vulkano a bit simpler because it doesn't have to deal with the different ways the user may want to specify formats. `ImmutableImage` for example has an `F` type parameter for the format, which is unneeded if `Format` is always used. Some of the traits in the `format` module may no longer be needed either.

### LINKED ISSUE #1524: Image copy commands set multiple aspect bits (validation error)
The Vulkan spec says [here](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBufferImageCopy.html#VUID-VkBufferImageCopy-aspectMask-00212), regarding image copy commands:

> The `aspectMask` member of `imageSubresource` **must** only have a single bit set

This is violated by Vulkano in `copy_image_to_buffer_dimensions`, if an image has both a depth and a stencil aspect:

https://github.com/vulkano-rs/vulkano/blob/c2d404c6697781b01f3581b1ca70447b11e802a6/vulkano/src/command_buffer/auto.rs#L1152-L1156

In the reverse command, buffer to image, depth and stencil aspects are left unimplemented, so the issue doesn't arise there. A proper fix for this would be to give the user control, by allowing to specify which aspect of the image to copy. This would be needed for image-to-buffer copy, but also buffer-to-image and image-to-image copies.

Rua:

This fixes #1518.

EDIT: I've added another commit that fixes up some of the image aspect related types a bit, and adds aspects and planes to `Format`. Consequently this now fixes #1524 too.


Eliah-Lakhin:
@Rua Works fine. Merged. Thank you for your work!

GPU memory objects such as `StorageImage` without (imo) unnecessary `Format` parameter are now becoming more straightforward and produce lesser confusion for the user, which is good. As usual my Idea Rust plugin can't expand `Format` type from the macro impls properly, but again this is an issue of the plugin.

PR #1559, opened 2021-04-18
Add `exact_entrypoint_interface` flag to shader! (Fixes #1556)

### LINKED ISSUE #1556: shader! will no longer generate descriptor information for variables only used in unreachable code.
The changelog says
Breaking shader! will no longer generate descriptor information for variables that are declared but not used in a shader.
But that also happens if they are used in unreachable code. This is very annoying because I often put return at the beginning of a shader for debugging purposes and this fact makes it so this becomes pretty much impossible since nothing will compile.

Arc-blroth:

See #1556 for the motivation for this PR.

This PR reverts the default behavior for finding descriptors introduced in #1497, and adds a new option `exact_entrypoint_interface` to re-enable strict descriptor search.


Eliah-Lakhin:
@Arc-blroth I'm sorry for late review. Both options verified. Works good. Merged.

Thank you very much for this fix!

PR #1562, opened 2021-04-21
Fix bug where AttachmentImages can't be used in multiple descriptor sets at the same time

### LINKED ISSUE #1557: Lots of conflicts with reused uniform buffers in 0.22
With 0.22 I get a *lot* of validation errors when recording draw commands. The offending code is always similar to this:

```rust
for mesh in &obj.meshes {
    let set = Arc::new(
        PersistentDescriptorSet::start(layout.clone())
            .add_buffer(light.uniform_buffer())
            .unwrap()
            .add_buffer(model_ub.clone())
            .unwrap()
            .build()
            .unwrap(),
    );

    builder
        .draw_indexed(
            shadow_pipeline.clone(),
            &dynamic_state,
            mesh.vb.clone(),
            mesh.ib.clone(),
            set,
            (),
            vec![],
        )
        .unwrap();
}
```

i.e. some loop that records draw commands and re-uses the same uniform buffer(s) over and over again. The error is always of the form

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SyncCommandBufferBuilderError(Conflict { command1_name: "vkCmdBindDescriptorSets", command1_param: "Buffer bound to descriptor 0 of set 0", command1_offset: 4, command2_name: "vkCmdBindDescriptorSets", command2_param: "Buffer bound to descriptor 0 of set 0", command2_offset: 8 })', src/main.rs:533:42
```

The error is returned by the `draw_indexed` call.
This code worked in 0.21 but stopped working when I upgraded to 0.22. I tried upgrading to 0.22 because I had similar issues (also conflicts but with images instead of buffers and only in a new piece of code that isn't involved here) and thought the overhauled image code in 0.22 could solve my problems. But now I have buffer problems :D

Is there already a fix for this? I'm also fine with patching up vulkano myself and just remove some validation code for now. Would be nice if someone can give me a hint where this could come from.

You can probably reproduce this issue when porting the examples to 0.22.

faulesocke:
This fixes #1557 It seems to work just fine with all the tests I did.

### DIFF
--- a/vulkano/src/command_buffer/synced/base.rs
+++ b/vulkano/src/command_buffer/synced/base.rs
@@ -733,7 +733,9 @@ impl SyncCommandBufferBuilder {
                                 //
                                 unsafe {
                                     let from_layout = if is_layout_initialized {
-                                        actually_exclusive = true;
+                                        // TODO why was this introduced? Everything seems to work
+                                        // better without this
+                                        //actually_exclusive = true;
                                         initial_layout_requirement
                                     } else {
                                         if img.preinitialized_layout() {


Rua:
A layout transition is a write operation, which means that concurrent access to the same image should not be allowed. Pipeline barriers protect against access on the same queue, but not another queue. The change you propose would introduce a data race if the image is concurrently accessed by another queue.

The `lock_submit` and `check_image_access` functions further down rely on the `exclusive` value to provide the correct locking for the image. If `exclusive` is false, then this locks the image only for reading, which allows a command buffer on another queue to be executed while this image is in use, without the insertion of a semaphore. If the first queue is actually writing to the image at that time, that's a synchronisation error.

faulesocke:
Oh. I wasn't aware that the `exclusive` flag is being used for cross-command buffer dependencies. I will think of a different fix.

Eliah-Lakhin:
@faulesocke Please ping me when it's ready for the final review. Thanks!

Eliah-Lakhin:
@faulesocke Any updates regarding this PR?

faulesocke:
@Eliah-Lakhin sorry I haven't had time for this and probably won't have time for another month or two.

vE5li:
Is there any update on this?

vE5li:
@fayalalebrun Does the commit #a03cfa0 mean this can be closed?

fayalalebrun:
@vE5li No, that's for a fork of Vulkano and doesn't address the underlying issue.

vE5li:
Ah, my bad :)

PR #1564, opened 2021-04-25
Change all flags conversions to use `From`

Rua:

Various places in the Vulkano code have functions to convert to/from raw Vulkan flags. These functions have a variety of names, not always consistent. But since Rust has the standard `From` trait, I converted all these functions into `From` implementations. These were all internal Vulkano functions, so it shouldn't break anyone's code.

I also renamed `AccessFlagBits` to `AccessFlags`, as the addition of "bits" was strange and unnecessary. This is a breaking change, but only for users that actually use this type. This type is used in unsafe functions and types, so only the users that use these directly will be affected.


Eliah-Lakhin:
@Rua I can confirm there are no regressions(at least I didn't detect any).

Regarding the implementation, on one hand I like it because we are making conversions more convenient from the Rust idiomaticity point of view(also, the new names look to me more clean too), on the other hand, some of converted functions implicitly became public. Which is maybe not a big deal, but I suppose they were private for a reason. For example, to hide transparency between Vulkano and VkSys. As such both projects are now closer bound to each other from the public API point of view. If someone start using these conversions, it would be an issue for us in the future when/if we decide to migrate from VkSys. Another feature that we are loosing(generally speaking) is that if intercommunication is hidden it enforces users to contribute to Vulkano when they need to implement a feature that requires VkSys direct usage.

Rua:
Hmm, that is true. It's a shame that you can't make trait implementations private. But I suppose if Vulkano had its own _private_ trait, then that could work? Something like `FromVulkan` and `ToVulkan`? It's not ideal to duplicate standard traits, but if keeping them private is a concern then this is the best I can think of.

Eliah-Lakhin:
@Rua I merged your PR, the work done quite well in general, I just highlighted one issue with implementation(probably not so critical). Will left it at your decision to improve.

 > However, we could also embrace the possibility that people will use them, since it gives users more freedom to interface with raw Vulkan if they want to. Vulkano's API isn't stable as it is.

I agree. But giving a possibility to hack is somewhat violates the idea of encapsulation too :)

PR #1566, opened 2021-04-27
Add `SwapchainBuilder` for easier (re)creation

Rua:

This adds a `SwapchainBuilder`, which provides sensible defaults for several of the values. It also means you don't have to remember the order of all of the parameters of `Swapchain::new`. The builder is used for recreating as well; in that case the values are copied from the existing swapchain. You can then use the builder to override anything you want to, e.g. dimensions.

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -243,10 +243,10 @@ pub struct Swapchain<W> {
     dimensions: [u32; 2],
     layers: u32,
     usage: ImageUsage,
-    sharing: SharingMode,
+    sharing_mode: SharingMode,
     transform: SurfaceTransform,
-    alpha: CompositeAlpha,
-    mode: PresentMode,
+    composite_alpha: CompositeAlpha,
+    present_mode: PresentMode,
     fullscreen_exclusive: FullscreenExclusive,
     fullscreen_exclusive_held: AtomicBool,
     clipped: bool,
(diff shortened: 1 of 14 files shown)


vulkano/src/swapchain/swapchain.rs
```diff
@@ -696,208 +876,107 @@ impl<W> Swapchain<W> {
...
-
-    /// Returns the number of images of the swapchain.
+    /// Sets the number of images that will be created.
     ///
-    /// See the documentation of `Swapchain::new`.
-    #[inline]
-    pub fn num_images(&self) -> u32 {
-        self.images.len() as u32
+    /// The default is 2.
+    pub fn num_images(mut self, num_images: u32) -> Self {
```
Eliah-Lakhin, line 882: @Rua Maybe put `#[inline]` to this and all similar builder's functions?

Rua: Done. Also made a few more improvements.

Eliah-Lakhin:
@Rua Merged. Thank you.

Regarding the builders topic. To my personal taste immutable builders are more convenient and concise when working with objects of multiple(and maybe optional) parameters. In particular the immutability feature allows the user to build the entire object without leaving a call-chain. So this improvement is a step forward in improving of Vulkano API ergonomics. And by the same reason I think it would be more convenient to have CommandBufferBuilders in immutable form too. At some point in the past they were turned to mutable form. Maybe consider changing it back?

Rua:
For command buffers it's a little different. The immutable builder pattern works well as a substitute for a large construction function, where there are many potential defaults and you build the object at the end of the call chain. But the recording process for command buffers is not really builder-like. It's much more involved and there's usually many other steps to be taken in between command buffer calls, some perhaps performed by various functions. You rarely record and build the command buffer in a single call chain.

That kind of pattern works much better with a mutable builder. With the previous immutable builder, if you wanted to record a single command, you had to move the builder out of somewhere (e.g. `Option<AutoCommandBufferBuilder>`), perform the operation, then store it back. Functions that took the command buffer had to return it by value too. This was especially cumbersome when recording commands repeatedly in a loop. The mutable `Option` ended up fulfilling the same role as a mutable borrow, except much more awkwardly.

Issue #1509, opened 2021-03-15
Should Vulkano try to support multiple Vulkan core versions?


Rua:
Vulkano is currently written to require version 1.1 of the spec. However, several parts of the code use extensions that were promoted to core in 1.1. In some cases Vulkano has alternative code paths depending on whether such a promoted extension is supported, for example in the case `VK_KHR_get_physical_device_properties2` in the instance creation code. The question I have is what to do with such cases now and in the future.

Choosing a specific Vulkan version makes work easier for Vulkano, as it can make certain assumptions about what is available and there is less to maintain. Alternative code paths such as the one mentioned above can be eliminated. On the other hand, allowing the user to choose is more flexible, and makes Vulkano less opinionated. But it also means there have to be alternative code paths to support different levels of support on the end user's machine. It may be infeasible to support several versions for other reasons I'm not aware of.

So is each Vulkano version intended for just one specific Vulkan version, or is it desirable to allow the user to choose in the future? Whatever is ultimately chosen, this should be documented for the user.

AustinJ235:
I have looked into this before, https://github.com/vulkano-rs/vulkano/issues/1365. We'd need someway to get the version to vulkano-shaders which is independent from vulkano pretty much.

Rua:
What do you mean by "get the version to vulkano-shaders"?

AustinJ235:
The vulkan version used by vulkano-shaders/shaderc is hardcoded atm. vulkano-shaders doesn't currently have a means of accessing the vulkano instance therefore no access to knowing what version is set. If you looked at the issue I linked, it contains the two locations where the vulkano version is set from. 

I purposed adding an additional optional field to the shader! macro. I feel this is kinda clunky though.

Rua:
In principle, the version passed to Vulkano-shaders is a compile-time constant. That means whatever version is given to Vulkano itself during Vulkan initialisation should also be a compile-time constant, in order to match the shaders. Is there some kind of user-specified configuration option that can be passed to the compiler of both crates? Something the user can set in their Cargo.toml to tell both Vulkano and Vulkano-shaders what Vulkan version to compile for?

Rua:
[...]

PR #1569, opened 2021-05-04
Add support for instance and device versions, check version in vulkano-shaders

Rua:

This PR adds the ability to query the API version of an instance or device. It also adds a method to query the highest supported instance version before creating an instance. To understand what's going on, let me explain.

Versions in Vulkan are a little more involved than simply specifying a number. Rather, there is a difference between desired version, supported version, and the final API version. Desired version is always 1.1 in Vulkano at the moment, is set when creating an instance, and applies to that instance and all devices using it. Supported version is intrinsic to a specific Vulkan library or physical device, and can differ between them (that is, library and device can support different versions). The final API version you end up getting for an instance or device is the minimum of the supported version and the desired version. So in Vulkano you can never get a 1.2 device, but you can get a 1.0 device! (Due to an oddity with Vulkan 1.0, you can never get a 1.0 instance, however)

That's why I also added some checks to vulkano-shaders to query the version of the device, and make sure the SPIR-V version of the shader is supported by the device. Previously, the user could create a device that supported only Vulkan 1.0, then provide it with a SPIR-V 1.3 shader, which is not allowed. Now, this error should be caught.

The desired version remains the same in this PR, always 1.1. However, I do want to look at allowing the user to change the desired version. That would require solving #1509 first.


Eliah-Lakhin:
@Rua Merged :+1:

Issue #1545, opened 2021-04-09
Segfault when creating a buffer with exportable fd on Linux after updating to 0.22


hakolao:
After updating to 0.22 I get segfault when creating a `DeviceLocalBuffer::raw_with_exportable_fd`.

*   Version of vulkano: 0.22
*   OS: Linux
*   GPU (the selected PhysicalDevice): GeForce RTX 2070 with Max-Q Design with 7.5 CUDA compute capability
*   GPU Driver: `460.32.03`
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: TODO

# Issue

Segfault after #1510. Segfaulting at:
`ptr_chain_iter`
![image](https://user-images.githubusercontent.com/12234005/114188581-f5ca2b00-9951-11eb-90d8-77e465a7d943.png)

It seems to be related to the allocation size somehow. I don't have much clue what the changes are for in above linked PR that broke the functionality for me, but just reporting this is happening. 34mb size seems to be enough to trigger this.

I'll use the previous versions meanwhile.

`main.rs:`
```rust
use vulkano::buffer::{BufferUsage, DeviceLocalBuffer};
use vulkano::device::{Device, DeviceExtensions};
use vulkano::instance::{Instance, InstanceExtensions, PhysicalDevice};

fn main() {
    // Add instance extensions based on needs
    let instance_extensions = InstanceExtensions {
        khr_surface: true,
        khr_xlib_surface: true,
        khr_get_physical_device_properties2: true,
        khr_get_surface_capabilities2: true,
        ..InstanceExtensions::none()
    };
    // Create instance
    let instance =
        Instance::new(None, &instance_extensions, None).expect("Failed to create instance");
    // Get most performant device (physical)
    let physical = PhysicalDevice::enumerate(&instance)
        .fold(None, |acc, val| {
            if acc.is_none() {
                Some(val)
            } else if acc.unwrap().limits().max_compute_shared_memory_size()
                >= val.limits().max_compute_shared_memory_size()
            {
                acc
            } else {
                Some(val)
            }
        })
        .expect("No physical device found");
    println!(
        "Using device: {} (type: {:?})",
        physical.name(),
        physical.ty()
    );
    let queue_family = physical
        .queue_families()
        .find(|&q| q.supports_graphics())
        .expect("couldn't find a graphical queue family");
    let device_extensions = DeviceExtensions::supported_by_device(physical);
    let features = physical.supported_features();
[...]

PR #1580, opened 2021-05-13
An improved fix for info pointer chain segfault

hakolao:

This should be a better fix for https://github.com/vulkano-rs/vulkano/issues/1545. At least the test in the issue report works for me in `cargo run` and `--release` without segfault.

The problem seemed to be that `&mut prev` was not null when `prev` was. Also, to achieve the same functionality `ash` had from where the function was copied I think the incoming `next` should be passed instead to `ptr_chain_iter`. And finally the `next_ptr` should become the new head.

1. Find end of `next`
2. Set end's next (which is null) to be `prev_head`'s next only if previous head was not null
3. Set prev_head (`self.allocate.pNext`) be new head `next_ptr`.

### DIFF
--- a/vulkano/src/memory/device_memory.rs
+++ b/vulkano/src/memory/device_memory.rs
@@ -224,8 +224,10 @@ impl<'a> DeviceMemoryBuilder<'a> {
             let mut prev_head = self.allocate.pNext as *mut BaseOutStructure;
             // Retrieve end of next chain
             let last_next = ptr_chain_iter(next).last().unwrap();
-            // Set end of next chain's next to be previous head
-            (*last_next).p_next = prev_head;
+            // Set end of next chain's next to be previous head only if previous head's next'
+            if !prev_head.is_null() {
+                (*last_next).p_next = (*prev_head).p_next;
+            }
             // Set next ptr to be first one
             prev_head = next_ptr;
         }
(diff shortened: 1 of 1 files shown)


dadleyy:
@hakolao thank you for looking into this!

I was wondering - since the `DeviceMemoryBuilder` owns it's `dedicated_info`, `export_info` and `import_info` as `Option<T>`'s, what benefits are we getting from using `push_next` in addition to mutating those fields? can we just create the `MemoryAllocateInfo` struct in `build` based on those properties instead of mutating `allocate` and having to rebuild the chain along the way?

I created a proof of concept of that approach [here](https://github.com/vulkano-rs/vulkano/compare/master...dadleyy:no-chain?expand=1).

hakolao:
@dadleyy  I must admit that looking at `dedicated_info`, `export_info` and `import_info` functions, your approach would be just fine, because only one of each can exist (otherwise `panic`). I didn't do the `DeviceMemoryBuilder` refactor though and was just trying to fix the pointer mistakes that were segfaulting for 0.22 (and still 0.23).

:adhesive_bandage: for :bug:

dadleyy:
@hakolao understood, thank you!!

Eliah-Lakhin:
@hakolao Thank you very much for digging inside the issue and fixing it! PR is merged. :+1:

Since Ash distributed under MIT, I think it's legal for us to grab even a part of their codebase under this license terms, especially since we put a link to original work.

PR #1581, opened 2021-05-13
Major refactor of shader interface and pipeline layout types

### LINKED ISSUE #1573: Allow graphics pipelines without a fragment shader
It is perfectly valid to have a graphics pipeline consisting of a vertex shader alone. You don't actually need a fragment shader. This just means that no colour output is generated, but you still get depth/stencil writes. However, Vulkano doesn't currently allow this combination, it unwraps the fragment shader in the builder assuming it is valid. It would be nice if this were allowed. I had a look at the code, and it's rather convoluted due to the presence of lots of type parameters on the builder...

Rua:

This is similar to the refactor to render passes that was done in the previous version, but because there are more types involved, the change is also bigger. In short, `ShaderInterfaceDef` and `PipelineLayoutDesc` are now structs, `PipelineLayout` has no more type parameters, neither do `ComputeEntryPoint` and `GraphicsEntryPoint`. A whole bunch of traits that were made superfluous by this were removed. The functionality was replaced by regular values, which could be (`unsafe`ly) provided at runtime if the user desires.

I've also renamed the `descriptor::pipeline_layout` module to `pipeline::layout`. Why it was categorised under descriptors I have no idea, the new location makes a lot more sense IMO.

In the end, all this fixes #1573 too. Because all the shaders and layout now have the same concrete type inside `GraphicsPipelineBuilder`, it is much easier to handle all possible combinations of shaders. The previous code had branches for every possible configuration, which was obviously not easy to extend: adding an optional fragment shader to the mix would have doubled number of the possible combinations!


CHANGELOG_VULKANO.md
```diff
@@ -3,7 +3,15 @@
     Please add new changes at the bottom, preceded by a hyphen -.
     Breaking changes should be listed first, before other changes, and should be preceded by - **Breaking**.
 -->
-
+- **Breaking** Changes to shader interfaces and pipeline layouts.
+  - The module `descriptor::pipeline_layout` has been renamed to `pipeline::layout`.
```
Eliah-Lakhin, line 7: From the naming "convention" point of view the Layout term is related to shader's Descriptors, not the pipeline itself. I don't like the initial module name too, but I think `pipeline::layout` could misleading too.

examples/src/bin/dynamic-buffers.rs
```diff
@@ -88,19 +88,11 @@ fn main() {
...
-                        .unwrap(),
-                    vec![(0, 0)], // The dynamic uniform buffer is at set 0, descriptor 0
-                )
-                .build(device.clone())
-                .unwrap(),
-            ),
+            {
+                let mut layout_desc = shader.main_entry_point().layout_desc().clone();
+                layout_desc.tweak(vec![(0, 0)]); // The dynamic uniform buffer is at set 0, descriptor 0
+                Arc::new(PipelineLayout::new(device.clone(), layout_desc).unwrap())
```
Eliah-Lakhin, line 94: Why is it became an Arc instead of a Box?

CHANGELOG_VULKANO.md
```diff
@@ -3,7 +3,15 @@
...
+  - The trait `PipelineLayoutDesc` has also been converted into a struct, with methods `descriptor_sets` and `push_constants` which return slices. The traits `PipelineLayoutSuperset`, `PipelineLayoutSetsCompatible` and `PipelineLayoutPushConstantsCompatible` have been integrated into this trait. The `EmptyPipelineDesc` trait has been removed. `RuntimePipelineDesc` is merged into `PipelineLayoutDesc`.
+  - Consequently, functions that took a value of these traits now take a plain `ShaderInterface` or `PipelineLayoutDesc`. Types that had a type parameter for it no longer have it, e.g. `VertexDefinition`, `GraphicsEntryPoint`, `GraphicsEntryPointAbstract`, `PipelineLayout`.
+  - Now that `PipelineLayout` has no more type parameter, the trait `PipelineLayoutAbstract` is removed. The layout type parameter is also removed from `ComputePipeline` and `GraphicsPipeline`.
```
Eliah-Lakhin, line 11: I believe that the initial design goal for Vulkano descriptors/pipelines API was to enforce the user to correctly use Descriptor Sets in compile time. That's why I think all of these `*Abstract` types and type type-params were introduced in first place.

This seem to be no longer true after some medium refactorings done in the past, but there still was a possibility to return back to initial goals before of this refactoring. So... I would say that this shift to runtime-checking and runtime building is not necessary bad, but perhaps we should discuss this approach at first, and take full responsibility and understanding that we, as a community, are turning Vulkano into runtime-configurable API instead of statically configurable, which is, in particular, involving some runtime overhead too. Again, I'm not saying this is going to be a bad approach in general, and personally I like the implementation proposed by @Rua, but I just want to highlight a general design contradiction that we should discuss.

@AustinJ235, @gurchetansingh @Arc-blroth, @stephan-cr, @faulesocke, @apessino I would like to hear your opinions on this topic. Thank you.

Eliah-Lakhin (commented):
@Rua Thank you for your hard work! I like this implementation in general, but I left a few questions to be addressed in this PR.

At this moment I only made a code review. Once the questions addressed, I'm also going to do some local code tests.

examples/src/bin/dynamic-buffers.rs
Rua: A pipeline layout needs to be passed on to certain commands in a command buffer, in much the same way as a full pipeline, image or buffer. For that reason I figured it would make the most sense to store it in an `Arc` within the pipeline, so that it can be easily cloned. Previous Vulkano code would clone the `Arc` that wraps the pipeline instead, which is a bit strange in cases where only the layout is needed.

CHANGELOG_VULKANO.md
Rua: But it is a _pipeline_ layout, distinct from a _descriptor set_ layout which is another type of object. I would expect to find a pipeline layout under pipeline, not under descriptor.

Eliah-Lakhin: @Rua ok, makes sense.

CHANGELOG_VULKANO.md
gurchetansingh: I like the change.  The project must evolve to maintain revelance.  We can focus of performance later, but most important is safety and code quality.

Eliah-Lakhin:
ok. I left the time for the community to participate in discussion regarding these changes. Since there were no objections, and there are even some acceptance, I think we should proceed with merging.

@Rua Please fix the merging conflicts with the changelog(again! :), and I will merge the Pull Request. Thank you for your hard work!

Eliah-Lakhin:
Merged :+1:

PR #1584, opened 2021-05-16
Allow specifying the target Vulkan and SPIR-V versions in vulkano-shaders

Rua:

This implements what was proposed in #1509. Vulkano-shaders now produces SPIR-V 1.0 code by default, which can be overridden with two new arguments on the `shader!` macro. This means that vulkano-shaders is now completely independent of the device Vulkan version at runtime. In theory, a PR could be made now to allow the user to specify their desired version when creating an instance. Of course, Vulkano itself doesn't make use of any Vulkan 1.1 or 1.2 features yet, but that would be a future step.

### DIFF
--- a/CHANGELOG_VULKANO.md
+++ b/CHANGELOG_VULKANO.md
@@ -3,8 +3,10 @@
     Please add new changes at the bottom, preceded by a hyphen -.
     Breaking changes should be listed first, before other changes, and should be preceded by - **Breaking**.
 -->
+
 - **Breaking** Vulkano-shaders now checks if the device supports the shader's SPIR-V version, when loading the shader.
-- Added `DeviceExtensions::khr_spirv_1_4`, which allows SPIR-V 1.4 shaders in Vulkan versions below 1.2.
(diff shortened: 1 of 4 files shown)


Eliah-Lakhin:
@Rua Thank you for your work. Maybe for back-compatibility purposes we can consider changing the default behavior when versions are not specified explicitly to set them to their previous values(1.1 for vulkan and 1.3 for spirv?). That would avoid confusion for the users when migrating to the next Vulkano release.

Rua:
That would make sense in the current state of things, but I have a followup PR ready that adds the ability for users to select their desired Vulkan version. Once that's merged, there's really nothing "special" about 1.1 anymore in Vulkano, other than that it was at one point the default. 1.0 is a better default in the long term (once the old default is long forgotten), because it's the base version, and also because that's exactly how `shaderc` works.

Eliah-Lakhin:
@Rua ok, I merged the Pull Request. We can make a decision to revert to other default values in the future based on the users feedback.

PR #1588, opened 2021-05-20
improved image layout transitions

### LINKED ISSUE #1578: `is_layout_initialized` always returns false for several image types
@treeki asked me for some troubleshooting with Vulkano, because their image was being garbled after uploading it. We tracked down the API calls with RenderDoc, and found that whenever an `ImmutableImage` is recorded first into a command buffer, Vulkano inserts an image memory barrier that transitions the layout from `Undefined` to `ShaderReadOnlyOptimal`. This happens despite the image already having been uploaded and initialised properly with a separate command buffer/future. I confirmed that Vulkano's "image" example displays this behaviour too.

The `ImageAccess` trait has a method called `is_layout_initialized`, which is called by `SyncCommandBufferBuilder` to figure out if the image needs to be transitioned into another layout on first use. If so, then it inserts the aforementioned barrier and calls the `layout_initialized` method on the image. But for some reason, `ImmutableImage` and several other image types leave these two methods at their default, meaning that the former always returns `false` and the latter does nothing at all. Consequently, it inserts this barrier each time you use the image in a command buffer.

This behaviour has apparently been there since the olden days of Vulkano. So I am quite puzzled how this _hasn't_ caused obvious problems until now, and I'm hesitant to make any changes in case it turns out it was made this way for a reason.

### LINKED ISSUE #1595: Multisampling may be broken
*   Version of vulkano: 0.23
*   OS: Void Linux - Kernel 5.12.5
*   GPU (the selected PhysicalDevice): AMD RADV VEGA10 (ACO)
*   GPU Driver: Mesa v21.0.3
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: msaa-renderpass

# Issue

Example fails to start at all, no output. In Basalt using MSAA results in an `ImageNotInitialized` on the multisampled `AttachmentImage` during the `begin_render_pass` call. Not sure if there was a recent change that effected the behavior of multisamping coming from version 0.20, but it was working on my system with that version.

I'll investigate this further, but I am done for the night. Bit of lengthy process updating Basalt to the newest version.

Timbals:
I ran into the same problems that issue #1578 described when using [RenderDoc][renderdoc] to replay Vulkan commands and traced it down to `vulkano` inserting image memory barriers at the beginning of the renderpass that transitionend my textures from `UNDEFINED` to `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`. This caused [RenderDoc][renderdoc] to fill the texture with [a pattern for recognising discarded resources][renderdoc-issue].

They do this because according to [the vulkan specification][spec-undefined-discard]:
> Layout transitions always apply to a particular image subresource range, and specify both an old layout and new layout. [...] If the old layout is VK_IMAGE_LAYOUT_UNDEFINED, **the contents of that range may be discarded**.

Although this doesn't actually cause issues on my machine (NVIDIA GPU) this should probably still be fixed for correctness and to remove unneccessary image memory barriers.
I removed the default implementation of the `layout_initialized` and `is_layout_initialized` functions of the `ImageAccess` trait and implemented these for all remaining `vulkano` types that implement `ImageAccess`.

This is a breaking change for anyone that currently implements the `ImageAccess` trait.

fixes #1578
fixes #1595

- **Breaking** the first `CommandBuffer` that uses images will now initialize their layout which means the execution of the first command buffer that uses an image should be finished before recording more command buffers that use the image
- **Breaking** the `ImageAccess` trait now requires the implementation of the `layout_initialized()` and `current_layout()` functions, removed the  `layout_initialized()`, `initial_layout_requirement()` and `forced_undefined_initial_layout()` functions and changed the `unlock()` function to always receive the image layout that a command buffer transitioned to
- **Breaking** the `check_image_access()` function should no longer ignore `Undefined` layouts
- fixed incorrect layout transitions that caused issues when using RenderDoc
- implemented `is_layout_initialized()`, `try_gpu_lock()` and `unlock()` correctly for all image access types

[renderdoc]: https://renderdoc.org/
[renderdoc-issue]: https://github.com/baldurk/renderdoc/issues/284
[spec-undefined-discard]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#synchronization-image-layout-transitions


Rua:
Thank you for taking the time to fix this!

I've been pondering this problem myself, and noticed that there is a subtle bug that can happen with the current code. `layout_initialized` is called when _recording_ a command buffer, which is fine as long as you submit the command buffers in the same order that you record them. But does this still work if you submit the command buffers in a different order, or worse, if you drop the buffer that does the initialization without submitting it altogether? This isn't criticism against your PR btw, it's just something that still needs considering for a future fix.

Timbals:
I looked through the code and didn't find any easy way to fix the bug.

What do you think about removing the call to `layout_initialized` for now?
That would mean images are only marked as initialized in `unlock` which is after the layout transition has been completed. It would still cause other command buffers that get recorded before the `unlock` to insert incorrect image memory barriers (which they currently always do anyway without this PR).

Additionally `try_gpu_lock` could be changed to take an `Option<ImageLayout>` where the `None` value indicates that the caller doesn't care about the layout and `Some(Undefined)` indicates the caller expects the image to still be in an undefined layout. That would enable safe tracking of the image layout but would probably require some effort to change the command buffer builder implementation and is probably out of scope for this PR.

Removing the call to `layout_initialized` should be enough for this PR.

Eliah-Lakhin:
@Timbals Thank you for your contribution.

Unfortunately, I can't fully wrap my head around this issue as I'm not fully understand synchronization mechanisms too, so I would left it for your decision(@Timbals and @Rua) and will test just the final result. Please let me know when it's ready for the final review. Thanks!

AustinJ235:
Just some thoughts. Ran into to this in my project.

So what I can manage to sort out is that there are 3 layouts that an image could be in, VK_IMAGE_LAYOUT_UNDEFINED, VK_IMAGE_LAYOUT_PREINITIALIZED, and VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL; talking just color ones that is. undefined and preinitialized sound to me like they would be write only in this state and it is only where we'd be reading that layout would truly matter. Could be wrong here though. Would it be possible to remove layout checks where we'd be writing? Although undefined hints that it doesn't have a layout is this important? Then make it so where the image/resource is being read from that it needs to be in an optimal state?

Rua:
`VK_IMAGE_LAYOUT_PREINITIALIZED` only works for linear tiling, where layouts don't matter. And I don't believe any of Vulkano uses it currently.

`VK_IMAGE_LAYOUT_UNDEFINED` indeed means that the image can only be written (following a layout transition to something sensible). But more specifically, it means that the image is considered garbage, and any part of the image that is to be read from must have been written first. In theory, you can write to part of an image, leaving the rest still uninitialised. Then read from that part and write to another part. I believe this technique can be used for generating mipmaps: you write the largest mipmap first, then read from it in order to generate the next, then read from that in order to generate the next, etc.

Moreover, it is in theory possible to _transition_ only part of an image. That is, different parts can have different layouts. I don't know how Vulkano would deal with such a scenario, since it would essentially have to keep track of which parts of an image are initialised with data and which are still garbage. That would be really complex to handle. It would be simpler if Vulkano enforced that the entire image must have a single layout, and that all of it must be initialised at once.

Vulkano doesn't currently have any way to detect/enforce initialisation with data though. `SyncCommandBufferBuilder` inserts a layout transition from `VK_IMAGE_LAYOUT_UNDEFINED` at the start of the command buffer, but it doesn't check at all if the first use is a read or a write, nor if that write covers the whole image. In the case of writes by a shader to a storage image it may be impossible to detect this, too. A shader could read/write to a storage image at will, and Vulkano has no hope of enforcing that the shader writes to every pixel, nor that it doesn't read from uninitialised pixels.

AustinJ235:
Another thing would be, is it worth the extra runtime checks for the small amount of safely? I mean the user is going to know if they are reading garbage or not. I don't think anything bad happens from reading garbage data.

Rua:
It is technically undefined behaviour, which is very evil in the world of Rust. But I don't know if UB on the GPU has the same kinds of consequences as UB on the CPU.

AustinJ235:
I think if we do implement any sort of checks that it should be once the image has any sort of write access it is assumed that it is read ready. That'd probably be enough to catch most common errors people make, and it wouldn't have that much overhead I wouldn't think.

AustinJ235:
Although we'd have to check if for example an AttachmentImage is transient or not, because in that case the gpu doesn't care what is stored before or after the renderpass.

Timbals:
@Eliah-Lakhin ready for review

Eliah-Lakhin:
@Timbals I merged the recent `master` onto your branch(locally) and triangle example is failing with the following error:

```
Using device: Intel(R) HD Graphics 630 (KBL GT2) (type: IntegratedGpu)
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AccessError { error: ImageNotInitialized { requested: Undefined }, command_name: "vkCmdBeginRenderPass", command_param: "attachment 0", command_offset: 0 }', examples/src/bin/triangle.rs:489:22
```

Same with AMD, and some other examples.

Eliah-Lakhin (changes requested):
Please see my comment above.

Eliah-Lakhin:
@Timbals fyi, I set your PR on priority. So, once all related issues will be fixed I will proceed with merging as soon as possible. I would like to include this fix in the upcoming release. Thank you.

Timbals:
I fixed the immediate issue.

There is still an issue with the `teapot` example (and probably others as well).
The command buffer for the first frame contains image layout transitions from `UNDEFINED` to the correct layouts. When the command buffer for the second frame gets recorded it also inserts transitions from `UNDEFINED` to the correct layouts because the first command buffer is still in flight and the layouts have not been changed yet. These additional transitions cause an error at submission time.

Eliah-Lakhin:
@Timbals Thank you for the fix. Would it be possible to provide fixes for other examples too? We are using these examples as a form of least manual tests to check if no regressions introduced.

Timbals:
@Eliah-Lakhin I've added a workaround to all affected examples. I'm not sure what a better long-term solution would look like because Vulkano can't know in which order command buffers get submitted (or if at all) by the user.

Rua:
Perhaps there could be an `unsafe` command added to the command buffer builder, that manually adds the layout transition (presumably only from `Undefined` to the image's standard one). The constructors of the various image types can then use this command. That way, it is no longer necessary for the builder to try to autodetect the state of the image. It needs to be unsafe, because if you transition an image and then don't fill it with data, you read garbage. So the user of this command has to promise to fill the image with data before using it.

Eliah-Lakhin:
@Timbals, @Rua  Honestly speaking I don't like the current workaround. It's unclear for the user why they need to wait for the first submission. Moreover, it looks like this is a behavior that should normally be encapsulated somewhere inside Vulkano. Why can't we implement synchronization under the hood, e.g. in the gpu futures mechanism?

Eliah-Lakhin:
Or at least we can add more explicit method for gpu future to check for required transitions. Kind of `future.ensure_image_layout_transition()`. And it will be up to the user to put this method based on the program logic.

Eliah-Lakhin:
@AustinJ235 May I ask you to help me with this Pull Request review, please?

AustinJ235:
@Eliah-Lakhin I am investigating a solution to this. I personally believe this pull request is the wrong approach. I think the user should have not have to think about transitioning the image.

AustinJ235:
@Eliah-Lakhin Seems I was having another possibly unrelated issue. I think the issues outlined in https://github.com/vulkano-rs/vulkano/issues/1578 seem valid. But in this pull request the user is having to keep track of if something is initialized, I don't like that one bit. I think in that specific case my fix for the issue I was having would solve specific issue.

I think the methods `is_layout_initialized` & `layout_initialized` should be reviewed more. There might be an issue there with memory barriers in the aforementioned issue.

Timbals:
I also don't like the current solution.
A better approach would probably be to transition images to a layout when they are created and require all command buffers to transition any image resource back to its specified layout after use.
The downside would be that image creation would involve a command buffer submission and waiting on the resulting future.

Rua:
> A better approach would probably be to transition images to a layout when they are created and require all command buffers to transition any image resource back to its specified layout after use.
> The downside would be that image creation would involve a command buffer submission and waiting on the resulting future.

Isn't that exactly how all the current images already work?

Timbals:
@Rua as far as I know no image layouts are currently initialized at image creation and instead get re-initialized in every command buffer which may discard all contents of the image. The change from the current implementation would be to do the initialization once at image creation.

Rua:
Unless I'm missing something, when an image is created with data to initialise it, the constructor first creates a `CpuAccessibleBuffer` and directly writes the image data to it. Then it creates a command buffer and adds a copy command to copy from the CPU buffer to the GPU image. This command buffer then gets executed on the queue, and its future is returned from the constructor.

The constructor itself doesn't do any image transition, but the command buffer that is being created here is supposed to recognise automatically that the image is uninitialised, and add the transition itself. It's this detection that's going wrong currently. In [this comment](https://github.com/vulkano-rs/vulkano/pull/1588#issuecomment-856530332) above, I proposed an alternative solution: instead of the command buffer trying to figure out when to add the transition, the image constructor does it itself, via a new command on the command buffer builder.

Timbals:
I implemented the solution that makes images initialize their layout when they are created.
That means the user doesn't have to worry about transitions anymore and the only change in the examples is that creating images now required passing in a `Queue` that will be used to execute the transition.

vulkano/src/command_buffer/auto.rs
```diff
@@ -1733,6 +1733,20 @@ impl<L, P> AutoCommandBufferBuilder<L, P> {
 
         Ok(self)
     }
+
+    /// Adds a memory barrier that transitions the `image` from the `Undefined` to `image.layout()`
+    /// layout.
+    ///
+    /// # Safety
+    /// The layout of the image must not have been transitioned to `image.layout()` before.
```
Rua, line 1741: This is actually not unsafe. When you transition from `Undefined`, the previous contents is ignored, so there's no problem. What the user _should_ do here, though, is make sure that the image is filled with valid contents before it's read.

Eliah-Lakhin:
@Timbals I proceeded with merging of #1608 from @AustinJ235 . I hope it would help resolving the issue at least in form of a workaround.

I would like to ask you and @Rua if it resolving the original issue too, and if you think that Austin's solution needs further investigations and improvements.

Rua:
@Timbals are you still working on this?

Timbals:
I don't have the time to work on this any more, so I'm closing the Pull Request.

PR #1583, opened 2021-05-15
vulkano/image: SampleCounts for each pixel in an Image.

prataprc:
Map u32-bit SampleCountsFlagBits type to boolean collection of SampleCounts struct type.

* [x] Added an entry to `CHANGELOG_VULKANO.md` or `CHANGELOG_VK_SYS.md` if knowledge of this change could be valuable to users
* [x] Updated documentation to reflect any user-facing changes - in this repository
* [ ] Updated documentation to reflect any user-facing changes - PR to the [guide](https://github.com/vulkano-rs/vulkano-www) that fixes existing documentation invalidated by this PR.
* [x] Ran `cargo fmt` on the changes

PR #1593, opened 2021-05-23
Migrate to using Ash as a base

### LINKED ISSUE #1500: Proposal: replace vk-sys with ash
While searching for possible improvements to Vulkano design and maintainability, I came across the [ash](https://github.com/MaikKlein/ash) crate. This is a very lightweight wrapper around the raw Vulkan API, designed to make it more Rusty but in a very unopinionated way. Meanwhile, vk-sys is Vulkano's own version of the bindings, which is much more C-like, not always up-to-date, and attempts to auto-generate it have stalled (#89).

Therefore, I propose migrating Vulkano over to ash, and removing vk-sys entirely. It would make Vulkano's low-level code a little nicer, and also resolve the maintainance problem of vk-sys by building on top of the excellent work of others. We can also decide to re-export some of their types and functions, if that is convenient. In the case of #1496 that means we can directly use ash's bitflags types.

Rua:

This migrates Vulkano to use Ash, which closes #1500. I went with Ash in the end because it has some design decisions I agreed with more, and because I became more familiar with it. But I don't think it will matter too much.

Despite being a massive change affecting large parts of the codebase, the number of breaking changes is very small. The most obvious breaking change is that I took #1583 to its logical conclusion and created a `SampleCount` enum, which is now used everywhere to represent sample counts. Vulkano previously used a simple `u32` for this, but this didn't match the actual Vulkan spec, which only allows power-of-two values and defines symbols for them. The Ash migration made the old Vulkano approach more difficult to deal with, so I went ahead and included this change.

The only other breaking changes are those to the raw function pointers, which Vulkano exports from `FunctionPointers`, `Instance` and `Device`. This is by necessity, since these were types and symbols exported directly from vk-sys, and had to be adapted to how Ash handles them.

The vk-sys library now isn't used by Vulkano at all anymore, but I've left it alone in this PR. It could be kept around for anyone to use, but since it's so incomplete I doubt it will be used much. I'll leave it to the maintainers to decide what to do with it.

I went for a relatively minimal conversion to Ash (if you can still call it that, with 101 files changed...), where not much is changed in Vulkano other than what's needed to get the code to work. This means I used Ash's raw structs rather than the accompanying builders, and I used the raw Vulkan functions rather than Ash's "idiomatic" wrappers. This is mainly a matter of code style, and can always be changed in the future if anyone wants to.


vulkano/src/command_buffer/pool/sys.rs
```diff
@@ -384,11 +393,15 @@ mod tests {
     #[test]
     fn check_maintenance_when_trim() {
         let (device, queue) = gfx_dev_and_queue!();
-        let pool = UnsafeCommandPool::new(device, queue.family(), false, false).unwrap();
 
-        match pool.trim() {
-            Err(CommandPoolTrimError::Maintenance1ExtensionNotEnabled) => (),
-            _ => panic!(),
+        // TODO: test doesn't work with Vulkan 1.1
+        if device.api_version() < Version::V1_1 {
```
Eliah-Lakhin, line 398: This is an indication of migration regression.

Eliah-Lakhin (commented):
@Rua I want to say thank you for this great work, probably one of the most significant change in Vulkano for the recent years. And I believe it moves Vulkano in right direction making it closer tied to the rest of the overall Rust CGI infrastructure.

At the same time, this is quite risky migration too. I pointed one of the issues that probably needs further investigation. In general, I think it would be reasonable to try to cover with more unit tests whenever you think it could reveal undiscovered migration issues. As usual I'm going to make the final review on the next Sunday, but I will keep an eye on this PR during the week.

vulkano/src/command_buffer/pool/sys.rs
Rua: It's not really a regression. The problem is simply that the extension was promoted to core in 1.1, so it is always available.

Eliah-Lakhin:
@Rua One last question before I proceed with merging. I noticed that `MainLayout` is no longer accessed in the shader's generated module. How the user is supposed to access it after this migration?

Rua:
Huh, did my PR actually change that? Where?

Eliah-Lakhin:
@Rua hm... no, it's not. But it appears somewhere in unreleased changes. I'm looking into it

Eliah-Lakhin:
@Rua ok, I can clear say this issue is happening after https://github.com/vulkano-rs/vulkano/pull/1581 merge.

This is not a blocker for this PR, but it will be a blocker for the next release. Please take a look at this issue as you have the best expertise in this field. Basically, shader's `MainLayout` type is no longer accessed/generated.

Eliah-Lakhin:
PR merged.

Rua:
> @Rua ok, I can clear say this issue is happening after #1581 merge.

Ah, well that makes sense then. That PR moves away from representing pipeline layouts as distinct types, and uses the single `PipelineLayoutDesc` type instead. So there is no more `MainLayout` type at all being generated by vulkano-shaders, instead it generates a `PipelineLayoutDesc` value with the same information. Is there anything in Vulkano that still refers to `MainLayout`?

Eliah-Lakhin:
@Rua Well, the problem(or maybe not a problem) is that we have users who generate shaders from spirv directly(from RustGPU to be more specific). There are several layouts, and they need to distinct between them somehow.

PR #1599, opened 2021-05-29
Auto-generate extensions and features, check requirements

Rua:
This PR relies on #1593, so that should be merged before this one.

Changelog entries:
```markdown
- **Breaking** `RawInstanceExtensions` and `RawDeviceExtensions` have been removed. The `Instance` and `Device` constructors now take `&InstanceExtensions` and `&DeviceExtensions` respectively. The `loaded_extensions` function returns these reference types as well.
- **Breaking** The restrictions of each enabled extension and feature are checked when creating an instance or device.
```

```markdown
- All Vulkan extensions supported by Ash are now provided in `InstanceExtensions` and `DeviceExtensions`. This includes all but the very newest extensions; new extensions should be added whenever Ash is updated to a new version.
- Every extension is now documented with a link to its Vulkan page and information about requirements, conflicts, promotion and deprecation.
- `InstanceExtensions` and `DeviceExtensions` now have a `From` implementation that takes an iterator of `&CStr`. There is also a `From` implementation for `Vec<CString>` that performs the reverse conversion.
- All Vulkan features supported by Ash are now provided.
```

Using the [vk-parse](https://crates.io/crates/vk-parse) crate, I wrote a little utility that reads the official vk.xml file and generates Vulkano definitions for all the extensions. This means that Vulkano's extension support can now be considered complete, at least long as they don't add new ones again. :) I intend to continue developing this utility and make it auto-generate physical device features and properties as well, then release it to Vulkano when I think it's ready.

In addition, this PR includes information about requirements for extensions, and a function that checks whether those requirements are met before a device is created. This was previously missing. This could be a breaking change, but if you didn't enable the requirements then your code was already incorrect.

I made this PR a draft because I have two questions that need to be answered.
- What to do about the "raw" versions of the extensions, where you can supply a list of strings to be passed to Vulkan directly. Presumably, they were added so that a user could enable an extension that wasn't in Vulkano's list, but now that they're all in the list and easy to keep up to date, that doesn't seem needed anymore. Moreover, Vulkano never considered these types unsafe, but they are, because enabling an extension can bring additional requirements, which Vulkano can't check if it doesn't know about the extension. So I would like to remove `RawInstanceExtensions` and `RawDeviceExtensions` altogether in this PR.
- Deprecated and promoted extensions. They are still allowed to be used in Vulkan, but it's discouraged and unnecessary if the replacement is available on the physical device. Vulkano could make this a hard rule: you aren't allowed to enable a deprecated or promoted extension if there's a replacement available. But this is of course a bit restrictive and perhaps somewhat surprising to users. Alternatively, Vulkano could automatically disable the deprecated/promoted extension and enable the replacement instead. This would be friendlier to the user, but less explicit. Of course, Vulkano could do nothing, and deal with whatever the user enables.

EDIT: I have now also included auto-generated features in this PR, and included a bunch more checks for both extensions and features. Features are not currently checked against enabled extensions, but this doesn't cause any problems, such features are just silently ignored. I still want to add this, either to this PR or to a future one, but implementing it is proving tricky.

EDIT 2: I've also included the tool I wrote. I named it `vulkano-gen` and it resides in a new directory in the project workspace. It includes a readme file with some instructions on how to use it and update the `vk.xml` file.


Eliah-Lakhin:
@Rua #1593 merged, so this PR is unblocked now. Regarding your questions. For the first one, I think it's fine to remove "raw" versions. For the second, I need a little more time to think about it.

AustinJ235:
> * Deprecated and promoted extensions. They are still allowed to be used in Vulkan, but it's discouraged and unnecessary if the replacement is available on the physical device. Vulkano could make this a hard rule: you aren't allowed to enable a deprecated or promoted extension if there's a replacement available. But this is of course a bit restrictive and perhaps somewhat surprising to users. Alternatively, Vulkano could automatically disable the deprecated/promoted extension and enable the replacement instead. This would be friendlier to the user, but less explicit. Of course, Vulkano could do nothing, and deal with whatever the user enables.

Perhaps maybe there is a solution in between. Would it be possible to emit a compiler warning about using a deprecated extension, so i.e. mark a field deprecated? Is that a thing? Otherwise, mention in a runtime console message that the extension in use is deprecated/promoted and in the case of a replacement being available, suggest it?

Rua:
I'm not a fan of outputting anything to the console. I don't think that's something Vulkano should ever be doing. The `log` crate may be ok, but that adds a dependency.

I don't know if it's possible to add a deprecation notice to a single struct field. This may not work anyway if the user bypasses the field in some way, for example by using the value returned by Vulkano directly (when querying for supported extensions) without accessing the fields. But I suppose it's worth experimenting? Do you want to try it?

Rua:
This PR is now ready. The question about promotion and deprecation is still unanswered, but for now the PR takes the default approach of doing nothing.

Eliah-Lakhin (changes requested):
@Rua Excellent work! We have one merge conflict, but in general it works fine. Another thing I have noticed is that on my local machine the generator is using tabs instead of whitespaces. Is it machine-specific?

From the general implementation point of view, code generation in src is usually considered as bad practice in Rust. While I'm fine with this approach, additional crate and the need of running generator manually increases maintainability difficulties. I propose to consider to try another approach instead: use Cargo build script in the main Vulkano crate to generate this `autogen.rs` file, put it to `OUT_DIR`(which is not under git) and import it back to the crate's modules preliminary. I believe this approach would be better from maintainability point of view and should work fine too with crates.io. They support build scripts as far as I know.

See build script docs: https://doc.rust-lang.org/cargo/reference/build-scripts.html

Rua:
A build script would run each time Vulkano is built, but it only needs to be run when the Ash/vk.xml version changes. My intention was that, whoever makes a PR to update Ash when a new version is released, also updates vk.xml and then runs the tool once. Running the tool each time in a build script would add a lot of extra compile time for users, with no real benefit.

I can however make the tool output autogen.rs somewhere else. I just don't know where that would be. Any ideas?

Eliah-Lakhin:
> Running the tool each time in a build script would add a lot of extra compile time for users, with no real benefit

Well, to my observation currently the script takes up to 1ms on my machine, which is tolerable. This is not the only thing we do each time during compile time(i.e. shaderc and macro expansion are not cheap too), and Vulkano doesn't have a goal to be super-fast in build time in general, so I don't think it would be a big deal to add such relatively fast script to the build phase even if the user would run it each time.

Eliah-Lakhin:
@Rua To the `OUT_DIR`(this is an environment variable set by Cargo to something inside target). Then you can link to this file directly in some of the src's modules like this: `include!(concat!(env!("OUT_DIR"), "/hello.rs"));` You can create dedicated empty module(with this "include!" line) right in the src for this purpose. See: https://doc.rust-lang.org/cargo/reference/build-script-examples.html

Rua:
I think I need more time to figure out how this is supposed to work. Would it be ok to merge this as it is, and I can add a build script in a future PR?

Eliah-Lakhin:
@Rua ok. Please fix the merging conflict too, and I will proceed with merging. Also, please note that this improvement is a blocker for the release. Thanks!

Rua:
I managed to do it anyway. Once it "clicked" how it works it wasn't so hard anymore! :D

Eliah-Lakhin:
@Rua Merged. Thank you!

One thing that could be improved(I will left it at your decisions) is to do `println!("cargo:rerun-if-changed=PATH")` that could enforce prevent of the build script to rerun each time if `PATH` hasn't been changed. You can use vk.cml here, I think.

PR #1601, opened 2021-06-04
Goodbye vk-sys

gurchetansingh:
It's dead code at this point.  Those wishing to continue to use it can use older releases of Vulkano.

2. [ X] Please put changelog entries **in the description of this Pull Request**

        - None
        -
        - None, but deleted CHANGELOG_VK_SYS


gurchetansingh:
CC: @Eliah-Lakhin @Rua

AustinJ235:
I don't know if we should completely remove it. There may be someone out there looking to maintain these bindings or aren't ready to switch to another alternative yet. I think it be a good idea to move this to another repository or just leave it dormant in this repo. Downside of switching it to another repo would be that all the commit history wouldn't show up on that repo.

@Eliah-Lakhin we do seem to have permission to create repo's under the org.

Eliah-Lakhin:
@gurchetansingh At this point I propose to keep hosting of vk-sys under the main Vulkano github repo. In case someone is still using it, and in case we(by some reasons) will have to revert back by vk-sys. At least until we release this migration to cargo, and get some feedback from the users, I would keep it's code here.

Eliah-Lakhin:
At some point maybe we will remove it and/or migrate somewhere, but I would prefer to do everything more carefully and iteratively :)

gurchetansingh:
Ack.  I'll keep the PR up until the time is right to remove it.

Rua:
The Ash migration is now 3 months old, and I don't think there have been significant issues. I think this may be worth revisiting?

gurchetansingh:
rebased in case it's time to merge

AustinJ235:
`vk-sys` has been moved! You can now find it here: https://github.com/vulkano-rs/vk-sys

This is a fork from the last commit that touched this code, and I stripped it down.

PR #1603, opened 2021-06-07
Implement `fmt::Display` for `DisplayMode` and clarify what the return value of `DisplayMode.refresh_rate` means

i509VCB:
Implements `fmt::Display` for `DisplayMode`. An example output of this implementation would be `1920×1080px @ 60.000Hz`.

The other part of the PR is describing the unit of the return value from the `refresh_rate()` function in `DisplayMode`. The return value is the refresh rate in Hz multiplied by 1000, so it's effectively millihertz (mHz).

  - Implement `fmt::Display` for `DisplayMode`.
  - Clarify return value is in millihertz for `DisplayMode.refresh_rate()`


Rua:
Wouldn't it be easier to divide the refresh rate by 1000 and display it as a float in Hz? Millihertz is just going to confuse a lot of people...

Eliah-Lakhin:
@i509VCB Thank you for your work! The code looks good, but I agree with @Rua 's comment. For Display-ing perhaps it would be better to improve an output format to be more convenient for the end-users.

i509VCB:
Yeah I can change the result to be in hertz

i509VCB:
Changed display to print in hertz. I also added a small test. The keen viewer will notice the huge block of text explaining weird driver stuff and how it influences what the test will print.

Eliah-Lakhin:
@i509VCB Thank you for the fixes. The added test is failing. Moreover, it doesn't test anything by design. Please remove the test, and I will proceed with merging.

i509VCB:
@Eliah-Lakhin done

Eliah-Lakhin:
@i509VCB Merged. Thank you! :+1:

PR #1605, opened 2021-06-08
Add auto-generated device properties

Rua:
This follows up on #1599, adding auto-generated properties as well. This is more of a breaking change, because the method of retrieving properties has changed to accommodate the new properties.

Changelog:
```markdown
- **Breaking** Physical device properties are now retrieved through a new `properties` method, returning a `&Properties` reference.
  - To allow for the possibility that not every property is known by the physical device, fields in `Properties` are wrapped by an `Option`.
  - The previous methods for retrieving properties, `name`, `ty`, `limits`, `driver_version`, `pci_device_id`, `pci_vendor_id`, `uuid` and `extended_properties`, have been removed.
  - The `api_version` method remains, but there is now a semantic difference between it and the version reported by `properties`: The raw property gives the maximum supported version that the driver itself reports, while the method on `PhysicalDevice` returns the version a logical device would support if it were created from this physical device (that is, restricted by the instance's `max_api_version`).
```


Eliah-Lakhin:
@Rua Merged. Thank you!

PR #1606, opened 2021-06-08
Add support for semaphore file descriptor export

fayalalebrun:
Thanks to the `EXT_external_objects` OpenGL extension, there now exists a possibility to interoperate between OpenGL and Vulkan. Specifically, this extension allows OpenGL to use Vulkan textures and semaphores.

This PR takes a first step towards this, by allowing Vulkano to export its semaphores to Linux file descriptors, in the same way already possible with memory. Next, I will be working on allowing Vulkano to export its texture memory.

An example of semaphores being shared with Glium can be found here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7a9fa8702ca3f3b22813101daf861416

The related PR to add import functionality to Glium can be found here: https://github.com/glium/glium/pull/1939

        - Add creation of semaphores with exportable Linux file descriptor on
        - Add method to export file descriptor corresponding to semaphore


vulkano/src/sync/semaphore.rs
```diff
@@ -51,25 +55,47 @@ where
...
-        Semaphore::alloc_impl(device, false)
+        Semaphore::alloc_impl(device, false, false)
     }
 
-    fn alloc_impl(device: D, must_put_in_pool: bool) -> Result<Semaphore<D>, OomError> {
+    /// Same as `alloc`, but allows exportable opaque file descriptor on Linux
+    #[inline]
+    #[cfg(target_os = "linux")]
+    pub fn alloc_with_exportable_fd(device: D) -> Result<Semaphore<D>, OomError> {
+        Semaphore::alloc_impl(device, false, true)
```
Eliah-Lakhin, line 73: The third parameter seem to be ignored in actual implementation.

vulkano/src/sync/semaphore.rs
```diff
@@ -51,25 +55,47 @@ where
...
     pub fn alloc(device: D) -> Result<Semaphore<D>, OomError> {
-        Semaphore::alloc_impl(device, false)
+        Semaphore::alloc_impl(device, false, false)
     }
 
-    fn alloc_impl(device: D, must_put_in_pool: bool) -> Result<Semaphore<D>, OomError> {
+    /// Same as `alloc`, but allows exportable opaque file descriptor on Linux
+    #[inline]
+    #[cfg(target_os = "linux")]
+    pub fn alloc_with_exportable_fd(device: D) -> Result<Semaphore<D>, OomError> {
```
Eliah-Lakhin, line 72: What was the reason to dedicate this function, and why it cannot be propagated to other platforms?

Eliah-Lakhin (changes requested):
@fayalalebrun Thank you for your contribution. I left a couple of questions to be addressed.

gurchetansingh:
Why not use a builder?  Checkout the `DeviceMemoryBuilder`.

I think Vulkano should adopt a more standardized interface, so coding it up would be a non-brainer to anybody.  Right now, there's a bunch of styles.

Eliah-Lakhin:
> Why not use a builder?

@gurchetansingh I believe we could as long as it resolves the issues mentioned above.

Also, please notice that Linux test is failing. It requires an additional check in the test that the extension is available on particular device during the test running.

vulkano/src/sync/semaphore.rs
fayalalebrun: I did this because most likely other handle types will have to be handled completely differently. This is a POSIX file descriptor so it won't work on Windows. You'd need an NT handle or one of the D3D specific types, and for the NT handle you also need to provide parameters, if I'm not mistaken.

fayalalebrun:
@gurchetansingh I've added a `SemaphoreBuilder` and modeled it as close as possible to `DeviceMemoryBuilder`.

As part of this, I added some specific errors for Semaphores, but I had to convert them to the error type used in the `Swapchain` type and I'm not sure if I handled this the right way. I wrap `SemaphoreError` in `AcquisitionError` but they are both already wrapping `OomError`.

vulkano/src/sync/semaphore/semaphore.rs
```diff
@@ -0,0 +1,345 @@
...
+use std::fmt;
+#[cfg(target_os = "linux")]
+use std::fs::File;
+use std::mem::MaybeUninit;
+#[cfg(target_os = "linux")]
+use std::os::unix::io::FromRawFd;
+use std::ptr;
+use std::sync::Arc;
+
+use super::ExternalSemaphoreHandleType;
```
Eliah-Lakhin, line 26: Better use full qualifier with `crate::...`

vulkano/src/sync/semaphore/mod.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2021 The vulkano developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or https://opensource.org/licenses/MIT>,
+// at your option. All files in the project carrying such
+// notice may not be copied, modified, or distributed except
+// according to those terms.
+
+pub use self::external_semaphore_handle_type::ExternalSemaphoreHandleType;
```
Eliah-Lakhin, line 10: Missing referenced type.

Eliah-Lakhin (changes requested):
@fayalalebrun Thank you for the fixes. The code looks fine at a first brief glance. But there seem to be missing file. I can't proceed with testing(and auto tests fail too).

fayalalebrun:
@Eliah-Lakhin 
There was a missing file, but I also broke one of the semaphore tests. I did not bother to check anything other than the semaphore export test, which was my mistake. With the latest commit, all tests should pass now.

vulkano/src/sync/semaphore/semaphore.rs
```diff
@@ -0,0 +1,345 @@
...
+    export_info: Option<ash::vk::ExportSemaphoreCreateInfo>,
+    create: ash::vk::SemaphoreCreateInfo,
+    must_put_in_pool: bool,
+}
+
+impl<D> SemaphoreBuilder<D>
+where
+    D: SafeDeref<Target = Device>,
+{
+    pub fn new(device: D, must_put_in_pool: bool) -> Self {
```
Eliah-Lakhin, line 58: @fayalalebrun As it was turned into a Builder, perhaps would be better to also turn `must_put_in_pool` parameter into a builder's function? Like `pub fn set_must_put_in_pool(mut self, flag: bool) -> Self` or something.

fayalalebrun: Good idea. I've added it.

vulkano/src/sync/semaphore/semaphore.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        Self {
+            device,
+            export_info: None,
+            create,
+            must_put_in_pool: false,
+        }
+    }
+    /// Configures the semaphore to be added to the semaphore pool once it is destroyed.
+    pub(crate) fn in_pool(mut self) -> Self {
```
Eliah-Lakhin, line 69: > pub(crate)

Is is hidden in public API by design?

fayalalebrun: I believe so. The way semaphores behaved before there was no way to let users allocate semaphores which are meant for the pool. They were only created when needed. That's why I set it to only be available in the crate. Users can get semaphores from the pool using the `from_pool` method.

Eliah-Lakhin (approved):

Eliah-Lakhin:
@fayalalebrun ok, all tests passed, I didn't detect any issues in my local tests too and the code looks fine. PR is merged :+1: Thank you very much for this improvement and for all the fixes!

vulkano/src/sync/semaphore/semaphore.rs
```diff
@@ -0,0 +1,355 @@
...
+    /// - Panics if the export info has already been set.
+    pub fn export_info(mut self, handle_types: ExternalSemaphoreHandleType) -> Self {
+        assert!(self.export_info.is_none());
+        let export_info = ash::vk::ExportSemaphoreCreateInfo {
+            handle_types: handle_types.into(),
+            ..Default::default()
+        };
+
+        self.export_info = Some(export_info);
+        self.create.p_next = unsafe { std::mem::transmute(&export_info) };
```
pwaller, line 87: This looks suspiciously similar to this other issue I raised:
https://github.com/vulkano-rs/vulkano/pull/1734#discussion_r771974391

Is the export_info referenced here living long enough? I'm unclear on that. I think this might need to refer to the `self.export_info.as_ref` or something similar.

fayalalebrun: @pwaller Yes, I think there are two errors here. The first is that the struct the reference points to falls out of scope, since in this case the original is used instead of `self.export_info` (A copy is placed into the self struct). So the struct is destroyed when the scope of the function ends.

The second error is that a reference to the option is used instead of a reference to the underlying object. So `self.export_info.as_ref().unwrap()` can be used (We can unwrap here because we know the Option will always contain something at this point).

Could you make Github issues for these bugs? That way they are easy to keep track of. As comments to merged pull requests they might be easily forgotten.

PR #1607, opened 2021-06-13
Fixed chaining of ffi feature structures.

funmaker:
Without `as_mut()`, `unwrap()`s simply performed copy leaving original structures unchained.


Rua:
I caught this bug already, and fixed it as part of #1605 for the new `PropertiesFfi`, just forgot to add it to `FeaturesFfi`.

Eliah-Lakhin:
@funmaker Thank you for the fix. So, is it still relevant?

funmaker:
@Eliah-Lakhin no, the issue has been fixed along with #1605. Thanks.

PR #1608, opened 2021-06-14
Allow use of uninitialized images when being written to

AustinJ235:
This pull request allows for uninitialized images to be used where they are be written to. This will resolve access errors of the image not being initialized. This introduces an extra context for images resources coming from `SyncCommandBufferBuilder`. If the image is used where it is being written to it will mark them image safe to be used uninitialized. `ImageAccess` `try_gpu_lock` method has changed slightly to allow this extra flag to get passed.

        - **Breaking** `ImageAccess` trait method `try_gpu_lock()` now has an additional argument to allow locking the image in an uninitialized state.`
        - Improve `ImageLayout` checks to prevent `AccessError::ImageNotInitialized` from occurring where the image is safe to use uninitialized.

- This does not provide protection where the image is partial written to. It is assumed that if you are writing to part of the image that it is initialized. Implementing safety around this would add a lot of complexity for little additional safety.
- I op'd for creating the struct `ImageUninitializedSafe` instead of a simple boolean as it makes code related to command_buffers easier to read. This shouldn't effect performance I wouldn't think.
- `msaa-renderpass` example still does not launch with this change. Thought that issue may be related since at that time, I was having issues in Basalt around multisampling returning `AccessError::ImageNotInitialized`.

This fixes issues I was having in Basalt with having to manual transition images myself. I don't know if this issue is related to https://github.com/vulkano-rs/vulkano/pull/1588 and would be curious how these two work together if they provide a more robust solution.

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -87,7 +87,7 @@ use std::slice;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
-use crate::command_buffer::ImageResourceContext;
+use crate::command_buffer::ImageUninitializedSafe;
 
 /// Note that command buffers allocated from the default command pool (`Arc<StandardCommandPool>`)
 /// don't implement the `Send` and `Sync` traits. If you use this pool, then the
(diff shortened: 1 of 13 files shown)


Eliah-Lakhin:
@AustinJ235 Merged. Thank you very much for the fix!

I didn't detect any issues with this implementation in Vulkano examples and in my own project tests both, but in fact I never faced described problem on my GPU too. Hopefully it will at least resolve current Basalt blocker.

PR #1614, opened 2021-06-20
Better feature and device selection in examples

### LINKED ISSUE #1611: Selecting which GPU to run on?
*   Version of vulkano: 629fff826ea5f2cef3867e83ee84e99ef91a6680 - master, Mon Jun 14 2021
*   OS: Pop_OS! 20.10
*   GPU (the selected PhysicalDevice): 3090 RTX, GTX 1650, intel integrated
*   GPU Driver: 460.27.04, CUDA 11.2

# Issue

I've tried to run the `cargo test --all` and then `cargo run --bin triangle`, but the thing runs on intel integrated graphics. Is there a way for it to (pre) select running on a desired graphics card, like in my case I'd rather run it on 3090 RTX? It's an external GPU enclosure, not that it should matter..

### LINKED ISSUE #1621: All examples: feature conflict on Device::new with NVidia 2080 SUPER, driver 632.54, Win10
Hello! I am in the process of updating to 0.24 and noticed that, running *any* of the examples, I get the following panic on device creation:

```
Using device: GeForce RTX 2080 SUPER (type: DiscreteGpu)                                                                          
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: FeatureRestrictionNotMet(FeatureRestrictionError { feature
: "attachment_fragment_shading_rate", restriction: ConflictsFeature("shading_rate_image") })', examples\src\bin\triangle.rs:139:6 
```

Windows 10
GTX 2080 SUPER running the latest Studio Driver

Is this a know issue? Thanks...

A=

Rua:
Instead of blindly enabling all features, it now enables none, unless the particular example requires it. Examples also have much more elaborate device selection, based on the extensions/features needed, and gives priority to certain device types over others.

This fixes #1611, both the original issue and the crash in the second post, and fixes #1621.

Changelog:
```markdown
- Examples now enable only the features they need instead of all of them.
- Examples have much smarter device selection logic. In the triangle example this is supplied with comments.
```

I noticed that all the examples copy the instance and device startup code pretty much verbatim. For maintainability, I think it would be good if this common code was moved into a `common.rs` module, which the examples can then import. This could be something for a future PR.

### DIFF
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -17,7 +17,7 @@ use std::sync::Arc;
 use vulkano::buffer::{BufferUsage, CpuAccessibleBuffer};
 use vulkano::command_buffer::{AutoCommandBufferBuilder, CommandBufferUsage};
 use vulkano::descriptor::descriptor_set::PersistentDescriptorSet;
-use vulkano::device::{Device, DeviceExtensions};
+use vulkano::device::{Device, DeviceExtensions, Features};
 use vulkano::instance::{Instance, InstanceExtensions, PhysicalDevice};
 use vulkano::pipeline::ComputePipeline;
 use vulkano::pipeline::ComputePipelineAbstract;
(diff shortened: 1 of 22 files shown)


Eliah-Lakhin:
@Rua Merged. Thank you!

PR #1615, opened 2021-06-20
Handle Result from Vulkan calls in device_memory.rs

Rua:
This fixes two compiler warnings about an unused `Result` in `device_memory.rs`. In both cases, I decided to simply `unwrap` the error. One of the cases was in a `drop` implementation, so there was really no way to handle an error there except by unwrapping. The other case is in a normal function which doesn't currently return `Result`, so I unwrapped there too so that no breaking changes would be introduced. This case could be changed in the future if someone wants, there is a TODO message.

### DIFF
--- a/vulkano/src/memory/device_memory.rs
+++ b/vulkano/src/memory/device_memory.rs
@@ -747,12 +747,13 @@ impl MappedDeviceMemory {
                 ..Default::default()
             };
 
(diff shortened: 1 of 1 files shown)


Eliah-Lakhin:
@Rua Merged. Thank you very much! It was very annoying compilation warning I want to fix long time ago :)

PR #1617, opened 2021-06-21
Move instance/physical_device.rs to device/physical.rs, rewrite and refactor some parts of it

Rua:
Changelog:
```markdown
- **Breaking** Refactoring of `PhysicalDevice` and related types.
  - The module `instance::physical_device` is moved to `device::physical`. Items are exported directly from there.
  - Functions on `PhysicalDevice` that previously returned a custom iterator type now return `impl ExactSizeIterator`.
  - `PhysicalDevice` now has new methods `supported_extensions` and `required_extensions`. The equivalent constructors on `DeviceExtensions` are deprecated, but still present.
- **Breaking** Renamed `Device::loaded_extensions` to `enabled_extensions`, to match the terminology used for `enabled_features` as well as the Vulkan standard.
- **Breaking** Renamed `Instance::loaded_extensions` and `loaded_layers` to `enabled_extensions` and `enabled_layers` respectively.
```

The biggest external change here is the renaming of the physical devices module. I believe the reason for the original placement is that a physical device is "owned" by the instance, and exists before a logical `Device` is created. However, I think it is more intuitive to treat it as a precursor of a `Device`. It clearly represents a specific device and not a generic Vulkan library instance; it has device extensions, device features, device properties, and everything else we associate with devices.

I decided to include methods to retrieve extensions to `PhysicalDevice`, because this is more consistent with the existing methods that report supported features, properties etc. I also renamed the functions for extensions and layers on `Device` and `Instance` for consistency, although this is separate from the `PhysicalDevice` changes.


Eliah-Lakhin:
@Rua Can you fix merging conflicts with `examples`, please?

Rua:
Done.

Eliah-Lakhin:
@Rua Tested and merged. Thank you.

PR #1623, opened 2021-06-25
Include image format on descriptor, if provided by shader

Rua:
Changelog:
```markdown
- Vulkano-shaders now provides the image format for descriptors, if the shader requires a specific format.
```

This is a small improvement that should make Vulkano generate more correct code. The logic for handling descriptor format restrictions already existed on Vulkano's side, but Vulkano-shaders did not actually ever provide the format, it was always `None`. Now the SPIR-V image format is correctly converted to a Vulkano format and included in the descriptor.


Eliah-Lakhin:

PR #1624, opened 2021-06-26
Convert Vulkano-shaders to use spirv_headers crate

Rua:
Changelog
```markdown
- Vulkano-shaders now uses the `spirv_headers` crate for some of its types.
```

This adds an external crate to Vulkano-shaders, to provide types related to SPIR-V. The motivation is more or less the same as with abandoning vk-sys: this crate is better maintained and kept up to date. In fact, it's maintained by many of the same people that work on Ash.

Spirv_headers is a simple crate with just definitions, but it's part of a larger project called Rspirv. Rspirv has a full SPIR-V parser that Vulkano can use, except that it is currently lacking a few features that would make migrating to it difficult, which I brought up in https://github.com/gfx-rs/rspirv/issues/201. Vulkano-shaders currently uses this to retrieve types from shader code; I don't see an easy way to do this with Rspirv yet. But a future migration can be considered.


Eliah-Lakhin:
@Rua Still testing...

As a maybe future idea we can also try to migrate from Shaderc to Naga(or something similar, but is Rust based). I'm not sure if Naga is well stabilized at this moment, but we could try to do the migration in safe way by implementing it as a temporary optional Vulkano feature(to let the user switch between implementations).

Eliah-Lakhin:
ok, the PR merged. :+1:

PR #1629, opened 2021-07-03
Descriptor set/pipeline layout refactoring

Rua:
Changelog:
```markdown
- **Breaking** `DescriptorSetDesc` is converted from a trait to a struct. `DescriptorSetLayout` can take this struct in its constructor, as alternative to the iterator which was already possible previously.
- **Breaking** `PipelineLayout` is now built from a list of descriptor set layouts and push constant ranges, like in plain Vulkan. It no longer constructs its own descriptor set layouts internally.
- **Breaking** `PipelineLayoutDescPcRange` is renamed to `PipelineLayoutPcRange`.
- **Breaking** `PipelineLayoutDesc` is no longer needed and is removed. Its functionality has been moved to `DescriptorSetDesc` and `PipelineLayoutPcRange`. This also applies to `ComputeEntryPoint`, `GraphicsEntryPoint` and `EntryPointAbstract`.
- **Breaking** `PipelineLayout` now has a `descriptor_set_layouts` method which returns a slice of all the layouts, replacing the previous `descriptor_set_layout` which only returned one at a time.
```

The primary purpose of this change is bringing Vulkano closer to Vulkan, by constructing pipeline layouts from descriptor set layouts. This is how Vulkan users will expect it to work, so it's less surprising. It also means that you can re-use descriptor set layouts for multiple pipeline layouts if you want, which was not possible previously.

The second major change is making `DescriptorSetDesc` a struct. This matches the change that was made previously with `RenderPassDesc` and `PipelineLayoutDesc` (which ironically has been made obsolete again by this PR).


Eliah-Lakhin:
@Rua Verified and merged. Thank you for your work!

I noticed that this PR has changes that require the latest Rust compiler version(with array into_iter features). I will add corresponding notes to the changelog.

PR #1634, opened 2021-07-07
Derive Clone for LayerProperties

prataprc:
Make LayerProperties as clonable, so that application structures holding on to `Vec<LayerProperties>` field can return them as owned vector to the caller instead of slice `&[LayerProperties]`


Eliah-Lakhin:
@prataprc Merged. Thank you!

PR #1638, opened 2021-07-12
Add `PipelineBindPoint`

Rua:
Changelog:
```markdown
- **Breaking** The boolean `graphics` parameter on the `bind_descriptor_sets` method of `SyncCommandBufferBuilder`, `UnsafeCommandBufferBuilder` and `StateCacher` has been replaced with a new enum `PipelineBindPoint`.
```

This change aligns better with the Vulkan spec, where there are more than just two possible bind points. Vulkano still supports only compute and graphics pipelines, but this change makes it easier to extend this in the future. It also makes the code clearer; `true` for graphics and `false` for compute wasn't very obvious for anyone looking at the code.


Eliah-Lakhin:
@Rua Looks good. PR verified and merged. Thank you! :+1:

Issue #1560, opened 2021-04-18
Explicit panic when submitting command buffer


dobrakmato:
# Template

*   Version of vulkano: 0.21.0
*   OS: Windows
*   GPU (the selected PhysicalDevice): GTX 1060

# Issue

I am having a strange issue with triggering `panic!()` call on this line https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/command_buffer/synced/commands.rs#L2683. 

> thread 'main' panicked at 'explicit panic', vulkano\src\command_buffer\synced\commands.rs:2272:25

I do not understand the vulkano internals very well, but I believe this part of code is supposed to track usage of images in recorded buffer and verify that they are used correctly (with respect to memory).

I also cannot reproduce the bug consistently in every run of the application, which may suggest that this is some kind of race condition bug. 

From what I gathered using the debugger the `image` function is usually called with low numbers such as 0, 1 or 2. But in sometimes it happens to be called with unusually high number (eg. 5), which triggers the `panic` call.

I am able to trigger this by alt-tabing from and into the application's window, but this is also not consistent. I am able to trigger it more consistently by resizing the application's window, however it happens without having to resize the window (so I think it is not related to how I handle the resize). 

As I am currently unable to find the cause of the issue I cannot provide MVP, so this is more of a question what would be the best way to debug this?

My render pipeline consists of two render passes. First is deferred render to HDR buffer (geometry pass and lighting pass) followed by skybox pass and another pass which resolves the HDR buffer to LDR buffer with tone mapping. 

```rust
  let render_pass = Arc::new(
            vulkano::ordered_passes_renderpass!(
                device.clone(),
                attachments: {
                    gbuffer1: {
                        load: Clear,
                        store: Store,
                        format: Format::A2B10G10R10UnormPack32,
                        samples: 1,
                    },
                    gbuffer2: {
                        load: Clear,
                        store: Store,
                        format: Format::R8G8B8A8Unorm,
                        samples: 1,
                    },
                    gbuffer3: {
                        load: Clear,
                        store: Store,
                        format: Format::R8G8B8A8Unorm,
                        samples: 1,
                    },
                    depth: {
                        load: Clear,
                        store: DontCare,
                        format: Format::D16Unorm,
                        samples: 1,
                    },
                    hdr: {
                        load: Clear,
                        store: DontCare,
                        format: Format::B10G11R11UfloatPack32,
                        samples: 1,
                    },
                    ldr: {
[...]

PR #1639, opened 2021-07-16
Change how resource conflicts are handled in `SyncCommandBufferBuilder`

Rua:
Changelog:
```markdown
- **Breaking** The `conflicts_buffer` and `conflicts_image` methods on the `BufferAccess` and `ImageAccess` traits are removed.
```
```
- Reworked some of the internals of `SyncCommandBufferBuilder`, so that a mutex is no longer needed and resource conflict detection is simplified.
```

Most of this PR involves internal changes to how `SyncCommandBufferBuilder` works. The old code was rather complex since it relied on every `BuilderKey` having access to the resources in each command, which necessitated a `Mutex`. I've introduced a new private `ResourceKey` enum that contains all the information that would otherwise be retrieved from commands directly in `BuilderKey`. This removes the need for shared access, and therefore the `Mutex`.

It also means that every buffer and image type is checked for conflicts using the same logic, instead of using different logic for each implementation of the `BufferAccess` and `ImageAccess` traits. This was actually a problem because it meant the check was not symmetric: `a.conflicts_image(b)` could give a different result than `b.conflicts_image(a)` if `a` and `b` were different concrete types with different method implementations. These methods are now removed in favour of a standard `Eq` implementation on the new `ResourceKey`.

I renamed base.rs to builder.rs and moved anything not builder-related to mod.rs. I did this because after working with this file a lot, I found it was too large and stuff was scattered in various locations within the file. This change doesn't affect users, it's just internal.

### DIFF
--- a/vulkano/src/command_buffer/synced/builder.rs
+++ b/vulkano/src/command_buffer/synced/builder.rs
@@ -7,9 +7,13 @@
 // notice may not be copied, modified, or distributed except
 // according to those terms.
 
+use super::FinalCommand;
+use super::ResourceFinalState;
+use super::ResourceKey;
+use super::ResourceLocation;
+use super::SyncCommandBuffer;
 use crate::buffer::BufferAccess;
 use crate::command_buffer::pool::UnsafeCommandPoolAlloc;
-use crate::command_buffer::sys::UnsafeCommandBuffer;
 use crate::command_buffer::sys::UnsafeCommandBufferBuilder;
 use crate::command_buffer::sys::UnsafeCommandBufferBuilderPipelineBarrier;
 use crate::command_buffer::CommandBufferExecError;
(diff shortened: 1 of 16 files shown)


vulkano/src/command_buffer/synced/builder.rs
```diff
@@ -0,0 +1,745 @@
...
+}
+
+// State of a resource during the building of the command buffer.
+#[derive(Debug, Clone)]
+struct ResourceState {
+    // Indices of the commands that contain the resource.
+    command_ids: Vec<usize>,
+
+    // Index of the resource within the first command in `command_ids`.
+    resource_index: usize,
```
nxfs, line 727: Just had a brief look at this, I think we may want this to be a vec of ResourceLocations rather than all sharing the same resource index, otherwise we could run into the same problem as described in https://github.com/vulkano-rs/vulkano/issues/1560

Rua: That's what your PR solves right?

nxfs: Yes

Eliah-Lakhin:
@Rua PR merged. Thank you! Looks like it also fixed one rare race-condition bug in my system related to resources locking that I have difficulties to catch :)

PR #1640, opened 2021-07-20
Store resource location in BuilderKey/CbKey (#1560)

nxfs:
1. [ ] Describe in common words what is the purpose of this change, related

        - `Entry 1.`
        - `Entry 2.`...

### DIFF
--- a/vulkano/src/command_buffer/synced/base.rs
+++ b/vulkano/src/command_buffer/synced/base.rs
@@ -286,12 +286,9 @@ struct ResourceLocation {
 struct BuilderKey {
     // Same `Arc` as in the `SyncCommandBufferBuilder`.
     commands: Arc<Mutex<Commands>>,
-    // Index of the command that holds the resource within `commands`.
-    command_ids: RefCell<Vec<usize>>,
+    resource_locations: RefCell<Vec<ResourceLocation>>,
     // Type of the resource.
     resource_ty: KeyTy,
-    // Index of the resource within the command.
-    resource_index: usize,
 }
 
 impl BuilderKey {
@@ -303,9 +300,8 @@ impl BuilderKey {
     ) -> CbKey<'static> {
         CbKey::Command {
             commands: final_commands,
-            command_ids: self.command_ids.borrow().clone(),
+            resource_locations: self.resource_locations.borrow().clone(),
             resource_ty: self.resource_ty,
-            resource_index: self.resource_index,
         }
     }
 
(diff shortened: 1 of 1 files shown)


Rua:
This PR will conflict with #1639, so you will probably want to hold off on it until it's merged.

nxfs:
Ack, thanks. Just submitting this to see if I am on the right track. I noted the possible conflict with #1639 and I have commented on that PR.

Eliah-Lakhin:
@nxfs Thank you for your work! Please rebase your changes on top of #1639 merged just recently, and let me know when it's ready for the final review.

nxfs:
I've spent some time looking at the code and I think that #1639 makes the problem much more begin. I am not 100% sure.
I've been unable to repro the issue with the new code, so far. I have been unable to test my project with #1639 because it has other dependencies that are affected by breaking changes in the dev branch so I will wait until 0.25 is released and dependencies are upgraded to test again there.

The original occurrences of the issue in #1560 where due to comparing CbKey/BuilderKey. Since we don't do that anymore, it eliminates the problem there. The minimal repro in https://github.com/funmaker/vulkano-issue-1560 relied on that so it's not relevant anymore. Moreover we remove the conflicts_* function which could potentially panic if the resource index is incorrect.

As far as I can tell there's only one place where we use a stored resource index that might be invalid, it's when a command collides in a way that is impossible to resolve, leadind to a `SyncCommandBufferBuilderError::Conflict` 

In that case we generate an error using the index stored in the entry, which might not be the correct one for the conflicting the command
```
  297                                             return Err(SyncCommandBufferBuilderError::Conflict {
  298                                                 command1_name: cmd1.name(),
  299                                                 command1_param: match resource_ty {
  300                                                     KeyTy::Buffer => {
  301                                                         cmd1.buffer_name(entry_key_resource_index)
  302                                                     }
  303                                                     KeyTy::Image => {
  304                                                         cmd1.image_name(entry_key_resource_index)
  305                                                     }
  306                                                 },
```
I think it's possible that ```cmd1.buffer_name(entry_key_resource_index)``` or ```cmd1.image_name(entry_key_resource_index)``` panics there because the entry_key_resource_index might be incorrect. But if we reach there we have a bad user error case anyways. Not even sure if the edge case can arise at all because the collision may always arise on the first colliding command, which will have a correct index stored.

So I'd say, possibilites are:
* A: do nothing, seems reasonable to me
* B: simplify things and remove command1_param from the error message, then we don't need to store the resource index at all in the entry I think? 
* C: store resource locations instead of command ids and a single resource index to ensure that the index corresponding to the stored command is always used

Let me know what's your preferred approach, if any; I personally think A or B are good enough.

Eliah-Lakhin:
@nxfs I'm sorry for delay with the answer. The initial purpose of this PR, as far as I understand, was fixing #1560. I asked @dobrakmato if it already resolved through the recent merge of #1639. In this case we probably can go with option A. :)

Eliah-Lakhin:
ok. We received a confirmation that the issue was resolved in #1639. So, I believe this PR can be closed safely.

@nxfs Thank you again for your efforts in helping Vulkano! And I'm sorry we didn't get your PR into upstream, but I believe your code gave an idea to @Rua on how to resolve the issue in even better way. :+1:

PR #1641, opened 2021-07-20
Add `DescriptorSetWithOffsets` type

Rua:
Changelog:
```markdown
- **Breaking** Draw and dispatch calls on `AutoCommandBufferBuilder` no longer have a parameter for dynamic offsets. Instead, they are provided as part of the descriptor sets parameter. They are added to each descriptor set individually using the new `offsets` method on the `DescriptorSet` trait. `SyncCommandBufferBuilder` and `StateCacher` likewise take dynamic offsets as part of the descriptor set.
```

This makes dynamic offsets a bit easier to deal with. Not only do you no longer need to add an empty parameter for them even if there are none, but they are also specified on a per-descriptor-set basis rather than for all sets at once. As an added performance bonus, descriptor sets are no longer rebound by `StateCacher` if any of the dynamic offsets change, only the sets whose offsets were changed are now rebound.

All of this works using a new type `DescriptorSetWithOffsets`. The `DescriptorSetsCollection`, which draw and dispatch calls take as input, is now built out of items implementing `Into<DescriptorSetWithOffsets>`. A plain descriptor set also implements this trait; it just adds no offsets. This means that existing code can stay the way it is, other than having to remove the old dynamic offsets argument. You only need to do anything if you actually need to provide offsets. You can see the changes made to the examples, in particular the dynamic-buffers.rs example.

### DIFF
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -166,7 +166,7 @@ fn main() {
         // `Arc`, this only clones the `Arc` and not the whole pipeline or set (which aren't
         // cloneable anyway). In this example we would avoid cloning them since this is the last
         // time we use them, but in a real code you would probably need to clone them.
-        .dispatch([1024, 1, 1], pipeline.clone(), set.clone(), (), vec![])
+        .dispatch([1024, 1, 1], pipeline.clone(), set.clone(), ())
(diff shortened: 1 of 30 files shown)


Eliah-Lakhin:
@Rua Looks good. PR merged. Thank you for the improvement! :+1:

PR #1642, opened 2021-07-21
Fix storage buffer detection to work for all SPIR-V versions

Rua:
Changelog:
```markdown
- Fixed detection of storage buffers in SPIR-V so that it works for all SPIR-V versions.
```

Vulkano-shaders was misdetecting storage buffers in SPIR-V 1.0 shaders, which was causing validation errors. This PR fixes that so that it works for all versions.


AustinJ235:
Thanks

PR #1644, opened 2021-07-21
Replace `VertexDefinition` trait with `VertexInput` struct, remove pipeline traits

Rua:
Changelog: Replace the two existing items relating to `VertexDefinition` and `BuffersDefinition` with
```markdown
- **Breaking** The various types for vertex definitions have been replaced with the new `VertexInput` type, which can handle any number of vertex and instance buffers. Examples of how it's used can be seen in the "instancing" and "teapot" examples.
- **Breaking** The `VertexDefinition` trait has been removed. All uses of it have been replaced with `VertexInput`.
- **Breaking** The `ComputePipelineAbstract` and `GraphicsPipelineAbstract` traits are removed. All uses of them are replaced with the plain `ComputePipeline` and `GraphicsPipeline` types.
```

This PR continues on from #1619, by removing `VertexDefinition` as a trait, which allows removing the type parameter of compute and graphics pipelines, eliminating those respective traits as well. The problem with vertex/instance counts hasn't been addressed by these changes, so that remains open.


Rua:
Closing this, because I realised there are some use cases that are no longer possible with this PR. I need to rethink it first.

Issue #1635, opened 2021-07-07
Synchronisation error with bind commands


Rua:
To illustrate this problem, an example case might be best. Imagine that I am recording a command buffer, with the following sequence of commands:

1. `draw_indexed` using index buffer `foo`.
  a. This calls `bind_index_buffer` on the underlying `SyncCommandBufferBuilder`...
  b. ...followed by `draw_indexed`.
2. `copy_buffer` with `foo` as destination.
3. `draw_indexed` using `foo` again. Since the index buffer is still bound, the state cacher skips the bind and only issues a `draw_indexed` on the underlying builder.

Note that this sequence reads `foo`, writes to it, then reads from it again, so pipeline barriers are needed to avoid access conflicts. However, when `SyncCommandBufferBuilder` is inserting barriers, it treats the _bind_ as accessing the buffer, but not the _draw_. Consequently, it will not insert any barrier between steps 2 and 3, which means that when step 3 is accessing the index buffer during the draw, step 2 might still be copying data to the buffer! It will insert a barrier between steps 1a and 2, but the draw in step 1b could be on either side of the barrier (the correct location would be before it).

The underlying problem here is clearly that `SyncCommandBufferBuilder` has the wrong idea of where resources are accessed. A bind command doesn't actually access the resource, it just slots it in place where future commands can find it. It is these future commands that do the actual reading and writing of the resources, so it is these commands that barriers should be based around.

The design of `SyncCommandBufferBuilder` currently doesn't allow for the possibility that a command will access a resource from a _previous_ command. It expects that the command itself provide the resources it accesses, which makes this difficult to solve. Ultimately, it comes down to the command buffer state: binding modifies this state, which then affects future commands, but the builder doesn't track this state, so it has no way of knowing which previous binds will affect the current draw command. Some redesigning of the builder will be needed so that it tracks bindings, and commands should then be able to specify which of the current bindings they access.

EDIT: I confirmed this indeed happens. A barrier is inserted between 1b and 2, which is correct, but perhaps by accident. No barrier is inserted between 2 and 3. The Vulkan synchronisation validation catches the hazard:

```
[...]

PR #1648, opened 2021-07-27
Add `VertexInput` type

Rua:
Changelog:
Replace the line about `VertexDefinition` with:
```markdown
- **Breaking** The `VertexDefinition` trait no longer has `AttribsIter` and `BuffersIter` as associated types. It instead returns a `VertexInput`.
```
Add:
```markdown
- Added `VertexInput` type, which contains the binding and attribute descriptions.
- Added a `vertex_input` method to `GraphicsPipelineAbstract`, which returns a reference to the vertex input.
```

The main purpose of this change is to allow other code to access the vertex input description. This is needed for the work I'm doing on #1635, since draw commands need to be able to know which of the bound vertex buffers they are going to access. However, it's also a step closer to being able to eliminate the last remaining type parameter of `GraphicsPipeline`, and therefore the `GraphicsPipelineAbstract` trait.

### DIFF
--- a/vulkano/src/pipeline/graphics_pipeline/builder.rs
+++ b/vulkano/src/pipeline/graphics_pipeline/builder.rs
@@ -474,36 +474,15 @@ where
 
         // Vertex bindings.
         let (binding_descriptions, binding_divisor_descriptions, attribute_descriptions) = {
-            let bindings = self
+            let vertex_input = self
                 .vertex_input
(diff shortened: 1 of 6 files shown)


AustinJ235:
Looks good

Rua:
I think you forgot to include the replacement for the removed text in the changelog?

AustinJ235:
Indeed. I misread the changelog part "Replace the line about VertexDefinition with:" as "Remove...".

PR #1650, opened 2021-07-28
Pipeline barriers are placed according to draw/dispatch calls rather than binding calls

### LINKED ISSUE #1635: Synchronisation error with bind commands

Rua:
Changelog:
```markdown
- Fixed synchronization bug where pipeline barriers were placed based on the binding of resources instead of the draw/dispatch calls that use them.
- `SyncCommandBufferBuilder` and `UnsafeCommandBufferBuilder` accept any `IntoIterator` where they previously only accepted `Iterator`.
```

This fixes #1635. In order to do this, quite a few changes needed to be made elsewhere, some of which were in separate PRs that have already been merged. This PR contains only internal changes, nothing user facing. The following changes were made internally to `SyncCommandBufferBuilder`:
- The builder now tracks which binding state was most recently set by which command.
- `FinalCommand` is gone, so the built command buffer now uses the regular `Command`.
- Each `Command` is now stored in an `Arc` instead of a `Box`, which allows them to be referenced from multiple locations. A future PR could probably change the remaining cases where a command index is used to simply clone the `Arc`.
- Using this, draw and dispatch commands can now hold references to the bind commands that set the state they use.

Now that `SyncCommandBufferBuilder` is tracking binding state, it partially overlaps in function with `StateCacher`. This could allow a future PR to make the latter obsolete entirely.

### DIFF
--- a/vulkano/src/command_buffer/synced/builder.rs
+++ b/vulkano/src/command_buffer/synced/builder.rs
@@ -23,10 +23,12 @@ use crate::command_buffer::CommandBufferExecError;
 use crate::command_buffer::CommandBufferLevel;
 use crate::command_buffer::CommandBufferUsage;
 use crate::command_buffer::ImageUninitializedSafe;
+use crate::descriptor_set::DescriptorSet;
 use crate::device::Device;
 use crate::device::DeviceOwned;
 use crate::image::ImageAccess;
 use crate::image::ImageLayout;
+use crate::pipeline::{ComputePipelineAbstract, GraphicsPipelineAbstract, PipelineBindPoint};
 use crate::render_pass::FramebufferAbstract;
 use crate::sync::AccessFlags;
 use crate::sync::PipelineMemoryAccess;
@@ -96,6 +98,9 @@ pub struct SyncCommandBufferBuilder {
         ImageUninitializedSafe,
     )>,
 
+    // State of bindings.
+    bindings: BindingState,
+
     // `true` if the builder has been put in an inconsistent state. This happens when
     // `append_command` throws an error, because some changes to the internal state have already
     // been made at that point and can't be reverted.
@@ -161,9 +166,10 @@ impl SyncCommandBufferBuilder {
             barriers: Vec::new(),
             first_unflushed: 0,
             latest_render_pass_enter,
+            resources: FnvHashMap::default(),
             buffers: Vec::new(),
             images: Vec::new(),
-            resources: FnvHashMap::default(),
+            bindings: Default::default(),
             is_poisoned: false,
             is_secondary,
         }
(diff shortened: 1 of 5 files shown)


AustinJ235:
Do barriers need to be in between other calls such as `copy_buffer()` & `copy_image()`? I don't understand completely where barriers need to be inserted.

Rua:
Barriers need to be inserted somewhere between two commands, if one of the commands writes to the resource. The problem that this PR fixes is that previously, the bind commands would be taken as accessing the resources, when in reality bind commands don't access resources, they just insert a link to them in some internal Vulkan table. The commands that actually access the bound resources are the draw/dispatch commands, but those were actually not marked as accessing the resources in Vulkano. Hence bug #1635. This PR fixes that so that draw/dispatch commands are marked as accessing previously bound resources, while the bind commands themselves are no longer marked.

AustinJ235:
Thanks for explanation, just wanted to make sure I understood before I merged.

PR #1651, opened 2021-07-28
Check binding numbers against limit

Rua:
Changelog:
```markdown
- Added limits check for vertex buffer binding numbers.
```

This wasn't previously being checked, so this PR adds it.


AustinJ235:
Another good check. Good catch.

Issue #1649, opened 2021-07-27
some properties *are* known by every device


incertia:
as per #1605
```
  - To allow for the possibility that not every property is known by the physical device, fields in `Properties` are wrapped by an `Option`.
```
i assume this is being done via [`vkGetPhysicalDeviceProperties2`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPhysicalDeviceProperties2.html), which returns a [`VkPhysicalDeviceProperties2`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceProperties2.html), which contains a [`VkPhysicalDeviceProperties`](khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceProperties.html). from my understanding, the remaining fields, e.g. [`VkPhysicalDeviceDriverProperties`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceDriverProperties.html) are optional but the core properties should be not `Option`, especially since the call to [`vkGetPhysicalDeviceProperties2`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPhysicalDeviceProperties2.html) cannot fail.

Rua:
In principle you are right, these should not be an `Option`. But accounting for that in the `properties!` macro is tricky. Do you have any ideas?

AustinJ235:
Macros are tricky never really understood them well, but this would eliminate unwraps where they are redundant.

Rua:
Perhaps, since this code is generated by a build script anyway, we can get rid of the macros entirely and generate the final code directly from the autogen code. True Rust code can be a lot more flexible about what it generates than macros.

PR #1654, opened 2021-07-30
add some logic for known properties

incertia:
makes #1649 go brrr. i haven't updated the examples yet but would something like this be ok?

```
- **Breaking** Known physical device properties (VkPhysicalDeviceProperties, VkPhysicalDeviceLimits, VkPhysicalDeviceSparseProperties) are now no longer `Option<T>`.
  - These properties are available with every `vkGetPhysicalDeviceProperties2` call
- Add some `Default` instances to make the auto trait happy
```

### DIFF
--- a/vulkano/src/version.rs
+++ b/vulkano/src/version.rs
@@ -40,9 +40,9 @@ impl Version {
 }
 
 impl Default for Version {
-  fn default() -> Self {
-    Self::V1_0
-  }
+    fn default() -> Self {
+        Self::V1_0
+    }
 }
 
 impl From<u32> for Version {
--- a/vulkano/src/version.rs
+++ b/vulkano/src/version.rs
@@ -39,6 +39,12 @@ impl Version {
     }
 }
 
+impl Default for Version {
+  fn default() -> Self {
+    Self::V1_0
+  }
+}
+
 impl From<u32> for Version {
     #[inline]
     fn from(val: u32) -> Self {
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -44,7 +44,7 @@ fn main() {
                 .find(|&q| q.supports_compute())
                 .map(|q| (p, q))
         })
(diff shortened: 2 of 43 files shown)


Rua:
While it's funny, unfortunately GitHub doesn't understand "makes #1649 go brrr" to link the two issues. :D

incertia:
any ideas for improving the macro would be greatly appreciated. i had to parse it as `tt` because `literal` would not work and had to duplicate some code in the macros but as long as it works i guess.

Rua:
The autogen module is essentially already a macro, it generates code. So what you could do is eliminate the `properties!` macro entirely, and generate the final code entirely in autogen. That gives you much more freedom, because you are generating code with full Rust instead of the macro language.

vulkano/src/version.rs
```diff
@@ -39,6 +39,12 @@ impl Version {
     }
 }
 
+impl Default for Version {
```
Eliah-Lakhin, line 42: Please fix the indentation.

Eliah-Lakhin:
@incertia Thank you for your work! This improvement looks good to me in general. Please fix the issue with indentation I mentioned above, and if possible maybe get rid of macro use(however, I can proceed with merging anyway, the improvement is useful!)

incertia:
done

incertia:
> The autogen module is essentially already a macro, it generates code. So what you could do is eliminate the properties! macro entirely, and generate the final code entirely in autogen. That gives you much more freedom, because you are generating code with full Rust instead of the macro language.

I think this should be addressed in a separate PR. It would be quite a large change but if you think we can do it here I can look into making the changes as well.

Rua:
It's fine. If you want to go ahead and do it, you can, but it's no problem if you leave this PR as it is. I or someone else can do the conversion some other time.

incertia:
Probably good to merge then. I'll look into emitting the correct code within autogen sometime this week I hope.

Eliah-Lakhin:
@incertia Pull Request merged. Thanks again! I have a number of pending PRs to review, so the review process is taking time, but I hope to clean up the rest soon too.

Issue #1652, opened 2021-07-29
ShaderInterfaceEntry.name is Cow<'static, str>


incertia:
is there any reasoning behind this? i am trying to dynamically parse spir-v modules into entries and this seems like an unfortunate design decision.

incertia:
on that note, inside `src/pipeline/vertex/single.rs`
```rust
unsafe impl<T> VertexDefinition for SingleBufferDefinition<T>
where
    T: Vertex,
{
    type BuffersIter = OptionIntoIter<(u32, usize, InputRate)>;
    type AttribsIter = VecIntoIter<(u32, u32, AttributeInfo)>;

    fn definition(
        &self,
        interface: &ShaderInterface,
    ) -> Result<(Self::BuffersIter, Self::AttribsIter), IncompatibleVertexDefinitionError> {
        let attrib = {
            let mut attribs = Vec::with_capacity(interface.elements().len());
            for e in interface.elements() {
                let name = e.name.as_ref().unwrap(); // <--- THIS LINE HERE
```
crashes when you pass in `None` for the name

incertia:
doing even more reading, it appears as if i will need to provide my own `VertexDefinition` to get around this, so the explicit unwrap is probably ok to stay

Rua:
Is the problem that you are being required to provide a name when you don't want/have one?

incertia:
i have some runtime allocated name which is not `'static` that i would like to use

PR #1655, opened 2021-07-31
let ShaderInterfaceEntry own its name strings instead of requiring static strings

incertia:
addresses #1652 but maybe not in the most ideal way. i can attempt to add a generic lifetime to the `Cow` but this may end with me stabbing myself with a rusty plastic spoon

```
- **Breaking**: ShaderInterfaceEntry.name is now Option<String> rather than Option<Cow<'static, str>>
```

### DIFF
--- a/examples/src/bin/runtime-shader/main.rs
+++ b/examples/src/bin/runtime-shader/main.rs
@@ -19,6 +19,7 @@
 // $ glslangValidator frag.glsl -V -S frag -o frag.spv
 // Vulkano uses glslangValidator to build your shaders internally.
 
+use std::borrow::Cow;
 use std::ffi::CStr;
 use std::fs::File;
 use std::io::Read;
@@ -170,12 +171,12 @@ fn main() {
             ShaderInterfaceEntry {
                 location: 1..2,
                 format: Format::R32G32B32Sfloat,
-                name: Some("color".into()),
+                name: Some(Cow::Borrowed("color")),
             },
             ShaderInterfaceEntry {
                 location: 0..1,
                 format: Format::R32G32Sfloat,
-                name: Some("position".into()),
+                name: Some(Cow::Borrowed("position")),
             },
         ])
     };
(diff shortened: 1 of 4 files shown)


Rua:
Can't you just use `Cow::Owned` with a `String`?

incertia:
i can look into making lifetimes great again

incertia:
not sure if `'a` is the correct lifetime on the `graphics_entry_point` but it seems to work

Rua:
What is the `'a` lifetime for, though? The point of `'static` is that it doesn't have to borrow anything at all. Does it not work for you?

Eliah-Lakhin:
@incertia As it was mentioned above one can use `Cow::from` to obtain owned value as well as static reference to a static string literal. However, it does not let you hold a non-static reference. But what is the purpose? Why do you think we need this feature?

incertia:
I think it is useful when people using vulkano would like to generate their own `ShaderInterfaceEntry`s dynamically, e.g. by parsing some SPIR-V at runtime. The only way to generate a value with a static lifetime is by [declaring it as a literal somewhere within your program](https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html#reference-lifetime), which does not really work when it is read directly from a file loaded at runtime.

Rua:
But the value here is of type `Cow`, which means it can be borrowed *or* owned. If it's borrowed, then indeed it must be `'static`. But you can just as easily give it an owned `String`, which bypasses the lifetime altogether because nothing is being borrowed anymore. So if you are generating them dynamically at runtime, `Cow::Owned(String)` is what you'd use.

incertia:
Looks like I got it to work. Do you have any better options for `Option<Cow<'a, String>> -> Option<Cow<'static, str>>`? `self.name.as_ref().map(|n| Cow::Owned(n.deref().clone()))` looks really ugly.

Rua:
Where are you getting `Option<Cow<'a, String>>` from?

incertia:
turns out I just needed to refactor my code

PR #1656, opened 2021-07-31
Add support for EXTENDED_SRGB_NONLINEAR_EXT color space

VictorBulba:
I got crashes on my AMD Radeon Pro 560X because of missing color space

Change-log:
- Add support for EXTENDED_SRGB_NONLINEAR_EXT color space
- Better error message for wrong color space value


Rua:
It looks good, but "nonlinear" is written as one word in the Vulkan constant, so I think it should be one in Vulkano as well.

Eliah-Lakhin:
@VictorBulba Thank you for the improvement! Pull Request merged.

PR #1658, opened 2021-08-01
Use DeviceSize for on-device sizes, offsets etc.

Rua:
Changelog:
```markdown
- **Breaking** Values that represent a size, offset or index on the device are now represented as `DeviceSize` (a type alias of `u64`).
```

Vulkano was using `usize` for most of these, which would have made sense if the data was being stored on the host (the CPU running the Rust code). But there's no guarantee that the graphics device uses the same size type that the CPU does, and Vulkan has a special `DeviceType` alias for these cases, which is defined as `u64`. Converting this to/from `usize` is obviously a lossy process, and would have prevented users from expressing certain large values that the Vulkan API/device technically do support.

This PR puts `DeviceSize` in the places where it's used by Vulkan, and where it would otherwise make sense to use it (because it refers to on-device data). The most important effect is that buffer sizes are now a `DeviceSize`. The `DeviceType` type itself is defined by Ash, and re-exported from the Vulkano crate root.


gurchetansingh:
Looks good to me, always found the conversions to and from usize quite annoying.

Eliah-Lakhin:
@Rua Please fix the merging conflicts.

Eliah-Lakhin:

Issue #1657, opened 2021-08-01
improve push constant ranges


incertia:
as per the [vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPipelineLayoutCreateInfo.html#VUID-VkPipelineLayoutCreateInfo-pPushConstantRanges-00292), each push constant range when creating a pipeline layout may not share a stage with any other range. therefore my understanding is as follows

1. the simplest is to use a single range per stage. e.g. we define a range for vertex shaders and fragment shaders and send them separately
2. we can attempt to union these. if two ranges share the same offset and size, then we can represent it as a single range with both stage bits OR'd together. a representation of `{ [0, 8), Vertex | Fragment }, { [8, 16), Vertex }` would be wrong. `PipelineLayoutPcRange` sort of does this, but not in the right way i think. it tests if two ranges share a stage and then outputs a range with the two OR'd stages and unifies the extents. this loses the per stage granularity that the vulkan API is supposed to provide.
3. `PipelineLayout::new` checks for overlapping ranges when it should instead verify again that no two ranges have intersecting stages. e.g. `{ [0, 16), Vertex }, { [4, 12), Fragment }` should pass the checks but `{ [0, 8), Vertex | Fragment }, { [8, 16), Vertex }` should not
4. representation within `GraphicsEntryPoint` can probably be changed to `Option<PipelineLayoutPcRange>` to better represent this constraint

Rua:
So if I understand correctly, the right way to represent it would be something like `HashMap<ShaderStage, PushConstantRange>`? This would ensure that the same stage can't appear in more than one range, because `HashMap` guarantees this uniqueness.

incertia:
If `ShaderStage` represents a single stage, then yes. It should only be ok to merge the `stageFlags` when the range itself is equal.

PR #1659, opened 2021-08-03
update push constant ranges to match the spec. overlapping ranges are OK but intersecting stage flags are not

incertia:
this is my attempt at addressing #1657. we also check that a `PipelineLayout`'s push constant ranges completely encompass any range that was passed in `superset`. If the two stage sets intersect, we ensure that our range starts before and ends after the other range.


Eliah-Lakhin:
@incertia Looks good. PR merged. Thank you for the fix!

Next time please don't forget to include changelog entries to the PR description :)

PR #1660, opened 2021-08-03
improve the debug interface a bit and update the error message

incertia:
this should be quick and easy. are `const fn`s the way to go or should we prefer to use associated constants?

changelog:
```
- Added `impl BitOr` for `debug::MessageType` and `debug::MessageSeverity` and some bit field generators, also turning them into `const fn`.
```


Eliah-Lakhin:
@incertia Merged. Thank you for the improvement!

PR #1662, opened 2021-08-04
More sensible representation of stencil faces

Rua:
Changelog:
```markdown
- **Breaking** `StencilFaceFlags` is renamed to `StencilFaces`, and its variants are also shortened.
- **Breaking** `DynamicStencilValue` now contains a `front` and a `back` value.
- **Breaking** The stencil state commands of `SyncCommandBufferBuilder` now take a `StencilFaces` value and a `u32` instead of `DynamicStencilValue`.
```

The way the dynamic state for stencil values was set up was wrong, perhaps due to a misunderstanding of how things work. For each dynamic state setting (compare mask, write mask and reference), there are always two values associated: the value applied to front faces, and the value applied to back faces. These values should both be set when drawing, but Vulkan lets you set both values independently or in one go.

The way Vulkano seems to have treated this previously is as if the face specifier is simply part of the dynamic state being set, but it's not; both values should always be specified. This is why `DynamicStencilValue` is changed so that it always contains both a front and a back value. For the set commands of `SyncCommandBufferBuilder`, I stayed closer to raw Vulkan, so that you specify a new value and the face(s) for which this new value should apply.

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -50,6 +50,7 @@ use crate::image::ImageAccess;
 use crate::image::ImageAspect;
 use crate::image::ImageAspects;
 use crate::image::ImageLayout;
+use crate::pipeline::depth_stencil::StencilFaces;
 use crate::pipeline::input_assembly::Index;
 use crate::pipeline::layout::PipelineLayout;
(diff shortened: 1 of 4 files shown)


Eliah-Lakhin:
@Rua Looks good. Merged. Thank you for the fix!

PR #1664, opened 2021-08-06
Better SPIR-V Capability checks

Rua:
```markdown
- Vulkano-shaders: added extension/feature checks for more SPIR-V capabilities.
```

This PR adds a whole lot more capabilities and their requirements. I also restructured how requirements are represented internally, because the old method didn't make sense for how they actually work. Not every SPIR-V capability is covered yet, but most now are. Some require some further improvements to the code, others are just not provided by `spirv_headers` yet.

I did all this work manually, but this data is present in vk.xml, so a future PR could make it auto-generated. The autogen code is currently placed in the main Vulkano crate, so Vulkano-shaders can't use it. On the other hand, there's nothing about the requirements checks specifically that means it needs to be generated by Vulkano-shaders, it could be performed at runtime in the main crate instead.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -262,33 +262,43 @@ where
                     type_id: _,
                 } => storage_class_requirement(storage_class),
                 Instruction::Capability(cap) => capability_requirement(cap),
-                _ => DeviceRequirement::None,
+                _ => &[],
             }
         };
 
-        match dev_req {
-            DeviceRequirement::None => continue,
-            DeviceRequirement::Features(features) => {
-                for feature in features {
-                    let ident = Ident::new(feature, Span::call_site());
-                    cap_checks.push(quote! {
-                        if !device.enabled_features().#ident {
-                            panic!("Device feature {:?} required", #feature);
-                        }
-                    });
-                }
-            }
-            DeviceRequirement::Extensions(extensions) => {
-                for extension in extensions {
+        let (conditions, messages): (Vec<_>, Vec<_>) = dev_req
+            .iter()
+            .map(|req| match req {
+                DeviceRequirement::Extension(extension) => {
                     let ident = Ident::new(extension, Span::call_site());
-                    cap_checks.push(quote! {
-                        if !device.enabled_extensions().#ident {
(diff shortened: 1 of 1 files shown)


Eliah-Lakhin:
@Rua I can't tell the exact reason why it's happening, but compile time has been increased exponentially with this change. For small shaders it's unnoticible, so all Vulkano examples pass fine, but I have relatively large compute shader(a few thousands lines of code with a number of includes) that never ends to compile. If I comment out significant part of the shader's code it takes about a minute to compile. Before this change it took a second or two to compile the entire shader.

Rua:
Hmm, that's not good. I don't immediately see anything that should be taking a lot of time in the new code. Profiling it might give some clues?

Eliah-Lakhin:
@Rua A minimal reproducing example would put before this line: https://github.com/vulkano-rs/vulkano/blob/master/examples/src/bin/shader-include/standard-shaders/common.glsl#L9 just the same copy-pasted line:

```glsl
arg = 2 * multiply_by_3(multiply_by_2(arg));
```

you will notice that the build time has increased from ~2 seconds to ~23 seconds(even more if you put more).

Rua:
I'll investigate what's going on later this week.

Eliah-Lakhin:
@Rua Digged inside a little bit. Setting `dump: true` to the macro showing that it inserts too many duplicates of the similar checks:

```
        } if! std :: array :: IntoIter :: new([]).all(| x | x)
        {
            panic!
            ("One of the following must be enabled on the device: {}", "") ;
        } if! std :: array :: IntoIter :: new([]).all(| x | x)
        {
            panic!
            ("One of the following must be enabled on the device: {}", "") ;
```

which came from this line: https://github.com/vulkano-rs/vulkano/pull/1664/files#diff-431678ba4963a372b83d23262783ab87ce5b35513acd5e8e5556d73aeda67479R298

I think that `(conditions, messages)` arrays need to be cleaned up more carefully.

Rua:
Thank you for finding that! I added a check to skip the rest of the code if there's no requirements. Does it work properly now?

Eliah-Lakhin:
@Rua Now works fine. Thank you for the fix! PR merged.

PR #1670, opened 2021-08-12
Generate full code directly in autogen instead of using macros

Rua:
Changelog:
```markdown
- The full Rust code is now generated from vk.xml by autogen directly, instead of using intermediate macros.
```

This implements what I proposed in #1654. I also changed a few implementation details: the code is generated using the `quote!` macro now instead of directly writing text to the file, and it's then formatted by `rustfmt` if it's available.


incertia:
Thanks for doing this.

Eliah-Lakhin:

PR #1671, opened 2021-08-12
Replace `VertexSource` with `VertexBuffersCollection`, remove pipeline traits

Rua:
Changelog:
```markdown
- **Breaking** The `VertexSource` trait is removed, and has been replaced with the new `VertexBuffersCollection` trait, which works analogously to `DescriptorSetsCollection`. Vertex buffers can now be passed in as a tuple, just like descriptor sets.
- **Breaking** Removed the vertex definition type parameter from `GraphicsPipeline`, which is no longer needed with the change above.
- **Breaking** The `ComputePipelineAbstract` and `GraphicsPipelineAbstract` traits are no longer needed and have been removed, with their methods made available on the base `ComputePipeline` and `GraphicsPipeline` types.
```

This refactoring further simplifies the API by making the way that the vertex buffers are passed to draw calls no longer dependent on the `VertexDefinition` being used. You can now always pass the vertex buffers as a single item, a tuple of various sizes, or a vec of trait objects, regardless of the vertex definition. This in turn allows the type parameter to be removed, which then makes the traits unnecessary as well.

### DIFF
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -21,7 +21,6 @@ use vulkano::device::physical::{PhysicalDevice, PhysicalDeviceType};
 use vulkano::device::{Device, DeviceExtensions, Features};
 use vulkano::instance::{Instance, InstanceExtensions};
 use vulkano::pipeline::ComputePipeline;
-use vulkano::pipeline::ComputePipelineAbstract;
 use vulkano::sync;
 use vulkano::sync::GpuFuture;
 use vulkano::Version;
--- a/examples/src/bin/deferred/frame/ambient_lighting_system.rs
+++ b/examples/src/bin/deferred/frame/ambient_lighting_system.rs
@@ -160,7 +160,7 @@ impl AmbientLightingSystem {
                 0,
                 self.pipeline.clone(),
                 &dynamic_state,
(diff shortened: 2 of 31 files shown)


incertia:
I was just going to complain about the `VertexDefinition: VertexSource<Vec<Arc<dyn BufferAccess + Send + Sync>>>` trait bound. Thanks for removing it! Would it make more sense to use `VertexBuffersCollection<I: IntoIterator<Item = Box<dyn BufferAccess + Send + Sync>>`?

Rua:
I'm not sure what you mean, `VertexBuffersCollection` doesn't have a type parameter in my code.

incertia:
Er, would it make more sense to just make everything `IntoIterator` instead of packaging everything into a `Vec`? I took a brief look at `draw()` etc and it didn't seem like anything really required it to be `Vec`.

Rua:
What would the return type of the `into_vec` function be then? I haven't figured out how to make that work yet, but if you want to give it a go after this merges, please do! :)

incertia:
I think my plan is to completely get rid of the trait bound `V: VertexBuffersCollection` and replace it with the correct `V: IntoIterator<Item = ...>` but we'll see how good my engineering skills are.

Eliah-Lakhin:
@Rua Good job! PR merged. Thank you! :+1: 

@incertia fyi.

PR #1674, opened 2021-08-17
Support for Runtime Descriptor Arrays

AustinJ235:
~Depends on https://github.com/vulkano-rs/vulkano/pull/1680~

        - **BREAKING** `BufferAccess` now requires Send + Sync. This cascades into `TypedBufferAccess`, `BufferSlice`, `ImmutableBufferInitialization`, `ImmutableBuffer`, `DeviceLocalBuffer`, `CpuBufferPoolSubbuffer`, `CpuBufferPoolChunk`, `CpuAccessibleBuffer`, and `BufferView` now requiring their types to be `Send + Sync`.
        - **BREAKING** `DescriptorSetLayout::new()` now returns `Result<Self, DescriptorSetLayoutError>` instead of Result<Self, OomError>`
        - **BREAKING** `DescriptorCompatibilityError` additional variant `VariableCount`.
        - **BREAKING** `GraphicsPipelineCreationError` additional variant `PipelineLayoutCreationError`.
        - **BREAKING** `PipelineLayoutCreationError` additional variant `SetLayoutError`.
        - **BREAKING** `FixedSizeDescriptorSetsPool` has been replaced by `SingleLayoutDescSetPool`.
        - **BREAKING** Set builders now return `&mut Self` instead of `Self` & methods take values wrapped in an `Arc`.
        - Descriptor sets now support variable count descriptors.
            - e.g. `layout(set = 0, binding = 0) uniform sampler2D textures[];`

This pull request replaces the current descriptor set creation. `PersistentDescriptorSet` & `FixedSizeDescriptorSetsPool`. to allow for variable count arrays. e.g. `layout(set = 0, binding = 0) uniform sampler2D textures[];`


Rua:
If you have some idea of what a common length is, you can always use `SmallVec` instead of `Vec`.

AustinJ235:
Did some testing with `SmallVec` it seems to be much slower if you account for a sensible limit to what a user might have in a set. My testing included 8 possible bindings and 32 total writes. The test case my tidal app caldera with basalt has 4 bindings, with the variable one having 2 currently.

Doing some pre-calculations with the context provided having `Vec` use `with_capacity` where possible yielded the best results. I personally have nothing against using `Vec`. In my testing the latest commit improves performance about 13% over the previous. Using `SmallVec` in place of `Vec` decreased performance by 223%. It should be worthy of note that this isn't a considerable amount of time to begin with. Previous commit took on average 5.9 microseconds whereas the current commit takes 5.1 microseconds. The `SmallVec` implementation took 13.2 microseconds.

I have it so when the user starts the builder they can specify the variable array capacity as a hint. Nothing bad happens if you go over that currently, but I am thinking ahead to when I do the pool implementation. If the user provides a capacity should it be an error if it goes over? I currently have the partial bound flag enabled which would allow us to allocate the capacity for each set and therefor have a consistent size and should in theory reach the same performance as fixed size set.

In the case of the user not specifying a capacity it could start off with say 8 and if the next allocation requires more it'd increase a power of 2 say for example. This would allow the user to not have to think about capacity and still be somewhat efficient in set reuse.

Rua:
Interesting and surprising! Maybe the use of `SmallVec` in other parts of Vulkano ought to be re-evaluated at some point then!

AustinJ235:
I think at one point in time `SmallVec` may have been faster in Rust's earlier days where optimization may have not been as good, but now days the compiler is probably pretty good at keeping things on the stack. This is beyond my level of comprehension though and maybe other cpu's could be different. This is using a ryzen 5600x.

One area to look at would be `vulkano/src/descriptor_set/sys.rs`:

```rust
        let mut buffer_descriptors: SmallVec<[_; 64]> = SmallVec::new();
        let mut image_descriptors: SmallVec<[_; 64]> = SmallVec::new();
        let mut buffer_views_descriptors: SmallVec<[_; 64]> = SmallVec::new();

        let mut raw_writes: SmallVec<[_; 64]> = SmallVec::new();
        let mut raw_writes_img_infos: SmallVec<[_; 64]> = SmallVec::new();
        let mut raw_writes_buf_infos: SmallVec<[_; 64]> = SmallVec::new();
        let mut raw_writes_buf_view_infos: SmallVec<[_; 64]> = SmallVec::new();
```

For most cases that is a lot of extra allocation.

AustinJ235:
Implementation of the pool is now complete.

- Pool creation takes `runtime_array_capacity` argument used as the base capacity for the runtime array descriptor.
     - `UnsafeDescriptorPool` creation uses a layout where the `array_count` member of `DescriptorDesc` for the runtime arrray desc or variable count descriptor is set to the `runtime_array_capacity`.
- If a set is built where the `array_count` of the runtime array exceeds the `runtime_array_capacity` used to create `UnsafeDescriptorPool` then the `runtime_array_capacity` is multiplied by two until it is able to fit this new count and a new pool is created.
- If a set is built and there are no available sets from the `UnsafeDescriptorPool`, then `UnsafeDescriptorPool` is recreated where the `max_sets` is equal to double the previous.

This pool implementation is specific to a layout similar to the fixed size pool implementation. In my testing creating a set from the runtime pool instead of a persistent runtime set is about 27% faster. (4.55 microseconds versus 6.21 microseconds also with much less jitter in timing)

AustinJ235:
It may be worth reviewing if this new runtime variant should replace the existing `PersistentDescriptorSet` and `FixedSizeDescriptorSetsPool` as the performance may be on par or better. This would be breaking however but would remove some similar code, so less upkeep. The new builder style returns an `&mut` instead of `self`. Also the builder takes values contained inside of an `Arc` instead of a reference. This is more favorable for having conditional building as it removes the weird type wonkyness.

Rua:
I assumed that your new version would replace the old one eventually. Other than compatibility, what reasons can you think of for having both?

AustinJ235:
I thought originally it'd be slower than the existing implementation, but it seems it may be faster possibly. Also the time here seems pretty negligible.

In the latest commit for adding support for buffer views I had to make some breaking changes. `BufferAccess` now requires `Send + Sync`. This cascades into `TypedBufferAccess`, `BufferSlice`, `ImmutableBufferInitialization`, `ImmutableBuffer`, `DeviceLocalBuffer`, `CpuBufferPoolSubbuffer`, `CpuBufferPoolChunk`, `CpuAccessibleBuffer`, and `BufferView` now requiring their types to be `Send + Sync`. This change was required for how this implementation stores resources. I think publicly for people using vulkano provided types this isn't that breaking, but for someone implementing their own this could potentially be fairly breaking.

AustinJ235:
I think things are mostly finalized aside from removal of the old implementations. I am keeping these here for now as I plan to do some performance regression testing tomorrow.

@Rua have any thoughts possible improvements or anything that is out of spec? I think everything should good spec wise.

Rua:
I haven't looked in too much detail, but what's in the changelog seems ok. I notice a bit of inconsistency in the pool module, with the type named "DescriptorSetPool" but many docs and comments calling it "DescriptorSetsPool" with an extra s. Also, it may be worth noting in the documentation of either the module or the type, that a DescriptorSetPool is a convenience wrapper provided by Vulkano, and that it shouldn't be confused with a DescriptorPool, which is a raw Vulkan object type with a very different function.

AustinJ235:
Only found a single instance of `DescriptorSetPool` with an extra `s`. That is a good point about the name though. For someone coming from vulkan that may be confusing. I added some further clarification to the doc for `DescriptorSetPool`.

```
/// `DescriptorSetPool` is a convenience wrapper provided by Vulkano not to be confused with
/// `VkDescriptorPool`. Its function is to provide access to pool(s) to allocate `DescriptorSet`'s
/// from and optimizes for a specific layout. For a more general purpose pool see `descriptor_set::pool::StdDescriptorPool`.
```

I have thought about another name, but I think it'd just make it more confusing. Before it was `FixedSizeDescriptorSetPool` which in itself was a bit deceiving. There was nothing really fixed about it other than the layout being fixed.

AustinJ235:
There is some performance regression in the current state:

```
Old PersistentDescriptorSet                4.022 microseconds
FixedSizeDescriptorSetPool                 0.933 microseconds
New PersistentDescriptorSet Fixed Array    5.700 microseconds
New PersistentDescriptorSet Runtime Array  6.389 microseconds
DescriptorSetPool Fixed Array              4.971 microseconds
DescriptorSetPool Runtime Array            5.461 microseconds
```

**Test Layout:** 2 input attachments, array of 4 sampled images

AustinJ235:
latest commit avoids recreation of DescriptorSetLayout. This doesn't effect `PersistentDescriptorSet`, but does however improve performance of `DescriptorSetPool`. This still is 3x the time of the old `FixedSizeDescriptorSetPool`.

```
DescriptorSetPool Fixed Array    3.090 microseconds  38% improved from previous
DescriptorSetPool Runtime Array	 2.933 microseconds  46% improved from previous
```

AustinJ235:
To remove possible confusion and some name spacing issues with the current pool mod, thinking it'd be better to rename `DescriptorSetPool`. Possible names I could think of would be:

- `FixedLayoutDescriptorSetPool`
    - May suggest the layout is truly fixed which isn't the case for runtime arrays. Those generate different layouts depending on capacity.
- `LayoutDependentDescriptorSetPool`
    - Kind of weird if I am going to be honest. Would suggest that the Pool depends on the layout which I guess is sort of true.
- `LayoutOptimizedDescriptorSetPool`
    - May suggest other layouts are supported, which is false.
- `LayoutDescriptorSetPool`
    - Pretty generic yet, gets the point across that this pool implementation only uses an single layout.

Rua:
`SingleLayoutDescriptorSetPool` maybe?

AustinJ235:
That seems to be better than any of my suggestions. I guess that is kind of a given when I say "LayoutDescriptorSetPool ... only uses an single layout"

AustinJ235:
I think things are mostly finalized here. The performance I think is decent unless you are wanting frame rates in excess of 100k which I think you'll become bound in other ways. I have yet to check the examples those might need some changes.

Eliah-Lakhin:
@AustinJ235 I will take a closer look today evening or tomorrow(bringing it into my project and do some tests too), but I don't want to block the PR merging too, so If you want to proceed please go ahead. Just a couple of briefly things that probably you noticed too:

1. Tests are failing. It looks like there are actual bugs in the code.
2. Is returning `&mut` instead of owned really necessary? Not too critical, but having owned values IMO was more convenient. Again, this is not a critical part, but anyway.
3. There is no example of descriptor arrays(did I miss something?). Would be nice to have one to demonstrate this new feature and show to the users on how to utilize it.
4. To the log entries I would also add a general description of the changes purpose("support of runtime descriptor arrays").

AustinJ235:
> 1. Tests are failing. It looks like there are actual bugs in the code.

Seems a merge brought in another test that used the old descriptor set.

> 2. Is returning `&mut` instead of owned really necessary? Not too critical, but having owned values IMO was more convenient. Again, this is not a critical part, but anyway.

This mimics how we build command buffers now. It makes building things conditionally easier i.e.

```
let mut builder = ...

for image in my_images {
    builder.add_image(image);
}
```

Users may find other ways to utilize this. It adds a couple lines of code. The alternative would be having an plural form of the functions, but this would then require the user to collect all their types which need to be generic. Say you have an array of different types of images, i.e. a StorageImage and an ImmutableImage you'd have to first cast them into `Arc<ImageViewAbstract + ...>` before adding them.

> 3. There is no example of descriptor arrays(did I miss something?). Would be nice to have one to demonstrate this new feature and show to the users on how to utilize it.

Runtime arrays work exactly the same as normal arrays in use.
 
> 4. To the log entries I would also add a general description of the changes purpose("support of runtime descriptor arrays").

I think it is better to break out the changes for this a bit more as it will help users migrating from previous versions.

vulkano/src/descriptor_set/single_layout_pool.rs
```diff
@@ -0,0 +1,492 @@
...
+            runtime_array_capacity,
+            set_count: 4,
+            target_layout,
+        })
+    }
+
+    /// Starts the process of building a new descriptor set.
+    ///
+    /// The set will corresponds to the set layout that was passed to `new`.
+    pub fn next<'a>(&'a mut self) -> Result<SingleLayoutDescSetBuilder<'a>, DescriptorSetError> {
```
Eliah-Lakhin, line 90: Can be simplified to:

```rust
pub fn next(&mut self) -> Result<SingleLayoutDescSetBuilder, DescriptorSetError> {
```

Eliah-Lakhin (changes requested):
@AustinJ235 I finished testing of this feature branch in my project, and I discovered one regression in implementation.

Shader's layout:
```glsl
...
// just an ordinary single sampler
layout(set = 0, binding = 0) uniform sampler2DArray TEXTURES;
...
```

Resource creation:
```rust
let (array, array_initializer) = ImmutableImage::uninitialized(
  ...
  ImageDimensions::Dim2d {
    width: ...,
    height: ...,
    array_layers: 1, // this value is a source of the bug
    ...
).unwrap();

// Then submitting CommandBuffer to initialize `array` resource. No issues so far.
```

Then describing an image:
```rust

let mut builder: SingleLayoutDescSerBuilder = ...;

builder
  .add_sampled_image(array_image_view.clone(), sampler_object).unwrap();

builder.build().unwrap(); // Panics to "ImageViewTypeMismatch { expected: Dim2dArray, obtained: Dim2d }"

```

Changing `array_layers` to a value greater than one "fixes" the issue. Prior to this migration Vulkano was smart enough to turn Dim2D to Dim2DArray automatically based on provided layout even if `array_layers` refers just a single image.

An issue with the newer implementation is that the user can't describe `sampler2DArray` resource with just a single layer, which is a correct array resource by itself(sometimes the user may want to switch between single-element arrays and multi-element arrays).

---

Aside from this, everything else work correct. I didn't detect any other issues or regressions. Also, I like the implementation in general. In turns out that the new(dynamic) interfaces are much more handy in practice. :)

And I also pointed out one minor code-clean target to fix.

vulkano/src/descriptor_set/single_layout_pool.rs
AustinJ235: Commit 4094a63

AustinJ235:
I'll look into the regression. Not exactly sure how that happens as the new implementation copies a large amount of code from the previous.

AustinJ235:
The `image_match_desc` method in `descriptor_set/builder.rs` is basically a direct copy from the original which was changed in https://github.com/vulkano-rs/vulkano/pull/1680. The only things that differs is that the errors types are converted. I am guessing that is the original source of the regression.

Investigating the code now.

AustinJ235:
@Rua I think this regression should be solved in commit ae3e278; however, I did not test it.

AustinJ235:
Don't merge yet.... Seems things might be broken.

AustinJ235:
Issue is with binding descriptor sets, `bind_descriptor_sets`. The pipeline layout needs to contain the correct amount of descriptors for a set in other words it needs to use the `DescriptorSetLayout` from the created `Set`.

Available options I can think of:
1. Provide `PipelineLayout::replace_descriptor(&self binding: u32, desc: Arc<DescriptorSetLayout>) -> PipelineLayout`
    - Requires the user to recreate the pipeline layout before using `bind_descriptor_sets` if they are using runtime arrays. The one provided by `pipeline.layout().descriptor_set_layouts()` is invalid for this use case.
2. Remove `pipeline_layout` argument from `bind_descriptor_sets()` and make the `DescriptorSet` trait be able to provide the layout. This requires the descriptor set constructors to use the entire `PipelineLayout` with an binding index instead. This is how was previous done I believe. The old docs, which are actually still in place, describe that behavior.
    - This would artificially limit use of runtime arrays to one set.
    - Kind of quirky when it comes to which layout to use from which set in the backend.
3. Maybe the best? Set the descriptor count to the limit of the device for descriptors with variable count.
    - Is there memory implications of this?

AustinJ235:
Commit 42a0341 provides a solution.

I am not sure if it is the correct one. The ash version of `DescriptorSetLayout` now defines `descriptor_count` as the limit of the device for `DescriptorDesc` that are `variable_count`. I am unsure if this will lead to excess memory usage. Allocation of descriptor sets will still use a separate count provided by `DescriptorSetLayout::descriptors_count`, so memory usage isn't a concern allocation wise. This method will panic if the `DescriptorSetLayout` is created a `variable_count` `DescriptorDesc` that does not define its `descriptor_count`. This value defaults to `0` from `vulkano-shaders`. This should only be possible if the end user is implementing their own allocation and descriptor sets hence why I opt'd for an `assert!` in the `descriptors_count` method to check if this count is undefined.

Another point of concern is that the layout used for allocations and ultimately provided to sets is technically different from what is created in the pipelines and ultimately in the bind commands. These I would assume are compatible.

Rua:
I'm not sure if I understand what's going on, but I'll try to say something useful anyway. As far as I can tell, `DescriptorsCount` is used for allocating descriptors from a pool. I'm not sure, but I assume that if you have a variable count, then the number of descriptors that's actually in a descriptor set will vary for each individual set, and not depend on the layout? If that's the case, then perhaps `DescriptorsCount` should become `Option<DescriptorsCount>`, or `DescriptorsCount` should only contain a minimum guaranteed count of descriptors, with the variable part left out and needing to be calculated from each descriptor set. The latter option means that as much of the work as possible counting descriptors is done up front, and only the variable counts need to be calculated when creating a descriptor set.

In your example, you correctly enable the two features needed for this to work, but these features only became part of core Vulkan in version 1.2, and require an extension for earlier versions. When creating the instance, you specify 1.2, but this is a *maximum* (note the documentation!), so you're simply promising not to use any features above 1.2. Whether the device itself supports it is another matter, so when enumerating physical devices, you need to require that devices *either* have an `api_version` of least 1.2, or support the `ext_descriptor_indexing` extension (which you must then add to `device_extensions`).

AustinJ235:
> I'm not sure if I understand what's going on, but I'll try to say something useful anyway. As far as I can tell, `DescriptorsCount` is used for allocating descriptors from a pool. I'm not sure, but I assume that if you have a variable count, then the number of descriptors that's actually in a descriptor set will vary for each individual set, and not depend on the layout? If that's the case, then perhaps `DescriptorsCount` should become `Option<DescriptorsCount>`, or `DescriptorsCount` should only contain a minimum guaranteed count of descriptors, with the variable part left out and needing to be calculated from each descriptor set. The latter option means that as much of the work as possible counting descriptors is done up front, and only the variable counts need to be calculated when creating a descriptor set.

This is a bit tricky how this all works.

Descriptors with a variable count act exactly the same as normal arrays do mostly (Aside from the extra pNext data in `DescriptorSetLayout`. This applies to layout creation, allocation, and binding. Both the persistent and the pool use the layout from the shader as a base. I have it so the `descriptor_count` coming from the shaders is set to `0`. Obviously you can't figure this out at runtime, so setting it to zero basically inhibits creation of sets without recreating of the `DescriptorSetDesc`. Both the pool and persistent recreate `DescriptorSetDesc`. In the persistent variant this is a fixed count set exactly the length. And in the pool recreates this every time the capacity changes. The pool allocates all of the capacity each time and partially binds descriptors too it. It'll will automatically adjust the capacity if a set is to be created in length that exceed the capacity. This all works good and I believe works to spec; however the tricky part comes to binding. Normally the binding methods take the layout directly from shader. It is kind of odd that we create this layout at shader creation as "most" seem to do this at the same time as descriptor set creation. `PipelineLayout` ideally needs to be recreated whenever the capacity of a any set changes (actually it true length, capacity as I have been talking about it). I can't think of a way that makes this possible without some overhauling to how `DescriptorSetLayout` is handled overall. This would probably include removal of its creation in the `ShaderModule` and then be created at runtime along with descriptor sets.

> In your example, you correctly enable the two features needed for this to work, but these features only became part of core Vulkan in version 1.2, and require an extension for earlier versions. When creating the instance, you specify 1.2, but this is a _maximum_ (note the documentation!), so you're simply promising not to use any features above 1.2. Whether the device itself supports it is another matter, so when enumerating physical devices, you need to require that devices _either_ have an `api_version` of least 1.2, or support the `ext_descriptor_indexing` extension (which you must then add to `device_extensions`).

I believe this only really works well with vulkan 1.2, so I think we'd go that route. I think the extension may work, but I believe there are some quirks with it. Didn't know that was the maximum though, so yeah should probably check that then.

Rua:
> Obviously you can't figure this out at runtime, so setting it to zero basically inhibits creation of sets without recreating of the `DescriptorSetDesc`. Both the pool and persistent recreate `DescriptorSetDesc`.

This part I don't understand. Why do you need to recreate `DescriptorSetDesc`? What part of it are you changing? According to the spec, "`VK_DESCRIPTOR_BINDING_VARIABLE_DESCRIPTOR_COUNT_BIT` indicates that this descriptor binding has
a variable size that will be specified when a descriptor set is allocated using this layout. The
value of descriptorCount is treated as an upper bound on the size of the binding." Note that the `tweak` method is explicitly used for aspects of a descriptor set layout that can't be inferred from shaders, but must be provided explicitly by the user. Currently this includes specifying which buffers are dynamic, but this upper bound is definitely also one of these things. I suggest renaming `tweak` to `set_dynamic_buffer`, and add a `set_variable_descriptor_count(binding: u32, descriptor_count: u32)` method. This can be used by the user to explicitly set descriptor count, and should error out for non-variable bindings.

> `PipelineLayout` ideally needs to be recreated whenever the capacity of a any set changes (actually it true length, capacity as I have been talking about it). I can't think of a way that makes this possible without some overhauling to how `DescriptorSetLayout` is handled overall.

Does this relate to the same stuff as above? Recreating pipeline layouts definitely seem excessive, and I don't know what in Vulkan actually requires it.

Eliah-Lakhin (approved):
All issues addressed. I'm approving this Pull Request :+1:

Eliah-Lakhin:
@AustinJ235 ah, sorry I missed to check provided example. And it is failing on my machine:

```
Using device: AMD RADV VEGAM (LLVM 10.0.0) (type: DiscreteGpu)
thread 'main' panicked at 'internal error: entered unreachable code', vulkano/src/descriptor_set/pool/standard.rs:125:67
```

A panic is happening on descriptor set creation unwrapping exactly: `let set = Arc::new(set_builder.build().unwrap());`

Eliah-Lakhin (changes requested):
There are still some issues with descriptors creation(see my comment above).

AustinJ235:
@Rua
- `DescriptorSetDesc` now has `set_variable_descriptor_count` method.
- `DescriptorSetLayout` is no recreated within the pool or persistent.
- When building a set the length must not exceed `descriptor_count`, but does not need be the correct length. (runtime arrays only)

I however did not rename the tweak method. Kind of seems like a needless breaking change? Although the name is sort of misleading.

Rua:
Looks good! Does that solve the issues you were having?

AustinJ235:
@Rua I mean the commit before did solve it, just in a brute force way... 🤣. I will say this feels more in line to what it should be. I still think creating a `PipelineLayout` manually is still ugly. Perhaps in the future we could create a builder, but I think this is probably outside the scope of this pull request.

Rua:
The `GraphicsPipelineBuilder::with_auto_layout` function takes a list of dynamic buffer indices that are then passed to the `tweak` method. This same principle could be applied to variable counts as well. That way, the builder does it for you automatically and you don't have to be creating pipelines manually. But instead of adding more and more parameters to the function for each kind of "tweak", I wonder if there's a generic solution that could work for any number of tweaks. Perhaps something with a user-provided closure that gets fed with `&mut [DescriptorSetDesc]` and can then make whatever tweaks it likes.

AustinJ235:
@Eliah-Lakhin I think this should work on your system now. The panic you were getting is suppose to be unreachable code, but I think the brute force method didn't fair your device well...

@Rua I was think more along the lines of something like this:

```
PipelineLayoutBuilder::start(...)
	.add_entry_point(vs.main_entry_point())
	.unwrap()
	.add_entry_point(fs.main_entry_point())
	.unwrap()
	.set_dynamic_buffers(0, 0, buffers)
	.unwrap()
	.set_variable_desc_count(1, 0, 4)
	.unwrap()
	.build()
	.unwrap()
```
`add_entry_point` would add all the shaders entry points you'd be using for this layout, you might not bind every descriptor in each shader (kind of awkward to merge these manually). The builder would make sure these shaders are compatible in terms of binding. The methods like `set_dynamic_buffers` and `set_variable_desc_count` would take a set number, and a binding number.

AustinJ235:
@Eliah-Lakhin This is ready for merge, I just want to make sure you don't have issues yet. I was unable to reproduce the issue on the commit that you had the issue on. I think it was just the brute force fix, but I am not sure.

This pull request works as intended on my systems:
- AMD Frontier Edition on Void Linux (Kernel 5.13 with RADV)
- Nvidia RTX 3060 Ti on Windows 11 (Build 22000)

Eliah-Lakhin (approved):

Eliah-Lakhin:
@AustinJ235 Thank you for fixing it. Yes, your fix resolved the issue on my end too. All works fine. This is ready for merge. :+1:

Rua:
@Eliah-Lakhin what about the issues I pointed out?

Rua (changes requested):

Eliah-Lakhin:
@Rua, @AustinJ235  Do you think guys, it would be reasonable to address these issues in dedicated PR?

Rua:
I'm ok with making a separate PR for it.

AustinJ235:
@Rua most of the checks are currently in builder.rs in the start() method

Rua:
That's not the right place though. When you start building an actual descriptor set, you are passing it a complete descriptor set layout. But these checks apply to the creation of the descriptor set layout already, so they should be put there.

AustinJ235:
> That's not the right place though. When you start building an actual descriptor set, you are passing it a complete descriptor set layout. But these checks apply to the creation of the descriptor set layout already, so they should be put there.

Literally just pushed a commit that solved that LOL

AustinJ235:
@Rua I think all of your issues that you addressed should be solved. `variable_count` however is still in `DescriptorDesc` and not `DescriptorSetDesc`. I think that change would just add additional complexity.

Rua:
Fair enough. I think it's ready for merging.

Rua:
@AustinJ235 why did you add the `Send + Sync` requirement to `BufferAccess` in this PR?

AustinJ235:
@Rua I believe I tried to work around this, but I couldn't find a solution to keeping the trait as is. Can't remember the specifics of it, but ultimately it led to `BufferAccess` requiring `Send + Sync`. Is there a usage out there that this change would break?

Rua:
Since this PR requires an `Arc` around its resources, I was thinking about making that a general requirement across Vulkano. I wondered if the `Send + Sync` requirement was a part of that. But I realise now that `ImageViewAbstract` didn't require `Send + Sync` in your PR, so there must be some other reason?

AustinJ235:
Think there was some quirkiness with `BufferView` that required `BufferAccess` to have `Send + Sync`. Noteworthy though is that I require `Send + Sync` with `ImageViewAbstract` provided, so one you could simply have that trait require that. I think most of the code ultimately assumes it to be `Send + Sync` as requirement. Even in the previous state of vulkano, is it even possible to have an image or a buffer not be `Send + Sync` and then use it?

I am most certainly not against a change to make this a requirement and in fact it probably should be done anyways. Vulkan itself is inherently multithreaded.

Issue #1675, opened 2021-08-18
`vulkano::buffer::BufferUsage::storage_buffer()` implementation missing


itzmeanjan:
# Template

*   Version of vulkano: 0.25.*
*   OS: GNU/Linux
*   GPU (the selected PhysicalDevice): [Intel HD Graphics](http://vulkan.gpuinfo.org/listreports.php?devicename=Intel%28R%29+HD+Graphics+5500+%28BDW+GT2%29)
*   GPU Driver: TODO
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: TODO

# Issue

Hello, first of all thanks for creating `vulkano`, enabling comfortable interaction with vulkan API, compared to writing vulkan in C. 

I was wondering why in [BufferUsage](https://docs.rs/vulkano/0.25.0/src/vulkano/buffer/usage.rs.html#19-32)::`storage_buffer` not implemented as it's done for other buffer usage patterns like [transfer{src,dst}](https://docs.rs/vulkano/0.25.0/src/vulkano/buffer/usage.rs.html#71-76) ?

Rua:
I don't see a reason why it couldn't be added. Do you want to make a PR?

itzmeanjan:
> Do you want to make a PR?

I've raised one [here](#1676)

Rua:
This is merged now.

PR #1676, opened 2021-08-19
Add: missing method `vulkano::buffer::BufferUsage::storage_buffer()`

itzmeanjan:
Added missing method `vulkano::buffer::BufferUsage::storage_buffer()`, which I mentioned of in [issue](#1675). 

It doesn't break anything & minimal user targeted documentation, I've added. This method enables using allocated buffer only as storage buffer.


AustinJ235:
Easy enough.

Issue #1633, opened 2021-07-06
Debug builds on windows result in "thread 'main' has overflowed its stack"


seratonik:
*   Version of vulkano: Git master branch at 965c9f152450f8f67f129f51111e6264c93d2ce1
*   OS: Windows 10 Pro 64bit Build 19042.1052
*   GPU (the selected PhysicalDevice): NVIDIA GeForce RTX 3090
*   GPU Driver: 471.11

# Issue

Something that changed after commit 3850a923f9353f9afba86d7c854bf54e5dc331d3 seems to be causing a stack overflow error when I attempt to run my game that uses Vulkano, but it only happens in Windows and only when compiling in debug mode. Compiling on OSX or in Release mode on Windows fixes the issue.

I don't have a small reproducible example at the moment, I just wanted to report it when I saw it. I also had issues with this latest version crashing rust-analyzer's macro system in VS Code in both Windows/OSX when using `vulkano_shaders::shader!`

Eliah-Lakhin:
@seratonik The commit you referred to is just a readme update. If the issue didn't exist before, it looks more like some "caching" issue or an issue with dependencies at a first glance. Also it unlikely that any recent changes in vulkano-shaders could lead to crashes. Can you try to find an exact commit from which your project works? Or at least at which Vulkano version.

funmaker:
I have the same issue in my project. I think the cause is surprisingly large size of GraphicsPipelineBuilder struct. In 0.23 it was 352 bytes, 1072 in 0.24, and 84232 in 0.25. In debug builds, each invocation of builder methods adds another 82kb to the stack. This can quickly overflow even in regular usage.

Reproduction:

```rs
use vulkano::pipeline::GraphicsPipeline;

fn main() {
	GraphicsPipeline::start()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back()
	                 .cull_mode_back();
}
```

Chris--B:
I tracked down some sizes of things and I think the issue is a `SmallVec` optimization in `GraphicsEntryPoint`.

`GraphicsEntryPoint` looks like this:
```rust
[...]

PR #1681, opened 2021-08-23
Decreased size of GraphicsPipelineBuilder

nalane:
Reduced size of GraphicsPipelineBuilder from ~84kb to ~1kb. This fixes the stack overflow described in [this issue](https://github.com/vulkano-rs/vulkano/issues/1633). Fix made primarily by @Chris--B.

### DIFF
--- a/vulkano/src/pipeline/shader.rs
+++ b/vulkano/src/pipeline/shader.rs
@@ -26,7 +26,6 @@ use crate::pipeline::layout::PipelineLayoutPcRange;
 use crate::sync::PipelineStages;
 use crate::OomError;
 use crate::VulkanObject;
-use smallvec::SmallVec;
 use std::borrow::Cow;
 use std::error;
 use std::ffi::CStr;
@@ -231,7 +230,7 @@ pub struct GraphicsEntryPoint<'a> {
     module: &'a ShaderModule,
     name: &'a CStr,
 
-    descriptor_set_layout_descs: SmallVec<[DescriptorSetDesc; 16]>,
+    descriptor_set_layout_descs: Vec<DescriptorSetDesc>,
     push_constant_range: Option<PipelineLayoutPcRange>,
     spec_constants: &'static [SpecializationMapEntry],
     input: ShaderInterface,
@@ -344,7 +343,7 @@ impl GeometryShaderExecutionMode {
 pub struct ComputeEntryPoint<'a> {
     module: &'a ShaderModule,
     name: &'a CStr,
-    descriptor_set_layout_descs: SmallVec<[DescriptorSetDesc; 16]>,
+    descriptor_set_layout_descs: Vec<DescriptorSetDesc>,
     push_constant_range: Option<PipelineLayoutPcRange>,
     spec_constants: &'static [SpecializationMapEntry],
 }


AustinJ235:
@Rua starting to think the use of SmallVec should be reconsidered in general. Especially if we plan to support platforms that have low stack size. It seems sort of questionable if there is even a performance benefit.

AustinJ235:
Not that pipeline creation is generally done in a performance critical path, but can we assume a minimum capacity. Creating a `Vec` with `with_capacity` can avoid several allocations. Using a `Vec` here we don't have to worry about stack overflows as they are on the heap instead.

Rua:
4 is the minimum number of descriptor set slots that an implementation must support, and some indeed support no more. So that might be a sensible default capacity.

Chris--B:
In my experience of using `smallvec` and related crates, the biggest benefit comes from small, short-lived types that store a small-ish but dynamic number of smaller types in a collection and then die. Parsing with a million short strings, short lists of ints in Advent of Code problems, etc. The performance wins come from reducing many thousands of allocations per second to zero.

Maybe there's a middle ground here that doesn't allocate in the common case, but still keeps object size under control? I think most of that 84k from #1633 isn't used by most pipelines, since most pipelines are vs+fs only [citation needed]. A lot of those bytes end up unused behind an `Option` in the builder for pipeline stages that aren't used.
https://github.com/vulkano-rs/vulkano/blob/4f4580e749246ababfa5f1e01846989904e1e9a1/vulkano/src/pipeline/graphics_pipeline/builder.rs#L63-L80

Here are the field sizes from my investigation:
```
[graphics_pipeline/builder.rs:105] mem::size_of_val(&self.vertex_definition)       = 24
[graphics_pipeline/builder.rs:106] mem::size_of_val(&self.vertex_shader)           = 16776 // !!!
[graphics_pipeline/builder.rs:107] mem::size_of_val(&self.input_assembly)          = 32
[graphics_pipeline/builder.rs:108] mem::size_of_val(&self.input_assembly_topology) = 8
[graphics_pipeline/builder.rs:109] mem::size_of_val(&self.tessellation)            = 33552 // !!!
[graphics_pipeline/builder.rs:110] mem::size_of_val(&self.geometry_shader)         = 16776 // !!!
[graphics_pipeline/builder.rs:111] mem::size_of_val(&self.viewport)                = 32
[graphics_pipeline/builder.rs:112] mem::size_of_val(&self.raster)                  = 40
[graphics_pipeline/builder.rs:113] mem::size_of_val(&self.multisample)             = 48
[graphics_pipeline/builder.rs:114] mem::size_of_val(&self.fragment_shader)         = 16776 // !!!
[graphics_pipeline/builder.rs:115] mem::size_of_val(&self.depth_stencil)           = 100
[graphics_pipeline/builder.rs:116] mem::size_of_val(&self.blend)                   = 64
[graphics_pipeline/builder.rs:117] mem::size_of_val(&self.subpass)                 = 16
[graphics_pipeline/builder.rs:118] mem::size_of_val(&self.cache)                   = 8
```

`vertex_shader`, `tesselation`, `geometry_shader`, and `fragment_shader` are all huge, and optional (since it's a builer). When Nate moved the `DescriptorSetDesc`s onto the heap, those types all got smaller. Since they're optional in this struct, we could get the same effect by `Box`ing in the builder (`Option<T>` -> `Optional<Box<T>>`). I'm not sure what consequences that has on the final pipeline though.

The builder also has a `TODO` about optimizing around `ash` structs, so maybe this is all mute.

Is there a good reference workload that @nalane and I could run proposed fixes on and count allocations to get a better idea of what's helping and what's hurting perf?

Rua:
Is this ready for merge?

nalane:
If you're asking me, I don't have any changes I'd like to make unless one of you thinks there's something that needs to be changed. One maybe open question was about creating the `Vec`s with `with_capacity`, but as it stands, they're created in the constructor via `.into_iter().collect()`, so I don't think that really applies here.

AustinJ235:
I'll be reviewing all open pull request tonight around 2am central.

PR #1682, opened 2021-08-24
Fix using the right functions for retrieving physical device info

Rua:
Changelog:
```markdown
- Fixed bug where the wrong functions were used for retrieving physical device info.
- Fixed minor bug in retrieving features from the physical device.
```

Just a small fix for a bug that was giving validation errors.


AustinJ235:
Always like getting closer to the goal of being validation error free!

Issue #1666, opened 2021-08-08
No way of using a different format in an ImageView tahn the image's one


pac85:
As stated in the [official documentation](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkImageViewCreateInfo.html) "If image was created with the VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT flag, and if the format of the image is not multi-planar, format can be different from the image’s format, but if image was created without the VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT flag and they are not equal they must be compatible. Image format compatibility is defined in the [Format Compatibility Classes](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#formats-compatibility-classes) section. Views of compatible formats will have the same mapping between texel coordinates and memory locations irrespective of the format, with only the interpretation of the bit pattern changing."

This is important to be able to automatically convert between sRGB and linear RGB and has many other uses, however vulkano assumes that those two are always equal [here](https://docs.rs/vulkano/0.24.0/src/vulkano/image/view.rs.html#146) [here](https://docs.rs/vulkano/0.24.0/src/vulkano/image/view.rs.html#328) and possibly in other places.

Rua:
If I understand correctly, in order for this to be useful, there needs to be support for those image flags first?

pac85:
mutable format seems to be exposed already

PR #1687, opened 2021-08-31
Add generation of `Format` to autogen

Rua:
Changelog:
```markdown
- **Breaking** Changes to `Format`:
  - `Format` variants are now all uppercase with underscores. This avoids interpretation problems where the underscore is significant for the meaning.
  - The `ty` method and the `FormatTy` enum are removed. They are replaced with the `NumericType` enum, which concerns itself only with the numeric representation and not with other properties. There are now three `type_*` methods to retrieve it, for colour, depth and stencil respectively.
  - The `planes` method now returns a slice containing the equivalent single-plane formats of each plane.
- **Breaking** The `ImageAccess` trait no longer has the `has_color`, `has_depth` and `has_stencil` methods. This information can be queried using the `aspects` or `type_*` methods of `Format`.
```
```markdown
- Non-breaking `Format` additions:
  - Formats and their metadata are now auto-generated from vk.xml. This significantly expands the support for YCbCr formats in particular.
  - The `compatibility` method returns an object that can be used to check the compatibility of two formats, explained [here](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap43.html#formats-compatibility-classes).
  - The `components` method returns the number of bits in each colour component.
  - The `compression` method returns the compression scheme as a new enum, `CompressionType`.
  - The `requires_sampler_ycbcr_conversion` method returns whether the "sampler YCbCr conversion" must be enabled on an image view and sampler in order to use this format.
```

This PR adds a bunch of new formats that weren't supported yet, expands with new methods to query particular properties, and makes the existing methods more sane. I also wrote some better documentation for the `format` module. The ability to test for format compatibility allows #1666 to be solved, while all the other new methods fulfill other validity-testing use cases that are currently unimplemented. Sampler YCbCr conversion would certainly be worth a future PR.

The biggest, and perhaps most controversial breaking change is that the variants are now uppercase snake-cased like in C, rather than camelcased like is customary in Rust. But I found that this increases readability of the names a lot for complex formats. Take `G16_B16R16_2PLANE_420_UNORM` for example. In Rust camelcase, this would end up as `G16B16R162Plane420Unorm`, where the 16 and 2 are no longer separated, and no distinction is made between components in the same plane and components in different planes (G in one plane, B and R in the other). Similar problems occur with names like `G8B8G8R8_422_UNORM` and `PVRTC1_2BPP_UNORM_BLOCK`.


AustinJ235 (approved):

AustinJ235:
@Rua The amount of times I have had to closely inspect a format name... I agree with the underscore changes they are very important.

PR #1688, opened 2021-09-02
Add support for immutable samplers, some small changes

Rua:
Changelog:
```markdown
- **Breaking** Changes made to `DescriptorDescTy` to support immutable samplers.
- **Breaking** `DescriptorWrite::combined_image_sampler` now takes the sampler in an `Option`. Use `None` when the descriptor has immutable samplers.
- **Breaking** Changes to pipeline layout tweaks:
  - `DescriptorSetDesc::tweak` is renamed to `set_buffer_dynamic`, and now takes only a single binding index.
  - `DescriptorSetDesc::tweak_multiple` is removed.
  - The dynamic buffers parameter of `GraphicsPipelineBuilder::with_auto_layout` has been replaced with a closure that can be used to make tweaks to the descriptor set layouts as needed.
  - `ComputePipeline::new` has an additional closure parameter identical to the one described above.
```
```markdown
- Added support for descriptors with immutable samplers.
  - They are included as separate members on the appropriate variants of `DescriptorDescTy`.
  - Added a `set_immutable_samplers` method to `DescriptorSetDesc`, which lets you set the immutable samplers for a descriptor. This can be used together with the closure provided to pipeline constructors.
  - `add_image` can be used when building a descriptor set, to provide an image to a combined image sampler descriptor that has immutable samplers.
```

This adds support for immutable samplers, which are samplers that are integrated into a descriptor set layout rather than provided at descriptor set build time. They have been in Vulkan since 1.0, but were not available in Vulkano before. I've included an example showing how they are used, but it's literally the image example with five lines changed, so I suppose that demonstrates how easy they are to use. Since immutable samplers are required for sampler YCbCr conversion, if I make a PR for that later I'll probably "reuse" the example for that, so that there's more new concepts being demonstrated.


AustinJ235:
Perhaps we could only keep the basic examples and the add the "quirky" stuff to the site as snippets that are relevant.

⁣Get BlueMail for Android ​

Rua:
Do you mean that I should remove the example from this PR?

AustinJ235:
@Rua just a future thought, the website is very behind at the moment, but in the future we may want to consider having just basic examples and then have the more advance stuff on the site in snippets. Although having example code specific to a feature is a good test to make sure the feature works as intended. Just thinking about total cost of upkeep.

On the topic of this pull request, Looks good! Personally I feel like 99% of use cases would just need an immutable sampler, so kind of weird that we opted to do all runtime to begin with.

Issue #259, opened 2016-09-11
The shader compiler should import struct definitions from Rust


tomaka:
...and not the other way around.

tomaka:
I opened this issue from my phone.
To elaborate a bit, the problem is that it is "cleaner" and more convenient to have only a single definition of a struct in your program. In other words, you don't want to copy-paste the definition of your struct `Foo`  everywhere you use it.

But right now if you compile multiple shaders that use the same struct, not only you need to copy-paste the definition, but vulkano-shaders will also generate one version of that struct definition for each shader. These versions are of course not "compatible" in the sense that according to Rust they are not the same type.

Kurble:
The generator could of course check for problems in the rust struct that prevent it from generating valid glsl/spir-v code. This case should be an error. What I find more tricky is how these generated structs should be part of a shader. Should they be magically included before compilation?
This is also a slippery slope towards developing a shader tool that compiles SPIR-V code straight from rust. I think that should be a target instead of "just the structs"

PR #1694, opened 2021-09-04
Sharing generated Rust types between shaders

Eliah-Lakhin:
Changelog:

```
 - A new Vulkano-shaders macro option `shaders` to compile several shaders in a single macro invocation producing generated Rust structs common for all specified shaders without duplications. This feature improves type-safe interoperability between shaders.
```

This Pull Requests addresses an old-standing issue in Vulkano-Shaders of lack of a type-safe way of interoperability between shaders' interfaces.

Assume the user declares two independent shaders relaying on the same glsl struct `Foo` in their layouts(e.g. one compute shader writes to the buffer with this struct, another one reads from the same buffer). Prior to this PR the user would have to put two independent invocations of Vulkano Shaders macro producing two independent sets of `Shader`, `Layout` and a  `Foo` Rust structs in the `ty` submodules per each invocation. Direct interoperability between these two Rust `Foo` structs is impossible because Rust treat them as independent types in independent submodules(even though both structs are completely the same in their definitions, fields and names). The user would have to workaround this problem either by introducing explicit casting between generated Rust types, or maybe through the unsafe transmutations between types. In either way the user would have to deal with multiple duplicate Rust types growing multiplicatively when the number of shaders is grow.

This PR introduces a new feature in Vulkano Shaders macro to compile several shaders together. In this new mode the macro inspects all types mentioned in the sources of the shaders, checking if there are no signature clashes between them, and puts final generated Rust structs without duplications in a single shared `ty` submodule common for all specified shaders.

Note, this is not a Breaking change. This new feature is still optional for the users. Also, I provided an example that demonstrates how to use it in practice.

Related to #259.


Eliah-Lakhin:
@Rua, @AustinJ235 Can you review my Pull Request, please?

Rua:
At first glance it looks ok. A few small oddities and typos in the documentation:
- "Within these option" > "With these options"
- "in from of" > "in front of"
- "throws compile-time error" > "throws a compile-time error"

Eliah-Lakhin:
@Rua Thank you for review. I merged the PR. Going to review the next pendings...

PR #1695, opened 2021-09-04
Fix CommandBufferExecFuture resubmitting command buffer after flush

funmaker:
1. [ ] Describe in common words what is the purpose of this change, related

https://github.com/vulkano-rs/vulkano/issues/1294#issuecomment-912838204

This should prevent `CommandBufferExecFuture` from adding the command buffer to submission after being flushed, potentially fixing #1294. I've moved previous implementation of build_submission to build_submission_impl to avoid deadlock in case when flush() locks `submitted` mutex but then executes build_submission that attempts to check submission state using the same mutex.

        - Fixed `CommandBufferExecFuture` adding the command buffer to queue submission after being flushed.

### DIFF
--- a/vulkano/src/command_buffer/traits.rs
+++ b/vulkano/src/command_buffer/traits.rs
@@ -323,17 +323,14 @@ where
     finished: AtomicBool,
 }
 
-unsafe impl<F, Cb> GpuFuture for CommandBufferExecFuture<F, Cb>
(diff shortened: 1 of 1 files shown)


Eliah-Lakhin:
@funmaker ok, I didn't detect any regressions. Merged. Thank you for the fix!

PR #1696, opened 2021-09-04
Add support for `mutable_format` and `block_texel_view_compatible`

### LINKED ISSUE #1666: No way of using a different format in an ImageView tahn the image's one

Rua:
Changelog
```markdown
- Added support and validation for the `mutable_format` and `block_texel_view_compatible` image creation flags.
- Added `format` to the `ImageViewBuilder`, to choose a different format from the underlying image.
```

Closes #1666. I included support for `block_texel_view_compatible` as well, because it fits within the same logic.

### DIFF
--- a/vulkano/src/image/mod.rs
+++ b/vulkano/src/image/mod.rs
@@ -287,28 +287,47 @@ impl From<ash::vk::ImageFormatProperties> for ImageFormatProperties {
     }
 }
 
+/// Flags that can be set when creating a new image.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
 pub struct ImageCreateFlags {
+    /// The image will be backed by sparsely bound memory.
+    ///
+    /// Requires the [`sparse_binding`](crate::device::Features::sparse_binding) feature to be
+    /// enabled.
     pub sparse_binding: bool,
+    /// The image is allowed to be only partially resident in memory, not all parts of the image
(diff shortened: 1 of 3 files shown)


Eliah-Lakhin:
@Rua Looks good. Merged. Thank you!

PR #1700, opened 2021-09-08
Exportable memory as dedicated

hakolao:
- Added `export_fd` and `with_exportable_fd` to `AttachmentImage` and `StorageImage` as well as `mem_size` which is needed when using those images with Cuda
- **Breaking** Buffers and Images that have `with_exportable_fd` use dedicated allocation, thus requiring
`khr_get_memory_requirements2` and `khr_dedicated_allocation` on top of already needed `khr_external_memory` and `khr_external_memory_fd`.

# Why?

~~When exporting memory (buffer) multiple times, the file descriptor stays the same because the allocation uses memory from Vulkano's memory pool. This does not work well with `Cuda` interops and I think it's better to use dedicated allocations.~~
Edit: I suppose the file descriptor might  stay the same anyway... But I found my issues with the buffers and images disappear when using the dedicated allocation.

- Example: Let's say my first buffer is vertices, second indices. The indices would be pointing somewhere else than my cuda buffer pointer when filling the indices with cuda (offsetting didn't seem to help).

I had solved this for months by using my own pool and by allocating a chunk of memory for buffers, and then using that one chunk for my purposes managing offsets myself on the cuda pointers. This though is rather needless and does not work with images. I want images :D

All kinds of weird bugs appeared when trying to let Vulkano choose whether the allocation becomes dedicated or not. (Spent a long time on this with debugger).
- One of the interesting ones was that if I recreated `AttachmentImage` (on resize) it would sometimes result with a standard memory pool allocation and lose the `export_fd` handle type on its memory.

I think all in all the whole exportable memory feature would be best by forcing the memory to be dedicated. Less code also to maintain, and cleaner...

2. [x] Changelog added

- I've used the changes here on a fork at version `0.25.0` (`AttachmentImage` and `DeviceLocalBuffer`


hakolao:
Let me know if this is not a good idea at all. I can work on a fork or revise :)

Eliah-Lakhin:
@hakolao Thank you for your contribution! I don't see any significant blockers to proceed with this implementation, but I will be able to closely review it and test in the beginning of the next week only.

Also, could you please remove changes from the changelog file for your PR, and put them right to this PR description instead? We are maintaining changelog file manually. That avoids a lot of false-positive merging conflicts :)

Eliah-Lakhin:
@hakolao There probably not that many people who use Vulkano together with Cuda, and who use exportable memory features too. So, I would trust your expertise. I didn't detect any regressions or issues with other subsystems, so I merged your Pull Request. Thanks again!

hakolao:
Thanks @Eliah-Lakhin!

PR #1701, opened 2021-09-08
Add auto-generated SPIR-V parsing

Rua:
Changelog
Remove this line, as the `spirv` crate isn't a dependency anymore:
```markdown
spirv_headers 1.5 > spirv 0.2 (crate was renamed and reversioned)
```
Add:
```markdown
- Added a `spirv` module to the main crate, which contains an auto-generated parser for SPIR-V files, and various other utilities that can be used to analyze shaders at runtime.
```

This PR adds SPIR-V parsing types and functions, auto-generated from the SPIR-V grammar document, which is included. The grammar is a JSON document, so this adds `serde_json` as a build-dependency for parsing. Some of the code has been moved from Vulkano-shaders and reworked. In particular, any parse errors that are reported will include information on the affected instruction, where the error happened, and the SPIR-V words of the instruction for debugging.

This is still a draft; the main parsing feature is now done, but I want to add code to gather some additional data from the module after it's parsed. Vulkano-shaders is also a bit of a mess, because it's only haphazardly converted to the new version. That needs to be fixed up too.


Eliah-Lakhin:
> but I plan to add code to ShaderModule that automatically parses the binary at runtime

@Rua Then it probably would require transferring of Shaderc dependency, the most heavyweight dependency in the entire project, to Vulkano crate too. Which is maybe not that big deal as most of the users already use on Vulkano-shaders anyway(but not all, some use 3rd party tools e.g. RustGPU). I suggest to do so by introducing Crate feature.

If you want to work on advancing Shader generation system even further you can think of migrating from Shaderc to something native to Rust, for example to [Naga](https://github.com/gfx-rs/naga).

Rua:
Shaderc is used for compiling the code, while the rest of Vulkano-shaders analyses the compiled code and generates Rust types and other things. I have no intention of moving the compilation over to the main crate, I'm not sure if that's even feasible. Only some of the code analysis will be moved over, because it's useful for shaders that aren't loaded via Vulkano-shaders too. Compilation and the generation of Rust structures will remain where it is.

Eliah-Lakhin:
@Rua Excellent work! As always! Thank you, Rua. :) PR merged.

Issue #1699, opened 2021-09-07
Validation error whilst using variable count descriptor sets


AustinJ235:
So I introduced this...

`[ VUID-VkWriteDescriptorSet-dstArrayElement-00321 ] Object 0: handle = 0x1adb8000000005ab, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x90f9e073 | vkUpdateDescriptorSets() pDescriptorWrites[2] failed write update validation for VkDescriptorSet 0x1adb8000000005ab[] with error: Attempting write update to VkDescriptorSet 0x1adb8000000005ab[] allocated with VkDescriptorSetLayout 0xeb7a02000000058e[] binding index #2 array element 0 with 1 writes but variable descriptor size is 0. The Vulkan spec states: The sum of dstArrayElement and descriptorCount must be less than or equal to the number of array elements in the descriptor set binding specified by dstBinding, and all applicable consecutive bindings, as described by consecutive binding updates (https://vulkan.lunarg.com/doc/view/1.2.189.0/windows/1.2-extensions/vkspec.html#VUID-VkWriteDescriptorSet-dstArrayElement-00321)`

AustinJ235:
https://github.com/vulkano-rs/vulkano/blob/13b7188b711d7bec67bfb0267979f76b82030d5a/vulkano/src/descriptor_set/pool/sys.rs#L189-L194

needs `pNext` set to [`DescriptorSetVariableDescriptorCountAllocateInfo`](https://docs.rs/ash/0.33.2+1.2.186/ash/vk/struct.DescriptorSetVariableDescriptorCountAllocateInfo.html]) when a variable count descriptor is present in the set.

This would mean that sets that contain a variable count descriptor can not be allocated from `StdDescriptorPool`. Which means that `PersistentDescriptorSet` can not be created from `StdDescriptorPool` as that pool is designed to be generic. It would however be valid to use `build_with_pool` method to build from a user created pool. We'd have to somehow check this.

`SingleLayoutDescSetPool` as is specializes for a single layout will still work as long as we pass [`DescriptorSetVariableDescriptorCountAllocateInfo`](https://docs.rs/ash/0.33.2+1.2.186/ash/vk/struct.DescriptorSetVariableDescriptorCountAllocateInfo.html]).

PR #1702, opened 2021-09-09
Fix variable count descriptor set allocations

AustinJ235:
Fixes: https://github.com/vulkano-rs/vulkano/issues/1699

        - `DescriptorSetLayout` now has `variable_descriptor_count` which returns the descriptor count in a variable count descriptor if present.

### DIFF
--- a/vulkano/src/descriptor_set/layout/sys.rs
+++ b/vulkano/src/descriptor_set/layout/sys.rs
@@ -33,6 +33,8 @@ pub struct DescriptorSetLayout {
     desc: DescriptorSetDesc,
     // Number of descriptors.
     descriptors_count: DescriptorsCount,
+    // Number of descriptors in a variable count descriptor if present.
+    variable_descriptor_count: Option<u32>,
 }
 
 impl DescriptorSetLayout {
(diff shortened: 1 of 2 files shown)


AustinJ235:
@Rua @Eliah-Lakhin any idea why `alloc` is split into `alloc_impl` with the main `alloc` method is inlined? Only comment that is there is ``// Actual implementation of `alloc`. Separated so that it is not inlined.`` Is there a performance benefit here?

Rua:
I've seen that in a few places, but I don't get it either.

AustinJ235:
This is a very old thing... https://github.com/vulkano-rs/vulkano/pull/244

There is no specific mention of this separation. Only think I could think of is that there is some performance benefit from mapping `DescriptorSetLayout` in an inlined method versus one that isn't inlined. I wonder if this is some old compiler quirk.

Rua:
Am I correct that a `variable_descriptor_count` of `None` means exactly the same as `Some(0)`? If so, it may be nicer to remove the `Option`.

AustinJ235:
They aren't exactly the same. The method will return `None` when there are no variable count descriptors in a set; however, if there is a variable count descriptor in a set then this will always be `Some(_)`. Although yes this is fundamentally the same in terms of how it is used is used in allocation, but I implemented it this way as obtaining if the set is variable count descriptor or not.

Rua:
But can it not be assumed that if the count is 0, there are no variable count descriptors?

AustinJ235:
Sometimes I over think things.

Eliah-Lakhin:
@AustinJ235 ok. Looks good. PR merged. Thank you for the fix! :+1:

PR #1704, opened 2021-09-12
fix immutable_samplers index in builder

AustinJ235:
@Rua 

https://github.com/vulkano-rs/vulkano/blob/13b7188b711d7bec67bfb0267979f76b82030d5a/vulkano/src/descriptor_set/builder.rs#L387

I think should be:
```rust
if !image_view.can_be_sampled(&immutable_samplers[descriptor.array_element as usize]) {
```

**No changelog entries required as this fixes unreleased code.**


Rua:
Oh yes, I think you're right. Thanks for the fix!

PR #1707, opened 2021-09-12
Add an interactive fractal example

hakolao:
When diving into the Vulkano, I found that it can be a fairly slow process to wrap your head around all the concepts and how to use a renderer as a part of a project. The existing examples lack usable "wholes". E.g. deferred example is great, but it could even go as far to demonstrate something more useful rather than a triangle.

I've found that I prefer Vulkano over e.g wgpu and other libraries which is why I'd like to see nicer examples here.

I thought a simple interactive fractal explorer app made over the weekend could demonstrate Vulkano usage better inside a non-trivial project. Or just get new interested developers up to speed faster.

 It contains
 - Compute pipeline to calculate Mandelbrot and Julia fractals writing them to an image target
 - Graphics pipeline to draw the fractal image over a quad that covers the whole screen
 - Renderpass rendering that image over swapchain image
 - An organized `Renderer` with functionality good enough to copy to other projects
 - Simple `FractalApp` to handle runtime state
 - Simple Input system to interact with the application

Let me know if you prefer to keep the examples more "contained" around specific features or if I'm doing something wrong with vulkano here. I might consider adding more examples later.

Run in release mode or not, doesn't matter ;)
```
cargo run --bin interactive_fractal --release
cargo run --bin interactive_fractal
```

Usage:
```
    WASD: Pan view
    Scroll: Zoom in/out
    Space: Toggle between Mandelbrot and Julia
    Enter: Randomize color palette
    Equals/Minus: Increase/Decrease max iterations
    F: Toggle fullscreen
    Right mouse: Stop movement in Julia (mouse position determines c)
    Esc: Quit
```

Mandelbrot:

mandelbrot

Julia:

julia


Rua:
Ooh, very nice! This will make a great addition!

Eliah-Lakhin:
@hakolao This is amazing! Thank you so much for helping us in improving Vulkano examples. That would help new users a lot, and help in community grow. Examples improvements are always welcome. So if you want to contribute more, I will gladly merge it too. :+1:

Eliah-Lakhin:
Also to mention `VK_LAYER_LUNARG_standard_validation` didn't work on my machine. The layer is deprecated according to Vulkan docs. So it all works only when I explicitly comment it out. Maybe remove it?

Rua:
Yes, `VK_LAYER_KHRONOS_validation` is the official name for it now, on all platforms. But I recommend not requiring layers for examples, as layers aren't included with Vulkan drivers but are installed as an optional extra. The debug example already crashes on my system because of missing layers. They should be treated like extensions: enabled only when available.

hakolao:
One worked on my mac, other on my linux. I dont mind at all if you remove it @Eliah-Lakhin. Better not to cause crashes.

hakolao:
I can remove the layer.

Eliah-Lakhin:
@hakolao You can remove layers within the next PRs. It's not a blocker in general, just a thing to improve.

PR #1709, opened 2021-09-13
Fix import warnings on mac for conditionally used imports

hakolao:
I realized some import warnings on conditional imports on my mac as remnants from #1700. This should fix those.


AustinJ235:
Was almost emitting loads of warnings on windows. Thanks for the fix.

PR #1712, opened 2021-09-15
Add `inner` method to `AutoCommandBufferBuilder`

Rua:
Changelog:
```markdown
- Added an `inner` method to `AutoCommandBufferBuilder`, which returns the inner `SyncCommandBufferBuilder` and can be used to query the current state/bindings.
```

A small PR that allows users to retrieve the state when using `AutoCommandBufferBuilder`.

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -496,6 +496,12 @@ impl<L, P> AutoCommandBufferBuilder<L, P> {
             .unwrap()
     }
 
+    /// Returns the inner `SyncCommandBufferBuilder`, which can be queried for the current state.
+    #[inline]
+    pub fn inner(&self) -> &SyncCommandBufferBuilder {
+        &self.inner
+    }
+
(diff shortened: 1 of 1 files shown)


Eliah-Lakhin:
@Rua What is the purpose of letting the users to see this state through the inner encapsulated value?

Rua:
It allows users to see which descriptor sets, vertex buffers etc they currently have bound, what the viewport is set to, etc etc.

gurchetansingh:
That makes sense.  Vulkano currently can be considered as a utility wrapper, which tracks things the application would other-wise have to.

Eliah-Lakhin:
I think normally components should not expose their inner encapsulated state entirely, especially builders. But if you need this feature we can accept it as a temporary solution.

Rua:
I could have duplicated all the methods so they forward to inner without exposing it, but simply returning a reference is easier. Note that `SyncCommandBufferBuilder` is a publicly available type, so the user already has this information available if they use it directly.

PR #1716, opened 2021-09-21
Limit compatibility check for bound descriptor sets to the length of the current layout

Rua:
This fixes a bug introduced in #1684, where an out-of-bound access panic occurs if the newly bound descriptor sets contain set numbers beyond the end of pipeline layout used for the previous bind.


Eliah-Lakhin:
@Rua I apologize for delaying with review. I'm overloaded with work, but I keep an eye on all of pending PRs, and will do accurate review as soon as possible(hopefully tomorrow). Also, probably we will make 0.26 release on this week. There are a plenty of unrelease features.

AustinJ235:
Looks good!

PR #1717, opened 2021-09-21
fix array_element accounting for add_buffer_view

rroohhh:
Seems like nobody uses {Storage,Uniform} Texel Buffers, and this was missed in #1674


AustinJ235:
Oops! thanks for the catch! Might have released that...

PR #1719, opened 2021-09-26
Replace `Iterator` parameters with `IntoIterator` where possible, replace various iterator types with `impl Iterator`

Rua:
Changelog:
```
- Changed parameters requiring an iterator so that they now take `IntoIterator`, for more flexibility.
- Removed various custom iterator types in favour of returning `impl Iterator`.
```

This is a small usability improvement, by allowing not just iterator types, but also types that can be converted into an iterator (e.g. `Vec`, array).

For maintainability I also replaced some of the custom iterator types that Vulkano defines with an `impl Iterator` return type. Unfortunately, due to Rust bug https://github.com/rust-lang/rust/issues/42940 I wasn't able to replace `QueuesIter` this way. Because of the bug, the compiler assumes that the returned `impl ExactSizeIterator<Item = Arc<Queue>>` is bounded by all input lifetimes, which in the case of `Device::new` includes the `'a` lifetime of `QueueFamily`. This then leads to code breakage which could be fixed by the user, but for such a small change I didn't think it was worth it.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -216,7 +216,7 @@ pub(super) fn reflect<'a, I>(
     types_registry: &'a mut HashMap<String, RegisteredType>,
 ) -> Result<(TokenStream, TokenStream), Error>
 where
-    I: Iterator<Item = &'a str>,
+    I: IntoIterator<Item = &'a str>,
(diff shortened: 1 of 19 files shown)


AustinJ235:
Nice little improvement, thanks!

PR #1720, opened 2021-09-28
Add `Send + Sync` requirements to various traits

Rua:
Changelog:
Remove the "BREAKING BufferAccess now requires Send + Sync" line, and replace with:
```markdown
- **Breaking** Many traits now require `Send + Sync`: `BufferAccess`, `ImageAccess`, `ImageViewAbstract`, `FramebufferAbstract`, `AttachmentsList`, `MemoryPoolAlloc`, `DescriptorSet`, `DescriptorPoolAlloc`, `PrimaryCommandBuffer`, `SecondaryCommandBuffer`, `CommandPoolAlloc`. This further means that any type parameters of types implementing these traits also require `Send + Sync`.
```

Per the discussion at https://github.com/vulkano-rs/vulkano/pull/1674#issuecomment-928977049, this adds an explicit `Send + Sync` requirement to various traits that are generally accompanied by this requirement anyway. This makes it unnecessary to write `Send + Sync` all over the place, as it's now implied by the trait.

### DIFF
--- a/examples/src/bin/buffer-pool.rs
+++ b/examples/src/bin/buffer-pool.rs
@@ -328,7 +328,7 @@ fn window_size_dependent_setup(
     images: &[Arc<SwapchainImage<Window>>],
     render_pass: Arc<RenderPass>,
     viewport: &mut Viewport,
-) -> Vec<Arc<dyn FramebufferAbstract + Send + Sync>> {
+) -> Vec<Arc<dyn FramebufferAbstract>> {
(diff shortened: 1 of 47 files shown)


AustinJ235:
I am all good with this change. Personally can't think of anything that this would potentially break and it really does clean up the code a fair bit. There is however conflicts as some pull requests have been merged.

Rua:
Done.

PR #1721, opened 2021-09-28
Group command buffer state retrieval under a new `CommandBufferState`

Rua:
Changelog:
1. Remove "SyncCommandBufferBuilder now has methods to return the state set by previous commands" line.
2. Remove "Added an inner method to AutoCommandBufferBuilder" line.

Replace 1. with:
```markdown
- `AutoCommandBufferBuilder` and `SyncCommandBufferBuilder` now have a `state` method, which can be used to query the state set by previous "bind" and "set" commands.
```

This is a better solution to #1712, which doesn't expose the whole inner builder. It also doesn't clog up the documentation of `SyncCommandBufferBuilder` with state retrieval methods.


AustinJ235:
I like this approach better it seems much cleaner and may seem less clunky to the end user. Good work!

PR #1722, opened 2021-09-29
Track state in the pipeline, invalidate dynamic state when binding pipeline with fixed state

Rua:
Changelog:
```markdown
- **Breaking** `Compare` is renamed to `CompareOp` to match Vulkan.
```
```markdown
- Replaced the various dynamic state querying methods of `GraphicsPipeline` with two: `dynamic_state` to query a single dynamic state, and `dynamic_states` to iterate over all of them.
```

According to the Vulkan spec, if a pipeline contains _fixed_ state, then it overwrites any _dynamic_ state that was previously set. Vulkano didn't track this before, but it became an issue with #1684. This PR fixes that by tracking, for every graphics pipeline, which state is "baked into" the pipeline, which it expects to be given dynamically, and which state isn't in the pipeline at all. `SyncCommandBufferBuilder` in turn deletes all the overwritten state whenever you bind a new pipeline.

The changelog doesn't mention any of this, since it's internal behaviour. The only user-visible changes are the rename, and the different interface on `GraphicsPipeline` and the two new types I introduced for this purpose, `DynamicState` and `DynamicStateMode`.

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -58,6 +58,7 @@ use crate::pipeline::vertex::VertexBuffersCollection;
 use crate::pipeline::viewport::Scissor;
 use crate::pipeline::viewport::Viewport;
 use crate::pipeline::ComputePipeline;
+use crate::pipeline::DynamicState;
 use crate::pipeline::GraphicsPipeline;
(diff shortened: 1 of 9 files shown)


AustinJ235:
Looks good!

PR #1724, opened 2021-10-09
Lots of additional dynamic state, and some new pipeline state

Rua:
Changelog:
```markdown
- **Breaking** Major changes have been made to most parts of the graphics pipeline state.
  - Each state category is now provided by a separate type ending in `State` that lives in its own submodule of `pipeline`. Some of these were existing types that were renamed.
  - Most of the state-setting methods on `GraphicsPipelineBuilder` are deprecated. State is now set using a `_state` method, which is given one of the above state types.
- **Breaking** `DynamicStateMode` is replaced with a simple `bool`.
- **Breaking** The presence of dynamic state in the pipeline state is now expressed more explicitly with two new types, `StateMode` and `PartialStateMode`.
```
```markdown
- Added support for lots more dynamic state, most of which requires features or extensions to use.
- Added support for discard rectangles, in the `pipeline::discard_rectangle` module.
- Added support for line rasterization state, in the `pipeline::rasterization` module.
- Added support for color write enable state, in the `pipeline::color_blend` module.
- `GraphicsPipeline` now has methods to return each state object it was created with.
- `ash` has been updated to 0.33.3, which provides new features and extensions.
```

A very large PR this one! A lot of new states and features are added here. I decided to change how state is provided to `GraphicsPipelineBuilder` because there was starting to be too much of it, making the number of methods unreadably large. Now, instead, the builder simply takes state objects, each of which has a pure Vulkan equivalent. These state objects have convenience builder methods where appropriate, and sensible defaults. Examples of how this looks are in the... well, examples.

I added support for the `ext_extended_dynamic_state` and `ext_extended_dynamic_state2` extensions, which add "dynamicness" to a lot of existing pipeline state. In addition, I added three completely new pipeline states.

I hope I didn't miss out on any important/breaking changes. There's so many changes that it's hard to remember and list every little thing.

### DIFF
--- a/vulkano/src/pipeline/graphics_pipeline/builder.rs
+++ b/vulkano/src/pipeline/graphics_pipeline/builder.rs
@@ -396,7 +396,10 @@ where
                 // FIXME: must check that the control shader and evaluation shader are compatible
 
                 if !device.enabled_features().tessellation_shader {
-                    return Err(GraphicsPipelineCreationError::TessellationShaderFeatureNotEnabled);
+                    return Err(GraphicsPipelineCreationError::FeatureNotEnabled {
+                        feature: "tessellation_shader",
+                        reason: "a tessellation shader was provided",
+                    });
                 }
 
                 match tess.tessellation_control_shader.0.ty() {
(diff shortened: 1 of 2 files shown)


Rua:
`array::map` was stabilised in 1.55. I can rewrite it if you want, though I'm not sure if there's an equally clean way to do it without `map`.

Eliah-Lakhin:
@Rua ok, there probably was something with my toolchain installation. I reinstalled 1.55 and it works fine now.

As of Rust 2021. I feel positively in moving to the new edition once it releases. Do you have any concerns?

Rua:
I have no concerns, other than the user's ability to compile Vulkano.

Eliah-Lakhin:
@Rua ok, let's discuss it again once they release, and hear more opinions. Personally, I would migrate to Rust 2021 when its ready.

PR #1725, opened 2021-10-11
Add `RangeSet` utility type, check that all push constants needed have been set

Rua:
Changelog:
```markdown
- Added `CommandBufferState::push_constants` to retrieve the set of all push constant bytes that have been set.
- Added check to `AutoCommandBufferBuilder` to ensure that the push constants that are needed by the pipeline have been set.
```

This PR adds a little utility type called `RangeSet`, which is useful for storing byte/index ranges and efficiently checking if they cover a range. Using that, I've implemented a check that was previously missing for push constants. I think `RangeSet` will be useful for other things in the future too - I'm working on a PR to add support for push descriptors, and it will come in handy for that.

### DIFF
--- a/vulkano/src/range_set.rs
+++ b/vulkano/src/range_set.rs
@@ -17,11 +17,13 @@ pub struct RangeSet<T>(Vec<Range<T>>);
 
 impl<T: Ord + Copy> RangeSet<T> {
     /// Returns a new empty `RangeSet`.
+    #[inline]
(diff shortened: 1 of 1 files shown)


Eliah-Lakhin:
@Rua Merged. Thanks!

> I'm working on a PR to add support for push descriptors

That would be very awesome to have it in Vulkano! Thank you very much for working on this one! :100:

PR #1734, opened 2021-10-23
Fixes related to Image export & OpenGL Interop

fayalalebrun:
As a continuation of my previous work, where I added semaphore export  #1606, I've been working on adding the possibility to import Vulkano textures into OpenGL.

However, a few bugs in Vulkano currently prevent this functionality from working correctly:

- The memory builder was not populating its p_next structures at all, meaning memory was created neither as exportable or dedicated. I fixed this by getting rid of the old method of populating the p_next structures and instead using the Ash functionality.
- The exportable image must also be created with an additional structure [VkExternalImageCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkExternalMemoryImageCreateInfo.html). I added a method to `UnsafeImage` which does this, and I used it accordingly in `AttachmentImage`.
- Issue 17 of OpenGL [EXT_external_objects](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_external_objects.txt) demands that the mutable bit be set when some common extensions are present. I thus set it to true by default, but perhaps @hakolao can provide his input on whether this would make sense when sharing images with CUDA.

The result of these issues was heavy visual corruption on the AMD radeonsi driver. Interestingly, the Intel Iris driver does not suffer from visual corruptions even with the p_next creation structures missing.

The only remaining issue at the moment is a crash happening with the Intel Iris driver, when flushing the signal semaphore command from the OpenGL side. Running a debug version of the driver, it prints out `Failed to submit batchbuffer: Invalid argument`. Given that the driver does not pass the piglit `vk-semaphores-2` test, this is probably a problem within the Iris driver. I will submit a bug report to the Mesa repository.

I have created some [sample code](https://gist.github.com/fayalalebrun/8845ba4464af687895cf49ba96b105d6). This program draws to a shared texture with Glium from another thread, which is the drawn to the window by Vulkano. This texture oscillates between a black and green color. I'm interested to hear how well it works on Nvidia or with other Intel/AMD drivers.

Changelog:
- `Fix device memory builder not including p_next structures.`
- `Fix exportable image not created with VkExternalMemoryImageCreateInfo.`
- `Fix mutable bit not set on exportable image.`
[
Related Glium MR](https://github.com/glium/glium/pull/1965)


vulkano/src/image/sys.rs
```diff
@@ -604,7 +660,7 @@ impl UnsafeImage {
 
             let mut out = MemoryRequirements::from(output.memory_requirements);
             if let Some(output2) = output2 {
-                debug_assert_eq!(output2.requires_dedicated_allocation, 0);
+                //debug_assert_eq!(output2.requires_dedicated_allocation, 0);
```
Eliah-Lakhin, line 663: @fayalalebrun Why this assertion commented out?

fayalalebrun: On the RADV driver and probably the Nvidia driver as well (Based on the issues reported in #1700), when you create an exportable image, it requires dedicated memory allocation. Any idea why this assert was there in the first place? Perhaps back then there was no type of image which would require a dedicated allocation?

fayalalebrun:
I forgot to add the fix to StorageImage, but now that is there as well.

vulkano/src/image/sys.rs
fayalalebrun: @Eliah-Lakhin I believe I know why the assertion was there to begin with. The [Vulkan documentation](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkMemoryDedicatedRequirements.html) says that this field may be true if `VkImageCreateInfo` contained a `VkExternalMemoryImageCreateInfo` structure.

Since this structure was never used before this MR, the author of this line probably included this as a sanity check. But now that this structure is used, some drivers do request dedicated allocation for exportable images.

Eliah-Lakhin (approved):

Eliah-Lakhin:
@fayalalebrun I apologize for being slow in review. In general, I don't see significant issues in proceeding with merge, but my knowledge of this topic is too limited, so I would trust your experience.

However, before we proceed I would like to ask you to make the following changes:

1. If you think that the assert we discussed above is no longer relevant, you should either remove it, or put a comment in the code where you have commented it out of why it was commented out. It would help developers in the future in understading of the changes.
2. You linked a sample code in gist demonstrating this feature. I will be very appreciated if you include it in the `examples` directory of our repository. It would not only give other users an idea of how they can utilize these features, but it also helps in maintenance too, because we sometimes use examples for manual testing of new incoming changes. Also, I will be glad to try that new example on my local machine and give you my feedback(I have an AMD driver). Don't hesitate to add glium to the Cargo.toml of examples dir if needed. Examples crate don't affect other crates anyway, and examples aren't published on crates.io.
3. I noticed that you have replaced `ash::vk::MemoryAllocateInfo` with two in-place fields. Is it possible somehow to keep using ash capabilities to achieve desirable goal?

fayalalebrun:
@Eliah-Lakhin First two points are fixed. 

For the third point, my first intention was to keep an ash builder in the memory builder. However, the problem is that pushing p_next structures using ash needs a mutable reference bounded by the lifetime of the builder. Specifically, I need to store the p_next structures inside of the memory builder so that the lifetimes are equal to the builder. But also, I need to take a mutable reference of the p_next structures to push them into the builder. Thus, the memory builder would need to have a mutable reference to itself.

If there's a Vulkano struct somewhere implemented in the way you would like me to do it, let me know, and I'll be happy to do it. However, at the moment I can't think of a clean way to do it.

Rua:
@fayalalebrun You can store the ash builder and the ash extension struct both in the builder, and "connect" them only when you are ready to make the final API call.

Rua:
Sorry, I misread what you actually said before.

Eliah-Lakhin:
> I'm interested to hear how well it works on Nvidia or with other Intel/AMD drivers.

On my `AMD VegaM` it works perfectly fine! :)

These are logs output:
```
Using device: AMD RADV VEGAM (LLVM 10.0.0) (type: DiscreteGpu)
mesa: for the --simplifycfg-sink-common option: may only occur zero or one times!
mesa: for the --global-isel-abort option: may only occur zero or one times!
mesa: for the --amdgpu-atomic-optimizations option: may only occur zero or one times!
Source: ShaderCompiler          Severity: Notification          Type: Other             Id: 1
Shader Stats: SGPRS: 16 VGPRS: 8 Code Size: 88 LDS: 0 Scratch: 0 Max Waves: 8 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
Source: ShaderCompiler          Severity: Notification          Type: Other             Id: 1
Shader Stats: SGPRS: 8 VGPRS: 28 Code Size: 20 LDS: 0 Scratch: 0 Max Waves: 8 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
```

Eliah-Lakhin:
@fayalalebrun Pull Request is merged. Thank you for your hard work and all the fixes! Good job! :+1: 

I see that the glium's PR also merged, so I changed examples' Cargo dependency to master.

Eliah-Lakhin (approved):

vulkano/src/image/sys.rs
```diff
@@ -551,6 +600,13 @@ impl UnsafeImage {
                 ..Default::default()
             };
 
+            if let Some(mem_types) = external_mem_handle_type {
+                infos.p_next = std::mem::transmute(&ash::vk::ExternalMemoryImageCreateInfo {
+                    handle_types: mem_types.into(),
+                    ..Default::default()
+                });
+            }
```
pwaller, line 608: Thanks for your efforts. I am a random bystander who happened to be trying to use this for a Christmas project and encountered an issue: If I'm reading this right, it allocates a `ExternalMemoryImageCreateInfo` temporary which is immediately destroyed. Indeed, when I compile with `--release`, I get a segfault. Hoisting the allocation of `ExternalMemoryImageCreateInfo` outside of the if block so that it spans the `create_image` call makes this issue go away.

fayalalebrun: @pwaller That's indeed a bug. Could you submit an MR with the change?

pwaller: I cannot easily, I am am afraid, though the fix is trivial, apologies.

PR #1739, opened 2021-10-26
`Arc`s, `Arc`s everywhere

Rua:
Changelog:
```markdown
- **Breaking** Added an `Arc` to many parameters and return types:
  - Objects implementing `BufferAccess`, `BufferViewAbstract`, `ImageAccess`, `ImageViewAbstract` or `DescriptorSet` are now always constructed in an `Arc`, and parameters that take one of these types require the `Arc` wrapper.
  - The type parameters of `BufferView`, `ImageView` and `BufferSlice` do not contain this implicit `Arc`.
  - The types `DescriptorSetLayout`, `PipelineLayout`, `ComputePipeline`, `GraphicsPipeline`, `QueryPool`, `RenderPass` and `Framebuffer` are also always constructed in an `Arc`.
- **Breaking** `Framebuffer` no longer has a type parameter.
  - This made the `FramebufferAbstract` trait redundant, and it has been removed.
  - `CommandBufferLevel` and its nested types no longer have a type parameter either.
  - `AttachmentsList` is no longer needed and has been removed.
- **Breaking** The `dimensions` method has been removed as an inherent method from types that already implement `ImageAccess`, to avoid confusion between the inherent method and the method of the trait when they have different semantics.
```

This PR takes the changes made for descriptor set builders in #1674 and applies them more generally. Most of these types were either already constructed in an `Arc`, or required the user to wrap them in one manually in order to do anything useful with them. That is now no longer required, but it results in a lot of breakage because of all the changed function signatures.

Being able to rely on things like buffers and images to be wrapped in an `Arc` makes a lot of things easier to handle in `SyncCommandBufferBuilder`, since it can now keep around as many references to things as it needs instead of having to access everything indirectly via methods on the `Command` trait. It's something I will be relying on to implement descriptor requirement checking using `DescriptorRequirements` and finer granularity for push descriptors. The changes also remove the need for the awful `Box<Arc<_>>` construction.

Having images available as `Arc` allowed me to remove the type parameter from `Framebuffer` and store the images in a simple `Vec` of trait objects.


vulkano/src/command_buffer/auto.rs
```diff
@@ -1105,14 +1080,13 @@ impl<L, P> AutoCommandBufferBuilder<L, P> {
     }
 
     /// Adds a command that copies from a buffer to an image.
-    pub fn copy_buffer_to_image<S, D, Px>(
+    pub fn copy_buffer_to_image<S, Px>(
         &mut self,
-        source: S,
-        destination: D,
+        source: Arc<S>,
+        destination: Arc<dyn ImageAccess>,
```
Eliah-Lakhin, line 1086: @Rua I like this approach in general as it practically eliminates some boilerplate code. However, it also introduces new additional restrictions that in turn lead to a new "boilerplate" issues.

When the user creates an image with `StorageImage::with_usage` they receive an `Arc<StorageImage>`. This object then could be wrapped into `ImageView::new` receiving an `Arc<ImageView<StorageImage>>`. A view internally holds an `Arc` to original storage images, however from now on there is no access to the "Arc" reference of the underlying object from ImageView that could be passed e.g. to the `copy_buffer_to_image` function and similar functions. As such the user will have to keep ImageView and the underlying StorageImage references both(which is error prone and have extra performance costs too).

Perhaps, it could be fixed by changing `ImageView::image` signature to return a reference to the Arc instead of just dereferencing the Arc.

Eliah-Lakhin (changes requested):

vulkano/src/command_buffer/auto.rs
Rua: Absolutely! This is a case I missed, but I fixed with a new commit. Are there any other cases like this that need fixing?

Eliah-Lakhin (approved):

Eliah-Lakhin:
@Rua ok, now looks good. :+1: Thank you for improvements and fixes! :)

PR #1744, opened 2021-11-03
Check against DescriptorRequirements instead of pipeline layout at draw time

Rua:
Changelog:
```markdown
- **Breaking** Replaced `DescriptorDescTy` with `DescriptorType` and made further changes to the members of `DescriptorDesc`.
- **Breaking** Added a `Pipeline` trait to hold methods that are common to all pipeline types.
```
```markdown
- Descriptor resources are now checked against the shader requirements at the time of a draw/dispatch call, rather than at the time the descriptor set is created. Only the resources that are actually needed in the shader are checked, the other resources in a descriptor set are ignored and don't need to be valid.
```

At draw/dispatch time, the code will now check whether the bound descriptor sets are valid by using `DescriptorRequirements`. Previously, the pipeline layout object was used for this, but a pipeline layout isn't really the right place for it, since the same layout could be used with multiple shaders, each with their own specific requirements. Moreover, a shader might not actually use every descriptor that's declared in the pipeline layout, the layout can be a superset of what the shader needs. In this PR, only the resources the shader actually needs are checked. Descriptors that aren't used by the shader can still be bound, but are ignored in the check.

In theory this means that it's now safe to create a descriptor set that doesn't contain resources in all of its descriptors, because all that matters is that the resources needed in the shader are valid. That is required for allowing incremental bindings for push descriptors. But that isn't implemented yet; I'm still thinking of ways to do that. Descriptor set builders currently require you to provide a resource for every binding in the layout, which would need changing. A possibility I'm considering is removing descriptor set builders altogether, and letting the user provide `DescriptorWrite` objects instead.

### DIFF
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -19,7 +19,7 @@ use vulkano::descriptor_set::PersistentDescriptorSet;
 use vulkano::device::physical::{PhysicalDevice, PhysicalDeviceType};
 use vulkano::device::{Device, DeviceExtensions, Features};
 use vulkano::instance::{Instance, InstanceExtensions};
-use vulkano::pipeline::{ComputePipeline, PipelineBindPoint};
+use vulkano::pipeline::{ComputePipeline, Pipeline, PipelineBindPoint};
 use vulkano::sync;
 use vulkano::sync::GpuFuture;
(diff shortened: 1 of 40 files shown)


Eliah-Lakhin:
@Rua I'm sorry for being slow in review. PR verified and merged. Thanks!

PR #1746, opened 2021-11-06
Raytracing flags in `ShaderStages`

VictorBulba:
New types of shaders in `ShaderStages` for future ray tracing support

- Added support for `raygen`, `any_hit`, `closest_hit`, `miss`, `intersection`, `callable` flags to `ShaderStages`


Rua:
I happened to include the extra shader types as part of #1747, so this PR may no longer be necessary?

VictorBulba:
Oh yeah, then it can be closed

Issue #1673, opened 2021-08-16
Improve vulkano-shaders to actually become runnable at runtime and not just proc macros


incertia:
vulkano-shaders is able to do a lot of introspection into a spirv blob and it would be nice to pull out that information dynamically at runtime rather than being forced to use macros at compile time, although keeping the option to use compile time macros would be great. i find myself writing my own spirv blob parser and having to inspire myself a lot from the vulkano-shaders internals to understand what is actually going on.

Rua:
I think you're proposing the same as #1558?

incertia:
Yeah that seems about right. I'll close this.

PR #1747, opened 2021-11-08
Move shader analysis to Vulkano crate, make available for runtime shaders

### LINKED ISSUE #1558: Make shader data generation available at runtime
Vulkano-shaders does two things: it compiles the shader (if you're not using `bytes`) and it generates a bunch of data structures and types from the shader. Since the second step is applied to finished SPIR-V code, some of it could in theory be done for shaders that are loaded at runtime. Of course it is still unsafe to actually _load_ these shaders, but there's no good reason to deny users the convenience of having things like pipeline layouts generated for them.

Therefore I propose to add some way to call on Vulkano-shaders to generate shader information at runtime. The `shader!` macro currently defines new types and implements traits on them, which is not possible at runtime: the runtime function has to return a single known type that can represent any pipeline layout, descriptor layout, etc. This would make it similar to the change made to `RenderPassDesc` that was merged recently. However, in this case we have the option to keep `shader!` working as it does now, or also generate a runtime type like the proposed system here.

Rua:
```markdown
- **Breaking** Vulkano-shaders no longer generates a `Shader` struct, but instead provides `load` as a standalone function that returns `Arc<ShaderModule>` directly.
- **Breaking** Vulkano-shaders no longer generates a function for each entry point. Entry points are now retrieved using the `entry_point` method of `ShaderModule`.
- **Breaking** The `shader` module is moved to the crate root, and `spirv` is now a submodule of it.
- **Breaking** `ShaderModule::new` is renamed to `from_bytes` to match the existing `from_words`.
- **Breaking** `ShaderModule` now parses and analyses the shader provided to it on construction, checks whether the shader is compatible with the device, and gathers data such as descriptor sets and input/output interfaces directly from its code. To provide this data yourself as before, and avoid parsing, you can use the `_with_data` constructors.
- **Breaking** `ComputeEntryPoint` and `GraphicsEntryPoint` are removed, as are the functions on `ShaderModule` that generate them. Instead, you can look up an entry point using the `entry_point` method.
```
```markdown
- Added support for additional shader types in `ShaderStage`, `ShaderStages` and `PipelineStages`.
```

Fixes #1558, #1673. This moves most of the shader analysis code from Vulkano-shaders to the main Vulkano crate, so that it can be used for runtime-loaded shaders as well. `ShaderModule` will now analyse the code given to it by default, but it's still possible to provide the data manually if you want. This is done by Vulkano-shaders to avoid having to parse and analyse the code each time the program is run, so you don't have to pay a performance penalty for using Vulkano-shaders.

The new code also checks compatibility between the shader (SPIR-V capabilities and extensions) and the device. For this, I added more code to autogen that generates checks from vk.xml, which already contains information about device requirements for each SPIR-V capability and extension.

The rest of the changes are mainly rearranging and changes of APIs to facilitate this. Shader entry points can now be retrieved at runtime via a string, based on the data collected from the shader. Vulkano-shaders generates its code a little differently.


examples/src/bin/shader-types-sharing.rs
```diff
@@ -102,7 +102,7 @@ fn main() {
                 // by default and the macro by default producing unique
                 // structs(`MultSpecializationConstants`, `AddSpecializationConstants`)
                 shared_constants: true,
```
Eliah-Lakhin, line 104: If you change this flag to `false`, it will generate `multSpecializationConstants`/`addSpecializationConstants`(is it?), which is also violates naming convention.

examples/src/bin/teapot/main.rs
```diff
@@ -301,8 +307,8 @@ fn main() {
 /// This method is called once during initialization, then again whenever the window is resized
 fn window_size_dependent_setup(
     device: Arc<Device>,
-    vs: &vs::Shader,
-    fs: &fs::Shader,
+    vs: Arc<ShaderModule>,
```
Eliah-Lakhin, line 310: Maybe pass `&ShaderModule` here to avoid `vs`/`fs` cloning?

vulkano-shaders/src/codegen.rs
```diff
@@ -775,6 +343,19 @@ mod tests {
             .collect()
     }
 
+    #[test]
+    fn test() {
```
Eliah-Lakhin, line 347: Can you give more descriptive name to this test, please? In case it fails, it would be easier to find the test code.

vulkano/autogen/spirv_parse.rs
```diff
@@ -415,11 +415,30 @@ fn bit_enum_output(enums: &[(Ident, Vec<KindEnumMember>)]) -> TokenStream {
 
 fn bit_enum_members(grammar: &SpirvGrammar) -> Vec<(Ident, Vec<KindEnumMember>)> {
     let parameter_kinds = kinds_to_types(grammar);
-
-    grammar
+    let kinds_dedup: Vec<_> = grammar
         .operand_kinds
         .iter()
-        .filter(|operand_kind| operand_kind.category == "BitEnum")
+        .cloned()
```
Eliah-Lakhin, line 421: Here and in similar places perhaps cloning and twice collecting can be avoided by turning functional-style call chaining into imperative for-loop. The entire code would look shorter and will cost lesser performance and memory.

vulkano/src/pipeline/compute_pipeline.rs
```diff
@@ -49,8 +50,8 @@ impl ComputePipeline {
     /// to add dynamic buffers or immutable samplers.
     pub fn new<Css, F>(
         device: Arc<Device>,
-        shader: &ComputeEntryPoint,
-        spec_constants: &Css,
+        shader: EntryPoint,
```
Eliah-Lakhin, line 53: Why owned value instead of reference?

vulkano/src/shader/mod.rs
```diff
@@ -7,15 +7,15 @@
 // notice may not be copied, modified, or distributed except
 // according to those terms.
 
-//! Stage of a graphics pipeline.
+//! A program that is run on the GPU.
```
Eliah-Lakhin, line 10: More broadly speaking `on the Graphics Device` which could be in theory a CPU-backed emulator of Vulkan :)

Eliah-Lakhin (changes requested):

Eliah-Lakhin:
@Rua Thank you for your hard work! It's amazing!

I only did a brief code review at this moment, left a few comments to discuss and maybe to fix. I will take a closer look later and do some tests.

In addition to that I have a concern about the generated entry-point reusability. Previously the user could load the module, and then reuse it's main entry point every time they create a pipeline. WIll they be able to the same in he new implementation, or the entry point generation in place is unavoidable?

examples/src/bin/shader-types-sharing.rs
Rua: Ah, I missed that bit. I can convert the name to CamelCase using the `heck` crate that is already used by autogen.

examples/src/bin/teapot/main.rs
Rua: Ah, that was a leftover from an earlier experiment I did with how the entry points should work, that I ended up abandoning. I'll fix it.

vulkano-shaders/src/codegen.rs
Rua: Ok, fixed.

vulkano/autogen/spirv_parse.rs
Rua: Do you mean having a `mut kinds_dedup` and then `push`ing elements into it? I don't know if that would be any faster, and I've always specifically avoided that style in favour of `collect`. My impression is that Rust is quite smart when it comes to optimising iterators, but comparing the assembly output or doing a benchmark is probably the only way to know for sure.

As for the cloning, I suppose that could be avoided, I'll see what I can do.

vulkano/src/shader/mod.rs
Rua: Good point, fixed!

Rua:
`EntryPoint` is not very expensive to generate. The `entry_point` method will borrow the `ShaderModule`, look up the name in a `HashMap`, and generate a new `CString` (which admittedly could be done in the pipeline builder, but performance wise it doesn't matter where it's done).

vulkano/src/pipeline/compute_pipeline.rs
Rua: I noticed while working that `GraphicsPipelineBuilder` took an owned `GraphicsEntryPoint` while `ComputePipeline` took a borrowed `&ComputeEntryPoint`. I removed the borrow to make them both the same. They could both borrow instead, but I think that would be problematic for `GraphicsPipelineBuilder`, and for performance a borrow just adds one layer of indirection.

Eliah-Lakhin:
@Rua Can you fix merging conflicts, please?

Rua:
Done.

Eliah-Lakhin:
@Rua I adapted my project's code to these changes, but I'm receiving the following error:

```
SpirvExtensionNotSupported {
             extension: "SPV_EXT_descriptor_indexing",
             reason: RequirementsNotMet(
                 [
                     "Vulkan API version 1.2",
                     "device extension `ext_descriptor_indexing`",
                 ],
             ),
         }
```

There was no such error before, and it worked fine. If I raise the Instance version to `1.2` the error is gone, but what I don't understand is that it works regardless to the Vulkan version specified in the `vulkano_shaders` macro(i.e. it work with `1.1` and `1.2` both).

Rua:
Do you have the `ext_descriptor_indexing` extension enabled on your device? That's actually a requirement in the spec, listed [here](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap46.html#spirvenv-extensions) and in vk.xml. I guess that Vulkano-shaders did not previously check for this, but the PR is behaving correctly.

Eliah-Lakhin:
@Rua I just tried to enabled it(for Vulkan 1.1). And it asked to enable "khr_maintenance3"(`ExtensionRestrictionNotMet(ExtensionRestrictionError { extension: "ext_descriptor_indexing", restriction: RequiresDeviceExtension("khr_maintenance3") })`). So I did it too. Now the error is:

```
FeatureRestrictionNotMet(
             FeatureRestrictionError {
                 feature: "descriptor_indexing",
                 restriction: RequiredByExtension(
                     "ext_descriptor_indexing",
                 ),
             },
         )
```

> I guess that Vulkano-shaders did not previously check for this, but the PR is behaving correctly.

How so? I suppose, the Instance/Device of Vulkan 1.2 shouldn't let the user load the shader that requires Vulkan 1.2 and/or it's extensions, and explicitly specifies version 1.1.

Rua:
And if you enable the feature too, what happens then? The feature requirement is also according to spec; some extensions require a feature to be enabled when the extension is enabled. `ext_descriptor_indexing` requires the feature `descriptor_indexing`, you can see it in the Vulkano documentation: https://docs.rs/vulkano/0.26.0/vulkano/device/struct.Features.html#structfield.descriptor_indexing.

> How so? I suppose, the Instance/Device of Vulkan 1.2 shouldn't let the user load the shader that requires Vulkan 1.2 and/or it's extensions, and explicitly specifies version 1.1.

It's not really a matter of the shader version itself as far as I can tell. Rather, the shader is compiled with certain SPIR-V capabilities and extensions enabled, and those in turn require matching features and extensions on the Vulkan device. If you look at the SPIR-V assembly code of your shader, you will see the `OpCapability` and `OpExtension` instructions. Vulkano sees those when it parses the shader and checks the device to make sure the proper support for them is enabled. Those instructions are in turn generated by `shaderc` based on the glsl code you've written.

Eliah-Lakhin:
@Rua Sorry, what do you mean by enabling the feature?

Rua:
When you create a device, you pass in a `Features` struct. You enable a feature by setting the appropriate member of the struct to `true`.

Eliah-Lakhin:
@Rua Got it. Now it says that the feature is not supported. It looks very strange, because all of it works with Vulkan 1.2 and without explicitly enabling of these features and extensions.

Eliah-Lakhin:
Anyway, I don't think this is a blocker for the PR merging.

Rua:
I think I see where the problem lies. The `descriptor_indexing` feature must indeed be enabled if the extension is enabled:
> If `ppEnabledExtensions` contains `"VK_EXT_descriptor_indexing"` and the `pNext` chain includes a VkPhysicalDeviceVulkan12Features structure, then `VkPhysicalDeviceVulkan12Features::descriptorIndexing` must be `VK_TRUE`

However, `VkPhysicalDeviceVulkan12Features` is a Vulkan 1.2 structure, so it's not available on Vulkan 1.1 and therefore this VUID will never apply. The logic for checking this doesn't take the version into account, so that's a bug that still needs fixing.

Eliah-Lakhin:
Understood. ok, I thought there is something with my gpu driver.

Anyway, PR is merged. Thanks for all the fixes and explanations!

PR #1748, opened 2021-11-12
Two fixes for extension requirements

### LINKED ISSUE #1730: example basic-compute-shader  fails on Apple M1 Silicon
# System Config

*   Version of vulkano: 0.26
*   OS: macOS 11.6
*   GPU (the selected PhysicalDevice): Apple M1 (type: IntegratedGpu)
*   GPU Driver: Vulkan version  1.2.189 (based on MoltenVK)
*   Minimal complete `main.rs` file that demonstrates the issue: the example is taken from the example ```basic-compute-shader``` taken unchanged from the github repository

# Issue

I am able to build everything, but the test fails on a MacBook Pro (13-inch, M1, 2020) with the following messages:

```
./target/debug/basic-compute-shader

Using device: Apple M1 (type: IntegratedGpu)
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ExtensionRestrictionNotMet(ExtensionRestrictionError { extension: "khr_portability_subset", restriction: RequiresInstanceExtension("khr_get_physical_device_properties2") })', examples/src/bin/basic-compute-shader.rs:70:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### LINKED ISSUE #1750: Check for features that must be enabled with an extension doesn't take version into account
As detailed in https://github.com/vulkano-rs/vulkano/pull/1747#issuecomment-968082921. When checking for features that must be enabled in tandem with a particular extension, the check is applied regardless of what the device API version is. That means that Vulkano can end up requiring features that don't even exist on the device's current version. For example, the `descriptor_indexing` feature was added in Vulkan 1.2, and it must be enabled if `ext_descriptor_indexing` is, but only on Vulkan 1.2. Requiring it on earlier Vulkan versions is nonsensical.

Rua:
```markdown
- Fixed two bugs related to the requirements for enabled extensions:
  - For required extensions that have been promoted, the promoted version now also fulfills the requirement.
  - For features that must be enabled in tandem with extensions (e.g. `descriptor_indexing`), the requirement only applies to Vulkan 1.2 and above, since these features do not exist on earlier versions and thus cannot be enabled.
```

Fixes #1730, fixes #1750. It doesn't fix the fact that `khr_get_physical_device_properties2` needs to be enabled on the instance, whenever the device requires `khr_portability_subset`. That's basically impossible. But now it does consider promotions, so that you only actually need `khr_get_physical_device_properties2` if you're running a Vulkan 1.0 instance; Vulkan 1.1 will also satisfy the requirement.

### DIFF
--- a/vulkano/autogen/extensions.rs
+++ b/vulkano/autogen/extensions.rs
@@ -41,12 +41,17 @@ struct ExtensionsMember {
     name: Ident,
     doc: String,
     raw: String,
-    requires_core: (String, String),
-    requires_device_extensions: Vec<Ident>,
-    requires_instance_extensions: Vec<Ident>,
     required_if_supported: bool,
+    requires: Vec<OneOfDependencies>,
     conflicts_device_extensions: Vec<Ident>,
     status: Option<ExtensionStatus>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq)]
+struct OneOfDependencies {
+    api_version: Option<(String, String)>,
+    device_extensions: Vec<Ident>,
+    instance_extensions: Vec<Ident>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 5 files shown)


Rua:
Making this a draft so I can include a fix for #1750.

Rua:
I've now added a fix for the other issue.

Eliah-Lakhin, when you test this, please test these two bugs that you encountered while testing #1747, when you enable the `ext_descriptor_indexing` extension:
- On Vulkan 1.0, the `khr_maintenance3` extension should be required, but on Vulkan 1.1 and higher it should not be required.
- On Vulkan 1.0 and 1.1, the `descriptor_indexing` feature should not be required, but on Vulkan 1.2 and higher it should be required.

Eliah-Lakhin:
@Rua I tested this branch with my project's codebase. Here are results.

Instance extensions:
```
[VK_KHR_get_physical_device_properties2, VK_KHR_get_surface_capabilities2, VK_KHR_surface, VK_KHR_wayland_surface, VK_KHR_xcb_surface, VK_KHR_xlib_surface]
```

Device extensions:
```
[VK_KHR_storage_buffer_storage_class, VK_KHR_swapchain]'
```

Shader's enabled extensions(same as in our `runtime_array` example):
```glsl
#extension GL_EXT_nonuniform_qualifier : enable
```

Device features(taken from the `PhysicalDevice::supported_features()`): https://gist.github.com/Eliah-Lakhin/ddc6b0cd0addfa0f16ab85ab03110ef4

When I run Instance version 1.2 ans Ahder's Vulkan version 1.1 **it works without issues**(as I understood it should complain about missing `descriptor_indexing` extension?). This is similar to `runtime_array` example.

When I run Instance version 1.1 and Shader's Vulkan version 1.1 I get this error:
```
         SpirvExtensionNotSupported {
             extension: "SPV_EXT_descriptor_indexing",
             reason: RequirementsNotMet(
                 [
                     "Vulkan API version 1.2",
                     "device extension `ext_descriptor_indexing`",
                 ],
             ),
         }
```

But if I include `ext_descriptor_indexing` to the `device_extensions` it works with versions 1.1 fine.

In other words(unless I misread your comments), it works in opposite way.

Rua:
With Vulkan versions I mean specifically the device's API version, the minimum between the supported version and the version you give when creating the instance. The shader Vulkan version doesn't matter, that just tells the compiler what version of the spec to follow and what SPIR-V version to generate.

If you create an instance with version 1.1, then you need the `ext_descriptor_indexing` device extension, so the behaviour you are seeing seems correct.

If you create an instance with version 1.2, and do not enable `ext_descriptor_indexing`, then the `descriptor_indexing` feature should also not be required. So I think that is also correct. The feature should only be required if the extension is enabled and the version is 1.2.

Eliah-Lakhin:
@Rua Understood. Then, I think we are good to go. Thank you for the fixes and clarifications! :+1:

Rua:
Also, just so you know, the features that are reported by `supported_features` depend on the API version. Features that are not available below a certain version will be `false` even if the device theoretically supports them. This includes `descriptor_indexing`, which exists only on 1.2.

PR #1751, opened 2021-11-13
Allocate less than the maximum descriptor count when variable

### LINKED ISSUE #1740: Allocation of descriptor sets with variable counts always allocates maximum count
The current code in `UnsafeDescriptorPool` simply allocates `descriptor_count` variable descriptors. In other words, always the maximum number allowed. But it should be possible to allocate less than the maximum, which is the whole point of variable counts in the first place. For this, the iterator passed to `alloc` should include both a descriptor set layout and a variable count. The descriptor set builder should then provide this value when allocating a new set for a layout that has a variable count, based on the number of elements that the user inserts into the array.

I tried fixing this myself, but realised that it will break the logic inside `SingleLayoutDescSetPool`. It expects every descriptor set it allocates to have the same number of descriptors, and therefore be interchangeable, so that previously allocated descriptor sets can be recycled. But if there is a variable count, then sets with the same layout could still have differing numbers of descriptors.

The problem could be solved by simply disallowing variable count layouts in `SingleLayoutDescSetPool`, but that's kinda harsh. It could also be solved by grouping the recycled descriptor sets by their variable count. Unfortunately this would lead to fragmentation of the pool, since only future allocations with the exact right variable count could reuse the recycled sets.

Rua:
Changelog:
```markdown
- **Breaking** Changes to allow allocating variable descriptor counts:
  - The `alloc` method of the `DescriptorPool` trait now has a second parameter to specify the number of descriptors to allocate for the variable count binding. If there is no variable count binding in the layout, this should be 0.
  - The `alloc` method of `UnsafeDescriptorPool` now takes an iterator of `DescriptorSetAllocateInfo`.
```
```markdown
- For `PersistentDescriptorSet` when using a layout with a variable count binding, allocate only the number of descriptors needed instead of always the maximum.
```

Fixes #1740.

Unfortunately I wasn't able to retain variable count functionality for `SingleLayoutDescSetPool`. I thought it would be enough to split the internal reserve of available descriptor sets based on the number of variable descriptors. However, I noticed that this reserve is filled with new descriptor sets as soon as the previous one runs out, which causes problems for variable counts.

When the reserve is empty, and you call `next`, it creates a new internal `UnsafeDescriptorPool`, then immediately allocates as many descriptor sets as possible from it. It then returns one of those descriptor sets for you to build, and adds the rest in its reserve. In the case of variable count layouts, this would mean that it allocates not just one descriptor set with the desired variable count, but a whole bunch of them.

My concern is that this would lead to a lot of wasted descriptor sets, because those descriptor sets can only be used if the variable count is exactly the same. Perhaps the allocation strategy should be changed so that it allocates only one descriptor set at a time, and the reserve is only filled when allocated descriptor sets are dropped and returned to the pool. But I think that should be considered for a future PR.

### DIFF
--- a/examples/src/bin/runtime_array/main.rs
+++ b/examples/src/bin/runtime_array/main.rs
@@ -83,7 +83,6 @@ fn main() {
             shader_uniform_buffer_array_non_uniform_indexing: true,
             runtime_descriptor_array: true,
             descriptor_binding_variable_descriptor_count: true,
-            descriptor_binding_partially_bound: true,
             ..Features::none()
         },
         &physical_device
--- a/vulkano/src/command_buffer/synced/commands.rs
+++ b/vulkano/src/command_buffer/synced/commands.rs
@@ -1567,7 +1567,7 @@ impl SyncCommandBufferBuilder {
         let set_resources = match state
             .descriptor_sets
             .entry(set_num)
-            .or_insert(SetOrPush::Push(DescriptorSetResources::new(layout)))
+            .or_insert(SetOrPush::Push(DescriptorSetResources::new(layout, 0)))
         {
             SetOrPush::Push(set_resources) => set_resources,
             _ => unreachable!(),
--- a/vulkano/src/descriptor_set/builder.rs
+++ b/vulkano/src/descriptor_set/builder.rs
@@ -27,10 +27,12 @@ struct BuilderDescriptor {
 /// A builder for constructing a new descriptor set.
 pub struct DescriptorSetBuilder {
     layout: Arc<DescriptorSetLayout>,
-    in_array: bool,
-    descriptors: Vec<BuilderDescriptor>,
-    cur_binding: u32,
+    variable_descriptor_count: u32,
     writes: Vec<DescriptorWrite>,
+
+    cur_binding: u32,
+    descriptors: Vec<BuilderDescriptor>,
+    in_array: bool,
     poisoned: bool,
 }
 
@@ -61,10 +63,12 @@ impl DescriptorSetBuilder {
 
         Self {
             layout,
-            in_array: false,
+            variable_descriptor_count: 0,
+            writes: Vec::with_capacity(desc_writes_capacity),
+
             cur_binding: 0,
             descriptors,
-            writes: Vec::with_capacity(desc_writes_capacity),
+            in_array: false,
             poisoned: false,
         }
     }
@@ -75,15 +79,20 @@ impl DescriptorSetBuilder {
             return Err(DescriptorSetError::BuilderPoisoned);
         }
 
+        if self.in_array {
+            return Err(DescriptorSetError::InArray);
+        }
+
         if self.cur_binding != self.descriptors.len() as u32 {
-            Err(DescriptorSetError::DescriptorsMissing {
+            return Err(DescriptorSetError::DescriptorsMissing {
                 expected: self.descriptors.len() as u32,
                 obtained: self.cur_binding,
-            })
+            });
         } else {
             Ok(DescriptorSetBuilderOutput {
                 layout: self.layout,
                 writes: self.writes,
+                variable_descriptor_count: self.variable_descriptor_count,
             })
         }
     }
@@ -113,7 +122,7 @@ impl DescriptorSetBuilder {
     pub fn enter_array(&mut self) -> Result<&mut Self, DescriptorSetError> {
         self.poison_on_err(|builder| {
             if builder.in_array {
-                Err(DescriptorSetError::AlreadyInArray)
+                Err(DescriptorSetError::InArray)
             } else if builder.cur_binding >= builder.descriptors.len() as u32 {
                 Err(DescriptorSetError::TooManyDescriptors)
             } else if builder.descriptors[builder.cur_binding as usize]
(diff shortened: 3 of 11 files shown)


Eliah-Lakhin:
Thank you for the fixes. I will take a closer look on this Sun(Nov 21-22).

Eliah-Lakhin:
@Rua On this PR's branch my code is hitting segfaults from time to time that don't happening on `master`. I can't outline the exact scenario of when its happening yet(however, I keep debugging), but it seem to be related to the branch changes specifically.

In my layout setup I have two descriptor sets of exact size, and three descriptor sets with variable size descriptors. For the first two I create descriptors through the two independent `SingleLayoutDescSetPool`s. For the last three I create the descriptors through the `PersistentDescriptorSetBuilder`. The issue seem to be persist even if I remove other three descriptor sets.

Rua:
Do you get anything when you turn on the validation layer?

Eliah-Lakhin:
@Rua Nothing special expect this one `Validation Information: [ VUID-NONE ] Object 0: handle = 0x7f74d49cf970, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x745174a4 | Cannot open shader validation cache at //tmp//shader_validation_cache-1000.bin for reading (it may not exist yet)`, but probably irrelevant. (I'm using `VK_LAYER_KHRONOS_validation` layer)

Eliah-Lakhin:
It passes some of the Compute dispatches, and then fails. Always at unpredictable points.

Eliah-Lakhin:
@Rua ok, I hit the same issue with our `interactive_fractal` example. Always sigfaults for me on your branch, but works fine with `master`. Does it work on your end?

Rua:
I don't get any segfaults running on Linux with the RADV driver. But I do get validation errors for `interactive_fractal`: "vkUpdateDescriptorSets() pDescriptorWrites[0] failed write update validation for VkDescriptorSet 0xa2eb680000000026[] with error". Do you get this as well? The same validation errors appear both on master and on this branch, so this must be a bug that existed before and may not be related to this PR.

If you have the layer available, you could try `VK_LAYER_LUNARG_api_dump`. That will give you a full listing on all the Vulkan calls being made, and perhaps you can compare between master and branch.

Eliah-Lakhin:
> Do you get this as well?

Yes. And I confirm it exists on both branches, however on `master` it does not lead to segfaults. In my codebase I don't receive any Validation Errors. However, it also segfaults on PR's branch and doesn't segfault on `master`. I think that this error is probably not a cause.

Rua:
I'm kinda at a loss regarding the cause of the segfaults. The only thing I can still think of is running the code through a debugger and examining what happens, but that's a lot of work.

Rua:
Random hunch, but if you replace in file vulkano/src/descriptor_set/pool/sys.rs lines from:
```rust
let variable_desc_count_alloc_info = if self.device.api_version() >= Version::V1_2 || self.device.enabled_extensions().ext_descriptor_indexing {
```
into
```rust
let variable_desc_count_alloc_info = if variable_descriptor_counts.iter().any(|c| *c != 0) {
```
do you still get segfaults?

Eliah-Lakhin:
> do you still get segfaults?

Nope, the segfault gone! :)

Rua:
Ok, well that gets a big WTF from me then. I think this may be a bug in your driver: it doesn't like being given a variable count when the layout doesn't have one. As far as I can tell, it is totally fine to pass any number as a variable count if there isn't actually a variable count in the layout. Going by this part of the spec:

> VUID-VkDescriptorSetAllocateInfo-pNext-pNext
> `pNext` **must** be `NULL` or a pointer to a valid instance of VkDescriptorSetVariableDescriptorCountAllocateInfo

And then for VkDescriptorSetVariableDescriptorCountAllocateInfo itself:

> VUID-VkDescriptorSetVariableDescriptorCountAllocateInfo-pSetLayouts-03046
> If VkDescriptorSetAllocateInfo::`pSetLayouts`[i] has a variable descriptor count binding, then `pDescriptorCounts`[i] **must** be less than or equal to the descriptor count specified for that binding when the descriptor set layout was created

From this I understand that if the layout has no variable descriptor count binding, then this value is not used and can be anything. Vulkano always sets it to 0 in that case, which is always less than or equal to any unsigned integer anyway.

I think for now, I will include the code I posted above, just so that it works for you and anyone else that has this bug. But I'd still like to know what OS, GPU and driver you're using so I can see if this is a known problem. Also, perhaps there is a new version for your driver that does not have this problem anymore, that's probably worth trying out.

Eliah-Lakhin:
@Rua, you are Genius! You found a bug that was nearly impossible to find, and almost blindly in remote machine. This is just amazing! :100: 

My platform maybe not so rare. It's Intel NUC with discrete AMD GPU and Linux.

Platform: Linux/Ubuntu 18.04.6 LTS. 
GPU: Radeon RX Vega M GH
Vulkan Instance Version: 1.2.189
Driver: `Vendor: X.Org (0x1002). Device: AMD VEGAM (DRM 3.35.0, 5.4.0-90-generic, LLVM 10.0.0) (0x694c). Version: 20.0.8`.

This is the rest of API dump until the segfault point, from the PR's branch with my project's code run:

```
Thread 0, Frame 2, Time 482554 us:
vkBindBufferMemory(device, buffer, memory, memoryOffset) returns VkResult VK_SUCCESS (0):
    device:                         VkDevice = 0x7f829c4ae620
    buffer:                         VkBuffer = 0x7f829c97ef40
    memory:                         VkDeviceMemory = 0x7f826c000de0
    memoryOffset:                   VkDeviceSize = 672

Thread 2, Frame 2, Time 482806 us:
vkUpdateDescriptorSets(device, descriptorWriteCount, pDescriptorWrites, descriptorCopyCount, pDescriptorCopies) returns void
```

I can provide the full dumps too if it would help.

Eliah-Lakhin:
@Rua Thank you for all the fixes! PR Merged. Good job :+1:

> 20.0.8 is over a year old now

Well, yes, I'm going to update it(and the entire system too), but there is a chance not all users(especially the end-users) would use outdated versions of the Driver too. And while this is an issue of the Driver, it's nice that Vulkano is reliable for such faults too.

Issue #1752, opened 2021-11-14
Make it possible to load shader bytecode with identical entry point names for different execution models


incertia:
It is only [required](https://www.khronos.org/registry/SPIR-V/specs/1.0/SPIRV.html#OpEntryPoint) that `OpEntryPoint` have unique `(name, execution model)`, which lets us link vertex and fragment shader code into a single module, both with the name `"main"`.

We should extend #1747 so that entry points are parameterized by both values, rather than overwriting pre-existing entry points.

Eliah-Lakhin:
Issue closed as resolved in #1753.

PR #1753, opened 2021-11-14
Shader improvements

incertia:
My attempt at addressing #1752. It introduces some breaking changes and a bit of redundancy but I think it is worthwhile for those who like to link their SPIR-V together to produce a single shader package.

Whether or not we want to move forward with this particular API is still rather unclear because an alternative storage method is `HashMap<String, HashMap<ExecutionModel, EntryPointInfo>>`, which has the added benefit of being able to select the singular entry point with a particular name at the expense of slightly more storage.

changelog (UPDATED)
```
* `ShaderModule` can how handle multiple entry points with the same name but differing execution models. `ShaderModule::entry_point` will now return `None` if the entry point with the specified name does not exist or is ambiguous. To disambiguate, use `ShaderModule::entry_point_with_execution`.
```


Rua:
It's a little unfortunate that users will have to select the execution model up front now, when in the majority of cases there will be only one shader in the module. So I think your alternatige suggestion is good. I propose having two methods instead of one:
- The existing `entry_point` method. This should return `None` if there are multiple entry points with the same name.
- A new `entry_point_with_execution_model` or similar, which is more fine-grained.

incertia:
I'm not a huge fan of having the second parameter wrapped in `Option`. I think that defeats the purpose of keeping `entry_point` with a single argument. I pushed some additional code that does that but it's a bit inelegant as we need to collect the entry point `IntoIterator` before we are able to re-iterate through it. Technically `n^2` in number of entry points but I doubt anyone will have enough entry points where that will matter.

vulkano/src/shader/spirv.rs
```diff
@@ -27,6 +27,15 @@ use std::{
 // Generated by build.rs
 include!(concat!(env!("OUT_DIR"), "/spirv_parse.rs"));
 
+impl Eq for ExecutionModel {}
+impl Copy for ExecutionModel {}
+impl std::hash::Hash for ExecutionModel {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // there shouldn't be anything too wrong by directly hashing it as a string
+        format!("{:?}", self).hash(state);
```
Rua, line 35: I think it would make more sense to derive `Hash` directly through autogen.

incertia:
I moved the derive quoting into a separate match arm in case there are future custom derives we would want.

Rua:
At some point I think it would be nice to include those traits on whichever types it's allowed for, but for now I think your solution is fine.

vulkano/src/shader/mod.rs
```diff
@@ -162,10 +164,30 @@ impl ShaderModule {
...
+                    }).collect::<HashMap<_, _>>()
+                )
+            })
+            .collect();
+
         Ok(Arc::new(ShaderModule {
             handle,
             device,
-            entry_points: entry_points.into_iter().collect(),
+            //entry_points: entry_points.into_iter().collect(),
```
Eliah-Lakhin, line 189: Why commented-out code left in the final PR's commit?

Eliah-Lakhin:
@incertia, @Rua It looks good, and I didn't detect any regressions. Also I like the fact that in the end we make it in back-compatible way. Good job, guys! PR merged.

PR #1754, opened 2021-11-14
Add shaderc-debug feature to Vulkano-shaders

### LINKED ISSUE #1742: Debug information for shaders in vulkano-shaders
I was wondering if it'd be possible to add an option to `vulkano-shaders` to turn on debug information in `shaderc-rs`, `shaderc-rs` already has an option for it in `CompileOptions`, `set_generate_debug_info()`, so I was wondering if this could become an option in `vulkano-shaders` since it'd be very useful to be able to see exact lines in tools like NSight Graphics instead of the raw SPIR-V code.
I would do it myself but I'm not very good with procmacros so I'd likely write bad code :P

Rua:
Changelog:
```markdown
- Added a `shaderc-debug` cargo feature to Vulkano-shaders, to emit debug information in the generated SPIR-V code.
```

Fixes #1742. @ProtoByter can you let me know if this implementation is sufficient for your purposes?

### DIFF
--- a/vulkano-shaders/Cargo.toml
+++ b/vulkano-shaders/Cargo.toml
@@ -25,3 +25,4 @@ vulkano = { version = "0.26.0", path = "../vulkano" }
 
 [features]
 shaderc-build-from-source = ["shaderc/build-from-source"]
+shaderc-debug = []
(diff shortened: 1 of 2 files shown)


ProtoByter:
Yea that looks good

PR #1757, opened 2021-11-17
use repr(C)

### LINKED ISSUE #1755: use repr(C) in examples
I think (at least within the context of vertex buffers and input assembly), we should be explicit and use `repr(C)` in the examples whenever possible. In particular, `repr(Rust)` has the weird quirk that the rust compiler is able to re-order field members. In the below non-sensical example, we can see that `repr(Rust)` will reorder `c` to appear before `a` and `b`, so any input assembler relying on `a` being at offset `0` will fail.

```rust
struct MyStruct {
  a: u8,
  c: u32,
  b: u8,
}

#[repr(C)]
struct MyStruct2 {
  a: u8,
  c: u32,
  b: u8,
}

fn to_bytes<T>(t: &T) -> &[u8]{
  let p: *const T = t;
  let p = p as *const u8;
  unsafe {
    std::slice::from_raw_parts(p, std::mem::size_of::<T>())
  }
}

fn main(){
  dbg!(std::mem::size_of::<MyStruct>());
  dbg!(std::mem::size_of::<MyStruct2>());
  dbg!(to_bytes(&MyStruct { a: 0x55, b: 0xaa, c: 0xffffffff }));
  dbg!(to_bytes(&MyStruct2 { a: 0x55, b: 0xaa, c: 0xffffffff }));
}
```

which outputs

```
[test.rs:23] std::mem::size_of::<MyStruct>() = 8
[test.rs:24] std::mem::size_of::<MyStruct2>() = 12
[test.rs:25] to_bytes(&MyStruct{a: 0x55, b: 0xaa, c: 0xffffffff,}) = [
    255,
    255,
    255,
    255,
    85,
    170,
    0,
    0,
]
[test.rs:26] to_bytes(&MyStruct2{a: 0x55, b: 0xaa, c: 0xffffffff,}) = [
    85,
    0,
    0,
    0,
    255,
    255,
    255,
    255,
    170,
    0,
    0,
    0,
]
```

thoughts?

incertia:
fixes #1755. also updates two docstrings to briefly mention why you should use `#[repr(C)]`.

if it is advisable to touch some other examples that share more complicated data with the gpu i can do that as well.

### DIFF
--- a/examples/src/bin/buffer-pool.rs
+++ b/examples/src/bin/buffer-pool.rs
@@ -40,6 +40,7 @@ use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::{Window, WindowBuilder};
 
+#[repr(C)]
 #[derive(Default, Debug, Clone)]
 struct Vertex {
     position: [f32; 2],
--- a/examples/src/bin/deferred/frame/ambient_lighting_system.rs
+++ b/examples/src/bin/deferred/frame/ambient_lighting_system.rs
@@ -151,6 +151,7 @@ impl AmbientLightingSystem {
     }
 }
 
+#[repr(C)]
 #[derive(Default, Debug, Clone)]
 struct Vertex {
     position: [f32; 2],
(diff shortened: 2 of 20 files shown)


Rua:
If you use "fixes" instead of "address", Github will link the two issues so that the issue is closed when this PR is merged. See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue.

Eliah-Lakhin:
@incertia Looks good. Merged. Thanks for the improvement!

Issue #1667, opened 2021-08-10
Ideas on setting attribute format of vertex shader inputs


JennToo:
*   Version of vulkano: 0.25.0
*   OS: Linux (Ubuntu 20.04)
*   GPU (the selected PhysicalDevice): Intel(R) HD Graphics 4600 (HSW GT2) (type: IntegratedGpu)
*   GPU Driver: mesa-intel
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue:
https://github.com/Nitori-/vulkano/commit/3d12692bc7b2cf205ba63c5e750d36d2a1d24c92

# Issue
If you try to run the modified triangle example I linked above, you'll get an error like this:

`IncompatibleVertexDefinition(FormatMismatch { attribute: "color", shader: (R32G32B32Sfloat, 1), definition: (U8, 3) })'`

On its surface this isn't entirely unexpected. The shader looks like it wants a 3-float input format (`R32G32B32Sfloat`), but we're giving it integers.

However, what I really want to use is a normalized format like `R8G8B8Unorm`. As far as I know, there's not a way for me to indicate this in the GLSL shader source itself. To me it seems like the `shader!` macro needs a way to provide hints about what formats you actually want to use. I think it would still be possible for it to validate that you've chosen a compatible format (e.g. if I hinted that I wanted `R8G8B8Unorm` but the shader input was a `vec4` that could and should give an error).

Assuming that `shader!` would be the right place to fix this,I briefly peeked at the `shader!` source. But before I got too far on any kind of implementation I thought it would be good to see what people think about this problem. I'm thinking maybe roughly a syntax like this (but I'm far from an expert on macro ergonomics):

```rust
    mod vs {
        vulkano_shaders::shader! {
            ty: "vertex",
            src: "
                   #version 450

                   layout(location = 0) in vec2 position;
                   layout(location = 1) in vec3 color;

                   layout(location = 0) out vec3 v_color;

                   void main() {
                       gl_Position = vec4(position, 0.0, 1.0);
                       v_color = color;
                   }",
            format: {
                color: Format::R8G8B8Unorm
            }
        }
    }
```

Rua:
I think the problem lies in `VertexMemberTy::matches`. It doesn't seem to take into account that normalised integers in the vertex buffer can be converted into floats in the shader. The table at https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap43.html#_identification_of_formats should probably be taken into account.

[...]

PR #1758, opened 2021-11-20
Bring vertex input in line with other pipeline state, reorganise modules

Rua:
Remove this from the changelog:
```markdown
- **Breaking** Major changes have been made to most parts of the graphics pipeline state.
  - Each state category is now provided by a separate type ending in `State` that lives in its own submodule of `pipeline`. Some of these were existing types that were renamed.
  - Most of the state-setting methods on `GraphicsPipelineBuilder` are deprecated. State is now set using a `_state` method, which is given one of the above state types.
```

And replace it with:

```markdown
- **Breaking** The modules `compute_pipeline` and `graphics_pipeline` are shortened to `compute` and `graphics`, and are now public.
- **Breaking** `ComputePipeline`, `GraphicsPipeline` and `PipelineLayout` are re-exported from the base `pipeline` module, but other items are not.
- **Breaking** Major changes have been made to most parts of the graphics pipeline state.
  - Each state category is now provided by a separate type ending in `State` that lives in its own submodule of `graphics`. Some of these were existing types that were renamed.
  - Most of the state-setting methods on `GraphicsPipelineBuilder` are deprecated. State is now set using a `_state` method, which is given one of the above state types.
- **Breaking** `ShaderInterfaceEntry` now contains the start `location` and `component`, and specifies its type using a new `ty` member instead of as a format.
```

This PR brings vertex input state in line with the changes that were made in #1724. I've also reorganised the modules a bit, so that the modules for each state type are placed under `graphics` instead of directly in `pipeline`. This should make it neater when ray tracing pipelines are introduced. I've included some new documentation as well, especially for the `graphics` and `vertex_input` modules.

While working on this, I came across a few existing shortcomings that need to be addressed in future PRs:
- Specifying shader interface entries using a general numeric type instead of a `Format`, and allowing more flexible matching between shader and attribute format. This is issue #1667, and requires refactoring the `Vertex` trait and the code in `BuffersDefinition`.
- Allowing multiple variables to share a location, using the `component` decoration. This is made possible with `VertexInputState` as introduced in this PR, but it is not supported when using `BuffersDefinition`. It also requires rethinking of the `Vertex` trait.
- 64-bit input values. Because they take up two components per value, and must be matched exactly by the attribute format, they need special care. Unfortunately, my graphics driver doesn't support these at the moment, so I am unable to test it.


Eliah-Lakhin:
@Rua PR looks good. I didn't detect any issues or regressions. Merged. :+1:

It's taking longer to review, than it normally should, and I appreciate your patience. I plan to improve this situation from the next year, when I will have more time to be dedicated for the project.

PR #1760, opened 2021-11-21
Add `type_for_format` macro

Rua:
Changelog:
```markdown
- Added a `type_for_format` macro, which returns a Rust type that is suitable for representing a given `Format`.
```

A little utility macro for the convenience of users. It may come in handy for solving #1667 too.

### DIFF
--- a/vulkano/autogen/formats.rs
+++ b/vulkano/autogen/formats.rs
@@ -38,6 +38,7 @@ struct FormatMember {
     components: [u8; 4],
     compression: Option<Ident>,
     planes: Vec<Ident>,
+    rust_type: Option<TokenStream>,
     size: Option<u64>,
     type_color: Option<Ident>,
     type_depth: Option<Ident>,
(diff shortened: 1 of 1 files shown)


Eliah-Lakhin:
@Rua Is it possible to add a use example to some of our existing examples?

Rua:
I can't think of any at the moment, I think everything uses the `impl_vertex` macro?

Eliah-Lakhin:
@Rua For example the Mandelbrot fractal uses a buffer with colors: https://github.com/vulkano-rs/vulkano/blob/aa5a97481ebd17510a1317d7f3d6472715e1f453/examples/src/bin/dynamic-local-size.rs#L192 Maybe use it here?

I just want to keep some connections between newly introduced features and the examples or tests in our repo.

Rua:
I think the main use of this macro would be if the user wants to define a custom `struct` that can hold vertex data of a particular format. They can then do something like this:

```rust
struct Vertex {
    position: type_for_format!(R8G8B8_UNORM),
    tex_coords: type_for_format!(R8G8_UNORM),
}
```

I suppose this could be added to examples wherever a vertex struct is defined, but I think that would only be temporary, as the way vertex formats are defined needs to change anyway due to the aforementioned bug #1667.

Eliah-Lakhin:
Understood. So, basically this is temporary helper for a kind of debugging purposes.

Eliah-Lakhin:
@Rua Merged. :+1:

PR #1763, opened 2021-11-24
Use `General` layout for `storage_image` regardless of image type

### LINKED ISSUE #1762: Validation errors for interactive_fractal example
When I run the example through `vkconfig` with validation enabled, I get a series of validation errors like this one:
> VUID-VkWriteDescriptorSet-descriptorType-04152(ERROR / SPEC): msgNum: -1536361367 - Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04152 ] Object 0: handle = 0xa2eb680000000026, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0xa46cfc69 | vkUpdateDescriptorSets() pDescriptorWrites[0] failed write update validation for VkDescriptorSet 0xa2eb680000000026[] with error: Write update to VkDescriptorSet 0xa2eb680000000026[] allocated with VkDescriptorSetLayout 0x84c0580000000017[] binding #0 failed with error message: Attempted write update to image descriptor failed due to: Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_STORAGE_IMAGE is being updated with invalid imageLayout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL for image VkImage 0xc4f3070000000021[] in imageView VkImageView 0x301e6c0000000022[]. Allowed layouts are: VK_IMAGE_LAYOUT_GENERAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_STORAGE_IMAGE the imageLayout member of each element of pImageInfo must be a member of the list given in Storage Image (https://vulkan.lunarg.com/doc/view/1.2.198.0/linux/1.2-extensions/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04152)
    Objects: 1
        [0] 0xa2eb680000000026, type: 23, name: NULL

When updating a descriptor set, Vulkano gets the layout in [this line](https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/descriptor_set/sys.rs#L302). `descriptor_type` here is `StorageImage` while the image view wraps around an `AttachmentImage`. `AttachmentImage` has this implementation for `descriptor_layouts`:
```rust
    #[inline]
    fn descriptor_layouts(&self) -> Option<ImageDescriptorLayouts> {
        Some(ImageDescriptorLayouts {
            storage_image: ImageLayout::ShaderReadOnlyOptimal,
            combined_image_sampler: ImageLayout::ShaderReadOnlyOptimal,
            sampled_image: ImageLayout::ShaderReadOnlyOptimal,
            input_attachment: ImageLayout::ShaderReadOnlyOptimal,
        })
    }
```
So it's using the `ShaderReadOnlyOptimal` layout when writing the descriptor set, which is incorrect in this case. The problem could maybe be solved by using Vulkano's `StorageImage` type instead of `AttachmentImage`, but I think that errors like this shouldn't happen in safe code regardless.

Rua:
Changelog:
```markdown
- Fixed bug where the wrong image layout was used for images used as a storage image.
```

Fixes #1762. According to the spec, only `General` can be used for a storage image (and `VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR` but Vulkan doesn't support that yet anyway). I added some better documentation to the `ImageLayout` type as well.

### DIFF
--- a/vulkano/src/image/attachment.rs
+++ b/vulkano/src/image/attachment.rs
@@ -649,7 +649,7 @@ where
     #[inline]
     fn descriptor_layouts(&self) -> Option<ImageDescriptorLayouts> {
         Some(ImageDescriptorLayouts {
-            storage_image: ImageLayout::ShaderReadOnlyOptimal,
+            storage_image: ImageLayout::General,
             combined_image_sampler: ImageLayout::ShaderReadOnlyOptimal,
             sampled_image: ImageLayout::ShaderReadOnlyOptimal,
             input_attachment: ImageLayout::ShaderReadOnlyOptimal,
--- a/vulkano/src/image/immutable.rs
+++ b/vulkano/src/image/immutable.rs
@@ -447,7 +447,7 @@ where
     #[inline]
     fn descriptor_layouts(&self) -> Option<ImageDescriptorLayouts> {
         Some(ImageDescriptorLayouts {
-            storage_image: self.layout,
+            storage_image: ImageLayout::General,
             combined_image_sampler: self.layout,
             sampled_image: self.layout,
             input_attachment: self.layout,
--- a/vulkano/src/image/layout.rs
+++ b/vulkano/src/image/layout.rs
@@ -9,35 +9,72 @@
 
 use crate::descriptor_set::layout::DescriptorType;
 
-/// Layout of an image.
+/// In-memory layout of the pixel data of an image.
 ///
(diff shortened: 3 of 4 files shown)


Eliah-Lakhin:
@Rua Merged. Thank you for the fix, and also for improving the documentation too. It makes more clear the meaning of these values, which is good. :+1:

PR #1768, opened 2021-11-25
Fix gl-interop validation error & window not closing

### LINKED ISSUE #1764: Validation error for gl-interop example
@fayalalebrun I get this error:

> VUID-vkBindBufferMemory-bufferDeviceAddress-03339(ERROR / SPEC): msgNum: 1151829889 - Validation Error: [ VUID-vkBindBufferMemory-bufferDeviceAddress-03339 ] Object 0: handle = 0xcb3ee80000000007, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x44a78781 | vkBindBufferMemory(): If buffer was created with the VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT bit set, memory must have been allocated with the VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT bit set. The Vulkan spec states: If the VkPhysicalDeviceBufferDeviceAddressFeatures::bufferDeviceAddress feature is enabled and buffer was created with the VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT bit set, memory must have been allocated with the VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT bit set (https://vulkan.lunarg.com/doc/view/1.2.198.0/linux/1.2-extensions/vkspec.html#VUID-vkBindBufferMemory-bufferDeviceAddress-03339)

This may be an oversight or a missing feature.

Also a bug, but possibly unrelated: The example doesn't close when I click the X on the window, although the rendered image still changes so it's not frozen.

fayalalebrun:
Fixes #1764 

No changes to the underlying library have been made. Thus no changelog items to report.

The validation error was caused due to activating all supported device features. Specifically feature `buffer_device_address`. This means this feature is not properly supported and for it to function correctly support will have to be added to the memory allocator. This is out of the scope of this issue.

I'm not sure what exactly caused the window not closing, but changing the structure of the event loop to more closely mirror the `triangle` example fixed the problem.

### DIFF
--- a/examples/src/bin/gl-interop.rs
+++ b/examples/src/bin/gl-interop.rs
@@ -184,112 +184,6 @@ fn main() {
     let mut previous_frame_end: Option<Box<dyn GpuFuture>> = Some(Box::new(now(device.clone())));
 
     event_loop.run(move |event, _, control_flow| {
-        unsafe {
-            let mut builder = SubmitCommandBufferBuilder::new();
-            builder.add_signal_semaphore(&acquire_sem);
-            builder.submit(&queue).unwrap();
-        };
-
-        barrier.wait();
-        barrier_2.wait();
-
-        unsafe {
-            let mut builder = SubmitCommandBufferBuilder::new();
-            builder.add_wait_semaphore(
-                &release_sem,
-                PipelineStages {
-                    all_commands: true,
-                    ..PipelineStages::none()
-                },
-            );
-            builder.submit(&queue).unwrap();
-        };
-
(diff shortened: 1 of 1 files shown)


Rua:
I would say it's not a good idea to enable features you don't need. Features don't just enable new behaviours, but they can change existing ones as well. As you already discovered, enabled features can add extra requirements and disallow certain other things.

Eliah-Lakhin:
@fayalalebrun Merged. Thanks for the fix, and I apologize for the late review.

PR #1772, opened 2021-12-05
Upgrade to Rust 2021, update dependencies

Rua:
```markdown
- **Breaking** Updated to half 1.8, smallvec 1.7, winit 0.26, raw-window-handle 0.4.
```

```markdown
- Upgraded to Rust 2021.
```

Rust 2021 is a much smaller change than 2018 was, but it's still nice to have, especially for the ability to call `into_iter` on arrays properly now. This does mean that Vulkano 0.28 will require Rust 1.56 to compile. Because of how editions work, Vulkano can still be included in *other* projects using the 2018 or 2015 editions, compatibility is not broken.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "examples"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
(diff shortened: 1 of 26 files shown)


Eliah-Lakhin:
@Rua Looks good. No regressions detected. Merged :+1:

Issue #1743, opened 2021-11-02
How could Vulkano safely support PARTIALLY_BOUND descriptor bindings?


Rua:
The Vulkan spec says:
> `VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT` indicates that descriptors in this binding that are not *dynamically used* need not contain valid descriptors at the time the descriptors are consumed. A descriptor is dynamically used if any shader invocation executes an instruction that performs any memory access using the descriptor.

This means that descriptor sets can have "holes" in them, with some bindings being valid and some not. But note the language in the spec, it says "dynamically used". This means that descriptor set bindings can be statically declared inside the shader, but if the shader doesn't actually access a particular item when it's run, it doesn't have to contain a valid descriptor. It is impossible to predict which items a shader will access without executing the entire shader in advance somehow, which is completely infeasible for Vulkano.

This means that Vulkano cannot rely on declarations in the shader *at all* when checking the validity of resources bound to descriptors. It has no idea which items will be accessed, and therefore which must be valid at draw/dispatch time. This makes it pretty much impossible to implement this safely with proper validation. The official validation layer doesn't bother validating it either; if the binding has this flag is just skips validating the resources for the entire binding.

So, how can Vulkano support this flag? It's obviously very useful, but we have to accept that it's going to be `unsafe` somehow. Is there a way that we can minimise the unsafety?

DemiMarie:
> So, how can Vulkano support this flag? It's obviously very useful, but we have to accept that it's going to be `unsafe` somehow. Is there a way that the API can minimise the unsafety?

Could static analysis of the shader source be useful here?

Rua:
I'm not sure if that could eliminate all unsafety. Imagine that the shader has some complicated logic to figure out the index of an array element in a binding, that uses a variety of input variables. Inputs could be from vertex buffers, uniform buffers and even images. For Vulkano to figure out what index the shader will end up accessing, and therefore which descriptor needs to be valid at draw/dispatch time, it would have to know all these inputs *and* execute the code that calculates the final index from them. While this isn't completely impossible, it would be extremely impractical to attempt this in Vulkano, not only because of the complexity involved, but also because of the performance costs it would add to each draw/dispatch call.

DemiMarie:
Is there a reason to support this?

DemiMarie:
Could this be made safe by generating the Rust code and the shaders from the same source, so that the shaders are known to not use the descriptors that are not bound?

PR #1774, opened 2021-12-06
Remove `DescriptorSetBuilder`, use descriptor writes directly

Rua:
Changelog:
```markdown
- **Breaking** `DescriptorWrite` is renamed to `WriteDescriptorSet` to match Vulkan, and moved into the main `descriptor_set` module.
- **Breaking** The constructors of `WriteDescriptorSet` are no longer unsafe. Added convenience constructors for writing non-arrayed bindings.
- **Breaking** Descriptor sets are no longer constructed using a builder, but with an iterator of `WriteDescriptorSet` items. See in the examples how this is used.
  - `PersistentDescriptorSet::start` has been replaced with three constructors, `new`, `new_variable` and `new_with_pool`. Each of these takes an iterator of `WriteDescriptorSet`.
  - `SingleLayoutDescSetPool::next` takes an iterator of `WriteDescriptorSet` as well.
```

This PR changes how descriptor sets are constructed, which will probably break a lot of existing code. But it allows for some things that were not possible before:
- It is possible to leave some descriptors unset, if the shader does not use them. This is especially important if `PARTIALLY_BOUND` support is added later.
- `push_descriptor_set` can be applied incrementally, like it was intended by the spec.
- It makes it easier to add support for updating existing descriptor sets (after constructing/using them), which allows more efficient usage and is also needed for `UPDATE_AFTER_BIND`. For this to work, descriptor sets need some kind of locking mechanism to prevent updates when the descriptor set is in use.

I realised that `PersistentDescriptorSet` and `SingleLayoutDescSet` have a lot of code in common, so I created a new `DescriptorSetInner` type to handle the common behaviour. This type is not public, it's only an implementation detail.

After this PR, I will be investigating whether I can add support for the `VK_KHR_descriptor_update_template` extension, which was promoted to Vulkan 1.1. This extension introduces an alternative way to update descriptor sets that reads the data directly from a buffer on the host (CPU), which can be more efficient than constructing `WriteDescriptorSet` objects.


Eliah-Lakhin:
> The constructors of `WriteDescriptorSet` are no longer unsafe

The main threat of this change is that if the user skips some bindings in use in the shader's code, that would lead to UB, as far as I understand.

Rua:
It is legal to leave some descriptors unset, if the shader doesn't use them. And if the shader does use them, then the changes made to `check_descriptor_sets_validity` in #1744 should ensure that this error is caught.

Eliah-Lakhin:
@Rua Ok, then I have an opposite question. If the shader has some explicit binding in layout, and this binding is not in use dynamically in the shader's code, and we have a descriptor set with an empty record for this binding, does this situation considered as valid by the "check_descriptor_sets_validity" procedure?

Rua:
This would not be considered valid, so an error should occur. The check uses the `DescriptorRequirements` information that is generated from the shader (by Vulkano-shaders or at runtime). This information is based on static analysis of the shader code, where any *potential* usage means that the binding is required. It's is more or less the same as what the spec says on the subject:
> Descriptors in each bound descriptor set, specified via `vkCmdBindDescriptorSets`, **must** be valid if they are statically used by the `VkPipeline` bound to the pipeline bind point used by this command

The spec does allow a descriptor to remain unset if the descriptor is statically used (the shader accesses the binding, or explicitly declares it in its interface) but is not dynamically used (because the inputs cause the descriptor to not be accessed). But this is only allowed when the binding is created with the `PARTIALLY_BOUND` flag set, which Vulkano doesn't support yet (see #1743).

vulkano/src/descriptor_set/persistent.rs
```diff
@@ -111,137 +166,44 @@ where
     }
 }
 
-/// Prototype of a `PersistentDescriptorSet`.
-pub struct PersistentDescriptorSetBuilder {
-    inner: DescriptorSetBuilder,
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PersistentDescriptorSetCreationError {
```
Eliah-Lakhin, line 170: @Rua This error describes exactly the same thing as a `SingleLayoutDescSetCreationError`. Maybe join them together?

Rua: Ok, done.

Eliah-Lakhin:
@Rua ok, I almost finished testing, but I have one(hopefully) last thing to clarify.

If a have a binding with variable descriptors in my layout("runtime array"), and I set variable descriptors count during the pipeline creation to let say 2, and then I'm going to fill the array with only 1 descriptor. Also, during the pipeline creation I also specified two immutable samplers for this array.

What write operations should I preform for this binding? This configuration works on my end when I do a single `image_view_array` write only. But I was thinking that I also has to fill array's tail with the `none_array` writes, isn't it?

Rua:
If you only have 1 descriptor, then you don't need `none_array` for the second one. `none_array` is only needed for descriptors that are actually going to be used. Also, it should be invalid to give `none_array` to the same binding as `image_view_array`; the writes for each binding must be of the same (correct) type.

Eliah-Lakhin:
> So if you try this, there should be an error.

Confirmed. That's what I got. Thanks for clarifications!

Eliah-Lakhin:
@Rua PR merged. Excellent work! :rocket:

PR #1778, opened 2021-12-08
Always generate shader requirements for the exact entrypoint interface

Rua:
```markdown
- **Breaking** Shader reflection now always generates only the descriptor requirements needed for each given entry point, instead of for all of them. The `exact_entrypoint_interface` argument to the `shader!` macro is removed.
```

@Arc-blroth This effectively undoes #1559. This could be controversial, and I realise I was in favour of the original change back in April, but I have a few reasons for making this change:
1. Since #1729 and #1744, the information that is gathered from shaders is only what is *required* as a minimum by the shader. It no longer generates a full-blown descriptor set layout like it did before. In this light, I think it makes more sense that the requirements reflect the entry point chosen by the user, instead of every possible entry point.
2. As noted at https://github.com/vulkano-rs/vulkano/issues/1556#issuecomment-821658405, different entry points can have conflicting requirements. For example, one entry point could want a buffer on binding 0, another could have a texture on the same binding.
3. I want to improve checking whether a storage descriptor is mutable or not, by analysing whether a given entry point performs any write operations on it. This is what the official validation layer does, and it would remove the need for users to put `readonly` in their shaders. This check makes more sense on a per-entrypoint basis.
4. I would like to do more validation on shader requirements than currently, with additional fields on `DescriptorRequirements`. The official validation layer checks for atomic operations on images, which require `VK_FORMAT_FEATURE_STORAGE_IMAGE_ATOMIC_BIT`. It also makes a list of which images are used with which samplers, which is needed for checking if the image and sampler are compatible. This is needed for supporting the `VK_KHR_sampler_ycbcr_conversion` extension, which I'd like to implement at some point. This extra validation only makes sense on a per-entrypoint basis too.

Both `ComputePipeline` and `GraphicsPipelineBuilder` have ways to construct a pipeline with a user-provided pipeline layout, so the user already has the ability to make a layout with more sets and bindings than the minimum needed by the shader entry point. When using raw Vulkan, this is what the user would always have to do; the ability to create a minimal layout automatically based on the requirements of the shader is purely a convenience to the user.

There is, in fact, an advantage to creating an explicit pipeline layout that is compatible with multiple shaders: it allows descriptor sets to remain bound when switching between pipelines, as long as the shaders have compatible interfaces. So this may be something Vulkano should actually encourage the user to do (in documentation), and make it more explicit that creating a new layout based on the minimum requirements is only a convenience that's useful for certain situations, not generally.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -210,7 +210,6 @@ pub(super) fn reflect<'a, I>(
     words: &[u32],
     types_meta: &TypesMeta,
     input_paths: I,
-    exact_entrypoint_interface: bool,
     shared_constants: bool,
     types_registry: &'a mut HashMap<String, RegisteredType>,
 ) -> Result<(TokenStream, TokenStream), Error>
@@ -244,7 +243,7 @@ where
         quote! { &Capability::#name }
     });
     let spirv_extensions = reflect::spirv_extensions(&spirv);
-    let entry_points = reflect::entry_points(&spirv, exact_entrypoint_interface)
+    let entry_points = reflect::entry_points(&spirv)
         .map(|(name, model, info)| entry_point::write_entry_point(&name, model, &info));
 
     let specialization_constants = structs::write_specialization_constants(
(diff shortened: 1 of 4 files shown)


Eliah-Lakhin:
@Rua I feel like we will have complaints from the users regarding this feature again :) But let's move it forward. The motivation for revert is reasonable. :+1:

PR #1779, opened 2021-12-11
Added command `clear_attachment`

### LINKED ISSUE #1749: Image Clear Rect
Hello,

I would like to be able to clear rect of an image, currently we can't. I'm not sure if vulkan itself support this, I found command `vkCmdClearAttachments`, which seem to be commented code for it in:

https://github.com/vulkano-rs/vulkano/blob/eb7e0c9ce06f52cd1877e5274a93216bd910475f/vulkano/src/command_buffer/sys.rs#L689-L710

Will this function enable us to clear rect? Or is there any other way to achieve that.

Thanks

Amjad50:
fixes #1749

This command implements `vkClearAttachments`, most checks are done
except for:
- Handling attachments with `VK_ATTACHMENT_UNUSED`, `vulkano` doesn't
  support unused attachments for now.
- command buffer protection and protectedNoFault, which `vulkano` is not
  supporting now.

       -  Added `ClearRect` struct, which contains the region information to be cleared
       - Added `ClearAttachment` enum, which will contain the `ClearValue` and type of attachment
       - Implemented `AutoCommandBufferBuilder::clear_attachments`, which implements all possible checks from the vulkan standards to make it safe.
       - Implemented `SyncCommandBufferBuilder::clear_attachments` and `UnsafeCommandBufferBuilder::clear_attachments`.

       - Implemented `clear_attachments` command

    Not sure how to test this, `draw` is not tested as well, suggestions are welcome on the procedure of testing these kind of functions.

### DIFF
--- a/examples/src/bin/clear_attachments.rs
+++ b/examples/src/bin/clear_attachments.rs
@@ -242,7 +242,7 @@ fn main() {
                         },
                         // Relative offset and extent
                         ClearRect {
-                            rect_offset: [width as i32 / 2, height as i32 / 2],
+                            rect_offset: [width / 2, height / 2],
                             rect_extent: [width / 3, height / 5],
                             base_array_layer: 0,
                             layer_count: 1,
--- a/examples/src/bin/clear_attachments.rs
+++ b/examples/src/bin/clear_attachments.rs
@@ -1,0 +1,300 @@
+// Copyright (c) 2016 The vulkano developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or https://opensource.org/licenses/MIT>,
+// at your option. All files in the project carrying such
+// notice may not be copied, modified, or distributed except
+// according to those terms.
+
+use std::sync::Arc;
+use vulkano::command_buffer::{AutoCommandBufferBuilder, CommandBufferUsage, SubpassContents};
+use vulkano::device::physical::{PhysicalDevice, PhysicalDeviceType};
+use vulkano::device::{Device, DeviceExtensions, Features};
+use vulkano::format::ClearValue;
+use vulkano::image::attachment::{ClearAttachment, ClearRect};
+use vulkano::image::{view::ImageView, ImageUsage, SwapchainImage};
+use vulkano::instance::Instance;
+use vulkano::pipeline::graphics::viewport::ViewportState;
+use vulkano::pipeline::GraphicsPipeline;
+use vulkano::render_pass::{Framebuffer, RenderPass, Subpass};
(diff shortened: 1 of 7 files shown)


Rua:
It looks good at first glance, thank you for the work!

Eliah-Lakhin:
@Amjad50 Thank you for the improvement! It looks good to me too. Is it possible to add an example that will use this feature? Or maybe add it to some of the existing examples?

examples/src/bin/clear_attachments.rs
```diff
@@ -0,0 +1,301 @@
...
+    			#version 450
+
+    			layout(location = 0) out vec4 f_color;
+
+    			void main() {
+    				f_color = vec4(1.0, 0.0, 0.0, 1.0);
+    			}
+    		"
+        }
+    }
```
Amjad50, line 122: In this example actually we are not using the shaders, but not sure how to remove them

vulkano/src/command_buffer/auto.rs
```diff
@@ -881,6 +893,92 @@ impl<L, P> AutoCommandBufferBuilder<L, P> {
...
+            }
+            if multiview && (rect.base_array_layer != 0 || rect.layer_count != 1) {
+                return Err(ClearAttachmentsError::InvalidMultiviewLayerRange);
+            }
+            // TODO: handle checking `rect.rect_offset` (the start of the rect)
+            if rect.rect_offset[0] + rect.rect_extent[0] as i32 > dimensions[0] as i32
+                || rect.rect_offset[1] + rect.rect_extent[1] as i32 > dimensions[1] as i32
+            {
+                return Err(ClearAttachmentsError::RectOutOfBounds);
+            }
```
Amjad50, line 963: In here, I did some checks for the place of the rect. For now, if the offset is negative it won't return an error.
I put these checks because of this clause in the docs:
```
VUID-vkCmdClearAttachments-pRects-00016
The rectangular region specified by each element of pRects must be contained within the render area of the current render pass instance
```
In the example, if you resize the window to a very small size, it will crash, since we have one of the rects with constant size [100, 100]. I tried removing the error, and it works without problem.

I'm not sure if it will corrupt some VRAM memory if the rect bounds are bigger than the render area. What do you think?

Rua: If the Vulkan spec says that the rectangle needs to fit into the render area, then there should be a check in Vulkano that enforces that. Not doing so is undefined behaviour, so it could lead to an out of bounds access on the GPU, or something worse. Since it's undefined, the driver is allowed to do anything.

Amjad50: Ok, I updated the `ClearRect` to use `u32` instead of `i32`, since in this case we shouldn't get negative values.

Amjad50:
I was enforcing x11, was having some issues with wayland. you can ignore this.

Eliah-Lakhin:
@Amjad50 Looks good, merged. Thanks again! :+1:

PR #1782, opened 2021-12-19
Allow `copy_image`, `blit_image` and `copy_buffer` source and destination to be the same image

Amjad50:
This PR added a fix to allow `copy_image`, `blit_image` and `copy_buffer` to work with `src` and `dest` images/beffers being the same.

The doc: `srcImage and dstImage can be the same image or alias the same memory`.

This was achieved by locking only one image exclusively if both are the same. And in order for this to work safely, we need to add checks for overlapping regions, since regions cannot overlap in memory in any of the above mentioned commands.

```txt
- Allowed `copy_image`, `blit_image` and `copy_buffer` to work with `src` and `dest` images/beffers being the same.
But, regions from src to dest must not overlap.
```

I also made an example, a copy from `image` with some edits


Rua:
Nice fix! Can you make this work for buffer self-copies too? And image blits maybe?

Amjad50:
Yes sure

Amjad50:
I noticed in the docs, that in these commands, while its allowed to have the same buffer for `src` and `dest`, regions must not overlap. I'll work on adding some checks to make sure the regions are not overlapping.

Amjad50:
Now the implementation is complete, and is safe. I'll edit the example to include the other functions to showcase them.

Amjad50:
I think this is now ready for review.

Rua:
Not related to your PR directly, but I wonder if there's a neat way to demonstrate things without making a full Rust program for each one.

Amjad50:
Yes, I was thinking of similar thing. An idea is to make a nice flow for unit tests, we can create a mock device and try to replicate some gpu functionality on the cpu. if we got this working or something similar, we can do better unit tests (to handle all vulkan documentation) and we can use doctests easily.

Eliah-Lakhin:
@Amjad50 Thank you very much for your work! The code looks good and all examples work as expected. No regressions detected. PR merged :+1:

PR #1783, opened 2021-12-20
Fix undefined behaviour on external semaphore & image creation; Fix wrong device chosen on hybrid graphics

fayalalebrun:
Currently, whenever an exportable image or semaphore are created, the p_next structure used to specify the exportable handle types falls out of scope prematurely, which means segmentation faults can occur in certain situations, most likely when the stack memory is overwritten or is invalidated. 

In order to fix this, I've not only made sure to define these p_next structures in the appropriate scope, but I've also changed the implementations to use ash builders. This way, this issue would have been impossible to begin with, and two unsafe calls are removed.

@pwaller Please confirm this fixes the issues you encountered.

Using new functionality recently added in https://github.com/glium/glium/pull/1977, there is now UUID checking in the gl-interop example during physical device creation. This is useful since previously, if a computer had more than one graphics card or drivers, it was possible for Vulkan to choose a different device and/or driver than the GL context it is sharing resources with. This would result in undefined behaviour.

        - Fix undefined behaviour on external semaphore creation
        - Fix undefined behaviour on external image creation
        - Add uuid checking to gl-interop example

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -24,4 +24,4 @@ serde = { version = "1.0", features = ["derive"] }
 ron = "0.7"
 rand = "0.8.4"
 
-glium = { git = "https://github.com/glium/glium.git" }
+glium = { git = "https://github.com/glium/glium.git", rev = "60889a2" }
--- a/examples/src/bin/gl-interop.rs
+++ b/examples/src/bin/gl-interop.rs
@@ -57,6 +57,15 @@ fn main() {
         .build_surfaceless(&event_loop_gl)
         .unwrap();
 
+    let hrb_vk = glutin::ContextBuilder::new()
+        .with_gl_debug_flag(true)
+        .with_gl(glutin::GlRequest::Latest)
+        .build_surfaceless(&event_loop_gl)
+        .unwrap();
+
+    let display =
+        glium::HeadlessRenderer::with_debug(hrb_vk, glium::debug::DebugCallbackBehavior::PrintAll)
+            .unwrap(); // Used for checking device and driver UUIDs
     let (
         device,
         _instance,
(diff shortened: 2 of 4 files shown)


Rua:
What was the original issue?

fayalalebrun:
> What was the original issue?

Both were reported in the PRs that introduced them, #1606 and #1734, not properly as issues.

pwaller:
I reported the issue since I spotted it as a drive-by while reading the code, and I hit a crash while trying to use the external memory interop. I've since moved on to a completely different approach which doesn't use this, but I thought you would like to know since I spotted it in the code.

Eliah-Lakhin:
@fayalalebrun `gl-interop` example does not compile with the latest glium. Can you check the code, please?

fayalalebrun:
@Eliah-Lakhin The issue is that when you merged my last merge request, you removed the "rev" specification on the glium dependency, which I did not add back. Thus Cargo was content to use whichever version of glium's master branch you had cached.

I've now specified the correct commit in the Cargo.toml.

Eliah-Lakhin:
@fayalalebrun ok, but I thought the changes already merged to `master`(in https://github.com/glium/glium/pull/1977), don't they?

fayalalebrun:
@Eliah-Lakhin Yes they are, the rev points to the newest commit in Master. You can easily check, it is "60889a2". The problem is that Cargo doesn't check if you there have been updates to the master branch every time you build. The first time you build, the commit you pulled from is stored in Cargo.lock and used thereafter.

If you delete the lock file, and build again, then the newest version of the branch will be pulled and it will build correctly even without the commit specified.

Eliah-Lakhin:
@fayalalebrun ok, thank you for clarification! I totally forgot about the Cargo.lock. :)

All in all everything looks good. PR merged. Excellent work! :+1:

PR #1786, opened 2021-12-29
Improve shader analysis, add and refine DescriptorRequirements fields

Rua:
Changelog:
```markdown
- Added additional fields to `DescriptorRequirements`, and changed `mutable` to be on a per-descriptor basis.
```

I've changed the analysis of shader entry points so that it tracks the usage of certain SPIR-V instructions in `DescriptorRequirements`. Specifically, it now includes information about samplers that are used with specific sampling instructions, storage images that are used with atomic instructions, and which images are sampled by each sampler. This is needed for additional validation checks required by the spec that Vulkano doesn't currently implement. These checks are not included in this PR, but it at least makes it possible to implement them.

Additionally, the `mutable` field has been changed so that it applies for each index individually for multi-descriptor bindings. This means that different buffers or images in an arrayed binding can have different mutability. Furthermore, the mutability is determined based on whether any instructions write to the variable now, instead of whether it has the `readonly` qualifier. Consequently this qualifier is probably no longer needed (although it might help the driver optimise).

There are a few shortcomings still that are more difficult to solve:
- For arrayed/multi-descriptor bindings, the reflection code can only determine the descriptor index being accessed if the index is a constant within the SPIR-V code. Indices could be non-constant (with e.g. the `shader_sampled_image_array_dynamic_indexing` feature), in which case the reflection code can't figure it out.
- I'm unsure if atomic operations on storage buffer data count as "mutable" or not for the purposes of synchronisation. For now, I've left them as not mutable.


Eliah-Lakhin:
@Rua Looks good, I didn't detect any regressions. PR merged.

For the outlined shortcomings. Not sure how they could be solve in general, as the deeper code analysis leads to the code interpretation(which is, I think, GPU driver does at some extent during SPIR-V loading). Maybe a simple constant-analysis would be enough for us at this stage.

PR #1787, opened 2021-12-31
Sampler improvements

Rua:
```markdown
- **Breaking** `Sampler` is now constructed using a builder, using `start` instead of `new`. The shortcut functions `simple_repeat_linear` and `simple_repeat_linear_no_mipmap` are retained, but now return `Result`.
- **Breaking** `MipmapMode` is renamed to `SamplerMipmapMode` to match Vulkan.
- **Breaking** `SamplerAddressMode::ClampToBorder` no longer includes a `BorderColor` value. This is now provided with the new builder.
- **Breaking** The `ImageViewAbstract::can_be_sampled` method is removed, and replaced with `Sampler::can_sample`.
```

```markdown
- Added `Filter::Cubic`.
- Added support for filter reduction modes on samplers.
```

The main change here is that samplers have a builder now. `Sampler::new` previously had a lot of parameters, and it wasn't very easy to remember which parameters were for what. With a builder, the user's code is more self-documenting. A builder also helps with providing defaults and convenience functions for setting multiple values at once.

I added support for two extra sampler-related features. There's more to add, but these were simple enough to include in this PR without much thought.


Eliah-Lakhin:
@Rua Looks good. Merged :+1:

PR #1791, opened 2022-01-05
Fixed failing `basic_conflict` test and added replacement for it

### LINKED ISSUE #1789: Self-copy tests failing
@Amjad50 The test `command_buffer::synced::tests::basic_conflict` is now failing, because a self-copy is no longer triggering an error. That is of course by design, but it means that the test needs to be adjusted a bit. I suggests two tests, `self_copy_overlapping` and `self_copy_not_overlapping`, which should test respectively that the copy fails and succeeds.

Amjad50:
Fixes #1789

When we implemented self_copy for images and buffers (#1782),
`basic_conflict` test failed. So we removed it and added tests for
`overlapping` and `non_overlapping` ranges checks when using buffer self
copy.

* Entries for Vulkano changelog (not sure if tests are added in changelogs):
    - Added `buffer_self_copy_overlapping` and `buffer_self_copy_not_overlapping` as unit tests for self_copy feature (#1782).
    - Removed test `basic_conflict` (failed after #1782).

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -4178,9 +4178,11 @@ impl fmt::Display for AutoCommandBufferBuilderContextError {
 
 #[cfg(test)]
 mod tests {
+    use super::CopyBufferError;
     use crate::buffer::BufferUsage;
     use crate::buffer::CpuAccessibleBuffer;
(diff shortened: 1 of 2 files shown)


Amjad50:
Btw @Rua, There seem to be old duplicates of synced buffers tests here `vulkano/src/command_buffer/synced/tests.rs`, the tests are now in `mod.rs` so you can remove it.

Issue #1572, opened 2021-05-06
Allow the user to implement `VertexMember` for custom types


Rua:
The `VertexMember` trait is used for the `impl_vertex` macro, and is currently implemented by Vulkano for a bunch of array and tuple types. But there are several libraries out there that have custom types for vectors. For example, I use Nalgebra in my project, and its `Vector3<T>` type has the exact same layout as `[T; 3]`, so in theory this would be valid as a drop-in replacement. But because of Rust's orphan rules, the user is not allowed to implement Vulkano's `VertexMember` on Nalgebra's `Vector3`.

Is there some way around this, so that users can use types like Nalgebra's in their vertex types?

It looks like #1133 and #1326 tried to implement this or something similar, but were ultimately closed. Unfortunate. What would be needed to make such changes acceptable?

Eliah-Lakhin:
Adding another dependency(even optional) would complicate maintainability.

As an alternative we might consider looking at some interop basis libraries such as [Mint](https://docs.rs/mint/0.5.6/mint/index.html) for example. But this is controversial too.

PR #1792, opened 2022-01-08
Implement VertexMember for nalgebra Vectors and Points

TJYSunset:

Implemented `VertexMember` for `nalgebra::Vector*<T>` and `nalgebra::Point*<T>` where * ∈ [1, 4]. 

Tested to be also working with `nalgebra_glm::Vec*`, since they are aliasing to the same types as `nalgebra::Vector*<f32>`.

64-bit floats won't work, but vulkano [doesn't](https://github.com/vulkano-rs/vulkano/blob/0ebf05f35d44c9742654cb6905abc69dbce619cb/vulkano/src/shader/mod.rs#L860) [seem to](https://github.com/vulkano-rs/vulkano/blob/0ebf05f35d44c9742654cb6905abc69dbce619cb/vulkano/src/shader/mod.rs#L887) support those anyway. (?)

Related: #1572, #1133, #1326

        - `Entry 1.`
        - `Entry 2.`...

- Added support for nalgebra `Vector`s and `Point`s in vertex definitions.

Sorry, but I couldn't find the right place for tests.

If anyone could point that out, I'll add them.

I have to apologize again, but I don't know where are the docs located either.

### DIFF
--- a/vulkano/Cargo.toml
+++ b/vulkano/Cargo.toml
@@ -21,6 +21,7 @@ crossbeam-queue = "0.3"
 fnv = "1.0"
 half = "1.8"
 lazy_static = "1.4"
+nalgebra = { version = "0.30.0", optional = true }
 parking_lot = { version = "0.11", features = ["send_guard"] }
 shared_library = "0.1"
 smallvec = "1.7"
--- a/vulkano/src/pipeline/graphics/vertex_input/impl_vertex.rs
+++ b/vulkano/src/pipeline/graphics/vertex_input/impl_vertex.rs
@@ -164,6 +164,100 @@ where
     }
 }
 
+#[cfg(feature = "nalgebra")]
+unsafe impl<T> VertexMember for nalgebra::Vector1<T>
+where
+    T: VertexMember,
+{
+    #[inline]
+    fn format() -> (VertexMemberTy, usize) {
+        <T as VertexMember>::format()
+    }
+}
+
+#[cfg(feature = "nalgebra")]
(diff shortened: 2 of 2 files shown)


Rua:
In principle this PR is ok I think, although it does introduce a dependency on Nalgebra (something Vulkano was hesitant about in the past). But I have been looking at possible ways to eliminate the `VertexMember` trait altogether, or at least restructure it. When I come up with a solution, that may end up making this PR obsolete.

TJYSunset:
Is there an estimated date of when the refactor will happen? Not trying to be rude, but I'm kind of tired waiting for upstreams to fix their issues... (like, after all these years, Intellij Rust still sometimes autocompletes random crap)

Rua:
I have no timeline for it yet, that's why I think this PR is fine as it is. Just making you aware that it could end up being removed at some point.

Eliah-Lakhin:
@AustinJ235 I would like to left it at your decision to proceed or not to proceed with this PR. As far as I understand the earlier decisions were keeping Vulkano to be agnostic from any specific linear algebra lib.

AustinJ235:
I believe this feature was considered for merge before and I approved it then, but it never got updated. Is this trait something that can be implemented externally or does it need to be implemented within the crate? If it does need to be implemented inside I think it should behind an optional depenancy.

Eliah-Lakhin:
As far as I understand at this moment the only way to do it is by implementing the trait inside this crate. Since this feature is optional I think it's fine to proceed with merging.

@TJYSunset Thank you for implementation, and I apologize for long review. PR merged. :+1:

PR #1793, opened 2022-01-08
Update dependencies

Rua:
Replace this line:
```markdown
- **Breaking** Updated to half 1.8, smallvec 1.7, winit 0.26, raw-window-handle 0.4.
```
with this:
```markdown
- **Breaking** Updated to ash 0.35, half 1.8, smallvec 1.7, winit 0.26, raw-window-handle 0.4.
```
And add:
```markdown
- **Breaking** `ColorSpace::DciP3Linear` is renamed to `DisplayP3Linear` to match an equivalent change in the Vulkan naming.
- Updated vk.xml to 1.2.203.
```

Just keeping up to date. vk-parse had an update that fixed some errors, so an updated vk.xml can be used now too.


Eliah-Lakhin:
@Rua Merged. :+1:

PR #1797, opened 2022-01-15
Many changes to images, views and format

Rua:
Depends on #1793.

```markdown
- **Breaking** The spelling `mip_level` is used everywhere as a function and variable name for consistency with Vulkan naming.
- **Breaking** Changes to `UnsafeImage`:
  - It is now constructed with a builder.
  - It is no longer unsafe to construct, only to use.
  - Memory requirements are no longer returned on construction, but retrieved by calling `memory_requirements` on the `UnsafeImage`.
  - `format_features` now returns a reference.
  - `preinitialized_layout` is replaced by `initial_layout`, which returns an `ImageLayout` value.
- **Breaking** Changes to the `ImageAccess` trait:
  - Added a `format_features` method, and removed `supports_blit_source` and `supports_blit_destination`.
  - `preinitialized_layout` is replaced by `initial_layout`, which returns an `ImageLayout` value.
  - `current_layer_levels_access` was misnamed, and is now named `current_array_layers_access`.
  - `current_miplevels_access` is now named `current_mip_levels_access`.
- **Breaking** Changes to `ImageView`:
  - `UnsafeImageView` has been merged into `ImageView`.
  - Removed `with_` from all the builder functions.
  - The `ImageViewAbstract` trait now has `VulkanObject` as a supertrait, and no longer has an `inner` method.
- **Breaking** Changes to `Format`:
  - `properties` has been moved to `PhysicalDevice::format_properties`. It now supports the `khr_format_feature_flags2` extension.
  - `requires_sampler_ycbcr_conversion` has been renamed to `ycbcr_chroma_sampling` and returns `Option<ChromaSampling>`, where `Some` replaces `true`.
  - `size` has been renamed to `block_size`.
  - `block_dimensions` has been renamed to `block_extent`.
  - The members of the `CompressionType` enum have changed.
- **Breaking** The `BufferViewAbstract` trait now has `VulkanObject` as a supertrait.
- **Breaking** `ExternalMemoryHandleType` has been renamed to `ExternalMemoryHandleTypes`, and the singular name `ExternalMemoryHandleType` has been introduced as an enum.
- **Breaking** `ComponentMapping` and `ComponentSwizzle` have been moved to the `sampler` module.
```

```markdown
- Added `PhysicalDevice::image_format_properties`.
- Added `UnsafeImage::tiling`.
- The `ImageViewAbstract` trait now includes a `format_features` method.
- The `BufferViewAbstract` trait now includes `format` and `format_features` methods.
```

A big one for sure. The original intent was to make this into multiple separate PRs, but because they all overlap in scope, they depend on each other and would have a lot of merge conflicts. Putting it all in one was easier. The good news is that most of the breaking changes are in things that I expect only more advanced Vulkano users would use. None of the examples had to be changed.

The new vk.xml added in #1793 includes a new section with format information, so I rewrote the part of autogen that generates the formats to use the XML data instead of trying to parse it from the names. It's much cleaner this way. I changed some of the method names and other details on `Format` to match the terminology used in the XML.

I found that `UnsafeImage` was lacking a lot, so I basically rewrote the whole thing. It has a builder now, which helps when there's a lot of parameters and it also prevents having to break things when new parameters are added. More validation checks were also added, and some correctness issues were addressed. As far as I can tell, it's completely safe to construct now, so the builder is not `unsafe`. I think that the "safe" image types could be changed to use a builder as well, but I think it would be even nicer to have a single generic image type that can handle all use cases, instead of all the different ones that Vulkano has now.

Regarding the move of `ComponentMapping`: at the moment `ImageView` is the only thing that uses it, but I had a look at the Vulkan spec, and it seems that it's actually used quite frequently in extensions to the `Sampler` that Vulkano doesn't currently support, but may support in the future. Moreover, only image views that are used as sampled images can actually be used with a component mapping, storage images and attachment images can't, so it's already associated with sampling even if indirectly.


Eliah-Lakhin:
@Rua Can you please fix the merging conflict? I'm going to start the review.

Rua:
Ok, done.

vulkano/src/device/physical.rs
```diff
@@ -929,3 +1087,240 @@ impl From<ash::vk::ShaderCorePropertiesFlagsAMD> for ShaderCoreProperties {
...
+    pub sampled_image_filter_linear: bool,
+    /// Can be used with samplers or as a blit source, using the
+    /// [`Cubic`](crate::sampler::Filter::Cubic) filter.
+    pub sampled_image_filter_cubic: bool,
+    /// Can be used with samplers using a reduction mode of
+    /// [`Min`](crate::sampler::SamplerReductionMode::Min) or
+    /// [`Max`](crate::sampler::SamplerReductionMode::Max).
+    pub sampled_image_filter_minmax: bool,
+    /// Can be used with sampler YCbCr conversions using a chroma offset of
+    /// [`Midpoint`](crate::sampler::ycbcr::ChromaLocation::Midpoint).
```
Eliah-Lakhin, line 1155: @Rua Cargodoc complains about `ycbcr` submodule. Other than that I didn't find any issues.

Rua: I have a PR in the works that adds that submodule, but I put it on hold to work on other things first. You can ignore this error.

Eliah-Lakhin: Ok!

Eliah-Lakhin:
@Rua Pull Request merged :+1:

PR #1803, opened 2022-01-20
Fix support for 2D array images

TJYSunset:

Vulkano currently doesn't support the following:

+ **Generation of mipmaps for 2D array textures**. Only the first layer has proper mipmapping, the other layers have blank images as their non-first mipmap levels. (If you run the example without the fix, the star and asterisk would disappear when the window is resized small enough.)
+ <s>**Single-layer array**. Sounds stupid, but perfectly valid (?). In my particular use case, I used a `sampler2DArray[]` to store textures of different sizes/formats, with auto-generated descriptors and name-to-index lookup - which didn't quite work out at first because one of the arrays only have a single image. Yes, there will probably never exist a finished product containing single-layer array textures, but auto-inferred type that cannot be overridden is probably still a bad design. </s>

This pull request aims to fix the issues.

Notes:

+ I have zero idea what am I doing... It seems to work and the existing tests seem to have passed, but that could be an illusion.

+ I have not tested support for 1D/3D images yet. Maybe I will commit a example containing them in this PR later.

+ <s>Perhaps breaking the API is not a good idea?</s>

P.S.: I am not a native English speaker, the ambiguous terminology is making my head spin and I am submitting this PR in the middle of the night... this PR is about `sampler2DArray` in GLSL or [whatever the thing presented in this example is](https://github.com/SaschaWillems/Vulkan/blob/master/examples/texturearray/texturearray.cpp). Any ambiguous terms above are referring to `sampler2DArray`.

        - `Entry 1.`
        - `Entry 2.`...

+ Fixed mipmap generation for 2D array images.

Maybe after the reviews... (rust-lang/rustfmt#1324)

... um, [immutable.rs](https://github.com/vulkano-rs/vulkano/blob/189a58900268fa9d59f1a391dc1f98354266772f/vulkano/src/buffer/immutable.rs) and [view.rs](https://github.com/vulkano-rs/vulkano/blob/4c7cdb44fe312bd0977d4d3b8cda2c86f35404ea/vulkano/src/image/view.rs) does not seems to contain any tests in the first place?


Rua:
This doesn't seem like the right way to fix it at all. `ImageViewBuilder` already has a `ty` method, so I don't see the value of adding a `ty` value to the constructor as well. Just override the default by calling `ty` on the builder if you need to, that's what the builder is for.

TJYSunset:
Oops... Thanks for the heads up!

Rua:
There's merge conflicts, could you resolve them please?

Rua:
Thank you for the new example and the improved mipmap generation!

PR #1805, opened 2022-01-24
Add example for object-oriented deferred cube

TJYSunset:
Here is the example I promised. Hope it's more useful than misleading...

Caveats: 
- I came from C# and don't really understand why every Rust project wants to stuff everything into one single `.rs` file as if it's a header-only library. So this example is C#-structured.
- I do not speak English, and JetBrains' LanguageTool plugin doesn't work for Rust comments.
- I have zero prior experience in 3D graphics other than have modeled some crappy animal characters in Blender.
- I haven't proofread this thing. Not after fixing 100+ compiler errors in four hours and another hour of RenderDoc debugging... But it does match the ground-truth Blender render.

Sorry if my wording is inappropriate - I haven't slept well in the past few weeks and the lackluster Rust language support is driving me crazy...


Rua:
I'm not sure if the C# style is really appropriate/idiomatic for Rust. Some of your files are quite small and maybe only contain one function. Rust is not object oriented by design, so perhaps this is a case of trying to make it into something it isn't?

TJYSunset:
Sorry... but where could I find a guide on idiomatic rust project structure? I'm really desperate because Rust is anything but intuitive, and the editing experience is horrible. All the articles I could find are about the details, and reading Rust projects didn't help...

> Rust is not object oriented by design

Maybe the title is a bit misleading... I'm not using inheritance at all. I just separated the components by C#/Java's project structure convention.

Unrelated, but my personal reasoning on this style is navigating in the file browser is easier than trying to find a particular function in a file with 1000 lines, and splitting the editor isn't really an option:
Snipaste_2022-01-25_06-34-22
(I know I could turn of inlay type hints, but then I would have even less information on whether my code compiles...)

And I'm making empty `mod.rs` because these also make navigation a nightmare:
Snipaste_2022-01-25_06-39-21

For the files in question, `scene/light.rs` is meant to be extended by the user (this file only contains a point light model). The other two files in `scene` are probably too short, but um why are short files a bad thing? (I'm asking because I really don't know why)

TJYSunset:
Nevermind... Sorry for wasting everyone's time. I'll try to get better at Rust first.

PR #1806, opened 2022-01-25
Better validation of image view creation, descriptor set updates, draw-time resources and others

Rua:
Changelog:
```markdown
- Added `aspects` to the builder of `ImageView`, to choose which aspect(s) of an image to use in the view.
```

This pretty much continues on from #1797. `ImageViewBuilder::build` got a major rewrite, with more validation checks. I also added more validation to the descriptor set update code, and for descriptor set resources at draw/dispatch time. The rest is minor internal renames and tweaks. There should be no breaking changes, so if you find anything that doesn't work, that's either a bug or something I forgot.

To help myself and others relate the Vulkano code to the Vulkan specification, I have started to add comments to the code that state which VUID (valid usage ID) is being checked in that code, or some other requirement if there is no VUID. I plan to continue to do this with code I modify as I encounter it, if you find it useful.

Some of these changes, such as the one adding `aspects`, are needed to add YCbCr conversion support, which I will submit after this PR.


Eliah-Lakhin:
@Rua Looks good. Merged :+1:

Issue #1756, opened 2021-11-16
Vulkan Video in Vulkano?


lmarianski:
I'd like to use the `VK_KHR/EXT_video_*` extensions for video encoding, and i see that they're available in the DeviceExtensions struct (except `VK_EXT_video_encode_h265`), but i don't see how i can call the methods from the vulkan documentation, any help would be appreciated.

(Note: i am learning vulkan, so please excuse me if the answer is obvious etc.)
Thanks

Rua:
Vulkano is still very much a work in progress, so there's a lot of things that haven't been implemented yet. I can try to take a look at this sometime. Can you tell me a bit more about these extensions, what they do, how they work etc?

lmarianski:
Fair, here's the info on the extensions: https://www.khronos.org/blog/an-introduction-to-vulkan-video

Rua:
Thanks! It looks like these extensions are currently still regarded as "provisional", meaning their final API and inner details haven't been finalised yet. I think it would be better for Vulkano to wait until they become fully supported.

lmarianski:
Ok, thanks

Eliah-Lakhin:
@lmarianski I will close this thread as we decided not to work on this feature for now. Later it could be reopened. However, if you think you can work on a prototype now or in the future, it is very welcome! :)

PR #1807, opened 2022-01-25
Sampler ycbcr conversion

Rua:
```markdown
- Added `SamplerYcbcrConversion`, and the ability to attach it to a sampler or image view.
```

Depends on #1806.

A sampler YCbCr conversion is a Vulkan object that's used for sampling video frames. It's a bit more of a niche feature, and I myself have no use case for it, but it's just one more thing that brings Vulkano closer to being complete. It's also needed for video encoding/decoding, a still-experimental Vulkan feature that was requested in #1756. I don't know if it's wise to implement an experimental feature, but if we do, then the requirements are now done.


Eliah-Lakhin:
@Rua After the #1806 merge, there are merging conflicts.

vulkano/src/sampler/ycbcr.rs
```diff
@@ -0,0 +1,636 @@
...
+//! A sampler YCbCr conversion is an object that assists a sampler when converting from YCbCr
+//! formats and/or YCbCr texel input data. It is used to read frames of video data within a shader,
+//! possibly to apply it as texture on a rendered primitive. Sampler YCbCr conversion can only be
+//! used with certain formats, and conversely, some formats require the use of a sampler YCbCr
+//! conversion to be sampled at all.
+//!
+//! A sampler YCbCr conversion can only be used with a combined image sampler descriptor in a
+//! descriptor set. The conversion must be attached on both the image view and sampler in the
+//! descriptor, and the sampler must be included in the descriptor set layout as an immutable
+//! sampler.
```
Eliah-Lakhin, line 21: Maybe provide an example of the sampler use?

Rua: Ok, how is this?

Eliah-Lakhin: Excellent!

Eliah-Lakhin:
@Rua Merged :+1:

PR #1810, opened 2022-01-30
Release Vulkano 0.28

Rua:
A lot of breaking changes have accumulated already, I think this may be a good moment for a new release?

I've renamed the changelog file because now that vk-sys is gone, there is only one changelog, so there is no need to make the name longer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
     Pull Request merge. 
 -->
 
+# Version 0.28.0 (2022-01-30)
+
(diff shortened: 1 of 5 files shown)


AustinJ235:
All examples ran as expected with the exception of the indirect on Windows 11 / RTX 3060 Ti. I pushed a commit on the pull request to resolve the unused import on windows builds. I am not sure if this example was broken in previous releases, but I don't see this is as blocking.

Will publish upon CI build completion.

PR #1598, opened 2021-05-26
add support for `VK_KHR_multiview`

Timbals:
Supporting the `VK_KHR_multiview` extension enables rendering multiple views in a single renderpass.

Entries for Vulkano changelog:
- Added support for the `VK_KHR_multiview` extension
- Vulkano-shaders: Added support for MultiView SPIR-V capability.
- `multiview` example added showing how to utilize the `VK_KHR_multiview` extension to render to multiple layers of a framebuffer at once

fixes #1596

PR #1811, opened 2022-01-30
fix not being able to create multi-layer framebuffers

Timbals:
I accidentally introduced a [bug](https://github.com/vulkano-rs/vulkano/pull/1598/files#diff-18589b8e2f8ac1459a2328526e37714b13a1c4badaab289b0741dc188a5d3a30R294) that prevents the creation of framebuffers with multiple layers when the multiview feature is *disabled*. This PR fixes that bug.

Entries for changelog:
- fix not being able to create multi-layer framebuffers

### DIFF
--- a/vulkano/src/render_pass/framebuffer.rs
+++ b/vulkano/src/render_pass/framebuffer.rs
@@ -334,7 +334,7 @@ impl FramebufferBuilder {
             }
         }
 
-        let mut layers = 1;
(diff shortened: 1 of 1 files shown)


Rua:
Oops!

Eliah-Lakhin:
@Rua Since the bug was introduced in one of the previous releases(0.24), I added a changelog entry into upcoming release logs.

PR #1812, opened 2022-01-30
A multi-window game of life application example

hakolao:
We need a classic game of life example. Why not multi-window? :). One you can draw on!

image

A multi windowed game of life application. You could use this to learn:
- how to handle multiple window inputs and draw on multiple windows
- how to draw on a canvas
- how to organize compute shaders with graphics
- how to do a cellular automata simulation using compute shaders

Lots more.

I wrote this very quickly (Sunday afternoon), so there might be mistakes. Only tested on Ubuntu today.


hakolao:
3D next? :)

examples/src/bin/multi_window_game_of_life/game_of_life.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    int alive_count = 0;
+    if (life_out[get_index(up_left)] == 1) { alive_count += 1; }
+    if (life_out[get_index(up)] == 1) { alive_count += 1; }
+    if (life_out[get_index(up_right)] == 1) { alive_count += 1; }
+    if (life_out[get_index(right)] == 1) { alive_count += 1; }
+    if (life_out[get_index(down_right)] == 1) { alive_count += 1; }
+    if (life_out[get_index(down)] == 1) { alive_count += 1; }
+    if (life_out[get_index(down_left)] == 1) { alive_count += 1; }
+    if (life_out[get_index(left)] == 1) { alive_count += 1; }
```
hakolao, line 203: I didn't check if the index can go out of range at borders. Intuition says a crash would occur, but it does not.

hakolao:
Uh. Will fix the build error…

Rua:
Do you think this makes the old multi-window example obsolete? If so, you could remove it.

Also, slight nitpick, but it's 2022. :)

hakolao:
I don't know if it does. The old multi-window is plain and simple. In a way nothing wrong with it. I'll consider it 

I'll change the year...

hakolao:
I truly can't believe it's 2022

Issue #1808, opened 2022-01-27
Proposal: use `non_exhaustive` (instead of builders) for all extendable Vulkan structs


Rua:
Vulkan has a neat, if somewhat cumbersome mechanism for adding new functionality to existing APIs, by means of the `pNext` pointers. The goal behind this design is to allow extending the API without breaking existing code. Preventing code breakage is also a goal for Vulkano, and one idiomatic way to do that is by using builders. You can always add new functions to a builder without breaking existing code, as long as the builder behaves the same as before the new function was added.

I therefore propose that, in principle, whenever Vulkan allows for extension using `pNext`, Vulkano maps this to a builder. Existing object types not currently using a builder would be converted to use one, such as `Instance`, `Device`, `UnsafeBuffer`, `ComputePipeline` and others.

Eliah-Lakhin:
Can you show an example of how e.g. ComputePipeline's builder will look like?

Rua:
Something like this:
```rust
ComputePipeline::start(device)
    .shader(cs.entry_point("main").unwrap())
    .layout(layout) // would be optional, defaults to a layout generated from the shader
    .build().unwrap()
```

So it would be similar to how the `GraphicsPipeline` builder works, just with a lot less options.

Eliah-Lakhin:
ok, looks good!

gurchetansingh:
Yes, that sounds good!  I think DeviceMemory already uses one.  Feel free to change the API if you find a cleaner way to do it generically!

Rua:
An alternative design I thought of, that is a bit further from the current typical builder design but closer to Vulkan, is this version:
```rust
ComputePipeline::new(
    device,
    ComputePipelineCreateInfo::default()
        .shader(cs.entry_point("main").unwrap())
        .layout(layout)
    ).unwrap()
```

In this version, the "builder" only helps to gather the creation info, and is then passed to `new`, where the construction itself happens. There's not necessarily an advantage or disadvantage to this style, it's just different, and closer to the C Vulkan style.

[...]

PR #1814, opened 2022-02-03
Add `InstanceCreateInfo` and `DeviceCreateInfo`

Rua:
Changelog:
```markdown
- **Breaking** `Instance` creation parameters are given using `InstanceCreateInfo`.
- **Breaking** `ApplicationInfo` and the `app_info_from_cargo_toml!` macro are removed, their functionality is now integrated into `InstanceCreateInfo`.
- **Breaking** `Device` creation parameters are given using `DeviceCreateInfo`.
```

This implements #1808 for instances and devices. The breaking changes should be relatively straightforward to fix.

Note the changes to how queues are specified when creating a device. Vulkan has at least one flag and one extension struct that can be applied to the queue family as a whole, so I created a small builder for queues with the option to add these missing features to it later without breakage. Also note that you now get a single queue for the specified queue family by default, so you if that's all you need, you don't need to specify priorities. examples/src/bin/multi_window_game_of_life/vulkano_context.rs has a more elaborate example of queue creation if you want to see.

When creating an instance, the highest known API version (to Vulkano) is now automatically picked for the `max_api_version` as default, so if you want to use all Vulkan versions you don't need to mess with this value anymore. Specifying the application info from the Cargo.toml is also done via the builder, so the macro is no longer needed and has been removed.


Rua:
Converting to draft until a decision is made on #1808.

vulkano/src/instance/instance.rs
```diff
@@ -1,692 +0,0 @@
...
-/// - Panics if the required environment variables are missing, which happens if the project
-///   wasn't built by Cargo.
-///
-#[macro_export]
-macro_rules! app_info_from_cargo_toml {
-    () => {{
-        let version = $crate::instance::Version {
-            major: env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap(),
-            minor: env!("CARGO_PKG_VERSION_MINOR").parse().unwrap(),
-            patch: env!("CARGO_PKG_VERSION_PATCH").parse().unwrap(),
```
Eliah-Lakhin, line 558: For the back-compatibility I think it would be reasonable to implement a `FromStr` trait for `Version` object that will be capable to parse a semver-alike string of the `env!("CARGO_PKG_VERSION")` in particular.

vulkano/src/device/mod.rs
```diff
@@ -636,11 +676,75 @@ impl Eq for Device {}
...
+    /// The default value is a single queue with a priority of 0.5.
+    pub queues: Vec<f32>,
+
+    pub _ne: crate::NonExhaustive,
+}
+
+impl<'qf> QueueCreateInfo<'qf> {
+    /// Returns a `QueueCreateInfo` with the given queue family.
+    #[inline]
+    pub fn family<'qf2>(family: QueueFamily<'qf2>) -> QueueCreateInfo<'qf2> {
```
Eliah-Lakhin, line 739: `'qf2` can be elided

vulkano/src/device/mod.rs
```diff
@@ -636,11 +676,75 @@ impl Eq for Device {}
...
+
+    /// The queues to create for the given queue family, each with a relative priority.
+    ///
+    /// The relative priority value is an arbitrary number between 0.0 and 1.0. Giving a queue a
+    /// higher priority is a hint to the driver that the queue should be given more processing time.
+    /// As this is only a hint, different drivers may handle this value differently and there are no
+    /// guarantees about its behavior.
+    ///
+    /// The default value is a single queue with a priority of 0.5.
+    pub queues: Vec<f32>,
```
Eliah-Lakhin, line 731: The previous implementation had only a single value for the entire family. As far as I understand the new implementation allows specifying priority per queue. Is it really needed?

In general I feel that dedicated `QueueCreateInfo` interface makes things too complicated. e.g. there is no simple way to specify a pair of `(Family, f64)` as we did in the past. The user can do something like `QueueCreateInfo { queues: (0..family.queues_count() + 1).map(|_| (<priority>)).collect() ..QueueCreateInfo::family(family)}`, but it will be too verbose. I suggest at least extend the `QueueCreateInfo::family` constructor with the priority parameter.

Eliah-Lakhin:
@Rua Looks good in general, and works without regressions on my end, but I left a few comments for the edge cases usages.

vulkano/src/device/mod.rs
Rua: The original implementation also allowed specifying a priority per queue. If you wanted to have multiple queues for the same queue family, you would do this in the old code:
```rust
[(family, 0.5), (family, 0.5)]
```
In the new code this becomes:
```rust
QueueCreateInfo {
    queues: [0.5, 0.5].into(),
    ..QueueCreateInfo::family(family)
}
```
It's a more verbose, but also more explicit. Adding the priority parameter to the `family` constructor would remove the functionality that was present in the old code, to be able to create multiple queues for one family.

The real reason why I added `QueueCreateInfo` though is extendability. In Vulkan, the `VkDeviceQueueCreateInfo` struct has a `pNext` field, which means it can be extended. And in fact there is already an extension struct for it, `VkDeviceQueueGlobalPriorityCreateInfoKHR`. Vulkano will probably want to add support for that at some point, and in the original implementation that would not be possible, in the new implementation it is possible. `VkDeviceQueueCreateInfo` also has a `flags` member with a flag that is currently not implemented by Vulkano, but the same consideration applies. For example:

```rust
QueueCreateInfo {
    queues: [0.5, 0.5].into(),
    protected: true, // the flag
    global_priority: QueueGlobalPriority::Low, // the extension struct
    ..QueueCreateInfo::family(family)
}
```

Note that the `protected` flag and `global_priority` field are specified _per family_ and not _per queue_. Only the queue priorities are specified per queue. In the old code, I don't really know what this would become.

vulkano/src/device/mod.rs
Rua: Fixed.

vulkano/src/instance/instance.rs
Rua: I added it.

vulkano/src/device/mod.rs
Eliah-Lakhin: ok. That makes sense. Thank you for explanation.

Eliah-Lakhin:
Merged :+1:

PR #1815, opened 2022-02-05
Add `UnsafeBufferCreateInfo` and `BufferViewCreateInfo`

Rua:
Changelog:
```markdown
- **Breaking** Changes to `UnsafeBuffer`:
  - Creation parameters are given using `UnsafeBufferCreateInfo`. It is no longer unsafe to construct, only to use.
  - Memory requirements are no longer returned on construction, but retrieved by calling `memory_requirements` on the `UnsafeBuffer`.
- **Breaking** Constructing a buffer of zero size now results in a panic (except for `CpuBufferPool` chunks).
- **Breaking** `BufferView` creation parameters are given using `BufferViewCreateInfo`.
- **Breaking** The following types are now considered non-exhaustive and must be constructed using a method call and struct update syntax: `BufferUsage`, `SparseLevel`.
```
```markdown
- Added a `Format::texels_per_block` method.
```

More builderfication, and adding missing validation checks as I go.


Rua:
Converting to draft until a decision is made on #1808.

Eliah-Lakhin:
> Constructing a buffer of zero size now results in a panic (except for `CpuBufferPool` chunks).

@Rua This new restriction fits Vulkan's documentation, but perhaps propagating it to the safe public API is not convenient. Vulkano is trying to be handy for the end-user through hiding internal complexities of Vulkan.

Assume the user has a generic code to crate an `ImmutableBuffer` from the `data` collection that sometimes may be empty:

```rust
let (buffer: Arc<ImmutableBuffer<...>>, buffer_submission) = ImmutableBuffer::from_iter(data.into_iter(), BufferUsage {...}, queue).unwrap();

buffer_submission
    .then_signal_fence()
    .wait(None)
    .unwrap();
```

Then they store the `buffer` variable somewhere, and then they write a descriptor to this buffer for actual dispatch/renderpass.

If the data can be empty, with the new implementation they will have to store some kind of an `Option<Arc<ImmutableBuffer<..>>>` type instead, and to alter descriptors creation depending on the `Option` state. This is feasible, but it makes the end-user code much more complex.

As the `ImmutableBuffer` is a high-level type, perhaps it would make sense to encapsulate the option into the type's struct. Though, I'm not sure how difficult would it be to make it this way in Vulkano. What do you think?

Rua:
`CpuBufferPool` works that way at least, it gives you an empty chunk. But with the other types I'm not sure if it would be right to do it that way. If the user asks for a buffer, should we give them a maybe-buffer? The previous code would sneakily create a buffer of size 1 when you asked for a buffer of size 0. We could go back to that, but it feels very wrong to "disobey" the user.

Eliah-Lakhin:
> The previous code would sneakily create a buffer of size 1 when you asked for a buffer of size 0

Interesting, I didn't realize that. ok, there could be different approach to resolve this situation. And each of them have pros and cons. Let's leave it at the user's decision then.

Eliah-Lakhin:
@Rua Looks good. Merged. :+1:

PR #1820, opened 2022-02-09
Added `overlapping_push_constant_ranges` to `PipelineLayout`

### LINKED ISSUE #1819: Unsound `push_constants` command implementation
The current implementation of `push_constants` in `auto.rs` has many problems:

1- Serious buffer overflow bug (increment offset with same size):
https://github.com/vulkano-rs/vulkano/blob/4f094396ad7f8452e44ff6a907bfe43603e5c37f/vulkano/src/command_buffer/auto.rs#L1874-L1877

2- When calling `push_constants` we don't know which stages will be pushed to:
If you have multiple stages, it will try to include as many as it can.
For example if you have these push_constant ranges:
```
fragment (0..4)
vertex (0..4)
```
It will push to the two of them, even if the content is different, for example:
```
fragment:
layout(push_constant) uniform PushConstantData1 {
    uint a;
} pc;

vertex:
layout(push_constant) uniform PushConstantData2 {
    uint b;
} pc;
```

3- Also there is a bug in selecting the stages, if you have two stages
 ```
fragment (0..4)
vertex (0..8)
```
And you pushed struct of size `8`, it will push to the two of them, which invalidates
```
VUID-vkCmdPushConstants-offset-01795
For each byte in the range specified by offset and size and for each shader stage in stageFlags, there must be a push constant range in layout that includes that byte and that stage
```

I'll create a PR on making it mandatory to supply the stage that is needed, and also fix the buffer overflow bug.

Any other suggestions to improve `push_constants` handling?

### PR DESCRIPTION
**Pull request by Amjad50**
Fixes #1819 

Added `overlapping_push_constant_ranges` in `PipelineLayout`, which helps us track overlapping ranges, and what stages belong into them. It will be used in the `push_constants` command to help fixing some issues regarding:
- VUID-vkCmdPushConstants-offset-01795
- VUID-vkCmdPushConstants-offset-01796
And also solidify the implementation.

CHANGELOG:
```
- Added `overlapping_push_constant_ranges` in `PipelineLayout` to keep track of overlapping ranges.
- Added tests for the algorithm on how to compute the `overlapping_push_constant_ranges` in `PipelineLayout`.
- Fixed buffer overflow bug in `AutoCommandBufferBuilder::push_constants`.
- Edited `AutoCommandBufferBuilder::push_constants` to push multiple times in case of range overlap (to accommodate VUIDs 01795 and 01796) 
```

### DIFF
--- a/vulkano/src/command_buffer/auto.rs
+++ b/vulkano/src/command_buffer/auto.rs
@@ -79,7 +79,6 @@ use crate::render_pass::Framebuffer;
 use crate::render_pass::LoadOp;
 use crate::render_pass::Subpass;
 use crate::sampler::Filter;
-use crate::shader::ShaderStages;
 use crate::sync::AccessCheckError;
 use crate::sync::AccessFlags;
 use crate::sync::GpuFuture;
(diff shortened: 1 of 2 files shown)


Rua:
It looks good, thank you. I think the name of the function is a bit of a misnomer though. I would expect the ranges that it returns to *not* overlap. So maybe `push_constant_ranges_non_overlapping` (I prefer the name in this order, but it's fine either way).

Rua:
Looks good and it runs without errors. Thank you!

PR #1822, opened 2022-02-09
fix queue 'supports_' methods with empty flags

Overpeek:
The methods prefixed with `supports_` in `vulkano::device::physical::QueueFamily` should use the `contains` method with queue_flags. This makes sure that if the tested flags are are empty, they are supported. This fixes a panic when for example: timestamping `PipelineStage::TopOfPipe`, `PipelineStage::BottomOfPipe` or `PipelineStage::AllCommands` where no flags are required for them to be supported.


Rua:
Thanks again for the fix!

PR #1823, opened 2022-02-10
fix shader! macro for geometry stages failing to compile

Overpeek:
Using the `vulkano_shaders::shader!` macro with geometry type would fail to compile. It seemed to use an earlier version of `ShaderExecution` enum. Also `GeometryShaderInput` would have to be manually imported for it to compile.

        - `fixed shader! macro failing to compile with geometry shaders`


Rua:
Thank you again for your fix! I don't know how you keep finding these, but it's very welcome!

PR #1824, opened 2022-02-12
Refactor VertexBuffersCollection to allow Arc<dyn BufferAccess>

AustinJ235:
This pull request makes it possible to `Arc<dyn BufferAccess>` in collections where `VertexBuffersCollection` is accepted.

**Changelog Entries**
```
- Refactor `VertexBuffersCollection` to allow `Arc<dyn BufferAccess>`
```

I am not sure if this is a breaking change or not. It ***shouldn't*** break existing code.


Rua:
It all looks good! To prevent future breakage it might be good to have a test that tries all the different combinations of concrete/trait object types.

PR #1827, opened 2022-02-15
Only include `SamplerReductionModeCreateInfo` if needed

### LINKED ISSUE #1826: Sampler reduction mode causes Sampler2DShadow to not work properly on some drivers
please add a condition at [this line](https://github.com/vulkano-rs/vulkano/blob/3ad03f85932a7c29add3ae13d711e66c03607e35/vulkano/src/sampler/mod.rs#L563). If compare is enabled then set it to None.

I don't think this is required by the spec, probably not, but it would work around an annoying bug for me (Intel WhiskeyLake-U GT2 [UHD Graphics 620], archlinux, linux 5.16.8-arch1-1, mesa 21.3.5-1)

Rua:
Changelog:
```markdown
- Fixed bug on certain drivers where samplers would behave oddly when minmax samplers are enabled.
```

Fixes #1826.

@pac85 Can you test this and see if it fixes the issue you were having?

### DIFF
--- a/vulkano/src/sampler/mod.rs
+++ b/vulkano/src/sampler/mod.rs
@@ -561,15 +561,10 @@ impl SamplerBuilder {
         }
 
         let mut sampler_reduction_mode_create_info =
-            if device.enabled_features().sampler_filter_minmax
-                || device.enabled_extensions().ext_sampler_filter_minmax
-            {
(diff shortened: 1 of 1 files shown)


pac85:
I tested it and it fixes the issue.
Thanks,

AustinJ235 (approved):

PR #1831, opened 2022-02-18
Include raw function pointer structs in documentation

Rua:
I noticed that these were missing from the documentation. The `fns` methods of `FunctionPointers`, `Instance` and `Device` do reference these structs, and they are useful for people who want to use functionality that Vulkano doesn't support yet.


AustinJ235 (approved):
Thanks!

PR #1832, opened 2022-02-19
Add `SwapchainCreateInfo`, various related changes

### LINKED ISSUE #1801: Swapchain does not expose color_space
Probably only a minor annoyance, but it is somewhat inconsistent.

Speaking of which, the "new" (compared to the last time I tried vulkano) triangle example does not deal with color space at all. Which is probably fine since the builder has a sane default of sRGB, but isn't the example supposed to show off all the options? (inferring from the even less used `CompositeAlpha` being in the example)

### LINKED ISSUE #1830: How to provide HMONITOR for full-screen exclusive mode on Windows?
Vulkano currently has support for full-screen exclusive mode with swapchains. But when implementing it, one detail seems to have been missed in the specification for `VkSwapchainCreateInfoKHR`:

> VUID-VkSwapchainCreateInfoKHR-pNext-02679
> If the `pNext` chain includes a [VkSurfaceFullScreenExclusiveInfoEXT](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceFullScreenExclusiveInfoEXT.html) structure with its `fullScreenExclusive` member set to `VK_FULL_SCREEN_EXCLUSIVE_APPLICATION_CONTROLLED_EXT`, and `surface` was created using [vkCreateWin32SurfaceKHR](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCreateWin32SurfaceKHR.html), a [VkSurfaceFullScreenExclusiveWin32InfoEXT](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSurfaceFullScreenExclusiveWin32InfoEXT.html) structure **must** be included in the `pNext` chain

The `VkSurfaceFullScreenExclusiveWin32InfoEXT` structure is not used in Vulkano at the moment, which means that creating a swapchain with `FullScreenExclusive::ApplicationControlled` is a spec violation if you use it on Windows. I'm not entirely sure how to solve this issue, though. This structure has a single field which is a `HMONITOR`, a handle for the Windows-specific monitor that should be made full-screen exclusive. This handle can be retrieved from Winit or whatever other API the user is using for their windowing, but Vulkano itself is agnostic to that.

But in what form should Vulkano ask for the `HMONITOR` handle when creating a swapchain? Ash [exposes the type directly](https://docs.rs/ash/latest/ash/vk/type.HMONITOR.html) so we don't have to worry about that part. Unfortunately, Vulkano has no way to check that the handle is valid, as far as I can tell. Is there a way to design it so that it's safe, somehow?

Rua:
Changelog:
```markdown
- **Breaking** Changes to `Surface`:
    - The constructors are renamed to match their Vulkan equivalents:
        - `from_display_mode` > `from_display_plane`
        - `from_anativewindow` > `from_android`
        - `from_ios_moltenvk` > `from_ios`
        - `from_macos_moltenvk` > `from_mac_os`
        - `from_vi_surface` > `from_vi`
        - `from_hwnd` > `from_win32`
    - `from_raw` now requires a `SurfaceApi` value.
    - `Surface::capabilities` has been moved to `PhysicalDevice` and split into three functions: `surface_capabilities`, `surface_formats` and `surface_present_modes`.
    - `Capabilities` has been renamed to `SurfaceCapabilities`.
    - The `is_supported` method has been moved to `QueueFamily` and renamed to `supports_surface`.
- **Breaking** Changes to `Swapchain`:
  - Creation parameters are given using `SwapchainCreateInfo`. Methods of `Swapchain` are renamed to match.
  - All names with `Fullscreen` and `fullscreen` have been renamed to `FullScreen` and `full_screen`, to match Vulkan word breaking.
  - `FullScreenExclusive::AppControlled` has been renamed to `FullScreenExclusive::ApplicationControlled` to match Vulkan.
  - Using `FullScreenExclusive::ApplicationControlled` on Windows now requires a `Win32Monitor`.
- **Breaking** Vulkano-win: `create_vk_surface` and `create_vk_surface_from_handle` have been renamed to `create_surface_from_winit` and `create_surface_from_handle` respectively.
```
```markdown
- Added an `api` method to `Surface` to return the windowing API that it was created from.
- Added a `create_info` method to `Swapchain` to return a `SwapchainCreateInfo` that contains all the parameters copied from the existing swapchain. This can be used for easy recreation of outdated swapchains.
- Added support for the `khr_get_surface_capabilities2` extension.
- Vulkano-win: Added `create_win32_monitor_from_winit` function.
```

Fixes #1801, fixes #1830.

More `CreateInfo`fication, along with some reorganising of code to make it more robust and better match Vulkan.


AustinJ235 (approved):
Looks good! The Win32 handling for `AppControlled` seems sound. Due to previously merged pull requests there are now conflicts. You can go ahead with a merge once those are resolved.

PR #1833, opened 2022-02-19
Implement is_layout_initialized for ImmutableImage

thanatos:
The default implementation always returns `false`. However, in
`ImmutableImage`'s case, by the time we are doing operations with
`ImmutableImage`, the layout should have already been initialized by
`ImmutableImageInitialization`.

Without this function, vulkano thinks the image referenced by the
`ImmutableImage` isn't initialized, and will insert a pipeline
transition from Undefined to `self.layout`, which isn't incorrect; the
command buffer created to initialize the image had a
`final_layout_requirement` of the final layout, and thus had already
transitioned the layout to the correct layout.

See, for example, the discussion in #1578.

### DIFF
--- a/vulkano/src/image/immutable.rs
+++ b/vulkano/src/image/immutable.rs
@@ -415,6 +415,11 @@ where
     }
 
     #[inline]
+    fn is_layout_initialized(&self) -> bool {
+        true
+    }
+
+    #[inline]
     fn initial_layout_requirement(&self) -> ImageLayout {
         self.layout
     }


thanatos:
I also want to address this comment from #1578 a bit:

>  So I am quite puzzled how this hasn't caused obvious problems until now, and I'm hesitant to make any changes in case it turns out it was made this way for a reason.

First, I've had good luck running the examples, and my best guess is that it is due to their simplicity that they're not affected. In my current (local) example, I have both a blit & a shader, and I think seems to be enough complexity to require that layout transitions happen correctly. (I originally started with just a blit, and even getting that to work properly was a bit of a struggle. There are more changes that need to happen to `ImmutableImage` for blits to work, but I wanted to keep the scope of this PR small.)

If without the implementation of `is_layout_initialized`, this is what I get:
without-method

With it:
with-method

(A note about the images: the part in the upper left corner is a blit. The blit slightly into the window and about 1/3 the way down is a pair of triangles textured to the same image with a shader.)

## About blits

Blits require the `transfer_source` usage, and there's no good way to set that with `ImmutableImage`'s interface currently. (You can generate mipmaps as a hacky way to do it.) `ImmutableImage` _also_ presumes that the layout will be ShaderReadOnlyOptimal, which if you want a blit, isn't what you want. (But, vulkano is smart enough to realize during the blit that the image is in the wrong layout & transition it, but this generates some unnecessary pipeline barriers if that's all you plan to use the image for.)

In my case, blits were the wrong solution for what I wanted to do (which is why my example here also now has a shader on it.)

Rua:
This looks ok in principle, but I have to ask what happens if the `ImmutableImageInitialization` hasn't been used yet?

Eliah-Lakhin:
@thanatos Thank you for your work. Can you please address @Rua's question?

thanatos:
> but I have to ask what happens if the `ImmutableImageInitialization` hasn't been used yet?

I'm not 100% sure how to answer this. "Nothing good" would be the short version. It's otherwise kinda complicated, I think.

The biggest problem, I think, would be that layout transitions would be wrong, if they were attempted (but see my notes below about `try_gpu_lock`; IDK if they would be). I think they're wrong in either case, though.

Prior to my PR, I think Vulkano would get `false` from `is_layout_initialized`. Any use of the image would then attempt to transition it to whatever layout is required. (Which would depend on the command the image appears in.) But on subsequent uses of the image, it would continue to return `false`, and still continue to transition the image out of `VK_IMAGE_LAYOUT_UNDEFINED` into the desired layout, and that violates [this part of Vulkan](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageMemoryBarrier.html), AFAICT:

> `VUID-VkImageMemoryBarrier-oldLayout-01197`
>
> If `srcQueueFamilyIndex` and `dstQueueFamilyIndex` define a [queue family ownership transfer](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageMemoryBarrier.html#synchronization-queue-transfers) or `oldLayout` and `newLayout` define an [image layout transition](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageMemoryBarrier.html#synchronization-image-layout-transitions), `oldLayout` must be `VK_IMAGE_LAYOUT_UNDEFINED` or the current layout of the image subresources affected by the barrier

Here, we're in the "or `oldLayout` and `newLayout` define an image layout transition" case.

The spec also mentions,

> The old layout must either be `VK_IMAGE_LAYOUT_UNDEFINED`, or match the current layout of the image subresource range. If the old layout matches the current layout of the image subresource range, the transition preserves the contents of that range. If the old layout is `VK_IMAGE_LAYOUT_UNDEFINED`, the contents of that range may be discarded.

So, "the contents of that range _may_ be discarded". Since uses of the image would continue to see that `false`, Vulkano would set up transitions from undefined to the target layout (if needed), each time discarding the contents. (But as the initializer is not used in this scenario, there aren't really any contents. I presume you'd get memory garbage.)

After my change, we'd always return `true` … so we'd either not transition the image properly (if the desired layout matched `self.layout` — we'd accidentally assume we were already in the correct layout) or we'd issue a transition from `self.layout` to the desired layout, which would also be wrong, as the image would still be in undefined. The common case for `ImmutableImage` is that `self.layout` is `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`, so if, say, we used it in a blit, we'd issue a pipeline barrier to set up a layout transition from `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` to something like `VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL`. But the image would be in undefined, so we'd again violate the spec's requirement that `oldLayout` be the old layout.

Honestly, I'm still figuring out how `try_gpu_lock` gets called & why. (It seems like it happens sometimes when I don't expect it.) But that function does do some checking of expectations in `ImmutableImage`, so I am wondering if it would catch this case, in particular, b/c `self.initialized` would still be `false`. So that might lead to a runtime error.

Rua:
I am working on overhauling the buffer and image synchronisation, so a lot of current things may change anyway. I merged your PR for the 0.29 release.

PR #1836, opened 2022-02-21
Vulkan 1.3 support, updated dependencies

Rua:
Changelog:
```markdown
- Added support for Vulkan 1.3.
- Added `Version::HEADER_VERSION`, a constant that contains the Vulkan spec version that Vulkano currently uses.
```

This updates Ash to 0.36, which includes support for Vulkan 1.3. I've added support for it to Vulkano where needed, including alternative function calls where extensions were promoted.


AustinJ235 (approved):
Awesome!

PR #1841, opened 2022-02-25
Add `CreateInfo` for `Event`, `Fence` and `Semaphore`.

Rua:
Changelog:
```markdown
- **Breaking** `Fence` and `Semaphore` no longer have a type parameter.
- **Breaking** `Event` creation parameters are given using `EventCreateInfo`.
- **Breaking** `Fence` creation parameters are given using `FenceCreateInfo`.
- **Breaking** `Semaphore` creation parameters are given using `SemaphoreCreateInfo`.
- **Breaking** `Semaphore::export_opaque_fd` is now `unsafe`.
```
```markdown
- Added `PhysicalDevice::external_semaphore_properties`.
```

Another relatively simple one.


Rua:
The MacOS run keeps failing on this one, but it's failing when trying to load the Vulkan library at the start of a test. I didn't change anything about the library loading process in this PR, so it's quite strange that it fails so consistently on this. I wonder what's going on.

Rua:
Now the issue also appears on Windows, and I'm even more confused.

Rua:
I've tracked it down and it seems to actually be a possible issue with all other tests *except* this one. All other tests use the `instance!` macro, which is internal to Vulkano and generates a basic `Instance` to use for tests. If this macro can't create an instance then it just aborts the test and shows it as passed. The `semaphore_export` test on the other hand uses `InstanceExtensions::supported_by_core()` and then explicitly creates an instance, unwrapping on errors in both. The CI environments seem to be missing the Vulkan library on both Windows and Mac, so this one test panics while all the other tests... get silently skipped on CI.

AustinJ235:
> I've tracked it down and it seems to actually be a possible issue with all other tests _except_ this one. All other tests use the `instance!` macro, which is internal to Vulkano and generates a basic `Instance` to use for tests. If this macro can't create an instance then it just aborts the test and shows it as passed. The `semaphore_export` test on the other hand uses `InstanceExtensions::supported_by_core()` and then explicitly creates an instance, unwrapping on errors in both. The CI environments seem to be missing the Vulkan library on both Windows and Mac, so this one test panics while all the other tests... get silently skipped on CI.

Not sure if there is any real implementations for Windows or MacOS on Github actions. There is this https://github.com/humbletim/setup-vulkan-sdk which I don't know if that would provide enough? Think these platforms may have to remain build only for most tests.

PR #1843, opened 2022-02-26
Add `UnsafeDescriptorSetCreateInfo`

Rua:
Changelog:
```markdown
- **Breaking** `UnsafeDescriptorPool` creation parameters are given using `UnsafeDescriptorPoolCreateInfo`.
- **Breaking** The `alloc` and `free` methods of `UnsafeDescriptorPool` are renamed to `allocate_descriptor_sets` and `free_descriptor_sets`.
- **Breaking** The `alloc` method of the `DescriptorPool` trait is renamed to `allocate`.
- **Breaking** `DescriptorSetLayout::descriptors_count` is renamed to `descriptor_counts`, and returns a hashmap instead of a separate type.
```

Not too complicated here either. I removed the `DescriptorsCount` type, because the same effect could be achieved more flexibly with a hashmap.


vulkano/src/descriptor_set/layout.rs
```diff
@@ -37,7 +37,7 @@ pub struct DescriptorSetLayout {
     bindings: BTreeMap<u32, DescriptorSetLayoutBinding>,
     push_descriptor: bool,
 
-    descriptors_count: DescriptorsCount,
+    descriptor_counts: FnvHashMap<DescriptorType, u32>,
```
AustinJ235, line 40: I think we should use std HashMap instead.

AustinJ235 (changes requested):
Overall looks good, but I think we should remove the use of the `fnv` crate. `FnvHashMap` was used in a time where the std implementation was slower than desired. Now days rust is using `hashbrown` since version 1.36 and is more performant. On top of the `fnv` hasn't been updated in 2 years.

AustinJ235 (approved):

PR #1852, opened 2022-02-28
arrays now implement `VertexBufferCollection`

AustinJ235:
Stumbled upon how to do this when I was implementing something similar in Basalt. Also added tests that use all variations. This does however create some dummy structs and dummy implementations. These are tucked away in a test mod though.

```
- `VertexBuffersCollection` now supports arrays.
```


Rua:
Looks fine, thanks!

PR #1853, opened 2022-03-02
Rework mapped memory, add `BufferContents` trait with bytemuck

### LINKED ISSUE #1484: CpuAccessibleBuffer::from_data is unsound
# Issue

Issue #1478 revealed that [`CpuAccessibleBuffer::from_data`](https://github.com/vulkano-rs/vulkano/blob/63aaa853ff894081a5f93e5cb3a098047d078094/vulkano/src/buffer/cpu_access.rs#L104) is unsound. It could be called like this:

```rust
let data_out: Vec<f64> = vec![0.0; data_in_size];
let data_buffer_out = CpuAccessibleBuffer::from_data(
    device.clone(),
    BufferUsage::all(),
    false,
    data_out
).expect("failed to create buffer");
```

The code example above compiles just fine, but it shouldn't. At runtime it invokes undefined behavior. This is because, `from_data` treats `data_out` as a flat memory structure, which can be read or written by simply taking its size and copying its bits.

The [documentation for `from_data`](https://docs.rs/vulkano/0.20.0/vulkano/buffer/cpu_access/struct.CpuAccessibleBuffer.html#method.from_data) states that

> Only allowed for sized data.

but `Vec` seems to be `Sized` as well.

Not sure which kind of types can safely be allowed for `from_data`, maybe all types which implement [`Copy`](https://doc.rust-lang.org/std/marker/trait.Copy.html).

### LINKED ISSUE #1809: `CpuAccessibleBuffer` methods `read` and `write` are unsound

Rua:
Changelog:
```markdown
- **Breaking** `MappedDeviceMemory` is now constructed separately with `new`, the `_and_map` variants of `DeviceMemory` are removed.
- **Breaking** Changed how `MappedDeviceMemory` handles CPU access; added `invalidate_range` and `flush_range` methods and `read` and `write` methods to get (im)mutable byte slices to the memory.
- **Breaking** Bytemuck dependencies:
  - All buffer types now require their data type parameter to implement `BufferContents`, which has `bytemuck::Pod` as a supertrait.
  - `Vertex` also has `Pod` as a supertrait.
  - Consequently, you must derive the `Zeroable` and `Pod` traits from bytemuck on any custom structures that are used as buffer contents.
  - To do this for structures generated by Vulkano-shaders, use the `types_meta` parameter as seen in the teapot example.
```

Fixes #1809. I decided against making all buffers be exposed to the user only as byte slices, since the typing of buffers does offer quite some convenience. This change can always be made in the future if decided. In the meantime, since buffers all contain `Pod` now, it's trivial to convert to and from bytes.


Eliah-Lakhin:
Review is in progress. One small comment so far. I noticed that you have reformatted imports. I like this change. Maybe we should add rustfmt config to the repository?

Eliah-Lakhin:
@Rua I finished testing. It works fine on my end.

Regarding the conversion to `Pod`. Do you think it would be safe to auto-implement `Pod` and `Zeroable` for all shader generated types? In this case I think we should advance vulkano-shaders macro as well.

At this moment if the user didn't specify `types_meta` in the macro, the macro auto-implements `Clone` and `Copy` traits only: https://github.com/vulkano-rs/vulkano/blob/master/vulkano-shaders/src/lib.rs#L268 Otherwise the user has to specify traits to implement explicitly, but there is currently no way to specify `unsafe impl`. The extension could be applied somewhere here: https://github.com/vulkano-rs/vulkano/blob/master/vulkano-shaders/src/lib.rs#L705

Eliah-Lakhin:
@Rua I merged it. The PR looks good, and it covers established goals. As of Vulkano shaders macro, I will do an addendum PR later. I think just having `unsafe impl` in the `types_meta` section should be enough.

Rua:
I don't think an `unsafe impl` is needed. Look at the teapot example for how it could be done.

I do agree though that it's preferable if Vulkano-shaders derives these types automatically. But because `bytemuck` is an external crate, I think it should be re-exported from Vulkano if this is done. Then Vulkano-shaders can import `Zeroable` and `Pod` from `vulkano::bytemuck` instead and the user doesn't need to add it as a separate dependency.

PR #1855, opened 2022-03-09
Fixed potential overflow in multiview vertex buffer validation

m3x3:
Fixed a potential integer overflow in the "valid" path of multiview vertex buffer validation when `max_instance_index` equals `u32::MAX`. In this case, the addition `max_instance_index + 1` on the right-hand side of the comparison overflows even for valid cases. The potential overflow in the error path (already marked "TODO") is not fixed by this.

I got hit by this issue when trying to run the multiview example on a Radeon 6700 XT with the amdgpu driver. The MESA driver apparently returns i32::MAX for `max_instance_index` and so it doesn't cause this overflow.


Rua:
Thanks for catching this!

PR #1861, opened 2022-03-15
Export enum MemoryMapError

tutan-chromium:
Export MemoryMapError which is required by MappedDeviceMemory::new()


Rua:
Thank you for the fix!

PR #1864, opened 2022-03-23
Support loading dynamic libraries from multiple paths and try more paths in default loader

ewencp:
Change the DynamicLibraryLoader::new interface to accept a slice of paths
instead of a single path. It tries to load them in order, using the first one it is
able to load. Update the auto_loader implementation to use the updated
version, including a larger set of paths for the Vulkan library.

```
 - `**Breaking** DynamicLibraryLoader::new now accepts a slice of paths.`
```

        - `Entry 1.`
        - `Entry 2.`...

### DIFF
--- a/vulkano/src/instance/loader.rs
+++ b/vulkano/src/instance/loader.rs
@@ -79,31 +79,49 @@ pub struct DynamicLibraryLoader {
 }
 
 impl DynamicLibraryLoader {
-    /// Tries to load the dynamic library at the given path, and tries to
-    /// load `vkGetInstanceProcAddr` in it.
+    /// Tries to load the dynamic library at the given paths, using the first successfully loaded,
+    /// and tries to load `vkGetInstanceProcAddr` in it.
     ///
     /// # Safety
     ///
     /// - The dynamic library must be a valid Vulkan implementation.
     ///
-    pub unsafe fn new<P>(path: P) -> Result<DynamicLibraryLoader, LoadingError>
+    pub unsafe fn new<P>(paths: &[P]) -> Result<DynamicLibraryLoader, LoadingError>
     where
         P: AsRef<Path>,
     {
-        let vk_lib = shared_library::dynamic_library::DynamicLibrary::open(Some(path.as_ref()))
-            .map_err(LoadingError::LibraryLoadFailure)?;
+        let mut errors = Vec::new();
+        let vk_lib = paths.iter().find_map(|path| {
+            shared_library::dynamic_library::DynamicLibrary::open(Some(path.as_ref()))
+                .or_else(|e| {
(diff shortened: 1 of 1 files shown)


ewencp:
For reference, the list of paths is the same as [used by SDL](https://github.com/libsdl-org/SDL/blob/release-2.0.20/src/video/cocoa/SDL_cocoavulkan.m#L40). This makes development on macos using the home-brew molten-vk package simpler since with this update the dylib can be found automatically (assuming paths are configured correctly). Otherwise you are forced to load the FunctionPointers yourself.

vulkano/src/instance/loader.rs
```diff
@@ -79,31 +79,49 @@ pub struct DynamicLibraryLoader {
...
+                        .symbol("vkGetInstanceProcAddr")
+                        // Arguably we could continue trying subsequent paths here, but we don't
+                        // expect to find a matching filename without the right contents, so better
+                        // to fail here, especially since an unknown library is now loaded into the
+                        // process.
+                        .map_err(|_| {
+                            LoadingError::MissingEntryPoint("vkGetInstanceProcAddr".to_owned())
+                        })?;
+                    mem::transmute(ptr)
+                };
```
AustinJ235, line 116: `DynamicLibrary` seems to implement `Drop`, so would it be safe to continue attempting paths?
https://docs.rs/shared_library/0.1.9/shared_library/dynamic_library/struct.DynamicLibrary.html#impl-Drop

I don't know much about loading library, so I don't know what is safe and what isn't.

vulkano/src/instance/loader.rs
```diff
@@ -79,31 +79,49 @@ pub struct DynamicLibraryLoader {
...
+
+                Ok(DynamicLibraryLoader {
+                    vk_lib: dl,
+                    get_proc_addr,
+                })
+            }
+            None => Err(LoadingError::LibraryLoadFailure(errors.join(" "))),
```
AustinJ235, line 123: I not sure I am a huge fan of the space separated errors. `shared_library` isn't very helpful with just returning a `String`. Perhaps the errors should be a `Vec<LoadingError { path: Path, error: String }>`?

AustinJ235 (commented):
Overall this is probably a good change for MacOS. Would other platforms also have multiple paths?

i.e. sdl's wayland code has this:
```c
#if defined(__OpenBSD__)
#define DEFAULT_VULKAN  "libvulkan.so"
#else
#define DEFAULT_VULKAN  "libvulkan.so.1"
#endif
```

Rua:
Vulkano uses Ash, and Ash has its own loading implementation. Perhaps these changes would be useful for them as well? I've been considering making Vulkano use Ash's loading code so then Vulkano would also benefit.

Issue #1417, opened 2020-09-26
Compile error in vulkano-win on Android


timwillett4:
The android build is broken for vulkano-win.  

It appears that this change from winit broke it: https://github.com/rust-windowing/winit/pull/1556

# Command line output
c:\Users\timwi\source\repos\vulkano\vulkano-win>cargo apk build --verbose
...
warning: unused import: `winit::platform::android::WindowExtAndroid`
   --> vulkano-win\src\winit.rs:134:9
    |
134 |     use winit::platform::android::WindowExtAndroid;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

**error[E0599]: no method named `native_window` found for reference `&winit::window::Window` in the current scope
   --> vulkano-win\src\winit.rs:136:56
    |
136 |     Surface::from_anativewindow(instance, win.borrow().native_window(), win)
    |                                                        ^^^^^^^^^^^^^ method not found in `&winit::window::Window`**

error: aborting due to previous error; 1 warning emitted

For more information about this error, try `rustc --explain E0599`.
error: could not compile `vulkano-win`.

Caused by:
[...]

PR #1868, opened 2022-04-04
fix for android surface acquiring

Dimkar3000:
fixes [issue#1417](https://github.com/vulkano-rs/vulkano/issues/1417#issuecomment-1087141236)
  - `Fix for android surface acquiring.`

1. [x] Run `cargo fmt` on the changes.
  - Running `cargo fmt` seams to have sorted some imports in [vulkano/src/command_buffer/mod.rs](https://github.com/vulkano-rs/vulkano/compare/master...Dimkar3000:android-fix?expand=1#diff-b0e244ebc3759b71d4b842e5d4ed912503552c51a5c40f7771f018b68f90a58d)

3. [ ] Update documentation to reflect any user-facing changes - in this repository.

### DIFF
--- a/vulkano-win/src/winit.rs
+++ b/vulkano-win/src/winit.rs
@@ -121,9 +121,13 @@ unsafe fn winit_to_surface<W: SafeBorrow<Window>>(
     instance: Arc<Instance>,
     win: W,
 ) -> Result<Arc<Surface<W>>, SurfaceCreationError> {
-    use winit::platform::android::WindowExtAndroid;
-
-    Surface::from_android(instance, win.borrow().native_window(), win)
+    use raw_window_handle::HasRawWindowHandle;
+    use raw_window_handle::RawWindowHandle::AndroidNdk;
+    if let AndroidNdk(handle) = win.borrow().raw_window_handle() {
+        Surface::from_android(instance, handle.a_native_window, win)
+    } else {
+        unreachable!("This should be unreachable if the target is android");
+    }
 }
 
 #[cfg(all(unix, not(target_os = "android"), not(target_os = "macos")))]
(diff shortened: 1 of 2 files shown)


vulkano-win/src/winit.rs
```diff
@@ -121,9 +121,13 @@ unsafe fn winit_to_surface<W: SafeBorrow<Window>>(
     instance: Arc<Instance>,
     win: W,
 ) -> Result<Arc<Surface<W>>, SurfaceCreationError> {
-    use winit::platform::android::WindowExtAndroid;
-
-    Surface::from_android(instance, win.borrow().native_window(), win)
+    use raw_window_handle::HasRawWindowHandle;
```
Rua, line 124: `raw_window_handle` is behind a feature gate in Cargo.toml. It's enabled by default, but if the user decides to disable it and use only the winit feature, this code won't work anymore.

Dimkar3000: interesting, the reason I chose this solution is because the other one was silently deprecated. This seams to be the way, the developer intended for this to work. I don't know how to fix this properly, maybe require the feature for android?

Rua: Since they suggested always using `raw_window_handle` (on any platform), I think removing `optional = true` from the dependency and having it always there makes the most sense. Then the code in winit.rs can use this for all platforms instead of Winit's various custom traits like `WindowExtUnix` and `WindowExtWindows`. It's a bit more of a change so if you want to do it only for Android that's fine, I can do it for the other platforms later.

Dimkar3000: I dont know why it is optional, maybe they have their reasons. On this end we could probably add the requirement in the README so everyone knows that the feature is required. I don't have access to all the platforms to properly test a change like that. I could do it for windows.

Rua:
I've merged it, and will look at the things I mentioned myself sometime. Thank you for the fix!

PR #1869, opened 2022-04-08
Allow `SyncCommandBufferBuilder` commands to specify the access range

Rua:
This changes the internals of `SyncCommandBufferBuilder` some more, making two changes:
1. Each resource is checked for conflicts before any resources are added to the builder.
2. Commands can now specify which part of the resource they use, instead of defaulting to the whole thing.

Point 1 means that "poisoning" is no longer needed and has been removed. Point 2 doesn't do much quite yet, but there are TODO items on each command; future changes should be made to make the declared resource ranges match what the command actually accesses. I've already been working on doing this for copy/blit commands, but it's needed for others too.


AustinJ235 (approved):
Nothing stands out and no regressions noted. Looks Good!

PR #1873, opened 2022-04-11
Rewrite copy commands

### LINKED ISSUE #1553: Allow multiple regions for copy operations
Vulkan's copy operations allow specifying multiple regions to copy over, and the `blit_image`, `copy_buffer`, `copy_image`, `copy_buffer_to_image` and `copy_image_to_buffer` methods of `SyncCommandBufferBuilder` already support this. But `AutoCommandBufferBuilder` doesn't support it yet. It would be nice if it did.

I investigated a bit whether I could make a PR for this. Ideally, the functions would take `IntoIterator` of some kind, but a few issues got in my way:
- Arrays don't implement `IntoIterator` yet. That's being worked on in https://github.com/rust-lang/rust/pull/65819, but until then, you have to use something else (`Vec`, `ArrayVec`, `SmallVec` or whatever else the user has on hand).
- The function has to validate the regions, which means using up the iterator. In order to pass the regions on to `SyncCommandBufferBuilder`, it then has to build a new container.
- It would be nice to have only one function, instead of separate `_dimensions` ones that Vulkano currently has. But for that to work, it should be ergonomic to use in the common case, where the dimensions are the whole image.

Rua:
Changelog:
```markdown
- **Breaking** The copy, blit, clear, fill and update commands are completely rewritten, and now use `Info` structs to specify the parameters. They now allow you to specify multiple regions at once to copy, give you fill control over the image subresources, and also let you select image layouts.
- **Breaking** The `transfer_source` and `transfer_destination` fields of `BufferUsage` and `ImageUsage` are renamed to `transfer_src` and `transfer_dst` to match Vulkan.
- **Breaking** `SubImage` has been removed.
- **Breaking** The `conflict_key` method on the `BufferAccess` and `ImageAccess` traits is removed.
```
```markdown
- Added `subresource_layers` and `subresource_range` methods to `UnsafeImage` and `ImageAccess` to easily generate these types from an image.
- Added support for the `khr_copy_commands2` device extension.
- Added the `resolve_image` command buffer command.
```

Depends on #1869, fixes #1553.

These commands now have much better support for all the functionality that Vulkan exposes. Additionally, they use the finer-grained synchronization that was added in previous PRs. I tried to make the API as ergonomic as possible, and some examples show how the API should be used now.


AustinJ235 (approved):
Looks good! Very welcomed change!

EDIT: there appears to be some conflicts with imports.

PR #1875, opened 2022-04-13
Add document detailing how much of Vulkan is currently covered by Vulkano

Rua:
Changelog:
```markdown
- Added `COVERAGE.md`, a document detailing how much of Vulkan is currently covered by Vulkano.
```

This document should be useful for users and developers alike to see at a glance what is in Vulkano and what is still missing. I decided not to include unsupported unpromoted extensions, because that's basically "everything else", and some of them are things we may never want to add support for.


AustinJ235 (approved):
Awesome!

PR #1878, opened 2022-04-16
Rename debug commands, add some new ones

### LINKED ISSUE #1872: No support for vk_ext_debug_utils during instance creation
# Template

If you dont understand something just leave it.
If you can provide more detailed information than the template allows for, please ignore the template and present all of your findings.

*   Version of vulkano: 0.29.0
*   OS: Ubuntu 20.04
*   GPU (the selected PhysicalDevice): nVidia RTX 2080
*   GPU Driver: TODO
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: The triangle example demonstrates it just fine.

# Issue

The library supports enabling debug output after the Vulkan instance has been created with code like this...

```
let debug_report_callback = DebugCallback::new(&instance, MessageSeverity::all(), MessageType::all(), |msg|{
    println!("Debug callback: {:?}", msg.description);
});
```

However, it doesn't support debugging output during _instance_ creation.  In C/C++ this is accomplished by passing the `VkDebugUtilsMessengerCreateInfoEXT` type in the `pNext` chain of the `VkInstanceCreateInfo`.  See [this](https://github.com/KhronosGroup/Vulkan-Samples/blob/master/framework/core/instance.cpp#L317) block of code for an example from the Khronos Vulkan-Samples repository.  

Instance creation failing is one of the hardest issues for new developers to deal with because many tutorials and examples simply assume success.

Rua:
Changelog:
```markdown
- **Breaking** The debug command buffer commands have been renamed to match Vulkan. They now take a `DebugUtilsLabel` value.
- **Breaking** `end_debug_utils_label` is now `unsafe`, as it requires validation checks that are not implemented yet.
- **Breaking** `DebugCallback` is renamed to `DebugUtilsMessenger` to match Vulkan, and now takes a `DebugUtilsMessengerCreateInfo` to construct. It is now `unsafe` to create, because the the callback cannot call any Vulkan API functions, which Vulkano is unable to check.
```

```markdown
- Added debug utils commands to `Queue`.
- Added `Instance::with_debug_utils_messengers`, to provide creation info for messengers that should be used at instance creation and destruction time. This function is also `unsafe`.
```

Fixes #1872. This PR brings some of the debug stuff up to spec and adds a few missing features. Not everything in this extension is implemented yet, but now at least there's more of it.


AustinJ235 (approved):
Looks good!

examples/src/bin/debug.rs
```diff
@@ -15,7 +16,10 @@ use vulkano::{
     format::Format,
     image::{ImageDimensions, ImmutableImage, MipmapsCount},
     instance::{
-        debug::{DebugCallback, MessageSeverity, MessageType},
+        debug::{
+            DebugUtilsMessageSeverity, DebugUtilsMessageType, DebugUtilsMessenger,
```
jherico, line 20: The advantage of having a language with namespace support is that you can avoid the naming verbosity of the C structs.  I'd argue that `MessageSeverity`, etc should have remained as the names.

PR #1880, opened 2022-04-16
Use only the selected range of buffer and image views in synchronization

Rua:
Changelog:
```markdown
- **Breaking** `ImageViewCreateInfo` now has a single `subresource_range` field instead of separate fields for aspect, mip levels and array layers.
- **Breaking** The `aspects`, `mip_levels` and `array_layers` methods of `ImageViewAbstract` are removed, and replaced with a single `subresource_range` method.
- **Breaking** The `current_mip_levels_access` and `current_array_layers_access` methods of `ImageAccess` are removed.
```

```markdown
- `BufferViewAbstract` now has a `range` method that returns the byte range of the underlying buffer that the view covers.
```

Command buffer synchronization now uses the specified ranges of buffer and image views instead of the whole buffer/image. I've also made it so that buffer and image ranges are passed correctly from a secondary command buffer to a primary one.

### DIFF
--- a/vulkano/src/buffer/view.rs
+++ b/vulkano/src/buffer/view.rs
@@ -42,22 +42,21 @@
 //! ).unwrap();
 //! ```
 
-use crate::buffer::{BufferAccess, BufferAccessObject, BufferInner};
-use crate::device::Device;
-use crate::device::DeviceOwned;
-use crate::format::Format;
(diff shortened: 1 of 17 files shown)


AustinJ235 (approved):
Look Good!

Issue #1879, opened 2022-04-16
Multi window game of life example crashes


Rua:
@hakolao This example crashes ever since additional validation checks were added in #1878. `DebugUtilsMessageSeverity::none()` is provided to the debug messenger, but that's not allowed, the value must not be empty.

PR #1883, opened 2022-04-19
Fix debug utils message severity error

hakolao:
Would this suffice for now? @Rua to #1879


Rua:
Yes, that's fine, thanks!

PR #1884, opened 2022-04-23
Improve render pass commands, more validation

Rua:
Changelog:
```markdown
- **Breaking** `begin_render_pass` now takes a `RenderPassBeginInfo` struct.
- **Breaking** `ClearRect` now has a single `Range<u32>` for array layers.
- **Breaking** The fields of `ClearAttachment::Color` are now named.
- **Breaking** The `ImageClearValue` trait is removed.
```

Depends on #1880. I improved the validation of render pass commands and added an info struct. There are some missing things that still need addressing in relation to render passes:
- The correctness of subpass dependencies isn't checked at all when beginning a render pass.
- Render pass attachments aren't checked for conflicts. For example, if the same image is used for multiple attachments, and they have different image layouts set, or if attachments are concurrently written.
- Draw commands don't track attachment usage.
- A few checks can't be performed in secondary command buffers, because they depend on knowing the parameters of the render pass in the primary command buffer. These need to be checked when the secondary command buffer is added to the primary.


AustinJ235 (approved):
Looks good! I went ahead a resolved the conflicts.

Issue #1219, opened 2019-07-16
Using dispatch with dimension 0 along any axis triggers a kernel crash, or freezes the OS


loovjo:
*   Version of vulkano: 0.13
*   OS: macOS Mojave
*   GPU (the selected PhysicalDevice): Intel HD Graphics 5000 1536 MB
*   GPU Driver: MoltenVK
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: https://gist.github.com/loovjo/cae4ed5fbda88c4d265c47d40bc00d40

# Issue

When calling `AutoCommandBufferBuilder::dispatch` with any of the three dimensions set to zero, the OS completely freezes, and then the kernel sometimes panics.

I'm not completely sure whether this is an issue with vulkano itself, if it is an issue in Vulkan, or if it is a problem with MoltenVK. Either way, as vulkano has a goal of preventing invalid API usage, this should either be checked for, or the function should be marked as unsafe with a note in the documentation.

As the check would be very inexpensive, I would suggest putting in the check either at the `dispatch` call, or at the execution of the command buffer.

(Some of this information might be incorrect, as having a kernel panic and restarting the computer is something quite time consuming, and as I don't have a lot of time at my hands, some of the conclusions I have drawn might be incorrect as it has been based on quite few tests. If someone would want to test this more exhaustively, that would be appreciated.)

Rua:
Should be fixed by #1886.

PR #1886, opened 2022-04-29
Disallow dispatch calls with dimensions of zero-length

ryco117:
1. [x] This is meant to address #1219 by adding an additional check to `check_dispatch()` for a zero-length dimension

        - Added new enum value `CheckDispatchError::ZeroLengthDimensions` to be returned when `dispatch()` is called with dimension(s) of length zero.

Let me know if there are any concerns! (and yes, this issue is pretty low-hanging fruit :P )


Rua:
Seems like a simple enough fix, thanks!

PR #1888, opened 2022-05-02
Raytracing shaders with vulkano-shaders

Atilogit:
This PR adds functionality to compile raytracing shaders.
This was part of a project to add partial raytracing support to vulkano but i didn't get around to finishing it.

The code in `vulkano/src/shader/reflect.rs` is only a workaround. I didn't have the time to get far enough to do it properly.

        - vulkano-shaders now supports `raygen`, `anyhit`, `closesthit`, `miss`, `intersection` and `callable` shaders


Rua:
Looks good, thank you! For future reference, the workaround in reflect.rs, what should actually be done there?

Atilogit:
Not sure. I didn't spend much time with the reflection part of vulkano

Issue #1885, opened 2022-04-26
then_signal_semaphore introduces VUID-vkQueuePresentKHR-pWaitSemaphores-03268


LeonMatthes:
# Template

*   Version of vulkano: "0.29"
*   OS: Fedora 35
*   GPU (the selected PhysicalDevice): GTX 1070 (desktop)
*   GPU Driver: nvidia
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: https://github.com/LeonMatthes/Vulkano-Hello-Triangle/commit/72720e934502dc967cc75694d326a7dd9277cb5a

# Issue

I've tried following vulkan-tutorial.com and translating the code to vulkano as closely as possible.
The code that errors should therefore be the minimal required to get to the triangle.
The problem occurs when I run the program in debug mode (with validation layers enabled).

At first, the triangle will render correctly, however, once the window is resized, the validation layers issue 2 instances of `VUID-vkQueuePresentKHR-pWaitSemaphores-03268` and the application becomes unresponsive.

```
VUID-vkQueuePresentKHR-pWaitSemaphores-03268(ERROR / SPEC): msgNum: 622825338 - Validation Error: [ VUID-vkQueuePresentKHR-pWaitSemaphores-03268 ] Object 0: handle = 0x55d11b9896b0, type = VK_OBJECT_TYPE_QUEUE; Object 1: handle = 0x980f360000000011, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x251f8f7a | vkQueuePresentKHR: Queue VkQueue 0x55d11b9896b0[] is waiting on pWaitSemaphores[0] (VkSemaphore 0x980f360000000011[]) that has no way to be signaled. The Vulkan spec states: All elements of the pWaitSemaphores member of pPresentInfo must reference a semaphore signal operation that has been submitted for execution and any semaphore signal operations on which it depends (if any) must have also been submitted for execution (https://vulkan.lunarg.com/doc/view/1.3.211.0/linux/1.3-extensions/vkspec.html#VUID-vkQueuePresentKHR-pWaitSemaphores-03268)
    Objects: 2
        [0] 0x55d11b9896b0, type: 4, name: NULL
        [1] 0x980f360000000011, type: 5, name: NULL
[...]

PR #1889, opened 2022-05-02
PresentFuture: Avoid flushing twice on flush error

LeonMatthes:
See #1885.
When a `then_swapchain_present` is preceded by a `then_signal_semaphore`, it will cause the application to freeze, if the `then_swapchain_present` call fails (i.e. by a window resize returning an `AcquireError::OutOfDate`).

The problem is that when flushing the `PresentFuture` fails, it will not be marked as either finished, or flushed.
In that case the `Drop` implementation of the PresentFuture will do another flush and then wait on the queue to finish.

The second flush is in itself incorrect, as we now have two `vkQueuePresentKHR` calls for one future.
Even more problematic is that both calls will wait on the previous semaphore.
As only one command can wait on a single Semaphore, the second command will never be signalled.
This then causes the application to freeze when the Drop implementation then waits for the queue to finish, as it never does.

This change always marks a `PresentFuture` as flushed when flush is called, even on failure.
It furthermore changes the drop implementation to only flush, when this hasn't been attempted before.

    - `Fix PresentFuture flushing twice if 'then_swapchain_present' fails`

**Disclaimer**: I'm unsure how to test this. It would require at least one window resize. Also we don't really want the test to freeze if the test fails :thinking:

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -1804,6 +1804,7 @@ where
             // If `flushed` already contains `true`, then `build_submission` will return `Empty`.
 
             let build_submission_result = self.build_submission();
+            self.flushed.store(true, Ordering::SeqCst);
 
             if let &Err(FlushError::FullScreenExclusiveLost) = &build_submission_result {
                 self.swapchain
@@ -1827,7 +1828,6 @@ where
                 _ => unreachable!(),
             }
 
-            self.flushed.store(true, Ordering::SeqCst);
             Ok(())
         }
     }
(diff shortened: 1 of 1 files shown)


LeonMatthes:
Open Question: Is it valid to `flush` a future twice and expect different results?

Rua:
I can't say immediately whether your changes have any flaws, as the `GpuFuture` code is one of the parts of Vulkano I'm least familiar with, and it also deviates quite significantly from how plain Vulkan does it. At some point I would like to bring it a bit closer to Vulkan's API, so that additions like timeline semaphores are easier to implement, but at the same time I don't want to sacrifice the ergonomics too much.

The [documentation of `flush` says](https://docs.rs/vulkano/latest/vulkano/sync/trait.GpuFuture.html#tymethod.flush):
> The implementation must remember that it was flushed. If the function is called multiple times, only the first time must result in a flush.

So I would assume that, if the first flush did anything, the second flush should do nothing.

PR #1891, opened 2022-05-06
Fixing Module comments that were accidentally changed

ryco117:
Whoops. I appreciate the speed with-which https://github.com/vulkano-rs/vulkano/pull/1890 was merged, but I notice now that I had accidentally changed the leading module comments to be outer-line doc comments.

This is a simple commit to fix those comments again.


Rua:
No problem, I appreciate the fix!

PR #1902, opened 2022-05-28
Add support for dynamic rendering

Rua:
Changelog:
```markdown
- **Breaking** The various `AutoCommandBufferBuilder` constructors for secondary command buffers have been merged into one `secondary` function, which directly takes a `CommandBufferInheritanceInfo` value.
- **Breaking** The `render_pass` values of `GraphicsPipelineBuilder` and `CommandBufferInheritanceInfo` have both been changed into an enum that selects between rendering with `begin_render_pass` and rendering with `begin_rendering`. They implement `Into` for easy conversion.
- **Breaking** Added the missing `rasterization_samples` field to `MultisampleState`, which must be provided when doing multisampled rendering.
```
```markdown
- Added support for dynamic rendering, and a `triangle-v1_3` example demonstrating how it's used.
```

Dynamic rendering is a new method of rendering where you don't need to create a render pass or framebuffer object, and instead specify the rendering parameters right when you actually begin rendering in the command buffer (and some of it also when creating the graphics pipeline). The usage envisioned by Khronos is that for simple rendering tasks that don't need multiple render subpasses, you use this, and for more complex things you use render pass objects.

This has been a pretty tough one, so I hope I got it all right. Not all the details are finished, but there are other unrelated issues that future PRs need to fix, and the missing checks for dynamic rendering might as well be included in that.


AustinJ235 (approved):
Holy molly, great work! Nothing really stands out to me took a bit to get through. Great work documenting all the VUID's as usual.

PR #1903, opened 2022-05-28
Remove `Default` from `NonExhaustive`

Rua:
Changelog:
```markdown
- Fixed a bug where `NonExhaustive` implemented the `Default` trait and was therefore still constructable by the user.
```

A small bug that I missed in the previous release.


AustinJ235 (approved):
Whoops

PR #1905, opened 2022-05-31
Fix #1881

### LINKED ISSUE #1881: Incorrect enumeration behavior
https://github.com/vulkano-rs/vulkano/blob/661f5713338e8ba03f7f54c085097d001585c250/vulkano/src/instance/extensions.rs#L62-L78

The implementation of an "enumerate" function as a simple "fetch the number, then populate the results with that number" is a common mistake in Vulkan code.  

While rare, code must deal with possibility that the number returned by the first call won't be valid by the time of the second call, and that the second call will have populated the `num` value with a different number than the original (in this case, the function should also have returned `VK_INCOMPLETE`)

Consider the implementation of the C++ bindings `vk::enumerateInstanceExtensionProperties` method:

```
    do
    {
      result = static_cast<Result>( d.vkEnumerateInstanceExtensionProperties( layerName ? layerName->c_str() : nullptr, &propertyCount, nullptr ) );
      if ( ( result == Result::eSuccess ) && propertyCount )
      {
        properties.resize( propertyCount );
        result = static_cast<Result>( d.vkEnumerateInstanceExtensionProperties( layerName ? layerName->c_str() : nullptr, &propertyCount, reinterpret_cast<VkExtensionProperties*>( properties.data() ) ) );
      }
    } while ( result == Result::eIncomplete );
```
This creates a loop that will almost invariably exit after the first pass, but which will still handle the case where the number of exposed extensions changes at runtime between the first and second calls.  This same logic should be applied to all the other functions with potentially return `VK_INCOMPLETE`.  Searching for `Result::eIncomplete` in [this](https://raw.githubusercontent.com/KhronosGroup/Vulkan-Headers/master/include/vulkan/vulkan_funcs.hpp) file gives a good list of them I believe.

Rua:
Changelog:
```markdown
- Fixed bug in various Vulkan calls where the returned data might be incomplete.
```

Fixes #1881.


AustinJ235 (approved):
Looks good!

PR #1906, opened 2022-06-02
fix errors I noticed in readme

coolbot123:
saw some errors while reading the readme then did a very _very_ quick proof-read/edit.


AustinJ235:
Thanks for the fixes. There was one instance of "contribution", that I am not why you capitalized, so I revered that.

PR #1909, opened 2022-06-05
Refactor query commands

Rua:
This PR refactors the query commands, in the same way that other command buffer commands have already been done. There are no breaking changes (other than to error types, which I haven't included in changelogs so far). After this, only the bind and the pipeline commands are left.


AustinJ235 (approved):
Looks good!

PR #1910, opened 2022-06-05
Add `is_signaled` to `FenceSignalFuture`

### LINKED ISSUE #1907: No ready() -> bool function in FenceSignalFuture
*   Version of vulkano: "0.29.0"

In the codebase I see that `Fence` has two associated functions: `ready(&self) -> Result<bool, OomError>` and `wait(&self, timeout: Option<Duration>) -> Result<(), FenceWaitError>`. When I'm constructing a future like in the examples I'm getting an instance of `FenceSignalFuture` (after calling `then_signal_fence_and_flush`). 

But I want to keep updating my scene on CPU, while checking each time if previous frame finished rendering and freed access to storage buffers so that I can upload new data to them. For that I wanted to have some function like `ready` to check if future is ready without blocking the cpu.

Can that be added? If no, then how can I get the behavior I described?
I can try to make a PR for that, but I'm not sure right now if I'm using the API in a correct way, so asking for a suggestion

Rua:
Changelog:
```markdown
- **Breaking** Renamed the `ready` method of `Fence` to `is_signaled`.
```

```markdown
- Added an `is_signaled` method to `FenceSignalFuture`.
```

Fixes #1907. I renamed the function to better fit Rust method naming guidelines and Vulkan terminology.

### DIFF
--- a/vulkano/src/command_buffer/submit/queue_submit.rs
+++ b/vulkano/src/command_buffer/submit/queue_submit.rs
@@ -313,14 +313,14 @@ mod tests {
             let (device, queue) = gfx_dev_and_queue!();
 
             let fence = Fence::new(device.clone(), Default::default()).unwrap();
-            assert!(!fence.ready().unwrap());
+            assert!(!fence.is_signaled().unwrap());
 
             let mut builder = SubmitCommandBufferBuilder::new();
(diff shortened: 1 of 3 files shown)


AustinJ235 (approved):
Simple enough and useful.

PR #1911, opened 2022-06-06
Refactor dynamic state commands

Rua:
Changelog:
```markdown
- **Breaking** The `set_depth_bounds` parameter is now a single `RangeInclusive` instead of two separate `f32`, to match the type used on the `DepthBoundsState` struct.
```

Mostly mechanical work in this one, and one breaking change.


AustinJ235 (approved):
Looks good!

PR #1927, opened 2022-07-18
Support for VK_KHR_portability_subset Devices

AustinJ235:
Adds support for `VK_KHR_portability_subset` devices.

**Changelog:**
```
- **Breaking** Removed `DeviceExtensions::required_extensions()`
- **Breaking** Removed `PhysicalDevice::required_extensions()`
- `InstanceCreateInfo` now has `enumerate_portability` field.
```


AustinJ235:
> What should Vulkano do with the `enumerate_portability` field if the `khr_portability_enumeration` extension isn't available? The previous default behaviour, before the extension was introduced, was equivalent to setting it to `true`. Should Vulkano enforce that the flag is always `true` for such cases, or should we explicitly document that Vulkan(o) behaves as if it's set to `true` and the user-set value is ignored?

Documentation add with https://github.com/vulkano-rs/vulkano/pull/1927/commits/a6843efc6fb06e16ed27e1f41813a6a32e30d20f

PR #1928, opened 2022-07-19
Move RangeMap into Vulkano

AustinJ235:
This pull request removes the `rangemap` dependency and moves the `RangeMap` type and needed supporting types into vulkano.

### DIFF
--- a/vulkano/src/lib.rs
+++ b/vulkano/src/lib.rs
@@ -89,7 +89,7 @@ pub mod instance;
 pub mod memory;
 pub mod pipeline;
 pub mod query;
-pub mod range_map;
+mod range_map;
 pub mod range_set;
 pub mod sampler;
 pub mod shader;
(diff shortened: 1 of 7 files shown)


vulkano/src/lib.rs
```diff
@@ -89,6 +89,7 @@ pub mod instance;
 pub mod memory;
 pub mod pipeline;
 pub mod query;
+pub mod range_map;
```
Rua, line 92: Does this need to be public? I don't know if there are any interfaces exposed to the user that use it currently, and it's better to be conservative in this regard.

AustinJ235: Now private as of [66c3f33](https://github.com/vulkano-rs/vulkano/pull/1928/commits/66c3f33ee9d7e0fe2ad8cb8df684a96dda020528)

PR #1932, opened 2022-07-24
Add `VulkanLibrary` as first initialization step before `Instance`

Rua:
Changelog:
```markdown
- **Breaking** Changes to `Instance` and Vulkan initialization:
  - `FunctionPointers` is renamed to `VulkanLibrary`, and now resides in a separate `library` module. It is re-exported from the crate root.
  - The `Loader` trait is now in the `library` module.
  - `Instance` now requires a `VulkanLibrary` object, which you must create beforehand.
  - The `auto_loader` function is removed.
  - Supported extensions and layers are now retrieved from the `VulkanLibrary` object. The old `layers_list` and `InstanceExtensions` methods are removed.
  - The deprecated methods of `DeviceExtensions` are removed.
  - Vulkano-win: `required_extensions` now requires a reference to `VulkanLibrary`.
- **Breaking** Changes to Vulkano-util:
  - Required instance extensions for surface creation aren't added until `VulkanoContext` is created.
  - The `instance`, `device`, `graphics_queue` and `compute_queue` methods of `VulkanoContext` now return a reference instead of an owned `Arc`.
```

This closes #1876 by partially reimplementing it, but differently.

This is still a draft, because I ran into a problem making Vulkano-util conform to the new API. `VulkanoConfig` has a `Default` implementation, but that no longer works now that the user must load the library themselves. @hakolao do you know of a good solution to this? If you think it's better, I can leave this PR as is, and give you the opportunity to fix it yourself later.


hakolao:
I’ll try to come up with something later today, cant think of a good solution right away.

hakolao:
@Rua How about something like [this](https://github.com/hakolao/vulkano/commit/de2254efa01f3d5a19dc81a12f559ea75ce83c34)?

So the required extensions can be appended to the inputs from config. That way `Default` can remain `None`, and user can still add the extensions that they want, which are not required extensions.

On top of that, I think that the `VulkanLibrary` should probably be accessible from `VulkanoContext`? Unless it can be retrieved through instance. Either would be fine I think.

Rua:
This is now ready for review.

AustinJ235 (approved):
Looks good!

Issue #1935, opened 2022-07-30
Add `from_handle` constructors for various Vulkan object types.


Rua:
On Discord, a wish was expressed for the ability to create Vulkano types from raw Vulkan object handles (whose types are defined by Ash). Although there are many object types to implement this for, I think this would be relatively easy for new contributors to implement. Some implementation guidelines:
- The new constructor, I'm naming it `from_handle` for the moment, should be `unsafe`.
- Whereas the normal `new` constructor usually takes a `Device` and some kind of `Info` struct, the `from_handle` constructor would take a `Device`, a raw handle of the appropriate Ash type, and the `Info` struct.
- The function must have a `# Safety` header explaining that the handle must be a valid Vulkan object handle, and that the parameters in the given `Info` struct must match those that were previously used to create the handle.
- Because the handle is assumed to be already valid, little to no validation will be needed inside `from_handle`. The data in the `Info` struct could probably just be moved into the Vulkano type without any checks.

If anyone wants to try implementing this, but is unsure about the number of changes required, you can start by implementing it only for a single, not too complex type (e.g. `Sampler`, `QueryPool`).

p3t3rix:
So i tried my hand at the QueryPool by liberally copying parts of the ctor, something along those lines:
```
/// Creates a new `QueryPool` from an ash-handle
/// # Safety
/// The `handle` has to be a valid vulkan object handle and 
/// the `create_info` must match the info used to create said object
pub unsafe fn from_handle(device: Arc<Device>,
                          create_info: QueryPoolCreateInfo,
                          handle : ash::vk::QueryPool
) -> Result<Arc<QueryPool>, QueryPoolCreationError>{
    let QueryPoolCreateInfo {
        query_type,
        query_count,
        _ne: _,
    } = create_info;
    return Ok(Arc::new(QueryPool {
        handle,
        device,
        query_type,
        query_count,
    }));
}
```
- Is this the right direction or am i missing something ? 
- Also should there be tests for these and if so how should they look like ?
- Should this even return an error as the object is already created ?

Rua:
Yes, I think that's the right direction. Personally I think the `handle` parameter should come before `create_info`, but that's just bikeshedding and I can't say why it seems better. There shouldn't be an error return, because there are no errors to report; like you say, the object is already created.

I don't think tests are necessary, because there isn't really anything to test, as far as I can think of.

p3t3rix:
[...]

PR #1938, opened 2022-07-31
 Add from_handle constructors for various Vulkan object types. #1935 

p3t3rix:
Add from_handle functions so there is a way to use ash directly.

### DIFF
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -182,6 +182,28 @@ impl UnsafeBuffer {
         Ok(Arc::new(buffer))
     }
 
+    /// Creates a new `UnsafeBuffer` from an ash-handle
+    /// # Safety
+    /// The `handle` has to be a valid vulkan object handle and
+    /// the `create_info` must match the info used to create said object
+    pub unsafe fn from_handle(
+        handle: ash::vk::Buffer,
+        create_info: UnsafeBufferCreateInfo,
+        device: Arc<Device>,
+    ) -> Arc<UnsafeBuffer> {
+        let UnsafeBufferCreateInfo { size, usage, .. } = create_info;
+
+        Arc::new(UnsafeBuffer {
+            handle,
+            device,
+
+            size,
+            usage,
+
+            state: Mutex::new(BufferState::new(size)),
+        })
+    }
+
     /// Returns the memory requirements for this buffer.
     pub fn memory_requirements(&self) -> MemoryRequirements {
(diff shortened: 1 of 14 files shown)


p3t3rix:
I started by looking for stuff that implements VulkanObject and had a new() function. Not sure if there is a common trait one could extract/implement to make this a bit more streamlined. 
Also as it is now i am not sure there is a consistent direction as for some things we let the user take all the responsibility and others where there is still validation inside (RenderPass for example).
Another thing where i am unsure is how should we handle things like ShaderModule or Buffer. UnsafeBuffer could be constructed but how do we (and should we) handle others that use it internally ?

Rua:
For `RenderPass`, you could extract the code that calculates `views_used` and use that directly, without any of the validation code that surrounds it. I think this code should do it:
```rust
let views_used = subpasses.iter().map(|subpass| u32::BITS - subpass.view_mask.leading_zeros()).max::<u32>();
```

For buffers and images, creating `UnsafeBuffer` and `UnsafeImage` is enough I think. Further adjustments may be needed later but we can do that when it comes to that point. `ShaderModule` probably needs more thought, so you can skip that one for now.

p3t3rix:
I am not sure if i forgot some files, for the following structs i was not sure what to do:
- ShaderModule 
- GraphicsPipeline
- ComputePipeline 
- PipelineCache (not needed as far as i am concerned)
- DeviceMemory
- Instance
- Device
- UnsafeDescriptorSet (has it already under new() but not public )
- UnsafeCommandPoolAlloc
- UnsafeCommandBuffer

for some cases i duplicated logic because the validate method returned stuff (e.g. UnsafeImage).

Rua:
This is very good already, thanks!

In the past, I made some validation functions return values, and in some cases the `Info` struct is also modified, but in hindsight I think this was a mistake. In the future we may want to add the ability to bypass the validation, and everything else should still work if the validation function isn't called. That means the validation functions should all be made read-only at some point, and return only the error. You don't have to do anything about this, just noting that it was a historical oversight on my part.

p3t3rix:
No problem, such is the life with codebases that grow, at some point you look back and think about all the things you would/could do better with the current knowledge 😄 
Also thank you for being so responsive and supportive.

How should we tackle the other points and did i miss some important structs ? Or should i mark this PR ready and the rest is handled in different issues?

Rua:
I think this can be merged as it is. The remaining types can be tackled another time, they probably warrant more individual attention and discussion.

PR #1947, opened 2022-08-07
Use libloading instead of shared_library

Rua:
Changelog (add to existing headers):
```markdown
- **Breaking** Changes to `Instance` and Vulkan initialization:
  - `VulkanLibrary::get_instance_proc_addr` is now `unsafe`, and returns `ash::vk::PFN_vkVoidFunction`.
```

This replaces the `shared_library` crate with `libloading`. The original library was created by Tomaka, but has not been maintained since. The replacement crate is more up to date and much more widely used.


AustinJ235 (approved):
Looks good!

PR #1951, opened 2022-08-10
Switch to `parking_lot`'s `Mutex`es

marcot1cs:
Title says all. I didn't bother with `CpuBufferPool` because of discussion on Discord about removing synchronization from it entirely.


Rua:
Seems like a good idea, thanks!

PR #1952, opened 2022-08-11
Refactor internal error handling

Rua:
This renames the internal `Error` type to `VulkanError` (so that it no longer conflicts with `std::error::Error`) and auto-generates it from vk.xml. The `check_errors` function and the `Success` enum are also removed. There is also a new `impl From<ash::vk::Result> for VulkanError`, which is tolerant to error codes it doesn't know about; it will wrap them in the `Unnamed` variant, whereas `check_errors` would panic before.

My intention for the future is to return `VulkanError` directly to the user. Currently, there are `From<VulkanError>` implementations for every error type, which translate known errors for that function. But this adds a lot of duplication, especially for each `Display` implementation that has to repeat the same error messages for standard Vulkan errors.

Moreover, these `From<VulkanError>` implementations currently panic if a Vulkan API function returns an error that is not expected for that function. Some Vulkan implementations will return errors that aren't in the spec, and Vulkano will therefore indiscriminately crash the user's program based on something they have no control over, which is not very helpful. So instead of trying to translate each known error and panicking on the remainder, returning `VulkanError` directly would be more tolerant of unusual Vulkan implementations, letting the user deal with the unexpected behaviour instead.


AustinJ235 (approved):
Awesome!

PR #1959, opened 2022-08-23
Opaque Windows handle external memory import support

idanraiter:
Adds support for importing OpaqueWin32 and OpaqueWin32Kmt handles. These are the handle variants that come from other Vulkan instances on Windows.

Changelog:
```
- Added support for importing OpaqueWin32 / OpaqueWin32Kmt memory handles.
```


Rua:
Thank you!

PR #1961, opened 2022-08-26
vulkano: export LoadingError

gurchetansingh:
       - This is useful if loading the VulkanLibrary fails.


gurchetansingh:
CC: @Rua

Rua:
Thanks for the fix!

PR #1963, opened 2022-09-02
Refactor draw/dispatch commands, render pass state, improve validation checks

Rua:
No changelog.

This is again a PR with a lot of internal rearranging, but hopefully that's the last of it for command buffers for the moment. I've finished the refactoring of draw/dispatch commands and improved/added validation checks there when needed. I also changed a bit how the render pass state is stored internally, to make these checks a little less messy.


AustinJ235 (approved):
Looks good!

PR #1964, opened 2022-09-05
Macrofy all Vulkan bitflag and enum types, check for device support

Rua:
Changelog:
```markdown
- **Potentially Breaking** Changes to bitflag structs and enums corresponding to Vulkan equivalents, as well as to `InstanceExtensions`, `DeviceExtensions` and `Features`:
  - All bitflag types now:
    - Have a `const fn empty()` constructor.
    - Have `const` methods `is_empty`, `intersects`, `contains`, `union`, `intersection`, `difference`, `symmetric_difference`.
    - Implement the operators `&`, `&=`, `|`, `|=`, `^`, `^=`, `-` and `-=`.
    - Implement `Clone`, `Copy`, `Debug`, `Default`, `PartialEq`, `Eq` and `Hash`.
    - Implement `From` to and from the corresponding Vulkan type (except `InstanceExtensions`, `DeviceExtensions` and `Features`).
  - All enum types now:
    - Implement `Clone`, `Copy`, `Debug`, `PartialEq`, `Eq` and `Hash`.
    - Implement `From` to the corresponding Vulkan type, and `TryFrom` from the corresponding Vulkan type.
  - The `all()` constructor has been removed from types that have it, as it has the potential to break forward compatibility: it can enable new flags that were not anticipated by old code. For the same reason, the `!` operator is not implemented.
  - All values are validated against the device API version and enabled extensions.
  - The constructor `::none()` is deprecated in favour of `::empty()`.
  - The method `is_superset_of` is deprecated in favour of `contains`.
  - `BufferUsage` and `ImageUsage` no longer have constructors for some specific combinations.
```

This makes everything much more consistent, by using macros to generate all these different types. Also important is that these types are now always validated when used, so if a particular bitflag or enum variant is only supported on later versions or with a particular extension, you should now get an error when you try to use them.


AustinJ235 (approved):
Great work!

PR #1965, opened 2022-09-06
Impl Present Wait Feature

AustinJ235:
This implements the `present_id` and `present_wait` features.
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_present_wait.html

```
**Breaking** Changes to swapchain presentation.
  - `swapchain::present` now takes `PresentInfo` parameter.
  - `swapchain::present_incremental` has been removed.
  - `GpuFuture::then_swapchain_present` now takes `PresentInfo` parameter.
  - `GpuFuture::then_swapchain_present_incremental` has been removed.
  - `SubmitPresentBuilder::swapchain` now takes `PresentInfo`.
  - Added support for `present_id` feature with `present_id` field on `PresentInfo`.
  - Added support for `present_wait` feature by adding method `swapchain::wait_for_present`.
```


Rua:
There isn't any validation for VUID-VkPresentIdKHR-presentIds-04999 as far as I can tell. It says:
> Each `presentIds` entry **must** be greater than any previous `presentIds` entry passed for the associated `pSwapchains` entry
This seems to mean that every ID may be used exactly once per swapchain, and they must be in strictly increasing order. Perhaps to ensure this, the swapchain could have an `AtomicU64` inside it. This counter is atomically incremented and returned to the user whenever the user presents with an ID, instead of requiring the user to pass in a `NonZeroU64` value.

AustinJ235:
> There isn't any validation for VUID-VkPresentIdKHR-presentIds-04999 as far as I can tell. It says:

Kind of overlooked that VUID, oops. It is an interesting one to enforce though. Thinking of a solution🤔...

AustinJ235:
@Rua Went ahead an implement `PresentInfoExt`. As this struct doesn't really represent the `VkPresentInfoKHR` structure, but rather is extending structures I added the `Ext` suffix. This removes the needed for the additional methods `_with_id` & `_incremental` and allows them to be used together. Also removed the `NonZeroU64` in favor of plain u64.

`wait_for_present` currently will return an error if `present_id` is zero, but the spec doesn't mention this usage. I think returning an error is this case is more helpful then doing a no-op and the user expecting it to do something.

Rua:
Everything works fine on my end, no problems detected. Thanks for the PR!

PR #1966, opened 2022-09-08
Better validation and errors

Rua:
After #1964 was merged I realised that there are still some shortcomings. This PR fixes those:
- Ability to specify instance extensions, or API version only, for the bitflags and enum macros.
- Ability to validate bitflags and enums with a physical device or an instance too.
- Standardised error types across Vulkano for reporting that something is required from the instance/device but not available (`RequirementNotMet`), and standardised the wording of the messages.
- Use a new `RequiresOneOf` struct type that can specify multiple possible requirements that may satisfy a requirement. This is useful for when extensions get promoted, and so there are multiple ways to get functionality.
- Some random formatting/style changes.

This is all internal stuff, other than the error types, but changes to errors haven't been listed in the changelog for a while now anyway.


AustinJ235 (approved):
Really like the `RequiresOneOf` type, great idea!

PR #1967, opened 2022-09-09
Make `PhysicalDevice` an independent object

Rua:
Depends on #1966.

Changelog:
```markdown
- **Breaking** Changes to `PhysicalDevice`:
  - `PhysicalDevice::enumerate` has been replaced with `Instance::enumerate_physical_devices`. This function returns `Arc<PhysicalDevice>`.
  - Enumerating physical devices multiple times now retrieves the list of devices each time, instead of only at instance creation. This makes it possible to handle devices that are added/removed during runtime.
  - `PhysicalDevice` now owns instead of borrows from its parent `Instance`, so it has no more type parameter.
  - Added `extension_properties`, `memory_properties` and `queue_family_properties` method to get the full properties directly.
  - `MemoryType`, `MemoryHeap` and `QueueFamily` have been removed. Where they were used, you now provide a `u32` index into the `memory_properties().memory_types`, `memory_properties().memory_heaps` and `queue_family_properties()` arrays.
  - `QueueFamily::supports_surface` has been replaced with `PhysicalDevice::surface_support`, to match the Vulkan structure.
  - `QueueCreateInfo` now implements `Default` (with a default `queue_family_index` of 0).
  - `Queue::family` has been renamed to `queue_family_index`.
```

This is based on a request that was made on the Vulkano discord. It doesn't make sense to keep the physical device data in the instance permanently, because devices can be added or removed while the program is running. So now, they are retrieved fresh each time you enumerate.

The changes to `MemoryType`, `MemoryHeap` and `QueueFamily` are made to bring the API a bit closer to Vulkan. While these types did add a layer of type system-level safety, they were also a bit cumbersome to use, and couldn't be stored anywhere because they had a lifetime parameter. So most of the time, you'd end up storing an index anyway.


AustinJ235 (approved):
Solved the git conflicts (Weren't actually any conflicts, just git being silly).

Great work!

PR #1968, opened 2022-09-10
Add sparse image querying functions

Rua:
Changelog:
```markdown
- Added two missing functions for querying about sparse images: `PhysicalDevice::sparse_image_format_properties` and `UnsafeImage::sparse_memory_requirements`.
```

These functions from Vulkan 1.0 were still missing, so this implements them. They'll be useful if anyone ever wants to implement sparse images properly.

I also moved some of the new types from the `physical` module to more appropriate places.


AustinJ235 (approved):
Looks good!

PR #1970, opened 2022-09-10
Fix validation for physical device surface methods

Rua:
This fixes some validation that I missed in #1967. Oops!


AustinJ235 (approved):
Oopsie

PR #1989, opened 2022-09-18
Treat atomic writes in shaders as write accesses

Rua:
Changelog (add to existing section):
```markdown
- Bugs fixed: 
  - Atomic writes in shaders are not treated as write access for the purposes of synchronization.
```


AustinJ235 (approved):
Thanks!

PR #1990, opened 2022-09-18
Release 0.31

AustinJ235:
### Update Versions & Changelog to 0.31

**Dependency Updates**:
```
half: 1 -> 2
vk-parse: 0.7 -> 0.8
```

### DIFF
--- a/vulkano/Cargo.toml
+++ b/vulkano/Cargo.toml
@@ -22,7 +22,6 @@ crossbeam-queue = "0.3"
 half = "2"
 libloading = "0.7"
 nalgebra = { version = "0.31.0", optional = true }
-once_cell = { version = "1.13", features = ["parking_lot"] }
 parking_lot = { version = "0.12", features = ["send_guard"] }
 smallvec = "1.8"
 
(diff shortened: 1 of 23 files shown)


vulkano/Cargo.toml
```diff
@@ -19,7 +19,7 @@ build = "build.rs"
 ash = "0.37"
 bytemuck = { version = "1.7", features = ["derive", "extern_crate_std", "min_const_generics"] }
 crossbeam-queue = "0.3"
-half = "1.8"
+half = "2"
```
Rua, line 22: This should probablt be "2.1" because the minor version can also introduce breaking changes.

AustinJ235: Once an application goes to a `1.x.x` version the first number becomes the major version, the second the minor, and the third revision.

[This guide uses the terms "major" and "minor" assuming this relates to a "1.0.0" release or later. Initial development releases starting with "0.y.z" can treat changes in "y" as a major release, and "z" as a minor release. "0.0.z" releases are always major changes. This is because Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible.](https://doc.rust-lang.org/cargo/reference/semver.html)

Rua: Never mind then.

PR #1992, opened 2022-09-19
Fix handling of runtime sized arrays in shaders

Rua:
Changelog:
```markdown
- Bugs fixed:
  - Incorrect check for descriptor set validity when the shader declares a runtime-sized array.
```

Reported on Discord by @winterle.


AustinJ235 (approved):
Thanks for the investigation/fix.

PR #1993, opened 2022-09-19
Hide queue operations behind a lock

Rua:
Changelog:
```markdown
- **Breaking** Changes to queues:
  - When doing operations on a queue, you must now first call `lock()` on the queue, which prevents concurrent access.
```

Previously, each queue operation would acquire and release the lock individually. If you had to perform multiple operations, then each one would acquire and release the lock, which is less efficient. Now, with a common lock, you can do multiple operations in sequence and only have to lock once.

I also moved the queue stuff into its own module.


AustinJ235 (approved):
Thanks

PR #1997, opened 2022-09-23
Memory allocation revamp

marc0246:
This PR aims to address some long standing issues around allocation, some more important than others:
- The standard memory allocator used to be stored inside the device, which brings with it some limitations. First, accessing it was inefficient, as a mutex needed to be acquired and an `Arc` needed to be cloned on each access. Furthermore, as the memory allocator contained a clone of the device-`Arc`, this caused some nasty issues with cyclic references, which couldn't be resolved without compromises. Second, the user was limited to using this allocator, which now that allocators must be created and managed by the user is resolved. The downside is that this is a little less convenient.
- Efficiency of the standard memory allocator has been improved.
- The user now has more flexibility when it comes to allocating with the built-in memory allocator, in that they can configure it and can choose the allocation algorithm.
- The type parameter `A` for the type of allocation has been removed from all buffers and images, as they never had a use in the first place because custom allocators were not usable. There is now only one allocation type, `MemoryAlloc`, no matter the allocator that is used. This will be very helpful for unifying the buffer and image types, and removing trait objects from types that need to store them, futher improving performance.
- If the user needs to write their own implementation of the `MemoryAllocator` trait for some reason, they do not have to write the suballocation algorithm(s), as the library provides all of them, except stack allocation because implementing one that can be used safely would be too complicated and nullify the performance benefits. With this, implementing the trait is a lot easier and much less error-prone. Implementing the `Suballocator` trait is not possible in any idiomatic capacity, which is a design decision. There really is no need to implement the trait. In fact I had trouble coming up with example docs for some of the ones that are already implemented, because their uses are so obscure.
- `PotentialDedicatedAllocation` has been removed, as all memory allocations share the same type.

Changelog:
```markdown
### Breaking changes
Changes to memory allocation:
- `MemoryPool` has been replaced with `MemoryAllocator` and `StandardMemoryPool` has been replaced with `StandardMemoryAllocator`.
- Removed `AllocFromRequirementsFilter`, `MemoryPoolAlloc`, `MappingRequirement`, `AllocLayout`, `PotentialDedicatedAllocation`, `StandardNonHostVisibleMemoryTypePool`, `StandardNonHostVisibleMemoryTypePoolAlloc`, `StandardHostVisibleMemoryTypePool`, `StandardHostVisibleMemoryTypePoolAlloc` and `StandardMemoryPoolAlloc`.
- Memory allocators must now be managed manually.
  - Buffer and image constructors now take an implementation of the `MemoryAllocator` trait.

### Additions
- Added the `Suballocator` trait and 4 of its implementors: `FreeListAllocator`, `BuddyAllocator`, `PoolAllocator` and `BumpAllocator`.
- Added `GenericMemoryAllocator` and its configuration `GenericMemoryAllocatorCreateInfo`.
- Added `MemoryAlloc`, `AllocationCreateInfo`, `AllocationCreationError`, `MemoryUsage`, `MemoryAllocatePreference`, `MemoryTypeFilter`, `SuballocationCreateInfo` and `SuballocationCreationError`.
```


marc0246:
It's 😙👌

vulkano/src/memory/allocator.rs
```diff
@@ -0,0 +1,4645 @@
...
+    NeverAllocate,
+
+    /// The allocator should always allocate `DeviceMemory`.
+    ///
+    /// This option is best suited if you are allocating a long-lived resource that you know could
+    /// benefit from having a dedicated allocation.
+    AlwaysAllocate,
+}
+
+/// Memory allocations are portions of memory that are are reserved for a specific resource or
```
Rua, line 524: "are are"

vulkano/src/memory/allocator.rs
```diff
@@ -0,0 +1,4645 @@
...
+
+    /// Invalidates the host (CPU) cache for a range of the allocation.
+    ///
+    /// You must call this method before the memory is read by the host, if the device previously
+    /// wrote to the memory. It has no effect if the memory is not mapped or if the memory is
+    /// [host-coherent].
+    ///
+    /// `range` is specified in bytes relative to the start of the allocation. The start and end of
+    /// `range` must be a multiple of the [`non_coherent_atom_size`] device property, but
+    /// `range.end` can also equal to `self.size()`.
```
Rua, line 718: It can only be `self.size()` if that is the end of the whole `DeviceMemory` object. If this allocation actually exists in the middle of a `DeviceMemory`, then it must align with the atom size.

vulkano/src/memory/allocator.rs
```diff
@@ -0,0 +1,4645 @@
...
+            OomError::OutOfHostMemory => AllocationCreationError::OutOfHostMemory,
+            OomError::OutOfDeviceMemory => AllocationCreationError::OutOfDeviceMemory,
+        }
+    }
+}
+
+/// Standard memory allocator intended as a global and general-purpose allocator.
+///
+/// This type of allocator should work well in most cases, it is however **not** to be used when
+/// allocations need to be made very frequently. For that purpose, use [`FastMemoryAllocator`].
```
Rua, line 1137: It may be good to clarify how frequent is frequent in this case. Once per frame? Lots of times per frame?

vulkano/src/memory/allocator.rs
```diff
@@ -0,0 +1,4645 @@
...
+    /// [`MemoryAllocatePreference::NeverAllocate`] however.
+    ///
+    /// The default value is `&[]`, which must be overridden.
+    pub block_sizes: &'b [(Threshold, BlockSize)],
+
+    /// The allocation type that should be used for root allocations.
+    ///
+    /// You only need to worry about this if you're using [`PoolAllocator`] as the suballocator, as
+    /// all suballocations that the pool allocator makes inherit their allocation type from the
+    /// parent allocation. In all other cases it doesn't matter that this is.
```
Rua, line 1985: that > what

vulkano/src/memory/allocator.rs
```diff
@@ -0,0 +1,4645 @@
...
+    ///
+    /// The default value is [`AllocationType::Unknown`].
+    pub allocation_type: AllocationType,
+
+    /// Whether the allocator should use the dedicated allocation APIs.
+    ///
+    /// This means that when the allocator dedices that an allocation should not be suballocated,
```
Rua, line 1992: decides

vulkano/src/memory/allocator.rs
```diff
@@ -0,0 +1,4645 @@
...
+        )
+    }
+}
+
+/// A [suballocator] that uses the most generic [free-list].
+///
+/// The strength of this allocator is that it can create and free allocations completely
+/// dynamically, which means they can be any size and created/freed in any order. The downside is
+/// that this always leads to horrific [external fragmentation] the more such dynamic allocations
+/// are made. Therefore, this allocator is best suited for long-lived allocations. If you need need
```
Rua, line 2384: "need need"

Rua:
This is absolutely amazing work! It's well documented and so much more flexible than anything we had before. Thank you so much!

I do find that the `allocator.rs` module got rather long, so it may be good to split it into several submodules for easier browsing. In particular a `suballocator.rs` module would be helpful, and a `generic.rs` for `GenericMemoryAllocator` and its two specializations.

marc0246:
I'm very glad to hear that. In the end what I hope is for this to be useful, I know I can't make use of *all* this offers, haha. But why not make it as flexible as possible while on it, right? You're totally right about splitting the files, totally forgot that! Also thanks for the corrections.

vulkano/src/memory/allocator/suballocator.rs
```diff
@@ -0,0 +1,3090 @@
...
+        let mut size = range.end - range.start;
+        let device_memory = self.device_memory();
+
+        // VUID-VkMappedMemoryRange-size-01390
+        if offset + size < device_memory.allocation_size() {
+            // We align the size in case `range.end == self.size`. We can do this without aliasing
+            // other allocations because the suballocators ensure that all allocations are aligned
+            // to the atom size for non-coherent host-visible memory.
+            size = align_up(size, atom_size);
+        }
```
marc0246, line 329: About the non-coherent alignment, it is explained here why a range of `0..alloc.size()` works as well.

Rua: Ah ok, that's fine then.

vulkano/src/memory/allocator/mod.rs
```diff
@@ -0,0 +1,1593 @@
...
+    /// You only need to worry about this if you're using [`PoolAllocator`] as the suballocator, as
+    /// all suballocations that the pool allocator makes inherit their allocation type from the
+    /// parent allocation. In all other cases it doesn't matter what this is.
+    ///
+    /// The default value is [`AllocationType::Unknown`].
+    pub allocation_type: AllocationType,
+
+    /// Whether the allocator should use the dedicated allocation APIs.
+    ///
+    /// This means that when the allocator decices that an allocation should not be suballocated,
```
Rua, line 1449: "decices"

I think you converted one Engrish to another by accident :D

marc0246: Oh wow.

PR #2000, opened 2022-09-24
Exclude markdown from actions

AustinJ235:
This prevents running build/test on markdown files mainly the changelog. Also updates the `chocolatey` and `checkout` versions.


Rua:
No objections here. Also, PR #2000!

PR #2001, opened 2022-09-24
Make queues own the resources of executing operations

Rua:
Changelog:
```markdown
### Breaking changes
Changes to queue operations:
- `Queue` now implements `VulkanObject` instead of `SynchronizedVulkanObject`, which is removed.
- `Queue` now takes ownership of resources belonging to operations that you execute on it, to keep them from being destroyed while in use.
- `QueueGuard` now has a `cleanup_finished` method, which does the same thing as on futures. Calling this method on a future will automatically forward it to its queue.
- If `Queue` is dropped, it will call `wait_idle` to block the current thread until all operations on it have completed.
````

This completes the step in making queues responsible for their own resources, rather than leaving that to futures. In theory, a queue could now be (`unsafe`-ly) used without needing a future at all. For now though, the queue operation methods are still `pub(crate)`.

### DIFF
--- a/vulkano/src/command_buffer/traits.rs
+++ b/vulkano/src/command_buffer/traits.rs
@@ -419,10 +419,7 @@ where
 
     #[inline]
     unsafe fn signal_finished(&self) {
-        if !self.finished.swap(true, Ordering::SeqCst) {
-            self.command_buffer.unlock();
-        }
-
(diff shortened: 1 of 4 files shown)


AustinJ235 (approved):
Looking good!

PR #2002, opened 2022-09-25
Remove `command_buffer::submit` module, make new `Queue` methods public

Rua:
Changelog:
Remove this line:

```markdown
- `SubmitAnyBuilder` and its wrapped types no longer have a lifetime parameter, as they own their data instead. Various methods of these types now take their arguments by value.
```

And add:

```markdown
### Breaking changes
Changes to queue operations:
- The `command_buffer::submit` module has been removed. The `SubmitAnyBuilder` enum is moved to the `sync` module, and no longer has a lifetime parameter.

### Additions
- Added `bind_sparse_unchecked`, `present_unchecked` and `submit_unchecked` methods to `QueueGuard`.
````

The new queue operation methods are now exposed. There may be some things that still need adjusting, and of course there need to be safe validated versions of them too.


AustinJ235 (approved):
Looks good!

PR #2003, opened 2022-09-25
Fix feature name casing

marc0246:
This fixes a small issue where all of the fields containing names for a feature would use snake case, which results in the generated link to the specification to not work correctly. This PR addresses that by storing the original (camel case) name of the feature in one of the fields.


marc0246:
Fixed device property links also.

Rua:
Looks good, thanks!

PR #2006, opened 2022-09-27
Cache results of some `PhysicalDevice` methods for faster future retrieval

Rua:
Changelog:
```markdown
### Additions
- Some methods of `PhysicalDevice` now cache their results, so that another call with the same arguments will retrieve them faster.
````

While some of these Vulkan API functions may be relatively fast to call on some drivers, we can't know for sure. I expect that looking up and cloning a hashmap entry is probably going to be faster as it does not depend on the driver implementation.


AustinJ235 (approved):
Looks good!

PR #2007, opened 2022-09-29
Rust 2018 idioms

marc0246:
Small fix regarding the 2018 idioms, namely mostly the elided lifetimes in paths. I've also added a warn attribute for the future.


Rua:
Huh, I thought removing the elided lifetimes was the way forward. Ah well.

PR #2008, opened 2022-09-30
#2007 fixes

Rua:
This fixes some issues that were left over after #2007:
- The `ensure_sendable` test was failing.
- The auto-generated code was triggering idiom warnings.
- Idiom warnings weren't enabled on the other crates in Vulkano, I did that now and fixed any warnings.


marc0246:
Thanks 🙏

AustinJ235 (approved):
Good catch!

Issue #2004, opened 2022-09-26
You can present swapchain images you didn't acquire


Rua:
Taking the triangle example, if you change the present command so that it always presents swapchain image index 0 instead of the one returned by `acquire_next_image`, then there is no error from Vulkano. You do get a message from the validation layer, though:

> VUID-VkPresentInfoKHR-pImageIndices-01296(ERROR / SPEC): msgNum: -945112042 - Validation Error: [ VUID-VkPresentInfoKHR-pImageIndices-01296 ] Object 0: handle = 0xe7e6d0000000000f, type = VK_OBJECT_TYPE_SWAPCHAIN_KHR; | MessageID = 0xc7aabc16 | vkQueuePresentKHR: pSwapchains[0] image at index 0 was not acquired from the swapchain. The Vulkan spec states: Each element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout at the time the operation is executed on a VkDevice (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-VkPresentInfoKHR-pImageIndices-01296)

AustinJ235:
Nothing prevents writing/reading to a swapchain image that isn't acquired either. `AccessError::SwapchainImageAcquireOnly` is defined, but never constructed. :thinking:

Need to somehow determine in a future chain if the swapchain image has been acquired and then allow it to be accessed and presented. This assumes that acquiring, accessing, and presenting all happen in the same future chain.

PR #2009, opened 2022-09-30
Swapchain acquisition checks

AustinJ235:
Fixes: https://github.com/vulkano-rs/vulkano/issues/2004

This provides a solution for checking to make sure when presenting that the swapchain image has been acquired and has not already been presented. There may be some short comings to this yet, but my initial tests it is working as intended.

- `SwapchainImageAcquireOnly` was renamed to `SwapchainImageNotAcquired` to be less confusing.
- This does not address accessing images via commands.

Changelog:
```markdown
### Breaking changes
Changes to `GpuFuture`:
- Added required method `check_swapchain_image_acquired`.
- `AccessError::SwapchainImageAcquireOnly` has been renamed to `SwapchainImageNotAcquired`.

### Bugs fixed
- Fix bug where a swapchain image could be presented without being acquired. #2004 
````


Rua:
It all looks good and no problems on my end. Perhaps this could do with a unit test to make sure the error actually does happen when you use the wrong index in various situations?

PR #2010, opened 2022-10-01
Add state tracking and validation to `Fence`

Rua:
Changelog:
Remove this line
```markdown
- When doing operations on a queue, you must now first call `lock()` on the queue, which prevents concurrent access.
```
and replace with

```markdown
- To do operations on a queue, you must now call `with` to gain access.
```

Remove this line
```markdown
- `QueueGuard` now has a `cleanup_finished` method, which does the same thing as on futures. Calling this method on a future will automatically forward it to its queue.
```

Add
```markdown
### Additions
- Fence methods are now validated and synchronized, so they take `&self`.
- When calling `Fence::is_signaled` or `Fence::wait`, if the fence is associated with a queue, any resources of the associated queue operation will be released.
````
This makes changes that allow `Fence` to be used safely, and also adds convenience by automatically signaling a queue.

### DIFF
--- a/examples/src/bin/gl-interop.rs
+++ b/examples/src/bin/gl-interop.rs
@@ -254,10 +254,9 @@ mod linux {
                     recreate_swapchain = true;
                 }
                 Event::RedrawEventsCleared => {
-                    unsafe {
-                        let mut queue_guard = queue.lock();
-                        queue_guard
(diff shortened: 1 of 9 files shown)


AustinJ235 (approved):
Looks good!

PR #2011, opened 2022-10-01
#1935 add from_handle to DeviceMemory

hsteinmueller:
Added `from_handle` constructor to `DeviceMemory`, see #1938.

I also tried to add the constructor to `Instance` and `Device` but I couldn't find a nice solution without refactoring too much.
If you could maybe provide a suggestion or hint, I might be able to add it for those too.


Rua:
Looks good! A while back, I changed the order of the parameters so that device is first. Can you change that here?

hsteinmueller:
> Looks good! A while back, I changed the order of the parameters so that device is first, and made some changes to the documentation. Can you change that here, so that it matches the others?

I missed that sorry. I also removed the call from `allocate` with the unnecessary `unsafe`.

PR #2016, opened 2022-10-03
`#[inline]`ing adjustments and minor syntax/formatting changes

marc0246:
Added the `#[inline]` attribute to functions that could use it, and removed it from functions where it could do harm, or wasn't doing anything. The attribute is only useful to allow for the compiler to inline functions across a crate boundry.

For reference:
- Putting `#[inline]` on methods of error types is always a bad idea, as these are supposed to be exceptional states. In this case it could actually do harm, as it would increase code size for paths that are almost never taken and slightly increase compile times (because the compiler won't be able to compile our crate and their crate in parallel, if they make use of the methods).
- I removed it from `impl Debug`s with similar reasoning, but this can do no harm since you don't need to debug-print in production (I hope).
- Generic methods and non-generic methods on generic types never need the attribute, because the methods need to be instantiated with a concrete type at the use-site, so they always need to cross the crate boundry and are always inlinable for rustc.
- Private methods never need the attribute as they can't cross the crate boundry.

I also made some syntactical changes here and there, mainly I removed needless dereferencing followed by `ref`-matching. For this I also added the `Copy` trait for all SPIR-V types, so we don't need to deal with a bunch of referencing and dereferencing. Kind of like how all `ash` types are `Copy`.

Other than that, I took the liberty to do some minor changes to formatting as I was looking through the codebase, hopefully you don't mind. Mainly I made sure that lines don't exceed the rustfmt limit of 100 columns, and that doc comments use correct formatting.

All in all the intention was to hopefully make the code a bit cleaner. It's all in one humongous commit because I didn't have anything specific in mind as I was going through it all other than the inlining, so my apologies. Hopefully it's not too big of a hassle to review.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -295,14 +295,12 @@ pub enum Error {
 }
 
 impl From<IoError> for Error {
-    #[inline]
     fn from(err: IoError) -> Error {
         Error::IoError(err)
     }
 }
 
 impl From<SpirvError> for Error {
-    #[inline]
     fn from(err: SpirvError) -> Error {
         Error::SpirvError(err)
     }
--- a/vulkano-util/src/context.rs
+++ b/vulkano-util/src/context.rs
@@ -46,6 +46,7 @@ pub struct VulkanoConfig {
 }
 
 impl Default for VulkanoConfig {
+    #[inline]
     fn default() -> Self {
         let device_extensions = DeviceExtensions {
             khr_swapchain: true,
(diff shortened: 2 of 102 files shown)


Rua:
Why did you remove `#[inline]` from trait methods too?

marc0246:
Which ones?

Rua:
For example on `BufferContents`, `BufferViewAbstract`, `PrimaryCommandBuffer`, `SecondaryAutoCommandBuffer` and many others.

marc0246:
Because those traits were implemented on generic types, which need to be instanciated at the place they are used, so those methods don't exist until then. They always are duplicated and always cross the crate boundry, so those attributes weren't doing anything.

Rua:
I see, thanks.

PR #2017, opened 2022-10-03
Replace generics with `impl Trait` in function arguments

marc0246:
This removes generic function arguments that were left over from The Before Times™ where argument-position `impl Trait` wasn't a thing.


Rua:
Thanks!

PR #2019, opened 2022-10-04
Improvements to caches and switch to `ahash`

marc0246:
I added `OnceCache` to abstract away our caches for implementation-specific properties, purely for cosmetic reasons, as it's hopefully a bit easier to look at the types this way. It might also come in handly when we need to refactor them, however. Other than that I switched all our `HashMap`s and `HashSet`s to use `ahash` as the hasher of choice, as discussed on Discord. `ahash` is twice as fast as SipHash, `std`'s default, so this should hopefully make the caches in particular a bit faster.


Rua:
Looks good, thank you!

PR #2021, opened 2022-10-04
More complete shield links

Rua:
This adds links to the crates.io and docs.rs pages of all four Vulkano crates. This should make it easier for people to find them. I also added a link to the website/guide.


AustinJ235 (approved):
Neat!

PR #2022, opened 2022-10-04
Better error message when source is too small for `ImmutableImage`

Rua:
After discussion on Discord, this provides the user with a better error message if the source data is too small for the image they're trying to create.


AustinJ235 (approved):
Thanks!

PR #2024, opened 2022-10-05
Add support for Unix and Windows external fences

Rua:
Changelog:
```markdown
### Additions
- Support for the `khr_external_fence_fd` and `khr_external_fence_win32` extensions.
````

I've added importing and exporting fences, as well as full validation as far as is possible. In part, this was a warm-up for me before I tackle making semaphores safe, which are a lot harder.


AustinJ235 (approved):
Looks good!

PR #2027, opened 2022-10-06
Fix physical device surface querying regression #2026 by removing surface caching

### LINKED ISSUE #2026: Surface-capabilities queries return out of date information due to caching
# Issue

`PhysicalDevice::surface_capabilities` and `PhysicalDevice::surface_capabilities_unchecked` return out of date information after they're first called due to the new caching functionality introduced in #2006.

# Reproduce

Run an example (e.g. triangle) and resize the window. Swapchain recreation always fails due to `SwapchainCreationError::ImageExtentNotSupported` because the surface capabilities in `Swapchain::validate` are out of date after the first call to `surface_capabilities_unchecked`.

This means caching surface queries is unfortunately not viable as surface properties can change at runtime for a given surface and physical device.

lilly-lizard:
Fixes #2026 

Fixed regression bug where new caching functionality introduced in #2006 caused surface queries to be out of date.

`surface_capabilities`, `surface_formats`, `surface_present_modes` and `surface_support` are no longer cached as it is proven in #2026 that surface properties can change at runtime for a given surface/physical device.

Also worth noting that requesting these structures doesn't happen too often so the performance impact will be minimal.


marc0246:
Are you sure we need to ditch caching all-together? On my system, querying this info from the implementation is 800X slower. We could invalidate the caches where needed instead perhaps?

lilly-lizard:
problem is I can't see a way to do that... the surface and physical device handles stay the same, but drivers seem happy to change the data returned by VK_KHR_surface commands depending on when you do the query e.g. VkSurfaceCapabilitiesKHR.min/maxImageExtent due to the window being resized.

you could assume that `surface_formats`, `surface_present_modes` and `surface_support` won't change at runtime, which is probably true but there's no guarantee of that. keeping those ones cached _probably_ won't cause any problems down the line but seems like a weird risk to take given the infrequency of the calls imo.

but basically `surface_capabilities` is the only one that needs to not be cached to fix #2026 so I'd be happy to re-add caching for `surface_formats`, `surface_present_modes` and `surface_support` if that's what you guys want.

marc0246:
I agree that caching the capabilities would be impractical, although possible. We would need to invalidate the cache on resize, which would pose no benefit as you likely only need that information when recreating the swapchain, so it would just overcomplicate things. About the other ones, I've looked through the specification repo but couldn't find anything specific about the lifetime of the info.

Rua:
Thank you for catching this one!

Rua:
There's an error on MacOS, can you fix it?

lilly-lizard:
whoops missed that. pretty sure macos/ios will work now

PR #2028, opened 2022-10-06
Check for empty aspects when creating `ImageView`

Rua:
A check that was missed, and that ended up causing more confusing errors further down the line.


AustinJ235 (approved):
Thanks

PR #2030, opened 2022-10-07
Buffer and image uploads take a command buffer instead of making one

Rua:
Changelog:

Remove this line (and remove the bullet from the previous one to merge it into the line above):
```markdown
- `DeviceLocalBuffer::{from_buffer, from_data, from_iter}` and `ImmutableImage::{from_iter, from_buffer}` now take an implementation of `CommandBufferAllocator`.
```

Add:
```markdown
### Breaking changes
Changes to buffer and image uploads:
- `DeviceLocalBuffer::{from_buffer, from_data, from_iter}` and `ImmutableImage::{from_iter, from_buffer}` now take a mutable reference to an `AutoCommandBufferBuilder` instead of a queue, and no longer return a future. The upload command will be recorded into the provided command buffer, which should be executed later.
````

Creating a new command buffer builder, recording a single copy command onto it, and then building and submitting a single queue command with just the one tiny command buffer is not very efficient. If a lot of things are uploaded, this can also lead to stack overflows when you try to flush this giant pile of futures all at once, as discovered on Discord. With these changes, the user has the freedom to manage this a bit better themselves, and they can group all uploads into a single command buffer if they wish.

### DIFF
--- a/examples/src/bin/debug.rs
+++ b/examples/src/bin/debug.rs
@@ -9,7 +9,9 @@
 
 use std::sync::Arc;
 use vulkano::{
-    command_buffer::allocator::StandardCommandBufferAllocator,
+    command_buffer::{
+        allocator::StandardCommandBufferAllocator, AutoCommandBufferBuilder, CommandBufferUsage,
+    },
     device::{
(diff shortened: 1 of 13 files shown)


AustinJ235 (approved):
Looks good!

PR #2031, opened 2022-10-08
Fix missing validation when binding memory to buffers with the  `shader_device_address` usage

marc0246:
Required for #1997.

Adds support for passing the `device_address` flag when allocating `DeviceMemory`, which is required when binding memory to a buffer with the `shader_device_address` usage in the `VK_KHR_buffer_device_address` extension (and core 1.2 which it was promoted to), but not its predecessor `VK_EXT_buffer_device_address`. Fixes the bug where `UnsafeBuffer::bind_memory` wasn't validating if the memory is allocated with the `device_address` flag, and also fixes the wrong function pointer in `BufferAccess::raw_device_address` when not using the outdated  `VK_EXT_buffer_device_address` extension.

Changelog:
```markdown
### Breaking changes
Changes to buffers:
- When binding memory to a buffer with the `shader_device_address` usage, the memory must now have been allocated with the `MemoryAllocateFlags::device_address` flag set.

### Additions
- Added `MemoryAllocateFlags`.

### Bugs fixed
- Fixed missing validation when binding memory to a buffer with the `shader_device_address` usage.
````

### DIFF
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -403,8 +403,8 @@ impl UnsafeBuffer {
     /// # Panics
     ///
     /// - Panics if `self.usage.shader_device_address` is `true` and the `memory` was not allocated
-    ///   with the [`device_address`] flag set and the [`ext_buffer_device_address`] device
-    ///   extension is not used.
+    ///   with the [`device_address`] flag set and the [`ext_buffer_device_address`] extension is
+    ///   not enabled on the device.
     ///
     /// [`device_address`]: crate::memory::MemoryAllocateFlags::device_address
     /// [`ext_buffer_device_address`]: crate::device::DeviceExtensions::ext_buffer_device_address
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -403,9 +403,11 @@ impl UnsafeBuffer {
     /// # Panics
     ///
     /// - Panics if `self.usage.shader_device_address` is `true` and the `memory` was not allocated
-    ///   with the [`device_address`] flag set.
+    ///   with the [`device_address`] flag set and the [`ext_buffer_device_address`] device
+    ///   extension is not used.
     ///
     /// [`device_address`]: crate::memory::MemoryAllocateFlags::device_address
+    /// [`ext_buffer_device_address`]: crate::device::DeviceExtensions::ext_buffer_device_address
     pub unsafe fn bind_memory(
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: vulkano/src/buffer/sys.rs [hunk 1]
```diff
@@ -435,6 +442,11 @@ impl UnsafeBuffer {
             }
         }
 
+        // VUID-vkBindBufferMemory-bufferDeviceAddress-03339
+        if self.usage.shader_device_address {
+            assert!(memory.flags().device_address);
+        }
```
Rua, line 448: Because the flag doesn't exist for the EXT version, I think this check should only apply if the EXT version not enabled?

vulkano/src/memory/device_memory.rs
```diff
@@ -692,6 +725,11 @@ pub struct MemoryAllocateInfo<'d> {
     /// The handle types that can be exported from the allocated memory.
     pub export_handle_types: ExternalMemoryHandleTypes,
 
+    /// Additional flags for the memory allocation.
+    ///
```
Rua, line 729: Per the previous Discord discussion, I think there should be this additional text:
> If not empty, the device API version must be at least 1.1, or the [`khr_device_group`](crate::device::DeviceExtensions::khr_device_group) extension must be enabled on the device.

And then in `DeviceMemory::validate` there should be a check to this effect.

vulkano/src/memory/device_memory.rs
```diff
@@ -466,6 +484,15 @@ impl DeviceMemory {
             allocate_info = allocate_info.push_next(info);
         }
 
+        let mut flags_info = ash::vk::MemoryAllocateFlagsInfo {
+            flags: flags.into(),
+            ..Default::default()
+        };
+
+        if !flags.is_empty() && !device.enabled_extensions().ext_buffer_device_address {
```
Rua, line 492: This should be added if `flags` is not empty, without the second check I think? At least for future-proofing if other flags are added.

vulkano/src/buffer/sys.rs
marc0246: I wanted to free the user of having to think about what extension is enabled when it comes to allocation, so the `DeviceMemory` must always be allocated with with the flag regardless of which extension is active. The extension is checked when allocating.

Rua: Ok, but that then makes the other things I commented on a bit more complicated.

marc0246:
Alright done.

PR #2034, opened 2022-10-12
Fixed queue family index count not being set in buffers and images

daigennki:

3. [X] Run `cargo fmt` on the changes.

4. [X] Please put changelog entries **in the description of this Pull Request**
   
5. [X] Describe in common words what is the purpose of this change, related

Changelog:
```markdown
### Bugs fixed
- Queue family count isn't being set in either `ash::vk::BufferCreateInfo` or `ash::vk::ImageCreateInfo`
````

In either case of buffers or images, the bug occurred when using more than 1 queue family (not exclusive use). `queue_family_index_count` in the `ash` structs was not being explicitly set and was left as `Default::default()`, which is 0, making it in violation of the Vulkan spec (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkBufferCreateInfo-sharingMode-00914, https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageCreateInfo-sharingMode-00942) and potentially causing undefined behavior.

### DIFF
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -226,10 +226,11 @@ impl UnsafeBuffer {
             flags |= sparse_level.into();
         }
 
-        let (sharing_mode, p_queue_family_indices) = match sharing {
-            Sharing::Exclusive => (ash::vk::SharingMode::EXCLUSIVE, &[] as _),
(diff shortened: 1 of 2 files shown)


Rua:
Thank you for the fix!

PR #2035, opened 2022-10-12
Add safety checks to semaphores, add importing and exporting

Rua:
Changelog:
```markdown
### Additions
- Support for the `khr_external_semaphore_fd`, `khr_external_semaphore_win32` and `fuchsia_external_semaphore` extensions.
- `Semaphore::export_fd` is no longer unsafe.
````

This adds checks to make semaphores safe to use, but only in a very limited fashion for now. To keep the checks simple, there can't be more than one signal operation and one wait operation pending at any given time. This limitation will hopefully be lifted in the future.

### DIFF
--- a/examples/src/bin/gl-interop.rs
+++ b/examples/src/bin/gl-interop.rs
@@ -153,16 +153,12 @@ mod linux {
             .unwrap(),
         );
 
-        let acquire_fd = unsafe {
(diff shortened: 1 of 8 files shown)


AustinJ235 (approved):
Looks good!

PR #2037, opened 2022-10-13
Rename `internal_object` to `handle`

Rua:
Changelog:
```markdown
### Breaking changes
Changes to the `VulkanObject` trait:
- The method `internal_object` is renamed to `handle`, and the associated type `Object` is renamed to `Handle`.
````

This is something I've been wanting to do for a while. In Vulkan terminology these are "handles" and we already use that terminology elsewhere: the private variables holding them are called `handle` and we have `from_handle` constructors now too.


AustinJ235 (approved):
Looks good, needs some conflicts addressed. You can go ahead with the merge when those are resolved.

marc0246:
I love this too, has been bothering me for a while haha!

PR #2038, opened 2022-10-13
Minor tweaks to `DeviceMemory`

marc0246:
Renamed `DeviceMemory::create` to `DeviceMemory::allocate_unchecked` and made it public. It also now returns `Result<DeviceMemory, VulkanError>`, so it's a bit more in-line to the new way of doing things. I also added `Copy` to some memory-related structs. This is useful for me to use in #1997.

### DIFF
--- a/vulkano/src/memory/device_memory.rs
+++ b/vulkano/src/memory/device_memory.rs
@@ -45,7 +45,8 @@ use std::{
 ///         memory_type_index,
 ///         ..Default::default()
 ///     },
-/// ).unwrap();
+/// )
+/// .unwrap();
 /// ```
 #[derive(Debug)]
 pub struct DeviceMemory {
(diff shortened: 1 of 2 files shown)


vulkano/src/memory/mod.rs
```diff
@@ -336,7 +336,7 @@ pub enum DedicatedAllocation<'a> {
 
 /// The properties for exporting or importing external memory, when a buffer or image is created
 /// with a specific configuration.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default)]
```
Rua, line 339: Non-exhaustive structs shouldn't be `Copy`, because that would make it a breaking change to add a non-copy type to the struct in the future.

vulkano/src/memory/device_memory.rs
```diff
@@ -426,12 +394,14 @@ impl DeviceMemory {
         Ok(())
     }
 
-    unsafe fn create(
-        device: &Device,
-        allocate_info: &MemoryAllocateInfo<'_>,
+    #[cfg_attr(not(feature = "document_unchecked"), doc(hidden))]
+    #[cold]
```
Rua, line 398: This doesn't seem correct. The function is called every time an allocation is made, that's not rare.

vulkano/src/memory/mod.rs
marc0246: Ah, I haven't thought about that. It's just a minor inconvenience anyway.

vulkano/src/memory/device_memory.rs
marc0246: I'd say allocating `DeviceMemory` is about as cold as it gets in a Vulkan application, excluding things like `Instance` and `Device` creation. I would prefer for this not to get inlined into an allocator. Seeing as LLVM's heuristic for if a function should be inlined is "yes", I can see how it could happen.

marc0246: Or should I use `#[inline(never)]`?

marc0246: Yeah you're right, it would be better to put `#[cold]` on a wrapper function, which is called from within the branches in an allocator.

marc0246:
Fixed

PR #2041, opened 2022-10-20
Move resource locking to `submit` function

Rua:
Changelog:
```markdown
### Breaking changes
Changes to command buffers:
- The `PrimaryCommandBuffer` and `SecondaryCommandBuffer` traits are renamed to `PrimaryCommandBufferAbstract` and `SecondaryCommandBufferAbstract`.
````

The resource locking code that was previously implemented in the `lock_submit` function of the `PrimaryCommandBuffer` trait, is now moved to the `Queue::submit` function. It's also implemented now as a mutex-guarded `CommandBufferState` type, similar to how state was already implemented for buffers, images, fences and semaphores.


AustinJ235 (approved):
Looks good!

PR #2044, opened 2022-10-23
Add checks for `khr_portability_subset`

Rua:
Changelog:
```markdown
### Bugs fixed
- Validation checks for `khr_portability_subset` devices.
````

This adds the missing validation checks for devices that expose the `khr_portability_subset` extension. This concerns MacOS and iOS devices, which do not fully conform to the Vulkan spec, and therefore do not support all the features that a conformant Vulkan implementation is expected to support. For such devices, certain features must be enabled on the device (if they are available) in order to use them.

### DIFF
--- a/vulkano/src/sampler/mod.rs
+++ b/vulkano/src/sampler/mod.rs
@@ -196,7 +196,7 @@ impl Sampler {
                 required_for: "this device is a portability subset device, and \
                     `create_info.mip_lod_bias` is not zero",
                 requires_one_of: RequiresOneOf {
-                    features: &["sampker_mip_lod_bias"],
+                    features: &["sampler_mip_lod_bias"],
                     ..Default::default()
                 },
             });
--- a/vulkano/src/sampler/mod.rs
+++ b/vulkano/src/sampler/mod.rs
@@ -187,6 +187,21 @@ impl Sampler {
             }
         }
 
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: vulkano/src/sampler/mod.rs [hunk 1]
```diff
@@ -187,6 +187,21 @@ impl Sampler {
...
+        // VUID-VkSamplerCreateInfo-samplerMipLodBias-04467
+        if device.enabled_extensions().khr_portability_subset
+            && !device.enabled_features().sampler_mip_lod_bias
+            && mip_lod_bias != 0.0
+        {
+            return Err(SamplerCreationError::RequirementNotMet {
+                required_for: "this device is a portability subset device, and \
+                    `create_info.mip_lod_bias` is not zero",
+                requires_one_of: RequiresOneOf {
+                    features: &["sampker_mip_lod_bias"],
```
AustinJ235, line 199: ``sampker`` :thinking:

AustinJ235 (changes requested):
Aside from the typo, looks good.

vulkano/src/sampler/mod.rs
Rua: Oops, thanks! Fixed now.

AustinJ235 (approved):

PR #2046, opened 2022-10-26
Make command buffer/descriptor set allocators `Sync` again

marc0246:
Made `StandardCommandBufferAllocator` and `StandardDescriptorSetAllocator` thread-safe using TLS.

At first I wanted to implement the TLS myself, using a simple `RwLock<HashMap<ThreadId, T>>`, but I noticed that Windows kept getting into deadlocks with that for some reason. So instead I opted for an established library, `thread_local`, which was added to the private dependencies. It's a very light-weight library and also more optimized than what I had in mind. For one it doesn't need acquiring any read locks, and also the entries in the TLS can be reused. This would not have been the case with what I have tried, as threads would always drop their allocator on exit and when a new thread would allocate it would create a brand new allocator for itself. This way those reentrant threads can reuse the allocators.

Another good thing that came from trying this library is that after benchmarking, there's almost no overhead to using this TLS over none at all. It's 6ns of overhead on my CPU, granted I also have a very dated CPU. Therefore, I think it makes no sense to have a non-`Sync` version at all, if we're going to use this library. The only performance implication this has is when a lot of threads need to access the TLS concurrently, which would lead to CPU-level contention on the `AtomicPtr`s that the TLS uses. But even with unreasonably high contention that I tested it with (all of my threads going at it) the times for allocation only doubled. If this is an issue the user still has the option to quite simply not share the allocators between threads.

### DIFF
--- a/vulkano/src/command_buffer/allocator.rs
+++ b/vulkano/src/command_buffer/allocator.rs
@@ -30,6 +30,7 @@ use crate::{
 use crossbeam_queue::SegQueue;
 use smallvec::SmallVec;
 use std::{cell::UnsafeCell, marker::PhantomData, mem::ManuallyDrop, sync::Arc, vec::IntoIter};
+use thread_local::ThreadLocal;
 
 /// Types that manage the memory of command buffers.
 ///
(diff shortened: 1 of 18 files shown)


vulkano/src/command_buffer/allocator.rs
```diff
@@ -100,31 +101,49 @@ pub unsafe trait CommandBufferAlloc: DeviceOwned + Send + Sync + 'static {
...
 ///
-/// A thread can have as many `StandardCommandBufferAllocator`s as needed, but they can't be shared
-/// between threads. This is done so that there are no locks involved when creating command
-/// buffers. You are encouraged to create one allocator per frame in flight per thread.
+/// The intended way to use this allocator is to have one that is used globally for the duration of
+/// the program, in order to avoid creating and destroying [`CommandPool`]s, as that is expensive.
+///
+/// Internally, this allocator keeps one `CommandPool` per queue family index per thread, using
+/// Thread-Local Storage. When a thread first allocates, an entry is reserved for it in the TLS.
+/// After a thread exists and the allocator wasn't dropped yet, its entry is freed, but the pools
```
Rua, line 109: exists > exits

vulkano/src/command_buffer/allocator.rs
```diff
@@ -100,31 +101,49 @@ pub unsafe trait CommandBufferAlloc: DeviceOwned + Send + Sync + 'static {
...
+}
+
+#[derive(Debug)]
+struct Pool {
+    inner: Arc<PoolInner>,
 }
 
+// This is needed because of the blanket impl of `Send` on `Arc<T>`, which requires that `T` is
+// `Send + Sync`. `PoolInner` is `Send + !Sync` because `CommandPool` is `!Sync`. That's fine
+// however hecause we never access the `CommandPool` concurrently, only drop it once the `Arc`
```
Rua, line 135: because

vulkano/src/descriptor_set/allocator.rs
```diff
@@ -69,12 +70,26 @@ pub trait DescriptorSetAlloc: Send + Sync {
...
-/// Internally, this implementation uses one [`SingleLayoutDescriptorSetPool`] /
-/// [`SingleLayoutVariableDescriptorSetPool`] per descriptor set layout.
+/// The intended way to use this allocator is to have one that is used globally for the duration of
+/// the program, in order to avoid creating and destroying [`DescriptorPool`]s, as that is
+/// expensive.
+///
+/// Internally, this allocator uses one [`SingleLayoutDescriptorSetPool`] /
+/// [`SingleLayoutVariableDescriptorSetPool`] per descriptor set layout per thread, using
+/// Thread-Local Storage. When a thread first allocates, an entry is reserved for it in the TLS.
+/// After a thread exists and the allocator wasn't dropped yet, its entry is freed, but the pools
```
Rua, line 80: exists > exits

Rua:
That library is a nice find! Maybe it will come in handy for other parts of Vulkano too.

Everything looks good, I'll merge when the typos are fixed.

marc0246:
Thanks for the corrections! Fixed.

PR #2050, opened 2022-10-28
Add new safe `Buffer` and `Image` types containing bound memory

Rua:
Changelog:
```markdown
### Breaking changes
Changes to buffers and images:
- `UnsafeBuffer` and `UnsafeImage` are renamed to `RawBuffer` and `RawImage`.
- `UnsafeBufferCreateInfo` and `UnsafeImageCreateInfo` are renamed to `BufferCreateInfo` and `ImageCreateInfo`.
- `ImageCreateInfo` now takes a `flags` instead of separate booleans.
- Replaced the various unsafe `*_linear_layout` methods of `UnsafeImage` with a safe `subresource_layout` method. The `LinearLayout` struct is renamed to `SubresourceLayout` to match.
- Added mostly-safe `bind_memory` methods to these two types. These take `self` by value, and return `Buffer` and `Image` on success.
- Added `Buffer` and `Image` types, which represent buffers and images that have had memory bound to them. This memory can be retrieved using the `memory()` method.
- Most previous uses of `UnsafeBuffer` and `UnsafeImage` now use `Buffer` and `Image` instead.

### Additions
- Added a `flags` field to `BufferCreateInfo`. This contains no flags yet, but will in the future.
- Added the `disjoint` flag to `ImageCreateFlags`. This flag is used in combination with multi-planar images, to bind separate memory to each plane of the image. It is not yet supported for the higher-level image types.
````

This is another step towards reducing the number of buffer and image types that are needed, and also makes binding memory mostly safe. Type safety is used to ensure that `Buffer` and `Image` always have memory bound to them. Once sparse binding is introduced, this will not fully be the case, since bindings can be made and unmade at any time; sparse resources will therefore require runtime checks.

@marc0246 There are some considerations in this PR that may affect your memory allocator code:
- The `disjoint` flag is added, which never allows dedicated allocations ([VUID-VkMemoryDedicatedAllocateInfo-image-01797](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkMemoryDedicatedAllocateInfo.html#VUID-VkMemoryDedicatedAllocateInfo-image-01797)). Images that were created with this flag don't have one single `MemoryRequirements`, but rather two or three (for each plane). The `memory_requirements` method for images therefore now returns a slice. The newly introduced `ImageMemory` type also has a vector to hold all the allocations.
- `MemoryRequirements` was missing the `requires_dedicated_allocation` flag, which this PR adds. The allocator needs to ensure that when this flag is set, either a dedicated allocation is made or it errors out.

### DIFF
--- a/vulkano/src/buffer/cpu_access.rs
+++ b/vulkano/src/buffer/cpu_access.rs
@@ -17,15 +17,15 @@
 //! or write and write simultaneously will block.
 
 use super::{
-    sys::UnsafeBuffer, BufferAccess, BufferAccessObject, BufferContents, BufferCreationError,
-    BufferInner, BufferUsage,
+    sys::{Buffer, BufferMemory, RawBuffer},
+    BufferAccess, BufferAccessObject, BufferContents, BufferError, BufferInner, BufferUsage,
 };
 use crate::{
-    buffer::{sys::UnsafeBufferCreateInfo, TypedBufferAccess},
+    buffer::{sys::BufferCreateInfo, TypedBufferAccess},
     device::{Device, DeviceOwned},
     memory::{
         allocator::{
-            AllocationCreateInfo, AllocationCreationError, AllocationType, MemoryAlloc,
+            AllocationCreateInfo, AllocationCreationError, AllocationType,
             MemoryAllocatePreference, MemoryAllocator, MemoryUsage,
         },
         DedicatedAllocation,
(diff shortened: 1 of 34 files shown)


marc0246:
Amazing! I love the that you renamed the `Unsafe*` types while on it, since that's not really a convention that exists. Granted those types were named long before anything resembling conventions materialized haha.

AustinJ235 (approved):
Looks great!

PR #2051, opened 2022-10-28
Add async future for Fence Signal

### LINKED ISSUE #1412: `GpuFuture` and `Future`
Hi,

I was wondering if you had any plans on making the `GpuFuture` trait compatible with the now standard `Future` trait? Being able to use `async`/`await` with vulkano would be great.

Joeoc2001:
Changelog:
```markdown
### Additions
- Add a spinning Future implementation to FenceSignalFuture.
````

Allows Fence futures to be `.await`ed by spinning. This is better than blocking in async applications since the async engine used can choose to progress some other future while the work for the fence is being completed on the GPU.

Resolves #1412

### DIFF
--- a/vulkano/src/sync/fence.rs
+++ b/vulkano/src/sync/fence.rs
@@ -24,6 +24,9 @@ use std::{
     sync::{Arc, Weak},
     time::Duration,
 };
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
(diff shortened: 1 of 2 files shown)


Rua:
Would it make sense to include this directly on `Fence` as well?

Joeoc2001:
> Would it make sense to include this directly on `Fence` as well?

Sure, I'll do that now

Joeoc2001:
@Rua done

Rua:
Thanks, looks good now!

PR #2052, opened 2022-10-28
Fix depth and stencil attachment check when building pipeline

insertt:
Changelog:
```markdown
### Bugs fixed
- Depth and stencil attachments are now properly checked when using dynamic rendering.
- Proper error (`NoStencilAttachment`) is now being returned if stencil attachment format is missing.
````

I've noticed these issues when tinkering with secondary buffers and dynamic rendering - there was no way to declare inheritance for depth attachments for secondary buffers, as it would result in pipeline validation error with or without it.


Rua:
Oops! Thanks for finding and fixing those!

PR #2061, opened 2022-10-30
Release Vulkano 0.32.0

Rua:


Rua:
Made it a draft for now because @marc0246 still wants to do some things.

PR #2062, opened 2022-10-30
Add support for the `requires_dedicated_allocation` flag to `GenericMemoryAllocator`

marc0246:


vulkano/src/memory/allocator/mod.rs
```diff
@@ -532,6 +532,12 @@ pub enum AllocationCreationError {
     /// enough memory in the pool.
     OutOfPoolMemory,
 
+    /// A dedicated allocation is required but was explicitly forbidden.
+    ///
+    /// This is returned when using [`MemoryAllocatePreference::NeverAllocate`] and the
+    /// imlementation requires a dedicated allocation.
```
Rua, line 538: implementation

PR #2073, opened 2022-10-31
Use actual integer bitflags for Vulkan bitflag types

Rua:
Changelog:
```markdown
### Breaking changes
Changes to all Vulkan bitflag types:
- They are now implemented as wrappers around an integer, instead of as a struct of `bool`s.
- The fields of the struct have been replaced with associated constants, using uppercase names as is the convention.
- All functions that take or return one of these types now take them by value (copy).
- Bitflag types that have a corresponding enum type now have the following additional items:
  - A `contains_enum` method that takes a value of the enum, and returns whether the bitflags contain the flag corresponding to that enum.
  - A `From` implementation to convert the enum to the bitflags type, with a single flag set.
  - A `FromIterator` implementation that collects an iterator of enums into a single bitflag value.
  - An `IntoIterator` implementation that does the reverse.

Changes to two swapchain bitflag types:
  - `SupportedSurfaceTransforms` is renamed to `SurfaceTransforms`.
  - `SupportedCompositeAlpha` is renamed to `CompositeAlphas`.
````

This is a huge breaking change, but it makes the types much smaller. They are now trivially copyable, and most operations become a lot more efficient since they are simple bitwise integer operations. Converting them to their Ash representation should be a no-op. I've made a custom `Debug` implementation for them as well, which should be a more readable representation than a random-looking integer.


AustinJ235 (approved):
Great work!

PR #2076, opened 2022-11-02
`CpuBufferPool` revamp

### LINKED ISSUE #1434: `CpuBufferPool` slower than glium
*   Version of vulkano: 0.19.0
*   OS: Arch Linux
*   GPU (the selected PhysicalDevice): GeForce GTX 1060 6GB
*   GPU Driver: nVidia 455.38 (Driver version 0x71c98000, Vulkan version 1.2.142)
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: Run Keeshond Doggymark example in latest git, reference code for the Vulkan renderer is here: https://gitlab.com/cosmicchipsocket/keeshond/-/blob/c91fbb2a011be18cb462a8173725730ce2052ceb/keeshond/src/renderer/vulkan.rs#L272
Sorry I can't provide anything more concise, but it was hard enough setting up vulkano for this project as-is.

# Issue

When using vulkano to draw instanced quads, the overhead for each draw is actually larger than glium, defeating the whole point of using vulkano in the first place.

I need low draw call overhead for my 2D sprite-based engine for scenarios where ordered draws involve lots of texture changes, as these can't be batched.

For each draw operation I do the following:

```
let chunk = self.instance_buffer.chunk(self.instance_buffer_src.clone()).expect("Failed to allocate buffer chunk");
let vertex_slice = self.vertex_buffer.clone();

builder.draw_indexed(self.pipeline.clone(), &self.dynamic_state,
                     vec![vertex_slice, Arc::new(chunk)],
                     self.index_buffer.clone(), (), ()).expect("Failed to draw buffer");
```

The problem is that 1. `chunk()` seems to be performing a lot of allocations, or is otherwise taking a long time to figure out which chunk to use and whether to allocate, and 2. I have to create a new `Arc` for every call to `draw_indexed()`. If I remove batching, callgrind reports 23.97% time spent in `vulkano::buffer::cpu_pool::CpuBufferPool<T,A>::try_next_impl` and 49.89% time spent in `__memcpy_avx_unaligned_erms` which is being called from `core::ptr::drop_in_place'2` which seems to be coming from dropping the `Arc`. Without this overhead I suspect vulkano would be quite fast, but right now it's blocking me from working on the rest of this renderer until this bottleneck is resolved.

I tried to do buffering myself using a Vec of `CpuAccessibleBuffer` objects, but I ran into #1429 and #1433 while trying to implement this.

Any help on this is *greatly* appreciated.

marc0246:
Fixes the performance bottlenecks of `CpuBufferPool`, now renamed to `CpuBufferAllocator`. The algorithm has been optimized to pretty much the theoretical limit, which required that the allocator is made `!Sync`. There still is the option to wrap the allocator in a mutex, if absolutely neccessary. I did a number of benchmarks to measure the throughput before and after for different sizes of inputs (using a logarithmic scale), and pretty consistenly observed a -67% reduction in overhead, or a 3X performance increase. I suspect that someone with better hardware than mine will observe better results, though. According to my profiling, all time spent in `CpuBufferAllocator::from_iter` is now spent on copying data to the mapping. As that is completely dependant on the hardware and driver, I hope that this closes #1434. There isn't much more we can do in that sense I don't think.

I also noticed that the allocator having a `T` type parameter didn't really make sense. Mainly for the reason that the whole point of having one buffer be suballocated is so that you can fit all kinds of data needed each frame into one buffer, which is not possible if the buffers have different types, unless you cast your data to `&[u8]`, in which case alignment goes ouf of the window on the other hand. Since there really was no reason for this constraint, I have moved the type parameter to the methods that allocate subbuffers. It also didn't make sense semantically, because the allocator doesn't own any `T`s, the allocated subbuffers do.

Changelog:
```markdown
### Breaking changes:
Changes to CPU buffer allocation:
- Replaced `CpuBufferPool` with `CpuBufferAllocator`, which is now marked `!Sync` and no longer has a `T` type parameter. The type parameter was moved to the methods, to allow one allocator to allocate as many types of buffers as needed.
- Merged `CpuBufferPoolChunk` and `CpuBufferPoolSubbuffer` into `CpuSubbuffer`.

### Additions
- Added `CpuBufferAllocatorCreateInfo`.

### Bugs fixed
- Fixed an issue with `CpuBufferPool<T>`, where the alignment of `T` was not being considered when allocating.
- Fixed an issue with `CpuBufferPool`, where the allocated subbuffers did not respect the non-coherent atom size for non-host-coherent memory types.


Rua:
Are you intending to implement some of the things we talked about yesterday, or is this good to go now?

marc0246:
All good.

Issue #1218, opened 2019-07-10
Waiting on `SwapchainAcquireFuture` without dropping it


LukasKalbertodt:
*  Version of vulkano: 0.13
*   OS: Ubuntu 18.04
*   GPU (the selected PhysicalDevice): Intel(R) HD Graphics 530 (Skylake GT2)

---

I'm developing an application where latency is more important than anything else. I want to block the CPU waiting for the GPU in several places. Unfortunately, I can't seem to wait for a `SwapchainAcquireFuture` on the CPU side. I can drop it, but that means I loose access to it. From the code of `drop()` it looks like one could easily expose a `wait()` method for that future. 

I tried two things: `acquire_future.then_signal_fence_and_flush()?.wait()?` but that leads to:

```
thread '<unnamed>' panicked at 'assertion failed: future.queue().is_some()', /home/lukas/.cargo/registry/src/github.com-1ecc6299db9ec823/vulkano-0.13.0/src/sync/future/fence_signal.rs:38:5
```

And since the `SwapchainAcquireFuture` already has an internal fence, it seems wasteful to add another one.

And as I said, I also tried dropping the future, assuming that afterwards acquisition is finished. Since the future is dropped and I don't have access to it anymore, instead of `acquire_future.then_execute(...)` I use `my_cb.execute()`. But this way I get:

```
access to a resource has been denied (AccessError { error: SwapchainImageAcquireOnly, command_name: "vkCmdBeginRenderPass", command_param: "attachment 0", command_offset: 1 })
```

So I'm not really reporting an *issue* but rather a feature request. *But* it also might be that I misunderstand something fundamentally, so just let me know if that's the case :P

faulesocke:
Why would you actually want to wait on the acquisition of a swapchain image and not the presentation? I recommend having a look at the different swapchain presentation modes vulkan offers, then deciding which one fits your purpose most and then waiting on the presentation future instead of the acquire future:

https://vulkan-tutorial.com/Drawing_a_triangle/Presentation/Swap_chain#page_Presentation-mode
https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/html/vkspec.html#VkPresentModeKHR

LukasKalbertodt:
Well... it probably boils down to "bad Intel drivers". Here a bit longer explanation, with the key points in bold:

[...]

PR #2080, opened 2022-11-04
Allow waiting on SwapchainAcquireFuture

phire:
Changelog:
````md
### Additions
- Allow waiting on `SwapchainAcquireFuture`
````

Fixes: #1218

It was previously theoretically possible to wait on `SwapchainAcquireFuture` by dropping it.

But if you dropped it, it became impossible to actually present the image, as `then_swapchain_present` requires `acquire_future` to have been joined.

#### Why might you want to wait for the swapchain acquire?

 * Because my machine (Nvidia on Linux) pegs the cpu at 100% if you don't (like, wtf???)
 * Because that's what Sascha Willems' samples do
 * Because it allows you to lower input latency when running vsynced. 

#### Is this the best design?

There was some discussion on discord about making acquire_next_image return some opaque object that did more things automatically. Or perhaps tracking the acquire state on the image instead of with this future. 

But this approach achieve much the same result, and doesn't require invasive refactoring.

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -1633,6 +1633,18 @@ impl SwapchainAcquireFuture {
     pub fn swapchain(&self) -> &Arc<Swapchain> {
         &self.swapchain
     }
+
+    /// Blocks the current thread until the swapchain image has been acquired, or timeout
+    ///
+    /// If timeout is `None`, will potentially block forever
+    ///
+    /// You still need to join with this future for present to work
+    pub fn wait(&self, timeout: Option<Duration>) -> Result<(), FenceError> {
+        match &self.fence {
+            Some(fence) => fence.wait(timeout),
+            None => Ok(()),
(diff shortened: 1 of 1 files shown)


Rua:
It looks good, thanks! Would any of the examples in Vulkano benefit from using this function?

phire:
Not an easy question to answer. 

You could just add it to all of them: They all seem to use `PresentMode::FIFO`, and you could argue it's a good idea to always wait after acquire when using `PresentMode::FIFO`, especially if there are driver implementations out there (like mine) that appear to freak out when you don't.

Or you could argue that unless you are actually accepting input, there is no difference between waiting before building the command buffer, or after command buffer submission.  

And with a quick skim though the examples, only one (interactive fractal) that actually accepts input.

Rua:
Shall I merge this as it is then?

phire:
Yeah probably.

It's not really in-scope of the examples to start worrying about input latency.

PR #2081, opened 2022-11-05
Refactor `DescriptorRequirements`

Rua:
Changelog:
```markdown
### Breaking changes
Changes to `DescriptorRequirements`:
- The struct has been split into two levels: the per-binding `DescriptorBindingRequirements`, and the per-descriptor-index `DescriptorRequirements`.
- The `descriptor_requirements` method of various types has been renamed to `descriptor_binding_requirements` to match.
- `DescriptorBindingRequirements` has a single `descriptors` member instead of multiple `HashSet` members. This member is a `HashMap` with `Option<u32>` as the key type. The key `None` holds requirements for non-constant indices, while requirements for constant indices are stored as `Some`.
- Validation checks against the requirements, and pipeline barriers, now always include the requirements of the `None` key as well. This may result in false positives, but prevents false negatives.
- `DescriptorRequirements` now has `memory_read` and `memory_write` members, which hold a `ShaderStages` value for the stages which read or write the resource. This is used for more fine-grained pipeline barriers.
````

Not much to say here, it's pretty much in the changelog already.

### DIFF
--- a/examples/src/bin/runtime_array/main.rs
+++ b/examples/src/bin/runtime_array/main.rs
@@ -348,7 +348,9 @@ fn main() {
 
     let pipeline_layout = {
         let mut layout_create_infos: Vec<_> = DescriptorSetLayoutCreateInfo::from_requirements(
-            fs.entry_point("main").unwrap().descriptor_requirements(),
+            fs.entry_point("main")
+                .unwrap()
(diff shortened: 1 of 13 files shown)


AustinJ235 (approved):
Looks good

PR #2089, opened 2022-11-15
Add new `CommandBufferBuilder`, and `*CommandBuffer` types

### LINKED ISSUE #1360: Manual pipeline barrier in AutoCommandBuffer for debugging and workarounds
Would it be possible to add the ability to force a pipeline barrier in an autocommandbuffer. Currently in issues such as [this](https://github.com/vulkano-rs/vulkano/issues/1297) one must play around with other commands to try force a barrier in the case vulkano isn't insterting one automatically.

Would be nice with buffer_barrier and image_barrier functions for debugging purposes.

### LINKED ISSUE #1917: Proposal for new command buffer types
At the moment, when you use `AutoCommandBufferBuilder`, it performs both validation of the command and automatic insertion of pipeline barriers. But it's quite possible that a user might want to do their own barriers manually. There may be situations where they can't be done automatically, or where the automatic method is inefficient.

Currently, the user's only option is to use `UnsafeCommandBufferBuilder`. But this type is very unsafe; by using it, the user loses out on the command validation and also loses the ability to automatically keep resources alive inside the command buffer. It would be useful to have access to the safety of `AutoCommandBufferBuilder` without automatic barrier insertion.

So my thought is to rearrange the current types into two new layers: a safety/validation layer and an automatic barrier insertion layer.
- The builder types `UnsafeCommandBufferBuilder` and `SyncCommandBufferBuilder` would go.
- There would only be two types for built command buffers, `PrimaryCommandBuffer` and `SecondaryCommandBuffer`. The corresponding traits would go.
- A new type `CommandBufferBuilder` representing the safety/validation layer. It would validate all commands and keep resources alive, like `AutoCommandBufferBuilder` does now. It would not insert any barriers automatically, instead having a `pipeline_barrier` command to manually insert a barrier. But crucially, this command would be _safe_ and every call would be validated just like other command buffer commands. `CommandBufferBuilder` would keep track of accesses and validate the barriers inserted by the user to ensure they got it right.
- `AutoCommandBufferBuilder` would do what it does now, but it would be implemented on top of `CommandBufferBuilder`, by adding the automatic barrier insertion. Because the `CommandBufferBuilder` still validates the barriers, any bugs in `AutoCommandBufferBuilder` would hopefully be caught by the lower layer. In principle, `AutoCommandBufferBuilder` would just be one possible auto-barrier layer, the user could build their own if they wanted to.

Rua:
Changelog:
```markdown
### Additions
- A new `CommandBufferBuilder` type, that provides validation and keeps resources alive, but requires manual synchronization commands.
- The `PrimaryCommandBuffer` and `SecondaryCommandBuffer` types.
````

Closes #1917, closes #1360.

This adds the new "manual" command buffer builder that was announced some time ago. At the moment it does not do any synchronization tracking or validation, so any commands that rely on it are still unsafe. `PrimaryCommandBuffer` and `SecondaryCommandBuffer` don't implement the `PrimaryCommandBufferAbstract` and `SecondaryCommandBufferAbstract` traits yet either, so at the moment they're a bit useless. I'll work on this in future PRs.

### DIFF
--- a/vulkano/src/command_buffer/allocator.rs
+++ b/vulkano/src/command_buffer/allocator.rs
@@ -7,14 +7,11 @@
 // notice may not be copied, modified, or distributed except
 // according to those terms.
 
-//! In the Vulkan API, command buffers must be allocated from *command pools*.
+//! Traits and types for managing the allocation of command buffers and command pools.
 //!
-//! A command pool holds and manages the memory of one or more command buffers. If you destroy a
-//! command pool, all of its command buffers are automatically destroyed.
-//!
-//! In vulkano, creating a command buffer requires passing an implementation of the
-//! [`CommandBufferAllocator`] trait, which you can implement yourself or use the vulkano-provided
-//! [`StandardCommandBufferAllocator`].
+//! In Vulkano, creating a command buffer requires passing an implementation of the
+//! [`CommandBufferAllocator`] trait. You can implement this trait yourself, or use the
+//! Vulkano-provided [`StandardCommandBufferAllocator`].
 
 use super::{
     pool::{
--- a/vulkano/src/command_buffer/mod.rs
+++ b/vulkano/src/command_buffer/mod.rs
@@ -7,43 +7,63 @@
 // notice may not be copied, modified, or distributed except
 // according to those terms.
 
-//! Commands that the GPU will execute (includes draw commands).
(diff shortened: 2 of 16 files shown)


AustinJ235 (approved):
Great work so far!

PR #2090, opened 2022-11-18
Implement IntoIterator for `Features`, `DeviceExtensions` and `InstanceExtensions`

ilya-zlobintsev:

Changelog:
```markdown
### Additions
- Implement `IntoIterator` for `Features`, `DeviceExtensions` and `InstanceExtensions`
````

This PR adds implements `IntoIterator` to auto-generated structs, allowing to get the list of features/extensions as an iterator. This is useful when the values need to be used in a non-programmatic scenario, such as showing the list of features to the user.

### DIFF
--- a/vulkano/autogen/extensions.rs
+++ b/vulkano/autogen/extensions.rs
@@ -450,7 +450,7 @@ fn extensions_common_output(struct_name: Ident, members: &[ExtensionsMember]) ->
 
             /// Returns the list of extensions as an array.
             #[inline]
-            pub const fn as_arr(&self) -> [(&str, bool); #arr_len] {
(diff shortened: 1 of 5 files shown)


Rua:
While the idea in principle makes sense, I think it would be more idiomatic to use the `IntoIterator` trait for this. It can be implemented with `type Item = &'static str`.

ilya-zlobintsev:
> What is the `bool` in the item used for?

It shows if the feature/extension is enabled or not. My use case for this is showing the list of features with a status if they're supported or not. If someone wants just the list of names of supported features, it's trivial to filter and map the iterator, so I think this should cover both use cases.

Rua:
As long as it's clear that it only shows the extensions and features that are known to Vulkano at the time it was compiled. The list can change if a new one is added.

PR #2099, opened 2022-12-06
Add basic synchronization tracking to `CommandBufferBuilder`

Rua:
This adds tracking of buffer and image synchronization, and the application of pipeline barriers. Other synchronization primitives (events, render passes) are not covered yet, and no validation checks are performed yet based on this data (because it's still incomplete).

The design is heavily based on the code used in the official validation layer. I did my best to document it clearly, but it's tricky stuff that I can't even fully follow right now.


AustinJ235 (approved):
Looking great!

PR #2101, opened 2022-12-10
Add synchronization tracking for render pass operations in `CommandBufferBuilder`

Rua:
Continuing on implementing the new command buffer, this adds most of the render pass operations, including:
- Load operations at the start of a subpass
- Use of attachments during a draw operation
- Resolve operations at the end of a subpass
- Store operations at the end of a subpass


AustinJ235 (approved):
Looking good

Issue #893, opened 2017-12-09
Dynamic vertex buffer layouts?


DiThi:
In all the examples we have something like this

```rust
struct Vertex { position: [f32; 2] }
impl_vertex!(Vertex, position);
```

but I want to load a bunch of different meshes, some with normals, some with one or more UV layers, colors, skeletal weights, etc...

Is it possible to specify the layout at run time?

I'm sorry if there's a simple solution with a Rust feature I've missed.

tomaka:
It is doable, but not easy yet. Basically you have to implement the `VertexDefinition` and `VertexSource` traits on a struct of yours.

There is an example here: https://github.com/tomaka/vulkano-examples/blob/81f5a4eb3c5f3fcc6a194d3c41e53b2bc66f7add/gltf/gltf_system.rs#L687-L795

Rua:
Would a partial solution be to make the `Vertex` trait take `&self`? Then it would be easy to make a custom type that returns runtime values in the `member` method.

Rua:
There is now the `VertexInputState` type that gives you the same freedom as plain Vulkan.

Issue #736, opened 2017-08-09
Runtime VertexAttributes


Elzair:
Can you create a `GraphicsPipeline` or a `Buffer` when you only know the vertex attributes at runtime? I see there is an example for loading shaders at runtime, but the Vertex data is still specified in the code.

Say someone wants to write a viewer for GLTF 2.0 scenes. The [spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes) says meshes can have any of eight different vertex attributes (POSITION, NORMAL, TANGENT, TEXCOORD_0, TEXCOORD_1, COLOR_0, JOINTS_0, WEIGHTS_0) in any order (for now, let's assume they are all interleaved in one buffer). Accounting for all arrangements would require defining `floor(8! * e) - 1 = 109,600` structs. There HAS to be a better way to do this!

tomaka:
For the graphics pipeline it is possible by writing your own implementation of [the VertexDefinition trait](https://docs.rs/vulkano/0.5.6/vulkano/pipeline/vertex/trait.VertexDefinition.html). This trait is a bit crappy though.

For the buffer, if everything is a float you can simply create a buffer of type `[f32]`. Otherwise I guess you would have to use `[u8]` or `[u32]`, and either transmutes or [the to_bits functions](https://github.com/rust-lang/rust/issues/40470) that will be stable in Rust 1.20. This is not really something that is specific to vulkano I guess. If you want to create a box whose content type is only known at runtime, you have the same problem.

Also note that for the moment the content of the vertex buffer is not checked. In other words you can use any buffer of any content in combination with any vertex definition and will not get an error. That eliminates a potential problem.

tomaka:
Slightly off-topic but I wanted to write a glTF loader as well, so I'll probably encounter the same problems.

trevex:
Let me piggy back this issue:

I was looking for a nice way to leverage runtime vertex attributes as well.
While I am new to Rust and Vulkan(o), I believe Vulkano changed a lot since the inception of this issue.

When I tried to implement runtime vertex attributes, I ended up with a `Mesh` structure roughly as follows:
```rust
#[derive(Hash, Eq, PartialEq, Debug)]
pub struct VertexAttribute {
    pub name: Cow<'static, str>,
    pub format: Format,
}

pub struct Mesh {
    attributes: HashMap<VertexAttribute, Arc<dyn BufferAccess>>,
    length: u64,
}
```

[...]

PR #2106, opened 2022-12-18
Refactor Vertex trait to allow user-defined formats

trevex:
This a draft PR containing changes required to make format definition using the Vertex trait more flexible.
While discussing this with @Rua in Discord I noticed this might be above my head, but decided to give it a go regardless. Any input is appreciated as I am a rust and vulkan/vulkano novice and keen to learn.

The "vision" of this PR is to have more control over the used attribute format. In code this might look as follows:
```rust
// `impl_vertex` macro still works but relies on VertexMember trait for formats.
pub struct MyVertex {
    pub position: [f32; 3], // inferred formats?
    pub uvs: [f32; 2],
    pub texture_id: u16,
}
impl_vertex!(MyVertex, position, uvs, texture);

// proc macro can be used and if no field-level attributes exist VertexMember-trait is used.
#[derive(Vertex)]
pub struct MyVertex {
    pub position: [f32; 3], // inferred formats?
    pub uvs: [f32; 2],
    pub texture_id: u16,
}

// field-level attributes can change format etc. This also removes the need to implement VertexMember if `format` is provided
#[derive(Vertex)]
pub struct MyVertex {
    #[format(R32B32G32_SFLOAT)]
    pub position: glam::Vec3,
    #[format(R16B16_SFLOAT)]
    pub uvs: [half::f16; 2],
    #[name("albedo_id")] // Could also support list of names for multiple aliases in shaders
    pub texture_id: u16,
    #[format(R32B32G32A32_SFLOAT)]
    pub matrix: [f32; 16], // sizeof(matrix) / format.size => num locs (no remainder allowed)
}
```

`VertexMember` trait has been changed to simply return `Format`. However we need additional information to properly match attributes to their location(s). From my understanding we need the following:
* Format: provided by `VertexMember`
* Number of components: can be derived from Format
* Number of elements/locations: `sizeof(struct-field) / format` (no remainder allowed)
* Offset taken from field

All information can be obtained from format and the field itself in the macro and are returned as `VertexMemberInfo`.

Changelog:
```markdown
### Breaking changes
Deprecations:
- `impl_vertex` superseded by deriving `Vertex` instead
- `VertexMember` is deprecated as deriving `Vertex` requires explicit format specification
Changes to `impl_vertex`, `VertexMember` and `VertexMemberInfo`:
- `VertexMember::format` now directly returns a `Format`
- `VertexMemberInfo` now holds format and `VertexMemberTy` was removed
- `impl_vertex` was updated to correctly use the new `VertexMember::format` signature

### Additions
- `Vertex` can now be derived and formats have to be explicitly specified on each field
- BufferDefinition matching logic was updated to work with the new `VertexMemberInfo` and now matches based on scalar type, number of components and number of elements allowing the use of formats such as `*_UNORM`.
- `GraphicsPipelineBuilder` validation was extended to make sure locations are not bound multiple times when a single attribute spans multiple locations (e.g. using double precision float formats)
````

fixes https://github.com/vulkano-rs/vulkano/issues/1667
fixes https://github.com/vulkano-rs/vulkano/issues/1572

### CODE REVIEW

**Comment by trevex** [comment]
The proc-macro with both attributes outlined above is implemented. I added tests both for `impl_vertex` and the `#[derive(Vertex)]` approach.

Before I dive into the writing user-facing docs, it would be great to receive some feedback and preliminary review.

File: vulkano/src/pipeline/graphics/vertex_input/vertex.rs [hunk 1]
```diff
@@ -36,7 +36,17 @@ pub struct VertexMemberInfo {
     /// Attribute format of the member. Implicitly provides number of components.
     pub format: Format,
     /// Number of consecutive array elements or matrix columns using format.
-    pub num_elements: u32,
+    pub num_locations: u32,
```
Rua, line 39: In principle, the number of components/locations is determined exactly by the format. Every element of a format (containing up to 4 components) takes exactly one location, unless it's a 64-bit format, in which case it takes up 2 locations if there's 3 or 4 components.

trevex: I renamed the field back to `num_elements` and explained the distinction in the comment. I also updated the `VertexDefinition` implementation of `BuffersDefinition`. A preliminary test I did using the triangle example with a format exceeding a single location worked.

There are several to dos in the implementation of `ShaderInterfaceEntryType`, so on the reflection side double precision seems to not be properly supported. I added a to do to `BuffersDefinition` to indicate that the `VertexDefinition` logic might need updates accordingly when the former is implemented.

Strictly speaking I could also make sure in the `BuffersDefinition` that a location following a location occupied by a format exceeding a single location is kept free. Is that something that should happen as part of VertexDefinition or something to be validated by the pipeline builder?

Rua: I think in principle, anything that could be specified directly in `VertexInputState` should be checked by the pipeline builder. That includes ensuring that the info is compatible with the shader.

vulkano/macros/src/derive_vertex.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+        for name in &names {
+            member_cases = quote! {
+                #member_cases
+
+                if name == #name {
+                    #format
+                    let field_size = {
+                        let m = core::mem::MaybeUninit::<#struct_name>::uninit();
+                        let p = unsafe { core::ptr::addr_of!((*(&m as *const _ as *const #struct_name)).#field_name) };
```
Rua, line 68: This unsafe stuff shouldn't be needed anymore, because you presumably have direct access to the field type as a `TokenStream` and can just give that to `size_of`.

vulkano/src/pipeline/graphics/vertex_input/buffers.rs
```diff
@@ -131,35 +131,39 @@ unsafe impl VertexDefinition for BuffersDefinition {
                         attribute: name.clone().into_owned(),
                     })?;
 
-            if !infos.ty.matches(
-                infos.array_size,
-                element.ty.to_format(),
-                element.ty.num_locations(),
-            ) {
-                // TODO: move this check to GraphicsPipelineBuilder
+            // TODO: ShaderInterfaceEntryType does not properly support 64bit.
```
Rua, line 134: What about it doesn't support 64-bit right now?

vulkano/src/pipeline/graphics/builder.rs
```diff
@@ -994,7 +994,7 @@ where
                         // TODO: Check component assignments too. Multiple variables can occupy the same
                         // location but in different components.
 
-                        let shader_type = element.ty.to_format().type_color().unwrap();
+                        let shader_type = element.ty.to_numeric_type();
```
Rua, line 997: `shader_type` should really be the `ShaderScalarType` value directly, and then the match below should match on that.

vulkano/src/pipeline/graphics/vertex_input/vertex.rs
trevex: Added additional validation logic to `GraphicsPipelineBuilder` in last commit.

vulkano/src/pipeline/graphics/vertex_input/buffers.rs
trevex: I am referring to https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/shader/mod.rs#L1034, which led me to believe the num_locations logic might change.

Rua: It would be nice to have a proper implementation for that function, but I think the interface is incomplete as it is. The shader interface should really state exactly which components and locations it needs, and what type of value goes in each one. You don't have to implement that though.

vulkano/src/pipeline/graphics/builder.rs
trevex: Changed in latest commit.

vulkano/src/pipeline/graphics/builder.rs
```diff
@@ -723,7 +723,23 @@ where
...
                 // VUID-VkPipelineVertexInputStateCreateInfo-pVertexAttributeDescriptions-00617
-                // Ensured by HashMap.
+                // Ensured by HashMap with the exception of formats exceeding a single location.
+                // When a format exceeds a single location the location following it (e.g.
+                // R64B64G64_SFLOAT) needs to be unassigned.
+                let unassigned_locations: Vec<u32> = attributes
+                    .iter()
+                    .filter(|&(_, attribute_desc)| attribute_desc.format.block_size().unwrap() > 16)
+                    .map(|(location, _)| location + 1)
+                    .collect();
```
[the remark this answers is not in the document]
trevex: Thanks, for catching that! Still new to rust so those minor things are a big help.

vulkano/macros/src/derive_vertex.rs
trevex: Done, removed both unsafe blocks from proc macro and calculate offset and field_size using field type.

vulkano/src/pipeline/graphics/vertex_input/buffers.rs
trevex: Shall I keep the comment or adapt it in some way?

Rua: You can keep it for now.

trevex:
I rebased to resolve merge conflicts and re-implemented support for cgmath and nalgebra including their matrix types.
A simple example and documentation was added.

I believe the PR is now ready for a proper review.

Rua:
Is your aim to deprecate `impl_vertex`, or have them both exist side by side? If they both do the same thing, or if one of them can do everything the other can, there may not be much point in keeping both.

trevex:
Personally I prefer the derive macro as it is more powerful. If that is something agreed upon I can mark `impl_vertex` as deprecated and update the examples.

It looks like CI is failing due to the proc-macro failing documentation tests. I will try to reproduce this locally later.

trevex:
Alternatively I can also fully remove `impl_vertex`.

Another thought that crossed my mind is the removal of `VertexMember` and making the format choice explicit. This is again a personal impression, but I tend to choose formats explicitly as I use f16 or s/unorm a lot.

Rua:
I think deprecating it would be good. You could also re-export the `Vertex` macro from the `crate::pipeline::graphics::vertex_input` module, so that users don't need to import the proc macro module separately in their project.

I've thought about removing `VertexMember` as well, but would `impl_vertex` still work without it? How would the `Vertex` macro be implemented without it?

vulkano/macros/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+authors = ["Pierre Krieger <pierre.krieger1708@gmail.com>", "The vulkano contributors"]
+repository = "https://github.com/vulkano-rs/vulkano"
+description = "Macros used by Vulkano"
+license = "MIT/Apache-2.0"
+documentation = "https://docs.rs/vulkano"
+homepage = "https://vulkano.rs"
+keywords = ["vulkan", "bindings", "graphics", "gpu", "rendering"]
+categories = ["rendering::graphics-api"]
+
```
Rua, line 13: If this macro is going to be re-exported from the main crate, I think this should have `publish = false` so that it doesn't get uploaded to crates.io?

trevex: Yes, you are correct it is re-exported and publish should be false.

trevex:
> I think deprecating `impl_vertex` would be good. You could also re-export the `Vertex` macro from the `crate::pipeline::graphics::vertex_input` module, so that users don't need to import the proc macro module separately in their project

The latter already happens.

> I've thought about removing VertexMember as well, but would impl_vertex still work without it? How would the Vertex macro be implemented without it?

`impl_vertex` would break or need a major refactor that most likely is not worth it. The Vertex macro would treat the format field level attribute as mandatory and error on the field otherwise.

Rua:
That actually seems like a better idea. So then `VertexMember` would be deprecated as well.

trevex:
Alright, I will deprecate `impl_vertex` and `VertexMember`, make the format mandatory and update all example. I will remove the example I added as it is irrelevant with those changes.

trevex:
All discussed changes/deprecations are now implemented and all examples updated.

There was an edge-case with proc-macros and inferring the crate name in doctests, see comment here in latest commit: https://github.com/vulkano-rs/vulkano/pull/2106/commits/aae1954e59446bd02572edace4e2660f335d6e13#diff-79f1e6c12099322d7e61268e112edce8a4a96ee6cc0b265b22f184cc7ebadd2fR132

This fixes the doctests, but I had to re-export crate self as vulkano.

Rua:
Everything looks good! I noticed a few open Vulkano issues that this seems to solve. If you agree, can you add "fixes #X" for each issue number to your first post? That will link them to this fix and close them together.

https://github.com/vulkano-rs/vulkano/issues/1667
https://github.com/vulkano-rs/vulkano/issues/1572
https://github.com/vulkano-rs/vulkano/issues/893
https://github.com/vulkano-rs/vulkano/issues/736

PR #2119, opened 2022-12-28
Better custom vertex buffer layouts support

trevex:
The goal of this PR is to make creating and using custom vertex buffer layouts easier and allow them to leverage a pre-existing `VertexDefinition` implementation. This will also allow vertex types derived from `Vertex` to interact with custom runtime created ones.

Here is some pseudo-code to illustrate the idea:
```rust
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, Zeroable, Pod, Vertex)]
struct InstanceData {
    #[format(R32G32_SFLOAT)]
    position_offset: [f32; 2],
    #[format(R32_SFLOAT)]
    scale: f32,
}

// There is no runtime vertex buffer builder helper yet, but just image there is :)
const ATTRIBUTE_POSITION: VertexAttribute = VertexAttribute::new("position", Format::R32G32_SFLOAT);
const ATTRIBUTE_COLOR: VertexAttribute = VertexAttribute::new("color", Format::R32G32B32_SFLOAT);

let (buffer, info) = RuntimeVertexBufferBuilder::new()
        .add(
            ATTRIBUTE_POSITION,
            vec![
                Vec2::new(-0.5, -0.25),
                Vec2::new(0.0, 0.5),
                Vec2::new(0.25, -0.1),
            ],
        )
       .build();

// Both can be used together and leverage the matching logic previously implemented by `BuffersDefinition`
let pipeline = GraphicsPipeline::start()
        .vertex_input_state([info.per_vertex(), InstanceData::per_instance()])
```

Changelog:
```markdown
### Breaking changes
Deprecates `BuffersDefinition`:
- Use `VertexBufferDescription` directly.

### Additions
- Introduction of `VertexBufferDescription` makes it easy to create custom description and leverage the matching logic previously provided by `BuffersDefinition`.
````

While this change does not directly implement runtime interleaved vertex buffers. There is nothing stopping users from implementing them as demonstrated in Discord by @marc0246 and on the `runtime-vertex-*` branches of my fork.

Fixes https://github.com/vulkano-rs/vulkano/issues/736

### CODE REVIEW

**Comment by trevex** [comment]
Updated the description and reimplemented BuffersDefinition, although it would be deprecated by these changes.

If you look at the code please careful review the names I have chosen for structs. I am not sure the `*Info` is a great idea considering they have no relation to Vulkan and might confuse users, but I have yet to come up with better ones. Opinions appreciated.

File: vulkano/src/pipeline/graphics/vertex_input/vertex.rs [hunk 1]
```diff
@@ -37,17 +40,30 @@ pub use vulkano_macros::Vertex;
...
-        None
+    fn info() -> VertexInfo {
+        VertexInfo {
+            members: VertexMemberMap::default(),
+            stride: 0,
+        }
     }
 }
 
+pub type VertexMemberMap = HashMap<String, VertexMemberInfo>;
```
Rua, line 58: Does this need to be a type alias? I feel like that just obscures to the user what's actually going on.

trevex: Not necessarily, this was just my rust-novice thinking, that this might save an `ahash` dependency on dependent crates. However this is not performance critical, so I am thinking maybe transparently using the `std`-HashMap is better.

vulkano/src/pipeline/graphics/vertex_input/vertex.rs
```diff
@@ -7,10 +7,14 @@
 // notice may not be copied, modified, or distributed except
 // according to those terms.
 
+use std::collections::HashMap;
+
 use crate::format::Format;
 use bytemuck::Pod;
 pub use vulkano_macros::Vertex;
 
+use super::{VertexInputBindingDescription, VertexInputRate};
```
Rua, line 16: Just a minor style nitpick, Vulkano doesn't put empty lines between the imports anywhere. Instead we just let `cargo fmt` sort out the order.

vulkano/src/pipeline/graphics/vertex_input/definition.rs
```diff
@@ -113,3 +116,105 @@ impl Display for IncompatibleVertexDefinitionError {
...
+unsafe impl VertexDefinition for &[VertexBufferInfo] {
+    #[inline]
+    fn definition(
+        &self,
+        interface: &ShaderInterface,
+    ) -> Result<VertexInputState, IncompatibleVertexDefinitionError> {
+        let bindings = self
+            .iter()
+            .enumerate()
+            .map(|(binding, buffer)| (binding as u32, (*buffer).clone().into()));
```
Rua, line 129: It seems like a bit of a waste if functions like `per_vertex` return `VertexBufferInfo` by value, this function borrows it, and then we end up cloning it here anyway. Is there an ergonomic way to avoid this clone?

trevex:
Thanks @Rua, I will go through your feedback.

As the changes of this PR make the implementation of runtime vertex formats possible I wanted to give it a try. The last commit implements a way to create interleaved runtime vertex formats and buffers. I added a simple test and example and seems to work fine.

Non-interleaved vertex buffers would already be possible either way, but this fixes the open issues regarding interleaved runtime buffers.

I expect the code to need careful review regarding allocations and the iterator logic.

Should I include those as part of this PR or move those changes elsewhere?

Rua:
The runtime stuff is probably better left to a separate PR.

If this is meant to deprecate `BuffersDefinition`, can you add a deprecation notice to it? And can you add documentation to the methods of `Vertex` and `VertexBufferInfo`?

vulkano/src/pipeline/graphics/vertex_input/vertex.rs
trevex: Will be fixed in next commit. Not sure why that happened. I am using `rust-tools.nvim` (with `rust-analyzer`) and this is how imports are formatted when using code completion to assign a type.

marc0246: Since rust-fmt's configuration for this (import grouping) is currently unstable, rust-analyzer has its own configuration for it. You can configure it like so when using packer:
```lua
  use {
    "simrat39/rust-tools.nvim",
    config = function()
      require "rust-tools".setup {
        server = {
          settings = {
            ["rust-analyzer"] = {
              checkOnSave = {
                command = "clippy",
              },
              imports = {
                group = {
                  enable = false,
                },
              },
            },
          },
        },
      }
    end,
  },
```
(I also recommend using clippy)

vulkano/src/pipeline/graphics/vertex_input/definition.rs
trevex: I am instantiating `VertexInputBindingDescription` directly now. But not sure if there is a much better way. I thought about making `VertexDefinition` consuming, but that would not be a great help either and make other cases e.g. using same `VertexBufferInfo` with multiple pipelines less ergonomic.

vulkano/src/pipeline/graphics/vertex_input/vertex.rs
trevex: Thanks, updated my config :+1:

trevex:
Before continuing to update the examples:
I would like to get some feedback on the struct-name `VertexBufferInfo`. Would `VertexFormat` or `VertexBufferFormat` be a better name?

Rua:
It seems to describe the contents of a vertex buffer, so how about `VertexBufferDescription`?

vulkano/src/pipeline/graphics/vertex_input/definition.rs
Rua: I suppose it could be solved by making `per_vertex` and co return a `'static` reference instead of a value, so the value is only constructed once and then reused. But I can't think of a way to do that without requiring the user to add a dependency to `once_cell` to their project. I wish they'd hurry up and stabilise it in the standard library already.

PR #2122, opened 2023-01-03
Optimize memory layout checks and fix allocation soundness

marc0246:
This introduces two new types and a new type alias: `DeviceLayout`, `DeviceAlignment` and `NonZeroDeviceSize`. Previously, allocation sizes and alignments were verified in the functions that allocate, and in functions that construct buffers. All that is now gone as those checks have been moved to `DeviceLayout`, which is a replica of std's `Layout`, but for Vulkan. Another issue was that previously, no checks were made whether the allocation size would overflow when rounded up, which would lead to undefined behavior. `DeviceLayout`'s size can not overflow `i64::MAX`, which eliminates that concern.

I've also marked some private functions `unsafe` and added more thourough safety documentation for them, and tried to document every possible case of arithmetic overflow as well. Hopefully this will make it easier to review the code and change it in the future.

```markdown
### Breaking changes
- `MemoryRequirements::{size, alignment}` fields have been replaced with a single `layout` field.
- `Suballocator::allocate_unchecked` has been removed.

### Additions
- Added `DeviceLayout` and `DeviceAlignment`.
- Added the type alias `NonZeroDeviceSize`.

### Bugs fixed
- Fixed possible arithmetic overflows when allocating memory with a size that's too large.
- Added missing check for the allocation type when binding memory.
```

### DIFF
--- a/vulkano/src/buffer/allocator.rs
+++ b/vulkano/src/buffer/allocator.rs
@@ -20,7 +20,8 @@ use crate::{
     memory::{
         allocator::{
             align_up, AllocationCreateInfo, AllocationCreationError, AllocationType,
-            MemoryAllocatePreference, MemoryAllocator, MemoryUsage, StandardMemoryAllocator,
+            DeviceAlignment, DeviceLayout, MemoryAllocatePreference, MemoryAllocator, MemoryUsage,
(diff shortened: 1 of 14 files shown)


Rua:
Very nice work!

PR #2127, opened 2023-01-07
Unify all buffer types

### LINKED ISSUE #867: How to read from CpuBufferPool/CpuBufferPoolChunk
I am currently using a compute shader to write to a buffer, I was previously using a `CpuAccessibleBuffer` but from the documentation this may be removed from Vulkano in future so I am moving to using a `CpuBufferPool`. 

I have a problem though, I cannot see any API methods which would allow me to read the data back out of the `CpuBufferPoolChunk`.

The documentation implies that reading from the `CpuBufferPool` on the CPU should be possible:

> if a buffer is under constant access by the GPU but you need to read its content on the CPU from time to time, it may be a good idea to use a DeviceLocalBuffer as the main buffer and a CpuBufferPool for when you need to read it. Then whenever you need to read the main buffer, ask the GPU to copy from the device-local buffer to the CPU buffer pool, and read the CPU buffer pool instead.

### LINKED ISSUE #962: CpuBufferPool array allocator
Hi,
  I've got a:
     buffer::device_local::DeviceLocalBuffer::<[u8]>::array(vdevice.clone(), voxelsize*voxelsize*voxelsize,...

I'd like to dump, and I was hoping to use a CpuBufferPool for that, but it seems to be missing an ::array() allocator, it looks like it'll only take a fixed type for allocations, for example:

174 |         let cpubuf = buffer::cpu_pool::CpuBufferPool::<[u8]>::download(self.vdevice.clone());
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `bulbvulk::vulkano::buffer::CpuBufferPool<[u8]>`
    |
    = note: the method `download` exists but the following trait bounds were not satisfied:
            `[u8] : std::marker::Sized`

error[E0277]: the trait bound `[u8]: std::marker::Sized` is not satisfied

I think it would be good to be able to allocate a CpuBufferPool in the same way that DeviceLocalBuffer lets you.
(Actually an even nicer thing would be to be able to pass a reference to my DeviceLocalBuffer and to create a CpuBufferPool of exactly the right size/type)
I guess I'll use a CpuAccessibleBuffer but that has the scary warning it might disappear.

### LINKED ISSUE #1281: Allow slicing on BufferAccess?
I have buffers containing different vertex formats, but I want to store them in a single data structure, which necessitates storing them as `dyn BufferAccess` to erase the type. However, this means that I am no longer able to create slices of the buffers, because that requires `TypedBufferAccess`.

It seems to me that the type parameter of buffers and slices isn't really needed, and is sometimes even counterproductive. Vulkano doesn't check the type of vertex inside the buffer against the pipeline layout, so you can put anything in the buffer and Vulkano will accept it. This makes sense, because you may want to load vertex information dynamically and construct a buffer of [u8] without having the "correct" static Rust type to use, and Vulkano shouldn't choke on that.

From the Rust side of things, all that really matters for slices is getting the element size correct to avoid cutting vertex elements in half. The Rust type of the buffer elements is totally irrelevant because they are never accessed from Rust code. So I think it would make sense if buffers simply store their element size instead of relying on the Rust type. That would allow them to become untyped, making `TypedBufferAccess` obsolete and allowing slicing directly with `BufferAccess`.

### LINKED ISSUE #1733: Panic in `CpuBufferPoolChunk::clone`
*   Version of vulkano: 0.24.0
*   OS: Win 10 x64
*   GPU (the selected PhysicalDevice): RTX 2070 Super
*   GPU Driver: 496.13

# Issue

I'm getting a panic when cloning a `CpuBufferPoolChunk`, where the corresponding `ActualBufferChunk` cannot be found. The only way I could find that this would be possible was for the entry in the `ActualBuffer` to be removed when there are still multiple references to it.

The panic in question happens at https://github.com/vulkano-rs/vulkano/blob/v0.24.0/vulkano/src/buffer/cpu_pool.rs#L602 

Additionally, the issue doesn't trigger when I run the debug build, but does when I run the release build. This suggests to me that there is some race condition, since the debug build runs slower than the release build, but I don't really understand what could cause this.

Is there a workaround or would the newer versions of vulkano fix this?

### LINKED ISSUE #2126: DeviceLocalBuffer API doesn't expose sharing mode
DeviceLocalBuffer's constructors don't expose SharingMode, so I didn't even know that this existed at first. It came to my attention when validation layers detected that it was not set correctly. It's set to Exclusive by default, but I need it to be set to Concurrent since I am uploading data to my buffers in a background thread on a background queue, then using them on the main thread in the main queue. 

Looking at the source code, this seems like a fairly easy fix, but adding a parameter is an API change so it warrants discussion before making a PR.

marc0246:
This PR aims to unify the types `CpuAccessibleBuffer`, `DeviceLocalBuffer`, `BufferSlice` and `CpuSubbuffer` (formerly two types, `CpuBufferPoolSubbuffer` and `CpuBufferPoolChunk`) into one: `Subbuffer`.

What this solves is the need for boxed trait objects, which can be detrimental to performance and might make the API harder to understand for newcommers. Another thing is that with the functionality spread like it is, some might not even be aware of what they can do. For example `BufferSlice`, was until now not that visible. With everything being in once place, people can discover things more easily. Another issue was that when slicing/indexing a buffer, since all the APIs expect an `Arc<dyn BufferAccess>`, you have to box every `BufferSlice`, which includes slices of slices, and would lead to a chain of `Arc`s as long as the number of slicings. That's a bummer because ideally all that needs to change is an offset and size in the existing one.

Another benefit of having only one type for buffers is type complexity: the `BufferAccess`, `TypedBufferAccess` and `BufferAccessObject` traits are gone and so are all the generics and trait bounds associated with them. Also, `BufferView` no longer has a generic type parameter and the `BufferViewAbstract` trait is also gone. I think this makes the API more approachable.

This PR also addresses the long-standing issue of the (then) `CpuBufferPool` only working for download. The (now) `SubbufferAllocator` can allocate subbuffers for any purpose.

Fixes #867, fixes #962, fixes #1281, fixes  #1733, fixes #2126.

Changelog:
```diff
  # Breaking changes
- Changes to CPU buffer allocation:
- - Replaced `CpuBufferPool` with `CpuBufferAllocator`, which is now marked `!Sync` and no longer has a `T` type parameter. The type parameter was moved to the methods, to allow one allocator to allocate as many types of buffers as needed.
+ Changes to buffer allocation:
+ - Replaced `CpuBufferPool` with `SubbufferAllocator`, which is now marked `!Sync` and no longer has a `T` type parameter. The type parameter was moved to the methods, to allow one allocator to allocate as many types of buffers as needed.

- - Merged `CpuBufferPoolChunk` and `CpuBufferPoolSubbuffer` into `CpuSubbuffer`.'
+ Changes to buffers:
+ - Merged `CpuAccessibleBuffer`, `DeviceLocalBuffer`, `BufferSlice`, `CpuBufferPoolChunk`, `CpuBufferPoolSubbuffer` into `Subbuffer`.
+ - Removed `BufferAccess`, `TypedBufferAccess`, `BufferAccessObject`, `BufferInner` and `BufferViewAbstract`.

  # Additions
  - Added `BufferAllocateInfo`.
```


vulkano/src/buffer/mod.rs
```diff
@@ -71,38 +71,881 @@
...
+
+    /// Returns the device address for this buffer.
+    // TODO: Caching?
+    pub fn device_address(&self) -> Result<NonZeroDeviceSize, BufferError> {
+        let device = self.device();
+
+        // VUID-vkGetBufferDeviceAddress-bufferDeviceAddress-03324
+        if !device.enabled_features().buffer_device_address {
+            return Err(BufferError::RequirementNotMet {
+                required_for: "`BufferAccess::raw_device_address`",
```
Rua, line 348: This should say `Buffer` now.

marc0246:
It's ready. I wish the Discord bot would pick up on this.

PR #2133, opened 2023-01-18
Document data layout in shaders

### LINKED ISSUE #1111: Mention that vec3[] still needs 4 elements
It took me a the whole day to debug why my shader did not receive the data I was sending:
```
layout(set = 0, binding = 0) uniform SomeData {
    vec3[] numbers;
} buf;
```
Turns out that instead of using `Buffer::fromIter<Vec<[f32; 3]>` I need to use a `Vec<[f32; 4]>` and fill the last element with some garbage (ie 0).

Might be good to have this somewhere in the docs (there is no example that sends a `vecX` to the shader.

Rua:
Changelog:
```markdown
### Additions
- Added documentation to the `shader` module to explain the layout of buffers, push constants and other data accessed by shaders.
````

Fixes #1111. This should hopefully reduce the confusion with users when shaders don't read the data they provided.

### DIFF
--- a/vulkano/src/buffer/mod.rs
+++ b/vulkano/src/buffer/mod.rs
@@ -90,9 +90,13 @@
 //! Using uniform/storage texel buffers requires creating a *buffer view*. See [the `view` module]
 //! for how to create a buffer view.
 //!
+//! See also [the `shader` module documentation] for information about how buffer contents need to
+//! be laid out in accordance with the shader interface.
+//!
 //! [`RawBuffer`]: self::sys::RawBuffer
 //! [`SubbufferAllocator`]: self::allocator::SubbufferAllocator
 //! [the `view` module]: self::view
+//! [the `shader` module documentation]: crate::shader
 
 pub use self::{subbuffer::Subbuffer, usage::BufferUsage};
 use self::{
--- a/vulkano/src/buffer/mod.rs
+++ b/vulkano/src/buffer/mod.rs
@@ -90,15 +90,6 @@
 //! Using uniform/storage texel buffers requires creating a *buffer view*. See [the `view` module]
 //! for how to create a buffer view.
 //!
-//! # A note on endianness
-//!
-//! The Vulkan specification requires that a Vulkan implementation has runtime support for the
-//! types [`u8`], [`u16`], [`u32`], [`u64`] as well as their signed versions, as well as [`f32`]
-//! and [`f64`] on the host, and that the representation and endianness of these types matches
-//! those on the device. This means that if you have for example a `Subbuffer<[u32]>`, you can be
-//! sure that it is represented the same way on the host as it is on the device, and you don't need
-//! to worry about converting the endianness.
-//!
 //! [`RawBuffer`]: self::sys::RawBuffer
 //! [`SubbufferAllocator`]: self::allocator::SubbufferAllocator
 //! [the `view` module]: self::view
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: vulkano/src/shader/mod.rs [hunk 1]
```diff
@@ -16,6 +16,86 @@
...
+//!
+//! In all three of these alignment rules, a primitive/scalar value with a size of N bytes has an
+//! alignment of N, meaning that it must have an offset in the buffer/memory that is a multiple of
+//! its size, like in C or Rust. So, for example, a `float` (like a Rust `f32`) has a size of 4 bytes, and
+//! an alignment of 4.
+//!
+//! The differences between the alignment rules are in how compound types (vectors, matrices,
+//! arrays and structs) are expected to be laid out. For a compound type with an element whose
+//! alignment is N, the scalar alignment considers the alignment of the compound type to be also N.
+//! However, the base and extended alignment are stricter:
```
marc0246, line 67:
```suggestion
//! However, the base and extended alignments are stricter:
```

vulkano/src/shader/mod.rs
```diff
@@ -16,6 +16,86 @@
...
+//!
+//! | GLSL type | Scalar         | Base           | Extended                                     |
+//! |-----------|----------------|----------------|----------------------------------------------|
+//! | primitive | N              | N              | N                                            |
+//! | `vec2`    | N              | N * 2          | N * 2                                        |
+//! | `vec3`    | N              | N * 4          | N * 4                                        |
+//! | `vec4`    | N              | N * 4          | N * 4                                        |
+//! | array     | N              | N              | N, rounded up to multiple of 16              |
```
marc0246, line 75: I think this is expressed more simply as:
```suggestion
//! | array     | N              | N              | max of N and 16                              |
```
The specification expresses things in a really convoluted way...

vulkano/src/buffer/mod.rs
```diff
@@ -90,15 +90,6 @@
...
-//! # A note on endianness
-//!
-//! The Vulkan specification requires that a Vulkan implementation has runtime support for the
-//! types [`u8`], [`u16`], [`u32`], [`u64`] as well as their signed versions, as well as [`f32`]
-//! and [`f64`] on the host, and that the representation and endianness of these types matches
-//! those on the device. This means that if you have for example a `Subbuffer<[u32]>`, you can be
-//! sure that it is represented the same way on the host as it is on the device, and you don't need
-//! to worry about converting the endianness.
-//!
 //! [`RawBuffer`]: self::sys::RawBuffer
```
marc0246, line 93: I think it's best to at least link to the shader module here, so that it's more discoverable.
```suggestion
//! See also [the `shader` module documentation] for information about how buffer contents need to 
//! be laid out in accordance with the shader interface.
//!
//! [`RawBuffer`]: self::sys::RawBuffer
//! [the `shader` module documentation]: crate::shader
```

vulkano/src/shader/mod.rs
Rua: That's not exactly the same though. If N were, for example, 24, then it would be rounded up to 32.

marc0246: Alignments must be a power of two though, which is why taking the maximum always works.

Rua: Ah right, good point.

marc0246:
Amazing, I think it's perfect now 😙👌

AustinJ235 (approved):

PR #2142, opened 2023-02-15
Update dependencies, clippy warnings

Rua:
Changelog:
```markdown
### Public dependency updates
- [ash](https://crates.io/crates/ash) 0.37.2
- [bytemuck](https://crates.io/crates/bytemuck) 1.13
- [nalgebra](https://crates.io/crates/nalgebra) 0.32
- [winit](https://crates.io/crates/winit) 0.28
````
There seems to be something wrong with a dependency conflict in the gl-interop example. Maybe @fayalalebrun should have a look at that.


fayalalebrun:
The problem is that winit was updated to 0.28, but glium is still stuck on 0.27. Therefore the solution is to import winit 0.27 under a different alias and then change the code so that vulkano windows and glium windows are created using the correct winit version.

Rua:
That doesn't seem possible, because the same `event_loop` variable is passed to both Glium and Vulkano-winit.

fayalalebrun:
I don't think it's necessary for both to have the same event loop. You'd probably need to run each event loop in its own thread however. It's also possible to run them in different processes, by sharing the FDs via a unix socket.

Rua:
I will leave the errors as they are for this PR, and then you can solve it whatever way works best afterwards?

fayalalebrun:
I implemented what I mentioned: https://github.com/fayalalebrun/vulkano/tree/gl-interop-fix

It's compiling now, but it seems like the example is now segfaulting on my machine. Even previously working versions segfault, perhaps due to a GPU driver update. It would be good to know whether it works on your machine.

Also, we're no longer using this extension for compositing, as we've since moved to DMABuf via Wayland. I wonder if anybody uses this extension with Vulkano. If there no users perhaps it is not worth it to support it.

AustinJ235 (approved):

Rua:
@fayalalebrun this is merged now. Can you make a followup PR?

fayalalebrun:
@Rua With that branch I linked? I'm not sure if it is correct, since it seems like even the older versions of the example which once worked no longer work on my system. I'm using RADV + RadeonSI, but sadly I have no easy access to a system with another driver to test.

PR #2148, opened 2023-02-25
Fix subbuffer alignment to the non-coherent atom size limitation

marc0246:
This fixes the limitation that a subbuffer must be aligned to the non-coherent atom size when reading or writing its contents from the host and its backing memory is non-host-coherent, by allowing any offset/size in the subbuffer, but still locking a range that is aligned to the non-coherent atom size, such that cache flushes and invalidations will not alias other subbuffers within the same buffer. The downside is that it is pretty opaque to the user that this is happening, and the error messages will be mysterious. Alas, I can't think of a solution that's both intuitive and transparent, and doesn't suffer from this big limitation, that the user can't read/write any size of subbuffer when they underlying memory is not host-coherent.

### DIFF
--- a/vulkano/src/buffer/mod.rs
+++ b/vulkano/src/buffer/mod.rs
@@ -874,15 +874,6 @@ pub enum BufferError {
         queue_family_index: u32,
         queue_family_count: u32,
     },
-
-    /// The memory is not host-coherent, and the [`Subbuffer`] bounds are not a multiple of the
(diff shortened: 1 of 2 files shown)


Rua:
Since as you say, this behaviour is pretty opaque, I think this behaviour should at least be mentioned in the documentation of `read` and `write`. The user should be able to understand why the subbuffer isn't always exactly what they asked for.

PR #2162, opened 2023-03-25
Fix potential panics during panics

marc0246:
Changelog:
```markdown
### Bugs fixed
- Fixed potential panics during panics.
```


vulkano/src/command_buffer/traits.rs
```diff
@@ -441,6 +442,10 @@ where
     F: GpuFuture,
 {
     fn drop(&mut self) {
+        if thread::panicking() {
```
Rua, line 445: I think here, this check can go inside the other `if` below?

Rua:
Yes.

PR #2174, opened 2023-04-01
Refactor `AllocationCreateInfo`

marc0246:
This refactors `AllocationCreateInfo` such that it can be used together with `BufferCreateInfo` and, in the future, `ImageCreateInfo`, to construct buffers and images, removing `BufferAllocateInfo`. Doing this before users start using this API in 0.33.

What is still unresolved is whether ignoring `BufferCreateInfo::size` when in these constructor functions is the right way.

Changelog:
Add:
```markdown
### Breaking changes
- `AllocationCreateInfo::{requirements, allocation_type, dedicated_allocation}` fields were removed.
- Added the parameters `requirements`, `allocation_type`, `dedicated_allocation` to `MemoryAllocator::allocate[_unchecked]`.
```

### DIFF
--- a/examples/src/bin/basic-compute-shader.rs
+++ b/examples/src/bin/basic-compute-shader.rs
@@ -14,7 +14,7 @@
 // GPU", or *GPGPU*. This is what this example demonstrates.
 
 use vulkano::{
-    buffer::{Buffer, BufferAllocateInfo, BufferUsage},
+    buffer::{Buffer, BufferCreateInfo, BufferUsage},
     command_buffer::{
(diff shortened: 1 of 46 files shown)


Rua:
I see two other options for `size`:
- Require it to be always zero.
- Pass it on to the buffer creation when it's not zero, allowing the user to override the size if they want.

Rua:
Let's go with that then. I think requiring a fixed value, especially one that is otherwise disallowed, is better than just ignoring the value. It allows us to change our mind later.

Rua:
Yeah I think a full assert would be better.

PR #2180, opened 2023-04-10
Add `root_path_env` property to shaders macro

Firestar99:
In order to load a shader generated by the build script one would need to specify the path from the `Cargo.toml` into the target directory, but one may not even now where exactly the target directory is located. It may be `./target` on a single crate, `../target` or `../../target` in a workspace environment or even in a totally different folder if the crate is used as a library from some other cargo project. As `src`/`bytes` are required to be string symbols one cannot specify them to `concat!(env!("OUT_DIR"), "path.spv")`, so with this PR I introduce the `root_path_env` property to the shader! macro.

```rust
//! ## `root_path_env: "..."`
//!
//! Instead of searching relative to your `Cargo.toml`, search relative to some other folder
//! specified by this env variable. The intended use case is using `OUT_DIR` to be able to load
//! shaders generated by your build script. Defaults to `CARGO_MANIFEST_DIR` corresponding to the
//! folder of your `Cargo.toml`.
```

Changelog:
```markdown
### Additions
- add `root_path_env` property to allow loading shaders generated by a build script
````

### CODE REVIEW

**Comment by Firestar99** [comment]
I'm not exactly sure how to go about adding unit tests. I could of course add a build script that copies an glsl file into the build dir and then use it in a test, but adding that to the shaders crate directly would also cause everyone using that crate to execute that unnecessary build script. I could also add a new crate just for testing this but that also seems a little overkill.
One could add such a "test" if you were to add  rust-gpu examples to your repo, but that would also bind your repo to the relatively large rust-gpu stack whenever you'd compile an example.

**Comment by Firestar99** [comment]
I just applied all your suggestions, don't really have anything to add to them

**Comment by Firestar99** [comment]
Would you deem it valuable to add documentation that one can specify custom rustc env vars in the build script?

`build.rs`
```rust
let out_dir = env::var("OUT_DIR").unwrap();
let shader_out_dir = format!("{out_dir}/../../../../spirv-builder/spirv-unknown-spv1.3/release/deps/example_shader.spvs/");
println!("cargo:rustc-env=SHADER_OUT_DIR={shader_out_dir}");
```

shader include:
```
vulkano_shaders::shader! {
	ty: "vertex",
	root_path_env: "SHADER_OUT_DIR",
	bytes: "image_shader-image_vs.spv",
}
```

PR #2189, opened 2023-04-23
Mildly optimize `RangeMap`

marc0246:
Following are the results of [my benchmarks](https://github.com/marc0246/vulkano-cmdbuf-builder-benchmarks/blob/main/src/main.rs) before and after:
```
test tests::descriptor_set    ... bench:  96,200,265 ns/iter (+/- 3,663,596)
test tests::dispatch_indirect ... bench:  86,317,442 ns/iter (+/- 1,821,521)
```
```
test tests::descriptor_set    ... bench:  67,867,435 ns/iter (+/- 1,519,747)
test tests::dispatch_indirect ... bench:  64,132,526 ns/iter (+/- 875,303)
```
Nothing spectacular unfortunately. I removed the `Gaps` iterator because I didn't have the mental capacity to re-imlement it with the new ways the ranges are stored, but I doubt we'll ever need that one anyway.

### DIFF
--- a/vulkano/src/range_map.rs
+++ b/vulkano/src/range_map.rs
@@ -2,10 +2,9 @@
 
 use std::{
     cmp,
-    cmp::Ordering,
     collections::{btree_map, BTreeMap},
     fmt::{Debug, Error as FmtError, Formatter},
-    iter::{FromIterator, FusedIterator, Peekable},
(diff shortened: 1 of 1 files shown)


Rua:
Still a welcome improvement, thanks!

PR #2191, opened 2023-04-24
Fix UB in `ShaderModule::from_bytes[_with_data]`

marc0246:
Changelog:
```markdown
### Bugs fixed
- Fixed UB in `ShaderModule::from_bytes[_with_data]`, where an unaligned reference could have been created.
```


Rua:
To make the code a bit more self documented, you could use `align_of::<u32>()` instead of the magic number `4`.

Issue #2192, opened 2023-04-24
i.MX8: image_format_properties_unchecked returns None


Fighter19:
*   Version of vulkano: c6a6a7fd2335cabdb8a355810ccb07fd7b1f5e98
*   OS: Linux (tdx-xwayland-glibc-x86_64-Reference-Multimedia-Image-cortexa35-colibri-imx8x-6.2.0)
*   GPU (the selected PhysicalDevice): Vivante GC7000L (NXP i.MX8 GPU)
```
GPU0:
VkPhysicalDeviceProperties:
---------------------------
        apiVersion        = 4202496 (1.2.0)
        driverVersion     = 25182211 (0x1804003)
        vendorID          = 0x10002
        deviceID          = 0x70006214
        deviceType        = PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU
        deviceName        = VeriSilicon 
        pipelineCacheUUID = 14620070-7d09-47bb-b5d1-0ac5c5b13c85

```
*   GPU Driver: V6.4.3.p4.398061 (Vulkan Instance Version: 1.2.182)
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: [see triangle example](https://github.com/vulkano-rs/vulkano/blob/master/examples/src/bin/triangle.rs)

# Issue

The triangle example doesn't work as image_format_properties_unchecked returns None, because "vkGetPhysicalDeviceImageFormatProperties2KHR(): Returned error VK_ERROR_FORMAT_NOT_SUPPORTED."

I assume this is caused due to the change b968451, which appears that it would have has_seperate_stencil_usage evaluate to false, while it seems it would evaluate to true now (see physical.rs in https://github.com/vulkano-rs/vulkano/commit/b958451#diff-5f03afc2104ee2633b0acea54a2cc12d07e4a4a626e820fb3b3fc8ff93ef64d0L1006).
Manually setting has_seperate_stencil_usage to false using a debugger, will make the call succeed.

See log including validation layer output here: 
[error.log](https://github.com/vulkano-rs/vulkano/files/11310330/error.log)
(src/main.rs corresponds to examples/src/bin/triangle.rs)

PR #2193, opened 2023-04-24
Revert determination of seperate stencil usage

Fighter19:
Fix crash on Vivante GPU (#2192)

N/A

See issue #2192.
Commit b968451 changed the way has_separate_stencil_usage is being determined, leading to it being evaluated to true, even when it was formerly false.
This change reverts that aspect of the commit.

Changelog:
```markdown
### Bugs fixed
- Crash of application during start on i.MX board due to vkGetPhysicalDeviceImageFormatProperties2KHR returning VK_ERROR_FORMAT_NOT_SUPPORTED.
````

### CODE REVIEW

**Comment by Fighter19** [comment]
I can imagine that the solution is flawed, as I am not entirely aware of how the caching nor Vulkan works.
I am not aware for example, why specifying the COLOR_ATTACHMENT bit is supposedly allowed for the stencil,
and why copying all of the bits is even valid. (I personally would have assumed, that COLOR_ATTACHMENT needs to be cleared and DEPTH_STENCIL_ATTACHMENT set).
I suppose my change does revert the benefit of treating both cases as the same.

At the same time it fixes the issue with has_seperate_stencil_usage always evaluating to true (in case ```stencil_usage.is_empty() || !aspects.contains(ImageAspects::DEPTH | ImageAspects::STENCIL)```)  and therefore adding a ImageStencilUsageCreateInfo to p_next, which the GPU or driver rejects.

In order to provide a better fix (optimally one that preferably omits the extra struct),
I'd need to know, if the way ImageStencilUsageCreateInfo is initialized is actually valid (passing the color attachment bit).

**Comment by Rua** [comment]
Looking again, I think the bug is just line 1070 having the condition reversed? It looks like it should be `!=`.

**Comment by Fighter19** [comment]
I think so too. I will try again using that.

I also found out that apparently the GPU rejects COLOR_ATTACHMENT being set, setting DEPTH_STENCIL_ATTACHMENT instead made the call succeed.

I temporarily tried the following:

```rs
*stencil_usage = (*usage).difference(ImageUsage::COLOR_ATTACHMENT).union(ImageUsage::DEPTH_STENCIL_ATTACHMENT);
```

Maybe that case could be handled, as not all GPUs see an issue with getting COLOR_ATTACHMENT passed, apparently?

PR #2197, opened 2023-05-01
Replace `ImageAccess::descriptor_layouts()`

Rua:
My attempt to get rid of one of the last methods of `ImageAccess`, to help with the image unification. An unsuccessful attempt so far, because the method is still there and is still being used, but at least I managed to reduce it to one single function. The current changes add the image layout to the `WriteDescriptorSet` element definitions, but there is no way for the user to provide the layout yet. That is what `descriptor_layouts()` is still used for as of now. I would like to make this user-settable now, but I'm not sure what would be a convenient API for the user to do this.

Adding `image_view_with_layout(_array)` and `image_view_sampler_with_layout(_array)` constructors to `WriteDescriptorSet` can work, but in the case of the latter the user would be required to provide an iterator of 3-tuples (image view, layout, sampler), which isn't the most intuitive. To keep the API close to Vulkan, it would be nice to ask for `DescriptorImageViewSamplerInfo` structs instead, but there is no equivalent struct for the `image_view_sampler(_array)` constructor, which doesn't take layouts.

All in all, still many design uncertainties for this one, so I'll leave it as a draft.

### DIFF
--- a/vulkano/src/command_buffer/commands/bind_push.rs
+++ b/vulkano/src/command_buffer/commands/bind_push.rs
@@ -17,10 +17,10 @@ use crate::{
         AutoCommandBufferBuilder,
     },
     descriptor_set::{
-        check_descriptor_write, layout::DescriptorType, sys::UnsafeDescriptorSet,
-        DescriptorBindingResources, DescriptorSetResources, DescriptorSetUpdateError,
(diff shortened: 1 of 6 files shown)


Rua:
I solved it in the end by making image + sampler a tuple of `DescriptorImageViewInfo` and `Arc<Sampler>`.

PR #2204, opened 2023-05-11
Add `Surface::from_window[_ref]`, deprecate vulkano-win

marc0246:
This rids us of a dependency on winit, since raw-window-handle is all that's needed, once we murk vulkano-win for good. It's also easier for the user since they don't need to add another crate, and the way a surface is created now is much more natural.

Closes #2088

Changelog:
```markdown
### Additions
- Added `Surface::from_window[_ref]` and `Surface::required_extensions`. This is now the preferred way of creating a surface, vulkano-win has been deprecated.
````


voidburn:
I love the way you think! This completely solves the issue I was having with separating the concern of building the window (something I want the user to do on their own terms) from that of initializing the Vulkan's instance and surface (something I want my framework to handle).

Thanks!

PR #2212, opened 2023-05-19
Replace `PipelineMemoryAccess` with `PipelineStageAccess`

Rua:
When making the late new command buffer, I introduced the crate-private `PipelineStageAccess` type to provide the ability to more finely specify the access types of individual commands. It seemed to be still useful for the existing command buffer, and possibly for task graphs later too, so I improved on it and replaced the older `PipelineMemoryAccess` struct with it.

### DIFF
--- a/vulkano/src/command_buffer/commands/pipeline.rs
+++ b/vulkano/src/command_buffer/commands/pipeline.rs
@@ -303,7 +303,7 @@ where
 
         let mut used_resources = Vec::new();
         self.add_descriptor_sets_resources(&mut used_resources, pipeline);
-        self.add_vertex_buffers(&mut used_resources, pipeline);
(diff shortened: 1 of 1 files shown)


Rua:
So weird that I don't get any of these errors.

Rua:
Here's a dump of me running the indirect example, with the API dump and validation layers enabled.

[dump.txt](https://github.com/vulkano-rs/vulkano/files/11652105/dump.txt)

Rua:
> Here's my dump: [dump.txt](https://github.com/vulkano-rs/vulkano/files/11668372/dump.txt)

This dump only contains the API calls, not the validation errors that occur as they are called. I really need both as a single stream to be able to correlate them together.

PR #2216, opened 2023-05-31
Disallows single image swapchains by default

### LINKED ISSUE #1159: Windows: Fullscreen makes acquire_next_image timeout
*   Version of vulkano: 0.11
*   OS: Windows 10
*   GPU (the selected PhysicalDevice): AMD Radeon HD 7800 Series
*   GPU Driver: 17.50.17.03-180131a-323831C-RadeonSoftwareAdrenalin
*   Vulkan Driver Version: 2.0.1
*   Vulkan API Version: 1.0.65
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: https://pastebin.com/3WqWux5P (same as triangle example, 4 lines modified)

This can be triggered by changing
```
let surface = WindowBuilder::new().build_vk_surface(&events_loop, instance.clone()).unwrap();
```
to
```
let surface = WindowBuilder::new().with_fullscreen(Some(events_loop.get_primary_monitor())).build_vk_surface(&events_loop, instance.clone()).unwrap();
```
in the triangle example.

By debugging I found out that after changing to fullscreen 1 or 2 frames may be rendered, but after that the function `vulkano::swapchain::acquire_next_image` will block the thread forever, freezing the rendering.

After that, if a timeout is not set, the window won't be responsible (the event loop is blocked as it's in the same thread) and you won't be able to close it or switch to another window unless you use CTRL+ALT+DEL and open the task manager - this will remove the focus from the window. Using CTRL+SHIFT+ESC to open the task manager won't work.

Not sure if I'm doing anything wrong or if this is a bug with Vulkano or Winit.

aedm:
Forcing swapchains to have at least two images so double buffering can be performed. Current Windows AMD drivers report `VkSurfaceCapabilitiesKHR.minImageCount` to be 1. Using that value to create a swapchain (like some of the Vulkano examples and vulkano-win itself) causes those apps to crash in fullscreen.

Closes #1159 

Changelog:
```markdown
### Bugs fixed
- Vulkano examples crash on AMD GPUs in fullscreen on Windows.
````

### DIFF
--- a/vulkano/src/swapchain/swapchain.rs
+++ b/vulkano/src/swapchain/swapchain.rs
@@ -131,10 +131,9 @@ impl Swapchain {
             clipped,
             full_screen_exclusive,
             win32_monitor,
-            allow_single_image,
+            allow_single_image: _,
             _ne: _,
         } = create_info;
-        let min_image_count = if allow_single_image { min_image_count } else { min_image_count.max(2) };
 
         let swapchain = Arc::new(Swapchain {
             handle,
@@ -232,10 +231,9 @@ impl Swapchain {
             clipped,
             full_screen_exclusive,
             win32_monitor,
-            allow_single_image,
+            allow_single_image: _,
             _ne: _,
         } = create_info;
-        let min_image_count = if allow_single_image { min_image_count } else { min_image_count.max(2) };
 
         let swapchain = Arc::new(Swapchain {
             handle,
@@ -284,7 +282,7 @@ impl Swapchain {
         create_info: &mut SwapchainCreateInfo,
     ) -> Result<(), SwapchainCreationError> {
         let &mut SwapchainCreateInfo {
-            min_image_count,
+            ref mut min_image_count,
             ref mut image_format,
             image_color_space,
             ref mut image_extent,
@@ -300,7 +298,6 @@ impl Swapchain {
             allow_single_image,
             _ne: _,
         } = create_info;
-        let min_image_count = if allow_single_image { min_image_count } else { min_image_count.max(2) };
 
         if !device.enabled_extensions().khr_swapchain {
             return Err(SwapchainCreationError::RequirementNotMet {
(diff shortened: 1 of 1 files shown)


aedm:
You're right, single buffering is super niche, I removed the flag.

> Is there actually a valid reason to set `allow_single_image` to true? Are there swapchains that really work properly with just one image?

To my surprise, single buffer swapchains do work in windowed mode. The desktop compositor is apparently able to work with them, I guess at the cost of some performance. They use less memory, but it really sounds unlikely that someone who picks Vulkan actually wants single buffering.

aedm:
> IMO what we should do is fix the examples such that they show the correct way of choosing the image count: pick the desired count e.g. 2 and clamp it between the min and max image counts from the surface capabilities. If desired, the user can also panic if the max image count is less than 2, which makes this more explicit and easier to teach, as it's right there in the example with comments to go along with it.

Fixing the examples a good solution, too, and it would likely prevent such errors. That was also my first approach. :)

My personal experience: I wrote an app using an NVidia GPU, and it crashed on my friends' computers who had AMD. I couldn't tell what the issue was and I ended up buying an AMD GPU just to fix it (and make this PR).

Even though I started from the Vulkano examples, I made a ton of tinkering, and I'm not certain `min_image_count.max(2)` would have survived that tinkering if the code runs perferctly on my machine without it. I consider `VkSurfaceCapabilitiesKHR.minImageCount = 1` to be a driver bug on AMD's part, and `.max(2)` forces Vulkano users to take a driver bug into account. It's not beginner friendly, and that's why lean towards this solution.

aedm:
@marc0246 I made PR #2218 that fixes the examples instead as you suggested. 
I prefer this one, but feel free to approve whichever you prefer and close the other one.

Rua:
I agree, I think this falls on the user rather than on Vulkano.

aedm:
@marc0246 @Rua Thank you both for your feedback, makes sense to me.

Closing this in favor of #2218

Issue #2208, opened 2023-05-15
Interactive Fractal example completely freezes when going full-screen on MacOS


voidburn:
* Version of vulkano: 0.33.0
* OS: Ventura 13.3.1 (a) (22E772610a)
* GPU: Apple M2 Pro (http://vulkan.gpuinfo.org/displayreport.php?id=20578#device)
* Vulkan SDK Version: 1.3.243.0 (29-Mar-2023)

# Issue

Going full-screen on while running the "Interactive Fractal" example freezes the screen in which the application is open (apparently indefinitely). Similar behavior is shown with the "Multi Window Game of Life" (with either window), but that one recovers after a while rendering at a lower resolution than the native one for the monitor.

No error is reported in the terminal where I launched the example, either in debug or release mode. The only solution is to kill the application (luckily I have a second monitor that remains working).

This is what it looks like once the application is frozen (the black parts are the rest of the monitor area, screenshot taken with the system utility):
![image](https://github.com/vulkano-rs/vulkano/assets/1438321/8fc06499-2f50-41d1-8a9e-068c099247be)

rauba-code:
I'm a game dev who is using Vulkano, and was experiencing the same issue. It happens because Vulkano examples don't follow the Winit recommendations of rendering (it re-renders every time after every single event and starves resources).

I tested this way of rendering in all Linux X, Wayland and MacOS:
```rust
*control_flow = ControlFlow::Poll; // Setting this to 'Wait' skips frames on X.
// ...
WindowEvent::MainEventsCleared => window.request_redraw(),
WindowEvent::RedrawRequested => {
// Rendering begins here
},
// ...
```

One may want to add more request_redraw() after the

rauba-code:
@marc0246 Checked out master, on debug this issue persists. Checking on release...

rauba-code:
On release, it still exists.

rauba-code:
Yes, these two permanently freeze.

The others freeze only for a second or so when going fullscreen.  Freezing and occasional flassing for a second also appears when hovering over the top of a menu bar. They do not lag at all when resizing a window.

rauba-code:
I fixed the issue for `interactive_fractal`. Check [this commit](https://github.com/rauba-code/vulkano/commit/d52d6edb031a94fded2d154b461db320cb1a7f6f).

When I finish with `multi_window_game_of_life`, I'd like to make a PR.

voidburn:
[...]

PR #2221, opened 2023-06-03
Fixed some examples freezing on fullscreen mode on MacOS for #2208

rauba-code:
This pull request solves the problem of freezing when going fullscreen for MacOS.
It was caused due to Winit behaving unexpectedly when using event loops that are managed by `run_return()`. Winit crate generally discourages the use of `run_return()` and recommends using `run()` instead.
(see: <https://docs.rs/winit/latest/winit/platform/run_return/trait.EventLoopExtRunReturn.html#tymethod.run_return>).

Changelog:
```markdown
### Bugs fixed
- Fixed the problem on MacOS of window completely freezing when setting the window to fullscreen mode for examples `interactive_fractal` and `multi_window_game_of_life`.
````

### DIFF
--- a/examples/src/bin/interactive_fractal/main.rs
+++ b/examples/src/bin/interactive_fractal/main.rs
@@ -75,7 +75,7 @@ fn main() {
     app.print_guide();
 
     event_loop.run(move |event, _, control_flow| {
-        *control_flow = ControlFlow::Wait;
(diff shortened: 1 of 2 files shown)


rauba-code:
@marc0246 Sure, I forgot it's the default.

PR #2222, opened 2023-06-04
Provide a default impl of CommandBufferResourcesUsage

michaelwu:
This makes it possible for `PrimaryCommandBufferAbstract` to be (minimally) implemented outside of vulkano.

I have a C library that generates command buffers and I want to execute them with futures, which require `PrimaryCommandBufferAbstract` to be implemented. Let me know if there's a better way of doing this.


Rua:
This is really a private type, that is only made public because of the command buffer traits. In the future we plan to get rid of these traits (and the type parameters of command buffers), which would make this change unnecessary.

michaelwu:
> This is really a private type, that is only made public because of the command buffer traits. In the future we plan to get rid of these traits (and the type parameters of command buffers), which would make this change unnecessary.

Sounds great, looking forward to it!

PR #2223, opened 2023-06-04
Reorganise render pass objects, add support for some extensions

Rua:
Changelog:
```markdown
### Breaking changes
Changes to render pass objects:
- Resolve attachments are now provided as `ResolvableAttachmentReference` values, as part of the `color_attachments`, `depth_attachment` and `stencil_attachment` values. In the `single_pass_renderpass` and `ordered_passes_renderpass` macros, the resolve attachment is indicated by a single arrow, and the resolve mode by a following colon, for example `[color -> color_resolve:Average]`.
- Depth and stencil attachments are now provided separately. They must still use the same attachment, but their layouts and resolve modes can differ.
- Input attachments are now provided as `InputAttachmentReference` values.
- If an attachment has a format with a stencil component, you must now provide `stencil_initial_layout` and `stencil_final_layout` to `AttachmentDescription`, in the same way as `stencil_load_op` and `stencil_store_op`.
- Renamed `LoadOp` and `StoreOp` to `AttachmentLoadOp` and `AttachmentStoreOp` to match the Vulkan names.
- In the `single_pass_renderpass` and `ordered_passes_renderpass` macros, the `load` and `store fields for attachments are renamed to `load_op` and `store_op`, and ordered below the `format` and `samples` fields, to match the fields of the `AttachmentDescription` structure.

### Additions
- Support for the `khr_depth_stencil_resolve` extension.
- Support for the `khr_separate_depth_stencil_layouts` extension.
- Added `flags` to `RenderPassCreateInfo`, `AttachmentDescription` and `SubpassDescription`.
````

### DIFF
--- a/examples/src/bin/async-update.rs
+++ b/examples/src/bin/async-update.rs
@@ -412,10 +412,10 @@ fn main() {
         device.clone(),
         attachments: {
             color: {
-                load: Clear,
-                store: Store,
                 format: swapchain.image_format(),
                 samples: 1,
+                load_op: Clear,
+                store_op: Store,
             },
         },
         pass: {
--- a/examples/src/bin/async-update.rs
+++ b/examples/src/bin/async-update.rs
@@ -420,7 +420,8 @@ fn main() {
         },
         pass: {
             color: [color],
-            depth_stencil: {},
+            depth: {},
+            stencil: {},
         },
     )
     .unwrap();
--- a/examples/src/bin/buffer-allocator.rs
+++ b/examples/src/bin/buffer-allocator.rs
@@ -211,10 +211,10 @@ fn main() {
         device.clone(),
         attachments: {
             color: {
-                load: Clear,
-                store: Store,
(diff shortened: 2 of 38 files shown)


Rua:
All the examples work for me, and I don't get any validation errors either. Do you get any?

Rua:
That should be fixed now. Still no idea what's going on with the indirect example though.

Rua:
No, it's there, create.rs lines 122, 273 and 375.

Rua:
Vulkano treats depth and stencil as entirely separate attachments, but they have to have the same attachment number. Vulkan treats them as a single combined attachment type from the get go, but lets you specify a separate layout for the stencil. The translation between them happens in the code after line 245.

Rua:
That's why I only did it because there's a model in dynamic rendering. Dynamic rendering already works exactly this way, and I suspect that if Khronos were to redesign the API they'd do it that way for render passes too.

Rua:
This is done now.

PR #2230, opened 2023-06-16
Add fallback library names

CodesOtakuYT:
### Additions
- Add fallback library names

This change doesn't break compatibility, it's only an implementation detail of trying to load the library with alternative names in case of failure instead of giving up.

The only detail that needs attention, is the fact that am returning the error of the last attempt of loading the library instead of all the errors which would require some breaking changes and more design decisions but provide the user with more information about all the attempts done.


CodesOtakuYT:
I got the library names from https://github.com/zeux/volk/blob/master/volk.c
And yes I did run cargo fmt

PR #2232, opened 2023-06-17
Prevent panic on overflow in `validate_push_constants`

Rua:
Fixes a small problem left by #2220.


Rua:
Since that's an unsafe function, I think it's ok for it to just panic if the user provides something too large.

PR #2233, opened 2023-06-18
Add `RequiresAllOf`, automatically enable required extensions and features

Rua:
Changelog:
```markdown
### Additions
- When creating an instance or device, you only need to specify the extensions and features you actually care about. Any extensions and features that are required by the extensions that you specified are now automatically enabled too.
````

This adds the ability to specify multiple requirements that must be enabled together, using `RequiresAllOf`. This is needed for some parts of the API, such as [`vkGetPhysicalDevicePresentRectanglesKHR`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDevicePresentRectanglesKHR.html). `RequiresOneOf` now contains a slice of `RequiresAllOf`, which in turns contains a slice of `Requires` enums. In other words, requirements now follow a [DNF](https://en.wikipedia.org/wiki/Disjunctive_normal_form) structure.

Additionally, I added the automatic enabling of required extensions and features. This is purely a convenience for the user, but hopefully one that makes dealing with extensions less frustrating, by letting the user focus on what they actually need.

### DIFF
--- a/vulkano/autogen/extensions.rs
+++ b/vulkano/autogen/extensions.rs
@@ -11,7 +11,7 @@ use super::{write_file, IndexMap, VkRegistryData};
 use heck::ToSnakeCase;
 use proc_macro2::{Ident, Literal, TokenStream};
 use quote::{format_ident, quote};
-use std::fmt::Write as _;
+use std::{cmp::Ordering, fmt::Write as _};
 use vk_parse::Extension;
 
 // This is not included in vk.xml, so it's added here manually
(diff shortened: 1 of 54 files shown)

### CODE REVIEW

**Comment by Rua** [comment]
I was confused that `Debug` seemed to print the same thing as `Display`, when the norm is for `Debug` to show the internal structure of the object. What do you think the output should say?

**Comment by Rua** [comment]
But how should it be formatted, now that there's multiple layers of requirements?

**Comment by Rua** [comment]
Actually looking at the code, I'm even more confused. There is a custom `Debug` implementation for `ValidationError` already.

**Comment by Rua** [comment]
Yes please.

**Comment by Rua** [comment]
Do you get a compile error? It looks normal on my end:
```rust
pub(super) fn check_requirements(
        &self,
        supported: &DeviceExtensions,
        api_version: crate::Version,
        instance_extensions: &crate::instance::InstanceExtensions,
    ) -> Result<(), crate::ValidationError> {
        if self.khr_16bit_storage {
            if !supported.khr_16bit_storage {
                return Err (crate :: ValidationError { problem : "contains `khr_16bit_storage`, but this extension is not supported by the physical device" . into () , .. Default :: default () }) ;
            }
            if !(api_version >= crate::Version::V1_1
                || instance_extensions.khr_get_physical_device_properties2)
            {
                return Err(crate::ValidationError {
                    problem: "contains `khr_16bit_storage`".into(),
                    requires_one_of: crate::RequiresOneOf(&[
                        crate::RequiresAllOf(&[crate::Requires::APIVersion(crate::Version::V1_1)]),
                        crate::RequiresAllOf(&[crate::Requires::InstanceExtension(
                            "khr_get_physical_device_properties2",
                        )]),
                    ]),
                    ..Default::default()
                });
            }
            if !(api_version >= crate::Version::V1_1) {
                return Err(crate::ValidationError {
                    problem: "contains `khr_16bit_storage`".into(),
                    requires_one_of: crate::RequiresOneOf(&[crate::RequiresAllOf(&[
                        crate::Requires::APIVersion(crate::Version::V1_1),
                    ])]),
                    ..Default::default()
                });
            }
        }
```
etc.

Rua:
Yes, go ahead.

Are you talking about the device extensions required by other device extensions? Those checks are gone because they're automatically enabled now.

Rua:
Ah never mind, I think I see what's going on.

Rua:
I made a fix. Whenever a device extension was one of the alternatives, since we're always able to enable that extension, we can treat that alternative as `true`, making the whole test true. So now, the check is only emitted when there is no device extension among the alternatives.

Rua:
Is there anything else I need to do?

Rua:
Yeah, it's fine.

PR #2235, opened 2023-06-24
Added required_subgroup_size to PipelineShaderStageCreateInfo

charles-r-earp:
Compute and Graphics pipelines can be created with a required subgroup size. 

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_subgroup_size_control.html

# To do
- Validation / validation errors. 
- Tests

# Limitations
For compute, can't validate max_compute_workgroup_subgroups without knowing the workgroup size, which is not currently reflected. 

This interacts with additional VK_EXT_subgroup_size_control flags for requiring full subgroups and allowing varying subgroups, but since those aren't currently implemented they were neglected. If implemented, they add additional constraints to validate.


Rua:
#2234 will make some changes to how `PipelineShaderStageCreateInfo` is validated, so it may be better to wait until that is merged.

vulkano/src/pipeline/compute.rs
```diff
@@ -226,7 +240,22 @@ impl ComputePipeline {
                 data_size: specialization_data_vk.len(),
                 p_data: specialization_data_vk.as_ptr() as *const _,
             };
+            required_subgroup_size_create_info =
+                required_subgroup_size
+                    .clone()
```
Rua, line 245: This clone is unnecessary, since you're not moving anything out of it.

vulkano/src/pipeline/compute.rs
```diff
@@ -191,12 +203,14 @@ impl ComputePipeline {
         let specialization_info_vk;
         let specialization_map_entries_vk: Vec<_>;
         let mut specialization_data_vk: Vec<u8>;
+        let required_subgroup_size_create_info;
```
Rua, line 206: All the variables holding data to be passed to Vulkan end in `_vk`. Could you include that as well please?

vulkano/src/pipeline/graphics/mod.rs
```diff
@@ -2678,6 +2691,8 @@ impl GraphicsPipeline {
             specialization_info_vk: ash::vk::SpecializationInfo,
             specialization_map_entries_vk: Vec<ash::vk::SpecializationMapEntry>,
             specialization_data_vk: Vec<u8>,
+            required_subgroup_size_create_info:
```
Rua, line 2694: Same here, please add `_vk`.

vulkano/src/shader/mod.rs
```diff
@@ -750,6 +750,19 @@ pub struct PipelineShaderStageCreateInfo {
     /// The default value is empty.
     pub specialization_info: HashMap<u32, SpecializationConstant>,
 
+    /// The required subgroup size.
```
Rua, line 753: A few points about the documentation style here:
- We prefer plain text rather than bullet points.
- We don't normally link to the Vulkan documentation, if we can document it ourselves.
- For requirements, the usual style is like this: https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/render_pass/mod.rs#L2282C7-L2284

vulkano/src/shader/mod.rs
```diff
@@ -761,11 +774,55 @@ impl PipelineShaderStageCreateInfo {
...
 }
 
+pub(crate) fn validate_required_subgroup_size(
+    device: &Device,
+    stages: ShaderStages,
+    workgroup_size: Option<u32>,
+    subgroup_size: u32,
+) -> Result<(), crate::ValidationError> {
+    if device.api_version() < crate::Version::V1_3
+        && !device.enabled_extensions().ext_subgroup_size_control
```
Rua, line 790: For clarity, we normally write this as:
```rust
!(device.api_version() >= crate::Version::V1_3 || device.enabled_extensions().ext_subgroup_size_control)
```

According to https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT.html, you can also enable `ext_shader_object` to enable this functionality.

Rua:
#2234 has been merged now.

Rua:
@charles-r-earp are you still interested in getting this merged?

charles-r-earp:
> @charles-r-earp are you still interested in getting this merged?

Yes

vulkano/src/shader/mod.rs
charles-r-earp: > According to https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT.html, you can also enable `ext_shader_object` to enable this functionality.

Based on https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkShaderCreateInfoEXT.html, VkPipelineShaderStageRequiredSubgroupSizeCreateInfo can be added to pnext for VkShaderCreateInfoEXT, but according to https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineShaderStageCreateInfo.html it requires the subgroupSizeControl feature to use with VkPipelineShaderStageCreateInfo.

Rua:
A few points:
- You forgot to include the `format!` macro when creating most of the error messages. That means that for example `{subgroup_size}` will appear literally in the message, not the value of the variable.
- Please put backticks ` around the variable names in the `problem` field (but not the `context` field).
- In a few places you wrote an error message instead of a VUID.
- VUID-VkPipelineShaderStageCreateInfo-flags-02759 still needs to be validated.
- Can you fix the review points I mentioned earlier?

vulkano/src/pipeline/mod.rs
```diff
@@ -301,7 +304,7 @@ vulkan_bitflags! {
 /// Specifies a single shader stage when creating a pipeline.
 #[derive(Clone, Debug)]
 pub struct PipelineShaderStageCreateInfo {
-    /// Additional properties of the shader stage.
+    /// Specifies how to create the shader stage.
```
Rua, line 307: Can you undo this change?

vulkano/src/pipeline/mod.rs
```diff
@@ -473,12 +493,99 @@ impl PipelineShaderStageCreateInfo {
...
+        .contains(stages)
+    {
+        return Err(Box::new(ValidationError {
+            context: "required_subgroup_size".into(),
+            problem: "shader stage {stage:?} is not in required_subgroup_size_stages".into(),
+            vuids: &["VUID-VkPipelineShaderStageCreateInfo-pNext-02755"],
+            ..Default::default()
+        }));
+    }
+    if subgroup_size.is_power_of_two() {
```
Rua, line 543: This is missing a ! operator.

charles-r-earp:
> * VUID-VkPipelineShaderStageCreateInfo-pNext-02756 still needs to be validated.

This is validated:

```  
if let Some(workgroup_size) = workgroup_size {
    let max_compute_workgroup_subgroups = properties
        .max_compute_workgroup_subgroups
        .unwrap_or_default();
    if max_compute_workgroup_subgroups * subgroup_size < workgroup_size {
        return Err(Box::new(ValidationError {
            context: "required_subgroup_size".into(),
            problem: format!("`subgroup_size` {subgroup_size} creates more than {max_compute_workgroup_subgroups} subgroups").into(),
            vuids: &["VUID-VkPipelineShaderStageCreateInfo-pNext-02756"],
            ..Default::default()
        }));
    }
}
```

However, workgroup_size is not available because it's not reflected from the shader, so it's currently None.

Rua:
Can you add the necessary reflection then?

vulkano/src/pipeline/mod.rs
```diff
@@ -475,10 +495,100 @@ impl PipelineShaderStageCreateInfo {
...
+            vuids: &["VUID-VkPipelineShaderStageCreateInfo-pNext-02755"],
+            ..Default::default()
+        }));
+    }
+    let stages: ShaderStages = [stage].into_iter().collect();
+    let properties = device.physical_device().properties();
+    if !properties
+        .required_subgroup_size_stages
+        .unwrap_or_default()
+        .contains(stages)
```
Rua, line 535: You can use `contains_enum` here and then you don't need to do the conversion above.

vulkano/src/shader/mod.rs
```diff
@@ -557,6 +557,41 @@ pub struct EntryPointInfo {
...
 }
 
+impl EntryPointInfo {
+    // The local size in Compute shaders, None otherwise.
+    // `specialization_info` is used for LocalSizeId / WorkgroupSizeId, using specialization_constants if not found.
+    // Errors if specialization constants
+    pub(crate) fn local_size(
+        &self,
+        specialization_info: &HashMap<u32, SpecializationConstant>,
+    ) -> Result<Option<[u32; 3]>, ValidationError> {
```
Rua, line 567: This function doesn't seem to ever return an error, so the `Result` wrapper may not be needed?

vulkano/src/pipeline/mod.rs
```diff
@@ -475,10 +495,100 @@ impl PipelineShaderStageCreateInfo {
...
+        return Err(Box::new(ValidationError {
+            context: "required_subgroup_size".into(),
+            problem:
+                format!("`subgroup_size` {subgroup_size} is greater than `max_subgroup_size` {max_subgroup_size}")
+                    .into(),
+            vuids: &["VUID-VkPipelineShaderStageRequiredSubgroupSizeCreateInfo-requiredSubgroupSize-02762"],
+            ..Default::default()
+        }));
+    }
+    if let Some(workgroup_size) = workgroup_size {
```
Rua, line 576: According to the VUID, this check is only needed for compute, mesh or task shaders. While we don't support mesh or task shaders yet, the code here could be made future proof by only checking in those cases.

charles-r-earp: workgroup_size is only Some for Compute, and Task and Mesh if / when the reflection is added. So I didn't check the stage because it's implied as WorkgroupSize is not applicable otherwise.

vulkano/src/shader/mod.rs
charles-r-earp: The error is due to invalid spec constants, either missing or not u32. I haven't implemented the error handling yet.

Rua:
It looks good so far, but I noticed that the required_subgroup_size test is causing an overflow when I run it:

---- pipeline::compute::tests::required_subgroup_size stdout ----
thread 'pipeline::compute::tests::required_subgroup_size' panicked at 'attempt to multiply with overflow', vulkano/src/pipeline/mod.rs:582:12

It looks like `max_compute_workgroup_subgroups * subgroup_size` can be too big to fit into an `u32`. Perhaps the expression can be rearranged so that an overflow can't happen.

charles-r-earp:
> It looks like `max_compute_workgroup_subgroups * subgroup_size` can be too big to fit into an `u32`. Perhaps the expression can be rearranged so that an overflow can't happen.

While `subgroup_size` is up to 128, `max_compute_subgroups` can be u32::MAX so I decided to use `checked_mul` for this. But I also realized that this was a problem for `local_size` too because while the product of `[x, y, z]` must be a u32, each dimension can be u32::MAX as well, so just casting to u64 seems simpler.

Rua:
> But I also realized that this was a problem for `local_size` too because while the product of `[x, y, z]` must be a u32, each dimension can be u32::MAX as well, so just casting to u64 seems simpler.

`u32::MAX * u32::MAX * u32::MAX` is still bigger than `u64::MAX` though!

charles-r-earp:
> `u32::MAX * u32::MAX * u32::MAX` is still bigger than `u64::MAX` though!

Nice catch!

vulkano/src/pipeline/mod.rs
```diff
@@ -512,15 +512,24 @@ impl PipelineShaderStageCreateInfo {
...
+                    let max_invocations = properties.max_compute_work_group_invocations;
+                    if workgroup_size > max_invocations {
+                        return Err(Box::new(ValidationError {
+                            problem: format!("the product of `workgroup size` {local_size:?} = {workgroup_size} is greater than `max_compute_work_group_invocations` {max_invocations}").into(),
+                            ..Default::default()
+                        }));
+                    }
+                    Some(workgroup_size)
+                } else {
                     return Err(Box::new(ValidationError {
```
Rua, line 525: I think, if the multiplication overflows, then `workgroup_size > max_invocations` must be true? You could do something like this:

```rust
if x.saturating_mul(y).saturating_mul(z) > max_invocations {
    // ...
}
```

vulkano/src/pipeline/mod.rs
```diff
@@ -512,15 +512,24 @@ impl PipelineShaderStageCreateInfo {
...
+                        problem: format!(
+                            "the product of `workgroup size` {local_size:?} overflowed"
+                        )
+                        .into(),
                         ..Default::default()
                     }));
                 }
-                Some(workgroup_size as u32)
             } else {
                 todo!()
```
Rua, line 534: What goes here in the place of the `todo!()`?

charles-r-earp: This isn't currently reachable because `EntryPointInfo::local_size()` only returns Some for Compute. So instead of the `if` it might be a `match` on the stage when / if additional stages are implemented.

vulkano/src/pipeline/mod.rs
charles-r-earp: As long as `max_invocations` isn't u32::MAX as well, because then you could saturate and have `workgroup_size == max_invocations`. Strangely according to <http://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxComputeWorkGroupInvocations&platform=all>, this only goes up to 67108896, but x/y/z can be u32::MAX. But that could change so I think the current impl is best.

vulkano/src/pipeline/mod.rs
Rua: Can you put that as a message inside `todo!()` then so that future contributors know what goes there?

Rua: Ah never mind you already did that.

vulkano/src/pipeline/mod.rs
Rua: Fair enough. But the error message in the `else` branch is not very helpful to the user. It could probably be the same message as above.

Rua:
Looks good now, thank you!

PR #2236, opened 2023-06-24
ValidationError-ify `Surface` and `Swapchain`

Rua:
Changelog:
```markdown
### Breaking changes
Changes to `Swapchain`:
- Swapchain creation no longer returns an error when the swapchain extent doesn't match the current surface extent. This requirement is ill-defined in the spec, as detailed here: [TOCTOU race condition on minImageExtent/maxImageExtent?](https://github.com/KhronosGroup/Vulkan-Docs/issues/1144).

### Additions
- Support for the `ext_surface_maintenance1` and most of the `ext_swapchain_maintenance1` extension.
````

Depends on #2233. I also reorganised the swapchain code a bit, by placing the main `Swapchain` type and related things in `swapchain/mod.rs`, and splitting off the acquire/present stuff into `swapchain/acquire_present.rs`.

Items of `ext_swapchain_maintenance1` that are still missing:
```markdown
- `vkReleaseSwapchainImagesEXT` / `VkReleaseSwapchainImagesInfoEXT`
- `VkSwapchainPresentFenceInfoEXT`
```
Both of these rely on proper (queue) sync logic, which we don't have yet.


Rua:
I don't know what the right way is anymore @.@

PR #2239, opened 2023-06-27
Improve debug formatting

marc0246:
A few small improvements to debug formatting. The most notable change is that `PhysicalDevice`s are no longer debug-formatted when debug-formatting a `Device`. That led to pages and pages of spam each time you debug-formatted a `Device` (and by extension each device-owned resource), and since #2227 the `Device` can even refer to multiple `PhysicalDevice`s. That is now fixed because only the `PhysicalDevice` ID is printed in `Device::debug`. I also removed some uninteresting fields from the debug output of `Device` and `PhysicalDevice`.

Changelog:
```markdown
### Additions
- Added the trait `InstanceOwned`.
- Debug formatting of Vulkan objects no longer follows references to other objects, instead only printing their handles, in order to alleviate the spam.
```


Rua:
Ooh, smart solution!

Rua:
Two object handles belonging to different devices can be identical while referring to different objects. Likewise, two devices belonging to different instances. Object handles really form a chain: [instance].[device].[object].

Rua:
For Vulkano-internal debugging, ids seem useful. For the user not so much. But we can't really differentiate, so they should probably be included.

Rua:
Yes. And for device-owned objects, also the device handle.

Rua:
I'll wait with merging this until after that then, so you can finish it.

Issue #2242, opened 2023-06-29
Condier providing a Device::set_debug_utils_object_name shortcut on individual objects


HadrienG2:
This could be done by adding a set_debug_utils_name method to the VulkanObject of DeviceOwned trait, with an appropriate where bound so that it is only available where the other trait is present.

The rationale for this proposal is that DeviceOwned objects, by definition, know which device they belong to. And from this perspective, needing to get my hand to a `&Device` just to set their debug utils name feels unnecessary ceremony.

Further, use of Vulkan requires creating a lots of objects, which in an ideal world would all be annotated with names in debugging/profiling mode so that tools provide more explicit output. Therefore, the ergonomics benefits of such a shortcut would quickly add up.

Rua:
Good idea! Do you think we still need the method on `Device` after this is implemented?

HadrienG2:
If we remove the method on Device, then the functionality will only be available via an extension trait, which may make it harder to discover by users.

Also, while playing with this in parallel, I realized that ergonomics of debug utils names could be further improved by making it possible to call set_debug_utils_name directly on a Vulkano object that maps to some underlying Vulkan object, like StorageImage or SubBuffer.

This could be an argument for having some sort of `BorrowVulkanObject` trait, that debug utils name setup would then accept as a bound instead of VulkanObject.

HadrienG2:
...also, I think I found a little use after free problem in the implementation :(

```rust
    /// Assigns a human-readable name to `object` for debugging purposes.
    ///
    /// If `object_name` is `None`, a previously set object name is removed.
    ///
    /// # Panics
    /// - If `object` is not owned by this device.
    pub fn set_debug_utils_object_name<T: VulkanObject + DeviceOwned>(
        &self,
        object: &T,
        object_name: Option<&str>,
    ) -> Result<(), OomError> {
        assert!(object.device().handle() == self.handle());

        // Building a CString, so far, so good...
        let object_name_vk = object_name.map(|object_name| CString::new(object_name).unwrap());
        let info = ash::vk::DebugUtilsObjectNameInfoEXT {
            object_type: T::Handle::TYPE,
            object_handle: object.handle().as_raw(),
            // ...but the CString will be dropped once this statement is done evaluating...
            p_object_name: object_name_vk.map_or(ptr::null(), |object_name| object_name.as_ptr()),
            ..Default::default()
        };

        // ...and thus the pointer will be dangling by the time we call Vulkan.
        unsafe {
[...]

PR #2243, opened 2023-06-29
Fix use-after-free in debug utils names setup

HadrienG2:
This fixes the use-after-free but that I discovered while playing around with debug utils names (first reported in #2242).

1. [x] Update documentation to reflect any user-facing changes - in this repository. => No docs change

2. [x] Make sure that the changes are covered by unit-tests. => UB cannot be covered by unit tests I think?

   Github Issues, and highlight important implementation aspects. => See above.

Changelog:
```markdown
### Bugs fixed
- `Device::set_debug_utils_object_name` no longer exhibits use-after-free UB.
````


Rua:
Thanks for spotting and fixing it!

PR #2247, opened 2023-07-01
Unify all image types

### LINKED ISSUE #108: Add StagingImage
Add a `StagingImage` wrapper around images with linear layouts.
This allows users to directly write the pixel data in the image, then copy it to another image.

### LINKED ISSUE #715: implement MutableImage
I am currently using ImmutableImage as if it were mutable. (Updating a text cache every frame)
It appears vulkano now has a check for this I am getting:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AccessError(ExclusiveDenied)', 
```
On this line:
```rust
        let future = previous_frame_end.join(acquire_future)
            .then_execute(queue.clone(), command_buffer).unwrap()
            .then_swapchain_present(queue.clone(), swapchain.clone(), image_num)
            .then_signal_fence_and_flush().unwrap();
```

I guess this means vulkano needs a MutableImage.

### LINKED ISSUE #1154: MipMapping not supported outside of ImmutableImage
*   Version of vulkano: 0.11.1
*   OS: any
# Issue
It's currently not possible to use mipmapping for anything else than immutable images.
As I wanted to implement a form of image based lighting, I need to be able to render to cubemaps and then blur the contents of that cubemap into it's mipmaps in order to simulate varying degrees of roughness.

Are there any plans on supporting mipmaps or is help wanted?

### LINKED ISSUE #1446: [Feature Request] ImageUsage parameter on ImmutableImage::from_iter
The `ImmutableImage::from_iter` internaly calls `from_buffer` which then creates the image with usage flags `transfer_destination` and `sampled`. However, sometimes one might want to specify more usage flags, for example `transfer_source` to use the image as a source for blit operations.

My suggestion is the following: There should be a `usage` parameter to both methods and it's fine to give `ImageUsage::none()` there. Internally the `from_buffer` method creates the correct usage with the following code:
```rust
let usage = ImageUsage {
    transfer_destination: true,
    sampled: true,
    .. usage
};
```

Of course this is a breaking API change so before I implement this and open a PR I would like to ask if it would be accepted for the next release.

Other options would be:
- provide two new methods `from_iter_with_usage` and `from_buffer_with_usage`. This would not be a breaking change.
- do nothing and leave this problem to the user

### LINKED ISSUE #1854: `ImmutableImage` isn't actually immutable
`ImmutableImage` is immutable in the sense that the user can't submit commands that modify the image contents. But there is one other kind of operation that is not considered in this: layout transitions. A layout transition is a write operation, and while `ImmutableImage` will normally be in the `ShaderReadOnlyOptimal` layout, doing a `copy_image` with the image as the source will transition it to `TransferSrcOptimal`. For the user this is invisible, but this matters a lot for synchronisation within the GPU, because such a write operation can't happen while another queue is accessing the image. This could create unexpected errors for the user when submitting command buffers, because the user thinks the image isn't ever written and thus concurrent access is possible, when it actually *is* written to, just invisibly to the user.

I have been thinking for a while about whether the immutable buffer and image types are even really needed in Vulkano. I suppose they signal the intent of the user, that they never intend to ever write to the buffer or image again, but to call it "immutable" is also misleading as demonstrated above. There may be a speed argument to be made, since the synchronisation for immutable resources is a little simpler, but that needs to be tested by comparing performance with the mutable resource variants.

### LINKED ISSUE #1988: [Feature Request] Pass ImageCreateFlags into ImmutableImage::from_buffer and from_iter
I ran across an issue while trying to create an ImmutableImage to back a CubeArray. I'm getting an error saying ImageNotCubeCompatible, which is explained in the [documentation](https://docs.rs/vulkano/latest/vulkano/image/view/enum.ImageViewCreationError.html#variant.ImageNotCubeCompatible):

> ImageNotCubeCompatible
> A cube image view type was requested, but the image was not created with the cube_compatible flag.

I see that `ImmutableImage::from_iter` is calling `ImmutableImage::from_buffer`, which has the flags [hard coded to be empty](https://github.com/vulkano-rs/vulkano/blob/8f99ceab3257d0a12a3c39ebe8206364c0229c7c/vulkano/src/image/immutable.rs#L278):
```rust
let flags = ImageCreateFlags::empty();
```
I believe that ImmutableImage is the appropriate type to use for CubeArray, because I just intend to write once and never again. It looks like as a workaround I could use `ImmutableImage::uninitialized` to specify the flags, but then I would have replicate the functionality of `ImmutableImage::from_buffer` to create a command buffer builder and perform the one-time write into the initializer, and also potentially generate mipmaps.

In summary, my request is that `ImageCreateFlags` should be added as a parameter to `ImmutableImage::from_iter` and `ImmutableImage::from_buffer`.

marc0246:
It is here, the prophecy foretold long ago, of the image unification. Things went a bit too smoothly, though I did make some decisions while consolidating features of the different image types that undermine some previous functionality, which are as follows.

### Constructors from iter/buffer

- `ImmutableImage::from_iter`

  Constructing an image from data on the host will always require a staging buffer, unless the image is ~~useless~~ linear. Hiding that this is going on is not helping anyone, quite the opposite. Creating a new staging buffer for each upload of an image for instance, when you have thousands, is insanity.

- `ImmutableImage::from_buffer`

  This is less tame, since there is no hidden buffer allocation. Yet it is still obstructing what is going on under the hood for no good reason. The image isn't created from the buffer. The function creates an uninitialized image. It's really awkward in my eyes, but worse still, recording a copy command is one line of code! This function has no utility.

### Exporting memory

- `{AttachmentImage, StorageImage}::new_with_exportable_fd`

  My issue with this is first and foremost that it is platform-specific, and if we indulged things like this we would very quickly have an explosion of constructors. Instead, I think whether or not memory is exportable is something that should be handled by the memory allocator or when allocating `DeviceMemory` manually.

  What one would do now is:
  1. Configure `StandardMemoryAllocator` to allocate memory with the desired external handle types: https://github.com/marc0246/vulkano/blob/bd68a6f601ec3e718ab3eb04f865d4d9fe27d6f4/examples/src/bin/gl-interop.rs#L576-L597
  3. Create an `Image` with the desired external handle types, using the regular `Image::new` constructor: https://github.com/marc0246/vulkano/blob/bd68a6f601ec3e718ab3eb04f865d4d9fe27d6f4/examples/src/bin/gl-interop.rs#L117-L137
    Using `MemoryAllocatePreference::AlwaysAllocate`, one can also ensure the image gets its own separate allocation.

  Alternatively one could allocate `DeviceMemory` with the desired external handle types, create a `MemoryAlloc` out of the `DeviceMemory`, and bind it to a `RawImage`.

- `{AttachmentImage, StorageImage}::export_posix_fd`

  Exporting memory is not specific to images either and therefore this functionality should be present on `DeviceMemory` only IMO.

- `{AttachmentImage, StorageImage}::mem_size`

  For suballocated images, having this be a method on the image itself doesn't make much sense. Instead one can get the allocation of the image and then the size of the `DeviceMemory` from that, albeit a bit more clunky.

Since @hakolao originally added these, hopefully he can chime in and tell me whether this sounds adequate. Also tagging @fayalalebrun because they added the GL interop example, so maybe they have a perspective on this as well.

### Importing memory

This only concerns `StorageImage::new_from_dma_buf_fd`.

Most important issues I see with this are same as above, namely that this is platform-specific and importing memory is not specific to images, and so encouraging these constructors could quickly spiral out of control. Instead I would like to err on the side of caution and request that the user imports memory using `DeviceMemory` directly. That said I wouldn't be completely opposed to the idea of adding platform-specific constructors to `Image` and `Buffer` for memory-importing, if such a shortcut is truly wanted, but it would add a lot to the maintenance burden especially with the amount of external handle types that there are.

What one would do now is:
1. Use `DeviceMemory::import` together with the file descriptor(s) to import the memory.
2. Create a `RawImage` with the desired explicit DRM format modifier(s).
3. Create `MemoryAlloc`(s) from the `DeviceMemory` and bind them to the `RawImage`.

Since @DavidR86 added this constructor, hopefully they can give me their opinion on the matter.

Fixes #1988, fixes #1854, fixes #1446, fixes #1154, fixes #715, fixes #108.

Changelog:
Remove:
```markdown
### Breaking changes
Changes to images:
- Removed the `ImageInner` type. The `inner` method of the `ImageAccess` trait now returns a reference to the inner image directly.
- Removed the `descriptor_layouts` method on the `ImageAccess` trait. All images by default now use the `General` layout for storage image descriptors, and `ShaderReadOnlyOptimal` for all other image descriptors.
```
Add:
```markdown
### Breaking changes
Changes to images:
- There is now only the single type `Image` to represent images. `ImageAccess`, `ImageInner`, `AttachmentImage`, `ImmutableImage`, `StorageImage` and `SwapchainImage` were removed.
- `ImageView` no longer has a type parameter, `ImageViewAbstract` was removed.
- Removed `ImageAccessFromUndefinedLayout`, `ImmutableImageCreationError`, `ImmutableImageInitialization` and `MipmapsCount`.
- `Image` was moved to the `image` module.
```


hakolao:
I think what you said about the exportable memory for images sounds good 👍🏻

Rua:
Do I understand correctly that, if the allocator is configured with external memory, then all buffers and images using it will also use external memory? It should probably be made clear in this case that you should use multiple allocators, one for normal resources and one for exportable ones. Having every single resource be exportable doesn't seem like a good idea.

vulkano/src/image/mod.rs
```diff
@@ -46,43 +46,863 @@
...
-    attachment::AttachmentImage,
-    immutable::ImmutableImage,
     layout::ImageLayout,
-    storage::StorageImage,
-    swapchain::SwapchainImage,
     sys::ImageError,
-    traits::ImageAccess,
     usage::ImageUsage,
-    view::{ImageViewAbstract, ImageViewType},
+    view::ImageViewType,
```
Rua, line 56: Re-exporting only `ImageViewType` here feels a bit silly. I think either `ImageView` should also be exported, or neither of them. Since it's a public module, I'm leaning towards neither.

Rua: What's your thought?

Rua:
Given how little those constructors actually did, I'm not opposed. For buffers though (which you haven't touched here), it may be a bit annoying for the user to have to calculate the size manually. A possible source of errors.

Rua:
The gl-interop example is giving me a validation error with this PR, that I don't get with the latest master:
> VUID-VkBindBufferMemoryInfo-memory-02726(ERROR / SPEC): msgNum: -1616898404 - Validation Error: [ VUID-VkBindBufferMemoryInfo-memory-02726 ] Object 0: handle = 0xcb3ee80000000007, type = VK_OBJECT_TYPE_BUFFER; Object 1: handle = 0xead9370000000008, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0x9fa0169c | vkBindBufferMemory2() pBindInfos[0]: The VkDeviceMemory (VkDeviceMemory 0xead9370000000008[]) has an external handleType of VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT which does not include at least one handle from VkBuffer (VkBuffer 0xcb3ee80000000007[]) handleType VkExternalMemoryHandleTypeFlags(0). The Vulkan spec states: If the value of VkExportMemoryAllocateInfo::handleTypes used to allocate memory is not 0, it must include at least one of the handles set in VkExternalMemoryBufferCreateInfo::handleTypes when buffer was created (https://vulkan.lunarg.com/doc/view/1.3.250.1/linux/1.3-extensions/vkspec.html#VUID-VkBindBufferMemoryInfo-memory-02726)

Oddly it's complaining about something related to buffers rather than images.

Rua:
Thanks, there's no more validation error now. Was it caused by the example code misusing unsafe code, or is it a bug in Vulkano itself?

PR #2252, opened 2023-07-04
ValidationError-ify images and image views, fix DRM format modifiers

Rua:
Changelog:
```markdown
### Breaking changes
Changes to images:
- `ImageDimensions` is removed. `ImageCreateInfo` now instead has `image_type`, `extent` and `array_layers` to match Vulkan. Some of the methods of `ImageDimensions` have been turned into bare functions in the `image` module.
- `ImageCreateFlags::ARRAY_2D_COMPATIBLE` is renamed to `DIM2D_ARRAY_COMPATIBLE` to better match the Vulkan name (but within the limitation that identifiers can't start with a digit).

### Additions
- Added `ImageCreateFlags::EXTENDED_USAGE` from the `khr_maintenance2` extension.
- Support for the `ext_image_drm_format_modifier` extension.
- Support for the `ext_image_2d_view_of_3d` extension.
````

DRM format modifiers are now properly supported, unlike the mess we had before. `ImageDimensions` is also gone and replaced with a more Vulkany API.


AustinJ235 (approved):
Looks good with the exception of a couple grammar things in the examples.

PR #2253, opened 2023-07-06
Box `ValidationError`

### LINKED ISSUE #2251: Consider boxing errors more
Clippy complains about my error enums based on Vulkano error types because they are very large (>= 160 bytes), and thus make Results very expensive to move around if the memcpy is not elided.

I think this is a valid performance concern from clippy's side, as I've observed my fair share of failed move optimizations from rustc/LLVM, and the performance impact that comes with them can indeed be quite significant.

On the other hand, I am also well aware that error boxing come with a large ergonomic hit. Among other things, it reduces ability to pattern match error types, adds annoying Box::new() everywhere, and if applied on the client side it breaks use of `?` unless you write all the `From<NotBoxedError>` impls yourself. The latter is an argument in favor of implementing boxing on the side of the code that throws errors, if it is to be implemented at all.

Overall, I'm a bit torn here, and could use extra opinions from other people. Are you fine with the performance risks of having large results, or would you rather have vulkano do the right thing from the performance PoV even if it comes at some ergonomic cost (and a major error API break if done today)?

Rua:
Changelog:

```diff
Changes to errors:
- - `VulkanError` is renamed to `RuntimeError`.
- - Added new types `ValidationError` and `VulkanError` (enum of `ValidationError` + `RuntimeError`) to return errors from any function. Some existing functions have been converted to use these types, others will follow later.
+ - Added new types `ValidationError` and `Validated` (enum of `ValidationError` + something else) to return errors from any function. Some existing functions have been converted to use these types, others will follow later.
````

Fixes #2251. I've also un-renamed `RuntimeError` back to `VulkanError`, as I wasn't quite happy with that change in the end. I made the combo enum generic, so that it can be used for error types other than `VulkanError`.


Rua:
I tried that, but it conflicts with the `From<ValidationError>` impl.

PR #2264, opened 2023-07-16
Switch to a more expressive way to specify memory usage

marc0246:
Previously there were these memory usages:
`MemoryUsage` | Ideal memory type
-|-
`DeviceOnly` | `DEVICE_LOCAL`
`Upload` | `DEVICE_LOCAL \| HOST_VISIBLE \| HOST_COHERENT`
`Download` | `HOST_VISIBLE \| HOST_CACHED`

However, almost all PC and laptop implementations also have a memory type that is `HOST_VISIBLE | HOST_COHERENT`, which was not possible to request using those, which is the biggest reason for this change. However, some other implementations have more quirky memory types as well, such a `DEVICE_LOCAL | HOST_VISIBLE | HOST_CACHED` memory type on MacOS and mobile devices. It was also not possible to distinguish between that and the non-device-local host-cached one.

Here's all the combinations possible now, using the presets available as associated constants of `MemoryTypeFilter`:
`MemoryTypeFilter` | Ideal memory type
-|-
`PREFER_DEVICE` | `DEVICE_LOCAL`
`PREFER_DEVICE \| HOST_SEQUENTIAL_WRITE` | `DEVICE_LOCAL \| HOST_VISIBLE \| HOST_COHERENT`
`PREFER_DEVICE \| HOST_RANDOM_ACCESS` | `DEVICE_LOCAL \| HOST_VISIBLE \| HOST_CACHED`
`PREFER_HOST` |
`PREFER_HOST \| HOST_SEQUENTIAL_WRITE` | `HOST_VISIBLE \| HOST_COHERENT`
`PREFER_HOST \| HOST_RANDOM_ACCESS` | `HOST_VISIBLE \| HOST_CACHED`

There are of course still some memory types that can't be differentiated, such as in the case of `HostAccessType::RandomAccess`, there might be both `HOST_VISIBLE | HOST_CACHED` and `HOST_VISIBLE | HOST_COHERENT | HOST_CACHED` around. If the user truly needs something that specialized, then they should specify the memory type filter manually.

Changelog:
```markdown
### Breaking changes
Changes to memory allocation:
- `AllocationCreateInfo::usage` and `SubbufferAllocatorCreateInfo::memory_usage` were replaced by a `memory_type_filter` field, to allow for a more flexible selection of the memory type. Additionally, `SubbufferAllocatorCreateInfo::memory_type_filter` defaults to `MemoryTypeFilter::PREFER_DEVICE` for consistency with `AllocationCreateInfo`, unlike the previous default of `MemoryUsage::Upload`.
- `SubbufferAllocatorCreateInfo::buffer_usage` is now empty by default for consistency with `BufferCreateInfo`.
```

### DIFF
--- a/vulkano/src/buffer/allocator.rs
+++ b/vulkano/src/buffer/allocator.rs
@@ -17,8 +17,8 @@ use crate::{
     device::{Device, DeviceOwned, DeviceOwnedDebugWrapper},
     memory::{
         allocator::{
-            align_up, AllocationCreateInfo, DeviceLayout, MemoryAllocator, MemoryAllocatorError,
-            MemoryUsage, StandardMemoryAllocator,
+            align_up, AllocationCreateInfo, DeviceLayout, HostAccessType, MemoryAllocator,
+            MemoryAllocatorError, MemoryLocationPreference, StandardMemoryAllocator,
         },
         DeviceAlignment,
     },
@@ -132,7 +132,8 @@ where
         let SubbufferAllocatorCreateInfo {
             arena_size,
             buffer_usage,
-            memory_usage,
+            location_preference,
+            host_access,
             _ne: _,
         } = create_info;
 
@@ -157,7 +158,8 @@ where
             state: UnsafeCell::new(SubbufferAllocatorState {
                 memory_allocator,
                 buffer_usage,
-                memory_usage,
+                location_preference,
+                host_access,
                 buffer_alignment,
                 arena_size,
                 arena: None,
@@ -265,7 +267,8 @@ where
 struct SubbufferAllocatorState<A> {
     memory_allocator: A,
     buffer_usage: BufferUsage,
-    memory_usage: MemoryUsage,
+    location_preference: MemoryLocationPreference,
+    host_access: HostAccessType,
     // The alignment required for the subbuffers.
     buffer_alignment: DeviceAlignment,
     // The current size of the arenas.
@@ -349,7 +352,8 @@ where
                 ..Default::default()
             },
             AllocationCreateInfo {
-                usage: self.memory_usage,
+                location_preference: self.location_preference,
+                host_access: self.host_access,
                 ..Default::default()
             },
             DeviceLayout::from_size_alignment(self.arena_size, 1).unwrap(),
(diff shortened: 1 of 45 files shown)


vulkano/src/memory/allocator/mod.rs
```diff
@@ -392,29 +392,39 @@ pub struct MemoryTypeFilter {
...
+            filter.not_preferred_flags |= MemoryPropertyFlags::DEVICE_LOCAL;
+        }
+    }
 
-        filter
+    match host_access {
+        HostAccessType::DontCare => {}
+        HostAccessType::SequentialWrite => {
+            filter.required_flags |=
+                MemoryPropertyFlags::HOST_VISIBLE | MemoryPropertyFlags::HOST_COHERENT;
```
Rua, line 418: I think `HOST_COHERENT` should be preferred, but not required?

vulkano/src/memory/allocator/mod.rs
```diff
@@ -392,29 +392,39 @@ pub struct MemoryTypeFilter {
...
+            filter.not_preferred_flags |= MemoryPropertyFlags::HOST_CACHED;
+        }
+        HostAccessType::RandomAccess => {
+            filter.required_flags |=
+                MemoryPropertyFlags::HOST_VISIBLE | MemoryPropertyFlags::HOST_CACHED;
```
Rua, line 423: Should `HOST_CACHED` be a strict requirement here?

vulkano/src/buffer/allocator.rs
```diff
@@ -413,10 +417,15 @@ pub struct SubbufferAllocatorCreateInfo {
     /// The default value is [`BufferUsage::TRANSFER_SRC`].
     pub buffer_usage: BufferUsage,
 
-    /// The memory usage that all buffers should be allocated with.
+    /// The preferred memory location the buffers should be allocated with.
     ///
-    /// The default value is [`MemoryUsage::Upload`].
-    pub memory_usage: MemoryUsage,
+    /// The default value is [`MemoryLocationPreference::Host`].
```
Rua, line 422: Having the default location preference for regular allocations be `Device`, but `Host` for subbuffers, could be rather surprising for the user when they're not aware of the difference.

vulkano/src/memory/allocator/mod.rs
Rua: Not every device is required to have `HOST_CACHED` memory, so asking for `RandomAccess` would fail on such a device, which is bad.

The problem I also see is that you're treating both flags as equally important, when they aren't necessarily. Is the caching more important, or being on the device?

vulkano/src/memory/allocator/mod.rs
Rua: The spec requires a `HOST_COHERENT` memory type to exist, but there is always the possibility of that memory running out. In that case, is a hard error preferred, or should the allocator fall back to something less ideal?

vulkano/src/buffer/allocator.rs
Rua: An empty usage by default makes more sense, because it's what we have for everything else already too.

Rua: Yes, I agree.

Rua:
Nice, I like it! Using Rust features in ways that don't work for a C API!

Rua:
Everything good, thanks!

PR #2268, opened 2023-07-24
ValidationError-ify `ShaderModule`

Rua:
Changelog:
```markdown
### Additions
- `ShaderModule` is now constructed with a `new` method, which takes a `ShaderModuleCreateInfo` struct. The old constructors `from_words` and `from_bytes` are deprecated.
````

Because having two different versions of `ShaderModuleCreateInfo`, one with bytes and one with words, would be difficult, I deprecated the old constructors. I also fixed an oversight in the runtime shader example: SPIR-V in *byte* form is always little-endian, so when reading the file you have to convert it.


Rua:
What error format do you mean?

PR #2273, opened 2023-07-25
Allow to be built without cargo

06393993:
aosp has introduced this project as a dependency, and aosp doesn't use cargo to build. `env!` will fail the compilation if the envs are not set. Replace with `option_env` to delay the panic to runtime.

2. [x] Make sure that the changes are covered by unit-tests: this change will be built by the aosp build infrastructure.

4. [x] Run `cargo fmt` on the changes.

Changelog:
```markdown
### Bugs fixed
- Allow compile without cargo
````

### DIFF
--- a/vulkano/src/instance/mod.rs
+++ b/vulkano/src/instance/mod.rs
@@ -1059,11 +1059,20 @@ impl InstanceCreateInfo {
     #[inline]
     pub fn application_from_cargo_toml() -> Self {
         Self {
-            application_name: Some(env!("CARGO_PKG_NAME").to_owned()),
+            application_name: Some(option_env!("CARGO_PKG_NAME").unwrap().to_owned()),
             application_version: Version {
-                major: env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap(),
-                minor: env!("CARGO_PKG_VERSION_MINOR").parse().unwrap(),
-                patch: env!("CARGO_PKG_VERSION_PATCH").parse().unwrap(),
+                major: option_env!("CARGO_PKG_VERSION_MAJOR")
+                    .unwrap()
+                    .parse()
+                    .unwrap(),
+                minor: option_env!("CARGO_PKG_VERSION_MINOR")
+                    .unwrap()
+                    .parse()
+                    .unwrap(),
+                patch: option_env!("CARGO_PKG_VERSION_PATCH")
+                    .unwrap()
+                    .parse()
+                    .unwrap(),
             },
             ..Default::default()
         }

### CODE REVIEW

File: vulkano/src/instance/mod.rs [hunk 1]
```diff
@@ -1061,9 +1061,18 @@ impl InstanceCreateInfo {
         Self {
             application_name: Some(env!("CARGO_PKG_NAME").to_owned()),
```
Rua, line 1062: I think this `env!` would also need to be replaced?

06393993: Thanks for the catch.

On the AOSP side, I was just trying to import the 0.33.0 version, where `application_name` is just `Some("vulkano".to_owned())`.

I also noticed that many `include!` with `env!` are introduced recently. If that in the next release breaks the AOSP build, I will try to raise questions to the Android Rust team to see the best way to solve this.

Rua:
Could your own build system not set these environment variables itself?

06393993:
> Could your own build system not set these environment variables itself?

According to the [document of Android.bp](https://ci.android.com/builds/submitted/10574911/linux/latest/view/soong_build.html) for `cargo_env_compat` and `cargo_pkg_version`, only `CARGO_CRATE_NAME`, `CARGO_BIN_NAME`, and `CARGO_PKG_VERSION` can be set.

Should I refactor the change to only use `CARGO_PKG_VERSION` or should I ask our build team to implement the support? Thanks!

Rua:
I think in the long term, adding support for those other variables in your build system would be more useful, especially if you intend to include other Rust projects in the future. But why not just use Cargo?

06393993:
> especially if you intend to include other Rust projects in the future

Sure. I am running an internal email thread to see if we can support those environment variables. Will update in this PR. On the other hand, the current way already allows us to import [many Rust crates](https://android.googlesource.com/platform/external/rust/crates/) already into AOSP.

> But why not just use Cargo?

[This blog](https://security.googleblog.com/2021/05/integrating-rust-into-android-open.html) summarizes the reason why Android doesn't nest cargo in the current Soong(and the future bazel) build system. [Chromium](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/tools/crates/README.md#generating-files-for-third_party-crates) and [Fuchsia](https://fuchsia.dev/fuchsia-src/development/languages/rust#cargotoml) follow the same practice. I am not at the position to judge or change this decision though.

06393993:
Update: we are likely to add support for `CARGO_PKG_VERSION_MAJOR`, `CARGO_PKG_VERSION_MINOR`, 
 and `CARGO_PKG_VERSION_PATCH` to the AOSP Soong build system.

Will close this PR once those changes landed. Feel free to close this PR early if you find it necessary. Thanks.

06393993:
With [this change](https://r.android.com/2679918) landed in Android. We don't need this change for the project to build in AOSP. Close the PR.

PR #2289, opened 2023-08-16
Memory allocator impl

vE5li:
This aims to fix some problems with usability when creating a custom memory allocator. Simple scenario is creating a struct wrapping a `MemoryAllocator`
```rust
pub struct FooAllocator(StandardMemoryAllocator);

unsafe impl MemoryAllocator for FooAllocator { /* ... */ }
```
When trying to create a `SubbufferAllocator` using an `Arc<FooAllocator>` it will not compile, because `MemoryAllocator` is only implemented for `Arc<GenericMemoryAllocator<S>>`. With these changes `MemoryAllocator` will be implemented for any `Arc<T> where T: MemoryAllocator`, making custom allocators easier to use.

While I was at it I also added the same blanket implementation for the `CommandBufferAllocator`.

### DIFF
--- a/vulkano/src/command_buffer/allocator.rs
+++ b/vulkano/src/command_buffer/allocator.rs
@@ -276,12 +276,12 @@ unsafe impl CommandBufferAllocator for StandardCommandBufferAllocator {
     }
 }
 
-unsafe impl CommandBufferAllocator for Arc<StandardCommandBufferAllocator> {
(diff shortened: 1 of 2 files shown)


vE5li:
Oh, I was actually looking for that impl earlier but couldn't find it. I'll be right on it :+1:

PR #2300, opened 2023-08-22
Merge `MappedDeviceMemory` into `DeviceMemory`, make `MemoryAlloc` reuse the logic

marc0246:
Changelog:
```markdown
### Breaking changes
- `Subbuffer::mapped_ptr` was replaced by `Subbuffer::mapped_slice`.
- `MemoryAlloc::new` no longer returns a `Result`, and doesn't map the `DeviceMemory` automatically anymore.
- `MemoryAlloc::mapped_ptr` and `MemoryAlloc::mapped_slice[_mut]` were replaced by `MemoryAlloc::mapped_slice`, which returns a pointer.
- `MemoryAlloc::{invalidate, flush}_range` now take a `MappedMemoryRange` as argument.

### Additions
- Added `DeviceMemory::{map, unmap, mapping_state, invalidate_range, flush_range}`, `MappedDeviceMemory` has been deprecated.
- Added `MemoryMapInfo`, `MemoryUnmapInfo`, `MappingState` and `MappedMemoryRange`.

### Bugs fixed
- Fixed potential UB when using `MemoryAlloc::try_unwrap`, where the allocation was mapped on contruction of the `MemoryAlloc` but not unmapped on unwrapping, allowing double-mapping.
- Fixed a bug in `GenericMemoryAllocator::allocate`, where the root allocations weren't created with the configured `AllocationType`.

### DIFF
--- a/vulkano/src/memory/allocator/mod.rs
+++ b/vulkano/src/memory/allocator/mod.rs
@@ -816,7 +816,7 @@ impl StandardMemoryAllocator {
 ///
 /// Every time a new `DeviceMemory` block is allocated, it is mapped in full automatically as long
 /// as it resides in host-visible memory. It remains mapped until it is dropped, which only happens
-/// if the allocator is dropped. In other words, all eligeble blocks are persistently mapped, so
+/// if the allocator is dropped. In other words, all eligible blocks are persistently mapped, so
 /// you don't need to worry about whether or not your host-visible allocations are host-accessible.
 ///
 /// # `DeviceMemory` allocation
--- a/vulkano/src/memory/allocator/mod.rs
+++ b/vulkano/src/memory/allocator/mod.rs
@@ -812,6 +812,13 @@ impl StandardMemoryAllocator {
 ///
 /// See also [the `MemoryAllocator` implementation].
 ///
+/// # Mapping behavior
+///
+/// Every time a new `DeviceMemory` block is allocated, it is mapped in full automatically as long
+/// as it resides in host-visible memory. It remains mapped until it is dropped, which only happens
+/// if the allocator is dropped. In other words, all eligeble blocks are persistently mapped, so
+/// you don't need to worry about whether or not your host-visible allocations are host-accessible.
+///
 /// # `DeviceMemory` allocation
 ///
 /// If an allocation is created with the [`MemoryAllocatePreference::Unknown`] option, and the
--- a/vulkano/src/memory/allocator/mod.rs
+++ b/vulkano/src/memory/allocator/mod.rs
@@ -1584,7 +1584,6 @@ unsafe impl<S: Suballocator> MemoryAllocator for GenericMemoryAllocator<S> {
             // - The memory can't be mapped already, because we just allocated it.
             // - Mapping the whole range is always valid.
             memory.map_unchecked(MemoryMapInfo {
-                flags: Default::default(),
                 offset: 0,
                 size: memory.allocation_size(),
                 _ne: crate::NonExhaustive(()),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: vulkano/src/memory/device_memory.rs [hunk 1]
```diff
@@ -370,6 +402,255 @@ impl DeviceMemory {
...
+    }
+
+    pub(crate) fn atom_size(&self) -> DeviceAlignment {
+        self.atom_size
+    }
+
+    /// Maps a range of memory to be accessed by the host.
+    ///
+    /// `self` must not be host-mapped already and must be allocated from host-visible memory.
+    pub fn map(&mut self, map_info: MemoryMapInfo) -> Result<(), Validated<VulkanError>> {
```
Rua, line 418: Functions like this that only validate and then call unchecked, usually have `#[inline]`.

Rua:
The [`VK_KHR_map_memory2`]https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_map_memory2.html extension provides updated versions of some of these functions. It's not used by anything in Vulkan yet, but at least Vulkano would be ready for anything that does in the future.

vulkano/src/memory/allocator/mod.rs
```diff
@@ -1546,20 +1552,41 @@ unsafe impl<S: Suballocator> MemoryAllocator for GenericMemoryAllocator<S> {
...
+        if self.pools[memory_type_index as usize]
+            .memory_type
+            .property_flags
+            .intersects(ash::vk::MemoryPropertyFlags::HOST_VISIBLE)
+        {
+            // SAFETY:
+            // - We checked that the memory is host-visible.
+            // - The memory can't be mapped already, because we just allocated it.
+            // - Mapping the whole range is always valid.
+            memory.map_unchecked(MemoryMapInfo {
```
Rua, line 1579: I think it would be good to explicitly document in `fn allocate` that the memory is automatically host-mapped in this case. Right now the user has no way to know this, it seems.

vulkano/src/memory/device_memory.rs
```diff
@@ -370,6 +397,304 @@ impl DeviceMemory {
...
+    pub fn unmap(&mut self, unmap_info: MemoryUnmapInfo) -> Result<(), Validated<VulkanError>> {
+        self.validate_unmap(&unmap_info)?;
+
+        unsafe { self.unmap_unchecked(unmap_info) }?;
+
+        Ok(())
+    }
+
+    fn validate_unmap(&self, unmap_info: &MemoryUnmapInfo) -> Result<(), Box<ValidationError>> {
+        let &MemoryUnmapInfo { flags: _, _ne: _ } = unmap_info;
```
Rua, line 531: `MemoryUnmapInfo` needs a `validate` too, even if only to check for future flags.

vulkano/src/memory/device_memory.rs
```diff
@@ -1507,6 +1509,14 @@ pub struct MemoryUnmapInfo {
     pub _ne: crate::NonExhaustive,
 }
 
+impl MemoryUnmapInfo {
+    pub(crate) fn validate(&self, _memory: &DeviceMemory) -> Result<(), Box<ValidationError>> {
+        let &Self { flags: _, _ne: _ } = self;
```
Rua, line 1514: Can you validate `flags`, so that it's not forgotten about when flags are added in the future?

vulkano/src/memory/allocator/mod.rs
```diff
@@ -812,6 +812,13 @@ impl StandardMemoryAllocator {
 ///
 /// See also [the `MemoryAllocator` implementation].
 ///
+/// # Mapping behavior
+///
+/// Every time a new `DeviceMemory` block is allocated, it is mapped in full automatically as long
+/// as it resides in host-visible memory. It remains mapped until it is dropped, which only happens
+/// if the allocator is dropped. In other words, all eligeble blocks are persistently mapped, so
```
Rua, line 819:
```suggestion
/// if the allocator is dropped. In other words, all eligible blocks are persistently mapped, so
```

Rua:
Everything is good now, thanks!

PR #2306, opened 2023-08-26
Fix validation and errors in `MemoryAllocator`

marc0246:
This fixes some extremely strange validation left over from #1997, and some other panics.

Seeing as allocation of `DeviceMemory` is not on the hot path whatsoever, I don't understand why the allocator didn't just call the validated allocation function before, but rather panicking. This also made it harder to implement a custom memory allocator as there were both checked and unchecked versions of the methods for no reason. I also removed some nonsensical validation that didn't do anything to begin with, and the panic when picking a memory type fails was replaced by an error variant.

Changelog:
```markdown
### Breaking changes
Changes to memory allocation:
- Merged `MemoryAllocator::allocate_from_type_unchecked` into `MemoryAllocator::allocate_from_type` and `MemoryAllocator::allocate_unchecked` into `MemoryAllocator::allocate`.
- Replaced `MemoryAllocator::allocate_dedicated_unchecked` with `MemoryAllocator::allocate_dedicated`.
```

### DIFF
--- a/vulkano/src/memory/allocator/mod.rs
+++ b/vulkano/src/memory/allocator/mod.rs
@@ -235,7 +235,8 @@ use super::{
 use crate::{
     device::{Device, DeviceOwned},
     instance::InstanceOwnedDebugWrapper,
-    DeviceSize, Requires, RequiresAllOf, RequiresOneOf, ValidationError, Version, VulkanError,
+    DeviceSize, Requires, RequiresAllOf, RequiresOneOf, Validated, ValidationError, Version,
+    VulkanError,
 };
 use ash::vk::{MAX_MEMORY_HEAPS, MAX_MEMORY_TYPES};
 use parking_lot::RwLock;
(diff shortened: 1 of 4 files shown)


vulkano/src/memory/allocator/mod.rs
```diff
@@ -742,37 +694,31 @@ pub enum MemoryAllocatorError {
...
-            Self::OutOfPoolMemory => write!(f, "the pool doesn't have enough free space"),
-            Self::DedicatedAllocationRequired => write!(
-                f,
-                "a dedicated allocation is required but was explicitly forbidden",
-            ),
-            Self::BlockSizeExceeded => write!(
-                f,
+        let msg = match self {
+            Self::AllocateDeviceMemory(_) => "allocating device memory failed",
+            Self::FindMemoryType => "finding a suitable memory type filed",
```
Rua, line 707:
```suggestion
            Self::FindMemoryType => "finding a suitable memory type failed",
```

Rua:
Thanks!

PR #2307, opened 2023-08-26
Wrap buffer values in `AccelerationStructureBuildGeometryInfo` in an `Option`

Rua:
Fixes #2304.

PR #2309, opened 2023-08-26
#2307 fixes

marc0246:


Rua:
Ah, that old mistake again. Thanks for spotting it!

PR #2314, opened 2023-08-28
Add an example for ray queries and acceleration structures

### LINKED ISSUE #2308: Add an example for `AccelerationStructure`
Introduced in #2213, but still not covered by any examples.

atynagano:
Fixes #2308.

The code is mostly from triangle-v1_3 and lacks compatibility check, comments, and may contains irregular coding styles.

### DIFF
--- a/examples/src/bin/ray-query.rs
+++ b/examples/src/bin/ray-query.rs
@@ -1,0 +1,1059 @@
+// Copyright (c) 2016 The vulkano developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or https://opensource.org/licenses/MIT>,
+// at your option. All files in the project carrying such
+// notice may not be copied, modified, or distributed except
+// according to those terms.
+
+// Welcome to the triangle example!
+//
+// This is the only example that is entirely detailed. All the other examples avoid code
+// duplication by using helper functions.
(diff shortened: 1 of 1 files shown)


atynagano:
I am not confident about the quality, but should I make it Ready for review?

Rua:
At first glance it looks ok, but acceleration structures aren't supported on my device yet (still waiting on Mesa 23.2) so I can't test it. The code should not crash when it's not supported, so that's a TODO for this PR still.

I notice that when you build an acceleration structure, you immediately submit a command buffer and then wait for it to finish. That doesn't seem very efficient.

atynagano:
@Rua Thanks for your review.

> I notice that when you build an acceleration structure, you immediately submit a command buffer and then wait for it to finish. That doesn't seem very efficient.

I know, but BLAS, TLAS, and pipeline have dependencies, and at least these cannot be executed asynchronously. I have referred to the following implementation, where they are executed synchronously.

https://github.com/KhronosGroup/Vulkan-Samples/blob/5e4e3d20b3a86f7814e026042341cf34ef530edd/framework/core/acceleration_structure.cpp#L218
https://github.com/KhronosGroup/Vulkan-Samples/blob/5e4e3d20b3a86f7814e026042341cf34ef530edd/framework/core/device.cpp#L621

Rua:
But is it necessary to *wait* for the submission? That means the whole execution stalls until the fence is signalled. A semaphore, or just putting them all in the same command buffer, seems less extreme.

PR #2322, opened 2023-09-07
Fix missing overflow check when constructing suballocators

marc0246:
Fixes an oversight in #2316, where a safety precondition no longer holds. Previously all allocations were bounded by the root allocation which couldn't exceed `DeviceLayout::MAX_SIZE`, but now that regions are separate from allocations, it has to be checked separately somehow. I added a `Region` type solely for the sake of documentation, but I still don't know if it's best like this, where the suballocator would panic on construction (which results in a bit of duplication and could be more error-prone for an implementor of the trait) or whether `Region` should have private fields much like `DeviceLayout` and have both a safe and unsafe way to construct it. The thing is though that creating a suballocator isn't on the hot path unlike creating a `DeviceLayout`.


Rua:
Is there a reason for not using `Range<DeviceSize>` to prevent overflow?

PR #2332, opened 2023-09-14
Make the insides of `AllocationHandle` private

marc0246:
Let's stay conservative, to for example allow adding more fields to it (in particular consider a thicc pointer) or turn it into a union (e.g. if the need for storing a `DeviceSize` ever arises).


Rua:
Curious why you made `AllocationHandle` transparent except in documentation?

Issue #1738, opened 2021-10-26
Validation of shader interface matching is insufficient


Rua:
Vulkano currently checks whether the interfaces of two shaders match, by checking that the values written to each interface location are the same. But the Vulkan spec has a stronger requirement, as described [here](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap15.html#interfaces-iointerfaces-matching). There is only a match if the types of the two variables, as defined with SPIR-V opcodes, match. It is possible for two shaders to have the same interface location representation but different actual types, so the current check in Vulkano seems to be not strict enough.

PR #2333, opened 2023-09-15
Fix overlapping attributes when using `dualSrcBlend` feature

Fuzzyzilla:
N/A, it was an internal bug

Looks like shader module as a whole is missing unit tests :<

3. [X] Run `cargo fmt` on the changes.

4. [X] Please put changelog entries **in the description of this Pull Request**
   
5. [X] Describe in common words what is the purpose of this change, related

This change allows for fragment shaders to make use of the `dualSrcBlend` feature of Vulkan 1.0, by assigning an `index` layout to color outputs. In vulkano 0.33.0, this causes a compile time error, and in the current `master` branch it causes a runtime error, due to detection of overlapping interface variables not taking `index` into account. This fixes that :>

Changelog:
```markdown
### Bugs fixed
- fragment shaders cannot use `dual_src_blend` device feature due to interface errors
````

### DIFF
--- a/vulkano/src/shader/mod.rs
+++ b/vulkano/src/shader/mod.rs
@@ -1314,6 +1314,10 @@ pub struct ShaderInterfaceEntry {
     /// The location slot that the variable starts at.
     pub location: u32,
 
+    /// The index within the location slot that the variable is located.
+    /// Only meaningful for fragment outputs.
+    pub index: u32,
+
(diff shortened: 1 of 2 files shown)


Rua:
The shader interface handling is due an overhaul anyway, to solve issues like #1738. But this is useful in the meantime, thanks!

PR #2345, opened 2023-10-02
Allow for setting the array length in a shader using a specialization constant

### LINKED ISSUE #1956: Cannot set array length with Specialization Constants
Hey,

I am trying to set an array length with a specialization constant. 
This is the shader code:
```glsl
#version 460
layout(constant_id=0) const uint my_array_length = 25;
struct Vertex{
    vec3 pos;
    vec4 data[my_array_length];
};
layout(set = 1, binding = 1) buffer VertexData{
    Vertex vertices[];
};
...
```

This should work but I get the following error:
```
error: proc macro panicked
   --> src/main.rs:259:5
    |
259 | /     vulkano_shaders::shader! {
260 | |         ty: "vertex",
261 | |         path: "path/to/shader.vert",
262 | |
...   |
266 | |         }
267 | |     }
    | |_____^
    |
    = help: message: failed to find array length
```

I think this is due to vulkano's array length checking that cannot handle const variables (see here)

# Info 

*   Version of vulkano: 0.30
*   OS: Linux
*   GPU (the selected PhysicalDevice): AMD Radeon R9 380 Series 
*   GPU Driver: amdgpu

rohdealx:
To allow for something like this using vulkano-shaders:
```glsl
layout(constant_id = 0) const uint n = 64;
layout(std430, set = 0, binding = 0) uniform layout_sample { float sample[n]; };
```

Fixes #1956

Changelog:
```markdown
### Addditions
- Vulkano-shaders: support for specialization-constant-sized arrays in structs (they are generated with the size specified as fallback in the specialization constant initializer).
```

### DIFF
--- a/vulkano-shaders/src/structs.rs
+++ b/vulkano-shaders/src/structs.rs
@@ -581,7 +581,7 @@ impl TypeArray {
 
         let length = length_id
             .map(|id| match shader.spirv.id(id).instruction() {
-                Instruction::Constant { value, .. } => {
+                Instruction::Constant { value, .. } | Instruction::SpecConstant { value, .. } => {
                     assert!(matches!(value.len(), 1 | 2));
                     let len = value.iter().rev().fold(0u64, |a, &b| (a << 32) | b as u64);
 


Rua:
That's already implemented by #2329.

Rua:
What about generating a struct with a const parameter for the length?

PR #2352, opened 2023-10-07
More #2317 docs fixes

marc0246:


vulkano/src/memory/allocator/suballocator.rs
```diff
@@ -1067,17 +1064,7 @@ struct BuddyAllocatorState {
 ///
 /// # Efficiency
 ///
-/// Allocation is *O*(1), and so is resetting the allocator (freeing all allocations). Allocation
-/// is always lock-free, and most of the time even wait-free. The only case in which it is not
-/// wait-free is if a lot of allocations are made concurrently, which results in CPU-level
-/// contention. Therefore, if you for example need to allocate a lot of buffers each frame from
-/// multiple threads, you might get better performance by using one `BumpAllocator` per thread.
```
hYdos, line 1074: i feel like this first paragraph is helpful for choosing an implementation

hYdos: ah, all good then

PR #2375, opened 2023-10-28
Improve the examples' directory structure

marc0246:
This is a collection of general improvements in regards to the directory structure of our examples that I think will benefit both users and developers, things that have been bothering me over the years.

In the first commit, I made each example its own workspace member. This has a couple of advantages:
- It should in theory allow cargo to parallelize the compilation of examples better. This is useful in particular when running `cargo test`, and when using rust-analyzer.
- You can easily remove individual examples from the workspace if you don't want them to build (e.g. while using rust-analyzer) by editing the workspace manifest. For instance you might want to have just one example in the workspace while working on something, so that incremental compilation is faster.
- It's clear which example uses which dependencies.
  - Mostly just less confusing to new users.
  - If you only want to build one example, you don't need to build the dependencies of all the other ones.
- Specifying files relative to the example's directory is cleaner, compare e.g. `src/bin/<example>/file.ext` vs. `file.ext`.

Unrelated to workspace members, I find these to be fine improvements as well:
- There's less nesting: compare `examples/src/bin/teapot/main.rs` to `examples/teapot/main.rs`. This makes it faster to browse.
- Each example is in a directory now unlike there being a mix of files and directories, which makes it easier to find the one you're looking for.
- The naming of the examples is consistent now, wheras before there would be a mix of `_` and `-`-separated names. I went with the `-` because that's what all our public crates use.

The second commit, besides adjusting the shader paths, makes the `runtime-shader` example work no matter where you are in the workspace. This was driving me *MENTAL*. Made me cuckoo I tell you. Institution-ready. Reason being that I like to run examples while having my cursor over them in vifm and hitting a shortcut to run it, which wouldn't work for this one.

The fourth commit, besides adjusting the shader paths, moves the `lib.rs` file from the `examples` directory to the `teapot` directory, and makes it a module of the example. Having this file in the `examples` directory was quite strange and confusing.

In the sixth and ninth commits I made all output files be stored in the directory of the example, rather than in your current working directory, so that it's clear which files come from where and don't pollute what they're not supposed to.


Rua:
I noticed that when running the `run_all.sh` script, it has to compile every example before running it individually, instead of compiling all of them in advance so they can all run in quick succession. That's a bit inconvenient.

Other than that, everything looks good.

Rua:
Yeah I get the same. As if the first compile wasn't enough or something?

Rua:
I find the script extremely helpful for quickly making sure all the examples work and aren't crashing or triggering validation errors. It's part of my standard routine before posting a PR. I never had any slowdown issues, so I'd just let it compile everything in the background and then quickly step through all the example programs.

Rua:
Much better now, thank you!

PR #2377, opened 2023-10-29
Fix Cargo manifests

marc0246:
A couple fixes to the Cargo manifests of our public crates:
- vulkano had both a `macros` and `vulkano-macros` feature, the latter of which is completely useless and can't be used. Now there's only `macros` so it's a bit less confusing hopefully.
- Each crate's manifest links to its subdirectory in the repo, instead of the repo as a whole.
- Each crate's manifest links to its own docs page, rather than vulkano's.
- Each crate now has the readme linked.
- Other miscellaneous improvements.


Rua:
The new repository URLs are all deadlinks, so I think something went wrong there.

PR #2391, opened 2023-10-31
Unify workspaces

marc0246:
This reverts the examples having their own workspace introduced in #2381, because it turned out to be more trouble than it's worth. It broke the `run_all.sh` script because the new `target` directory was generated in `examples`, also broke the manifest in the `examples` directory after the `target` directory is generated there.

The alternative would have been to create a `.cargo/config.toml` in the root of the repo with a `target-dir = "target"` so that both target directories are in the root, but again more trouble than it's worth and it would break the developer's global configs of the target-dir.


Rua:
With the workspace dependencies applying to every crate, this means that all the dependencies are going to be built regardless of which crate you're actually building?

Rua:
Because they are workspace dependencies, they apply to every crate in the workspace?

PR #2397, opened 2023-11-05
Improve examples scripts

marc0246:
This adds a PowerShell version of the `run_all.sh` script for those that want to test on Windows, as well as headless versions of both scripts which will come in handy for CI but can also be used by us. I also made the existing `run_all.sh` script more portable.


Rua:
I get an error when I try to run the script:
> rua@rua-desktop ~/code/vulkano/examples $ ./run_all.sh 
> ./run_all.sh: 2: set: Illegal option -o pipefail

Rua:
> rua@rua-desktop ~ $ sh --version
> sh: 0: Illegal option --

Eh? XD

Rua:
I assume it's done for compatibility reasons, and I guess it also stops people from using `sh` when they're not actually using commands compatible with `sh`. I have both `sh` and `bash`, and the latter is 5.1.16.

Rua:
Now it works, but the gl-interop example crashes (as it has been for a while), and the script terminates instead of continuing. Is there a way to make it continue?

Rua:
> Wait, is continuing on failure the _**expected**_ behavior?

It's what the script did before, at least. And there should be a way to skip over a problematic example so that you can continue with the rest, it shouldn't block you from everything else.

PR #2400, opened 2023-11-07
Descriptor set allocator API 2.0

marc0246:
This makes `DescriptorSetAllocator` object-safe, allowing us to get rid of the type parameter on `PersistentDescriptorSet`. This will be useful for when we remove the `DescriptorSet` trait.

Changelog:
```markdown
### Breaking changes
Changes to descriptor sets:
- `PersistentDescriptorSet` now takes an `Arc<dyn DescriptorSetAllocator>` on construction, the type parameter for the descriptor set allocator was removed.

Changes to descriptor set allocation:
- `DescriptorSetAllocator` no longer has an `Alloc` associated type in order to make the trait object-safe.
- There is now only the single type `DescriptorSetAlloc` to represent allocated descriptor sets. The `DescriptorSetAlloc` trait and `StandardDescriptorSetAlloc` type were removed.
- `DescriptorSetAllocator` has a new required method `deallocate`.
````


Rua:
If `DescriptorSetAllocator` is always required to be wrapped in an `Arc` now, would it make more sense to construct it already wrapped in one?

PR #2403, opened 2023-11-11
Command buffer allocator API 2.0

marc0246:
This makes `CommandBufferAllocator` object-safe, allowing us to get rid of the type parameter on the command buffer [builder] types. This will be useful for when we remove the `PrimaryCommandBufferAbstract` and `SecondaryCommandBufferAbstract` traits.

Since I needed to remove the associated types of `CommandBufferAllocator`, one of which being the type of iterator, I went ahead and made `CommandBufferAllocator::allocate` allocate a single command buffer at a time instead. This makes it consistent with `DescriptorSetAllocator`, and vulkano never allocates more than one command buffer at a time itself anyway. As for the distinction between a regular allocation and builder allocation, that's gone as well and all builders are now `!Send` while all finished command buffers are `Send + Sync`.

Changelog:
```markdown
### Breaking changes
Changes to command buffers:
- `AutoCommandBufferBuilder` and `UnsafeCommandBufferBuilder` now take an `Arc<dyn CommandBufferAllocator>` on construction.
- `AutoCommandBufferBuilder`, `PrimaryAutoCommandBuffer`, `SecondaryAutoCommandBuffer`, `UnsafeCommandBufferBuilder` and `UnsafeCommandBuffer` no longer have a type parameter for the type of allocator.

Changes to command buffer allocation:
- `CommandBufferAllocator` no longer has any associated types in order to make the trait object-safe.
- There is now only the single `CommandBufferAlloc` type to represent allocated command buffers. The `CommandBufferAlloc` and `CommandBufferBuilderAlloc traits and `StandardCommandBufferAlloc` and `StandardCommandBufferBuilderAlloc` types were removed.
- `CommandBufferAllocator::allocate` now returns a single `CommandBufferAlloc` on success and `Validated<VulkanError>` on failure.
- `CommandBufferAllocator` now has a required method `deallocate`.
```

### DIFF
--- a/vulkano/src/command_buffer/sys.rs
+++ b/vulkano/src/command_buffer/sys.rs
@@ -301,6 +301,12 @@ pub struct UnsafeCommandBuffer {
     inner: UnsafeCommandBufferBuilder,
 }
 
+// `UnsafeCommandBufferBuilder` is `!Send + !Sync` so that the implementation of
+// `CommandBufferAllocator::allocate` can assume that a command buffer in the recording state
+// doesn't leave the thread it was allocated on. However, as the safety contract states,
+// `CommandBufferAllocator::deallocate` must acccount for the possibility that a command buffer is
+// moved between threads after the recording is finished, and thus deallocated from another thread.
+// That's why this is sound.
 unsafe impl Send for UnsafeCommandBuffer {}
 unsafe impl Sync for UnsafeCommandBuffer {}
 


vulkano/src/command_buffer/auto/builder.rs
```diff
@@ -203,11 +182,11 @@ where
 
     #[inline]
     unsafe fn begin_unchecked(
-        allocator: &A,
+        allocator: Arc<dyn CommandBufferAllocator>,
         queue_family_index: u32,
         level: CommandBufferLevel,
         begin_info: CommandBufferBeginInfo,
-    ) -> Result<Self, VulkanError> {
+    ) -> Result<Self, Validated<VulkanError>> {
```
Rua, line 189: An `_unchecked` function shouldn't be returning validation errors I think?

vulkano/src/command_buffer/sys.rs
```diff
@@ -314,65 +297,51 @@ impl CommandBufferBeginInfo {
...
-    usage: CommandBufferUsage,
+pub struct UnsafeCommandBuffer {
+    inner: UnsafeCommandBufferBuilder,
 }
 
-impl<A> UnsafeCommandBuffer<A>
-where
-    A: CommandBufferAllocator,
-{
+unsafe impl Send for UnsafeCommandBuffer {}
```
Rua, line 304: I think this could do with some # Safety documentation, to explain why this is ok to do even if `CommandBufferAlloc` doesn't have these traits.

vulkano/src/command_buffer/auto/builder.rs
Rua: Hmmmmm...

Should there be an `allocate_unchecked`?

Rua: Ok, I don't think that matters for this PR, but maybe food for a future PR.

PR #2404, opened 2023-11-12
Descriptor set revamp

### LINKED ISSUE #2359: Allow updating descriptor sets after creation
This is a long-standing missing feature. Currently, descriptor sets can only be assigned contents when they are created. Updating them after creation is not possible. Adding this ability would allow better reuse of descriptor sets, especially if the user knows they only need to update a few things instead of everything.

Some things will be needed to make this possible:
- Descriptor sets will need state tracking and locking, similar to what buffers and images currently have. Only host writes and device reads need to be tracked.
- Recording a descriptor set into a command buffer may count as a read, depending on the `UPDATE_AFTER_BIND` and `UPDATE_UNUSED_WHILE_PENDING` flags.
- This state tracking will need to be granular on the individual descriptor level for `UPDATE_AFTER_BIND` and `UPDATE_UNUSED_WHILE_PENDING`, on the level of the whole descriptor set otherwise.

marc0246:
Following up on #2400, this removes the `DescriptorSet` trait, and adds the possibility to update an existing descriptor set, which fixes #2359.

Note that the updating of descriptor sets here is not at all what the issue described. That's because that approach is a waste of time, please don't attempt it. `UPDATE_AFTER_BIND` and `UPDATE_UNUSED_WHILE_PENDING` are flags that are used exclusively for bindless, where dynamic descriptor indexing is at play. There's no way to track which descriptors the shader does and doesn't access. On the other hand, in the bindful model, one overwrites entire bindings or even entire sets, where tracking individual descriptors would be overkill as well. Not only that, but in general we need *less* state sracking, not more! The way to make descriptor updates safe(ish) is to abstract descriptors away.

This PR instead introduces unsafe ways to update descriptors, as well as a safe way by virtue of a mutable reference to the descriptor set. This can still be used for bindless, so long as you double-buffer your descriptor sets. This has the downside that updating a descriptor set is duplicated, so 2X more writes to the descriptor sets overall. The updates should be small to non-existant in the first place so that shouldn't matter, though, unlike the more concerning issue of duplicating your descriptor memory. But even that should be relatively OK, what I'm more concerned about is that cache locality between frames is lessened this way. But as I say, the real approach is to abstract away descriptors entirely for this use case.

While making breaking changes to descriptors, I took the opportunity to rename `UnsafeDescriptorSet`.

Changelog:
```markdown
### Breaking changes
Changes to descriptor sets:
- There is now only the single type `DescriptorSet` to represent descriptor sets. The `DescriptorSet` trait and `PersistentDescriptorSet` type were removed.
- Renamed `UnsafeDescriptorSet` to `RawDescriptorSet`.

### Additions
- Ability to update existing descriptor sets.
````

### CODE REVIEW

File: vulkano/src/descriptor_set/mod.rs [hunk 1]
```diff
@@ -103,67 +103,225 @@ use std::{
...
 
     /// Returns the resources bound to this descriptor set.
-    fn resources(&self) -> &DescriptorSetResources;
+    #[inline]
+    pub fn resources(&self) -> RwLockReadGuard<'_, DescriptorSetResources> {
+        self.resources.read()
+    }
+
+    /// Updates the descriptor set with new values.
+    pub fn update(
```
Rua, line 190: This is not safe, because an `UnsafeCommandBuffer` could be holding a reference to the descriptor set internally. Updating a descriptor set when a command buffer holds a reference will invalidate the command buffer and make it UB to submit it.

Rua: But `UnsafeCommandBuffer` doesn't keep any of its resources alive, so that doesn't ensure safety here.

Rua: Because this unsafety works at a distance, and may catch the unwary. Moreover, it's not documented anywhere, not even in the safety documentation of `UnsafeCommandBufferBuilder` and `UnsafeCommandBuffer` (which, if you ask me, should be moved to the parent `sys` module).

Rua: Sorry, maybe I didn't explain it well. Imagine this scenario:
- User creates a descriptor set, which is now the only reference to the `Arc`.
- User records a command that uses it, into an `UnsafeCommandBuffer`. Because it doesn't hold a copy of the `Arc`, and the user is required to keep resources alive, there is still exactly one reference: the one the user holds.
- User calls `.get_mut().unwrap().update(...)` on the reference they hold, which succeeds because there are no other references to the `Arc`.
- User submits the `UnsafeCommandBuffer`, which is now invalid.

Rua: Fair enough. I will update the safety docs in a future PR, when I make `UnsafeCommandBuffer` usable again. That should be possible now that the type parameter is gone.

PR #2412, opened 2023-11-19
Update Surface documentation

ilyas-taouaou:
vulkano-win is deprecated, and I recommend using rwh feature when raw-window-handle compat issues arise

### DIFF
--- a/vulkano/src/lib.rs
+++ b/vulkano/src/lib.rs
@@ -15,8 +15,10 @@
 //!
 //! 3. If you intend to show graphics to the user on a window or a screen, create a [`Surface`].
 //!    A `Surface` is created from a window identifier or handle, that is specific to the display or
-//!    windowing system being used. The [`vulkano-win`] crate, which is part of the Vulkano
-//!    project, can make this step easier.
+//!    windowing system being used. You just need to install `raw-window-handle` crate too and then
+//!    use your favorite windowing crate to get the raw window handles.
+//!    Note: there is multiple versions of `raw-window-handle` which can be incompatible.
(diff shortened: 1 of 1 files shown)


Rua:
I don't think the `raw_window_handle` crate needs to be installed in your own project, you can just call `Surface::from_window`.

vulkano/src/lib.rs
```diff
@@ -15,8 +15,10 @@
 //!
 //! 3. If you intend to show graphics to the user on a window or a screen, create a [`Surface`].
 //!    A `Surface` is created from a window identifier or handle, that is specific to the display or
-//!    windowing system being used. The [`vulkano-win`] crate, which is part of the Vulkano
-//!    project, can make this step easier.
+//!    windowing system being used. You just need to install `raw-window-handle` crate too and then
+//!    use your favorite windowing crate to get the raw window handles.
+//!    Note: there is multiple versions of `raw-window-handle` which can be incompatible.
+//!    if you're using winit, you may want to use `rwh-X` feature to fix compatibility issues.
```
Rua, line 21: I can't find any feature in winit called `rwh-X`.

ilyas-taouaou: I have clarified in the new pushed version, let me know if it's better

freddycansic:
On the current version of Vulkano (0.34.1) I have to use the `rwh_05` on `winit` so that `EventLoop<T>` implements `HasRawDisplayHandle` (deprecated) instead of `HasDisplayHandle` since the current `Surface::required_extensions` function takes in a type implementing `HasRawDisplayHandle`.

So should the docs say that
> Currently, vulkano is compatible with `rwh_05`.

?

PR #2413, opened 2023-11-21
Support creating images that are not backed by vulkano managed memory

yshui:

# Description

  This need arose from OpenXR interoperation. We need to render into images returned from an OpenXR swapchain. The API doesn't specify what kind of memory will be backing those images, and AFAICT there's no vulkan API that can recover this information. Therefore the only option I can see is adding an opaque "external" image memory type, which is what I did here.

# Changelog

```markdown
### Additions
- Support creating images that are not backed by vulkano managed memory

````


Rua:
Looks good so far! I think it could work pretty much the same way for buffers?

yshui:
@rua should i make a PR for buffers as well?

Rua:
You can add it to this PR, that's probably easier?

yshui:
Added, and rebased.

Rua:
Your most recent commit makes me realise that this is not a problem particular to images, but to any object created from a handle. Vulkano will assume responsibility for destroying the object, and there's no way to stop it. Your current solution is not safe, because if the object is dropped before you take ownership, the handle will be destroyed. Vulkano should not be taking ownership of the handle in the first place, if it doesn't actually have ownership.

The only safe solution is to specify *when creating the object* that it should not be destroyed. This could be achieved by adding a `destroy_on_drop: bool` parameter to all the `from_handle` functions that currently exist in Vulkano, adding this value to the object in question, and then checking for it in `drop`. But this is better left for a separate PR.

yshui:
> This could be achieved by adding a take_ownership: bool parameter

Make senses, I can do this for images and buffers in this PR at least.

> Your current solution is not safe

I get what you are saying, but these functions are marked `unsafe` anyways

yshui:
hmm, i think this makes `bind_memory` unsafe.

yshui:
OK, I marked `bind_memory` unsafe.

yshui:
@marc0246 ah i forgot to change those, sorry!

PR #2414, opened 2023-11-21
Some shader parse refactoring and additions

Rua:
This is needed for some future changes I'm working on. `Spirv` now returns slices of instructions directly, instead of iterators, and the reflected SPIR-V internal types receive a few additional items.


Rua:
None.

Rua:
Yeah but `Spirv` is more of a Vulkano-internal thing anyway. It's used for Vulkano's own reflection, and is only public because Vulkano-shaders needs it.

PR #2421, opened 2023-12-07
Rename command buffer types

marc0246:
Reason being, that the "builder" pattern is what we call a helper that constructs some kind of object, most commonly field by field, using method calls. What these types are doing doesn't seem to align itself with that which is kind of confusing, instead they are recording commands to an underlying command buffer, not building some Rust struct. Furthermore, `UnsafeCommandBuffer` was renamed for consistency with the other raw types and because it's no longer completely unsafe.

Changelog:
```markdown
### Breaking changes
Changes to command buffers:
- Renamed `AutoCommandBufferBuilder` to `RecordingCommandBuffer` and `AutoCommandBufferBuilder::build` to `end`.
- Renamed `UnsafeCommandBufferBuilder` to `RawRecordingCommandBuffer` and `UnsafeCommandBufferBuilder::build` to `end`.
- Renamed `UnsafeCommandBuffer` to `RawCommandBuffer`.
````


Rua:
Why `CommandRecorder` rather than `CommandBufferRecorder`?

Rua:
True, but maybe also less clear to the user.

Also, the Vulkan function at the end of recording is called `vkEndCommandBuffer`, so maybe `end` rather than `finish` is more appropriate as a name.

Rua:
I think it's still helpful if the name indicates that there is a command buffer object underlying it. But that only applies if there is actually such an object, which I realise may not be the case anymore in the task graph. For the plain version that lets you record command buffers directly without any restrictions, I do think that `CommandBuffer` should be in the name.

Rua:
I'm not sure what you mean, what wrappers?

PR #2424, opened 2023-12-12
Fix StandardCommandBufferAllocator deallocation in release mode

lanquemar:
Description:

When an example is run in release build mode, `StandardCommandBufferAllocator` does not recycle its entries and ends up running out of memory:
```bash
$ cargo run --bin teapot --release
...
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', vulkano/src/command_buffer/allocator.rs:440:41
```
This is because the code line pushing back entries to the pools is within a `debug_assert!()` macro that is disabled in release mode. This PR replaces the `debug_assert!()` macro call to a simple `assert!()` so the behavior is the same but works for both debug and release modes.

Changelog:
```markdown
### Bugs fixed
- `StandardCommandBufferAllocator` does not run out of memory anymore when binary is built in release mode.
````


lanquemar:
Updated!

PR #2425, opened 2023-12-13
Merge `PrimaryAutoCommandBuffer` and `SecondaryAutoCommandBuffer` into `CommandBuffer`

marc0246:
Changelog:
```markdown
### Breaking changes
Changes to command buffers:
- Merged `{Primary,Secondary}AutoCommandBuffer` into `CommandBuffer` and the corresponding `RecordingCommandBuffer::{primary,secondary}` into `new`.
- `RecordingCommandBuffer` no longer has a type parameter for the command buffer level.

### Bugs fixed
- `RawRecordingCommandBuffer::end` being safe to call.
````

### DIFF
--- a/vulkano/src/command_buffer/sys.rs
+++ b/vulkano/src/command_buffer/sys.rs
@@ -204,12 +204,10 @@ impl RawRecordingCommandBuffer {
     /// Ends the recording, returning a command buffer which can be submitted.
     #[inline]
     pub unsafe fn end(self) -> Result<RawCommandBuffer, VulkanError> {
-        unsafe {
-            let fns = self.device().fns();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: vulkano/src/command_buffer/sys.rs [hunk 1]
```diff
@@ -169,7 +203,7 @@ impl RawRecordingCommandBuffer {
 
     /// Ends the recording, returning a command buffer which can be submitted.
     #[inline]
-    pub fn end(self) -> Result<RawCommandBuffer, VulkanError> {
+    pub unsafe fn end(self) -> Result<RawCommandBuffer, VulkanError> {
         unsafe {
```
Rua, line 207: This unsafe block is redundant now.

PR #2428, opened 2023-12-19
Improve clarity of `MemoryTypeFilter` docs

marc0246:
Hopefully this helps newcommers to not keep missing the relevant information. I moved the information about needing to combine e.g. `PREFER_HOST` with either `HOST_SEQUENTIAL_WRITE` or `HOST_RANDOM_ACCESS` to the top of the constant's docs and made it bold, whereas it was at the bottom before. And the corresponding constants have the same information duplicated at the top as well now.

### DIFF
--- a/vulkano/src/memory/allocator/mod.rs
+++ b/vulkano/src/memory/allocator/mod.rs
@@ -367,6 +367,8 @@ impl Debug for dyn MemoryAllocator {
 ///
 /// # Examples
 ///
+/// #### Direct device access only
+///
 /// For resources that the device frequently accesses, e.g. textures, render targets, or
 /// intermediary buffers, you want device-local memory without any host access:
 ///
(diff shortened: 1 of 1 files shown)


vulkano/src/memory/allocator/mod.rs
```diff
@@ -396,6 +398,8 @@ impl Debug for dyn MemoryAllocator {
 /// .unwrap();
 /// ```
 ///
+/// #### Sequential writes from host, indirect device access
+///
 /// For staging, the resource is only ever written to sequentially. Also, since the device will
 /// only read the staging resourse once, it would yield no benefit to place it in device-local
```
Rua, line 404:
```suggestion
/// only read the staging resource once, it would yield no benefit to place it in device-local
```

Typo that was there before, might as well fix it now.

PR #2433, opened 2023-12-26
Add support for the `ext_mesh_shader` extension

Rua:
Changelog:
```markdown
### Additions
- Support for the `ext_mesh_shader` extension.
````

This adds full mesh shader support for the EXT mesh shader extension. If we do want to support the NV version, that could be a little trickier, because it has its own `draw_mesh_tasks*` commands, which must be matched with the type of mesh and task shader. In other words, calling the EXT command requires EXT mesh shaders, and calling the NV command requires NV mesh shaders. Right now, the commands at least check to make sure you're using EXT mesh shaders. Other than the commands, NV mesh shaders are already handled properly so the commands are the only thing left, if we want to do it.

One item is still missing: `VK_QUERY_TYPE_MESH_PRIMITIVES_GENERATED_EXT`. I think this is better left to a separate PR because there is a non-mesh-shading equivalent to this as well that we're still missing, and they are better handled together.

I haven't added an example that uses mesh shaders, as I don't really fully know how they're supposed to be used. This is something we definitely want. Perhaps a mesh-shaderfied version of the triangle example to demonstrate the differences, and then a more elaborate example using task shaders as well?

### DIFF
--- a/vulkano/src/command_buffer/commands/pipeline.rs
+++ b/vulkano/src/command_buffer/commands/pipeline.rs
@@ -7,7 +7,7 @@ use crate::{
         auto::{RenderPassState, RenderPassStateType, Resource, ResourceUseRef2},
         sys::RawRecordingCommandBuffer,
         DispatchIndirectCommand, DrawIndexedIndirectCommand, DrawIndirectCommand,
-        RecordingCommandBuffer, ResourceInCommand, SubpassContents,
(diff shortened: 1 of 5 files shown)


Fuzzyzilla:
Seems to work great! Got a Task (just dispatches 4 meshes) + Mesh that outputs a couple rectangles. Yippee!

It required digging into the pipeline validation logic to see that when using Mesh shaders, input state and assembly state must be `None.` I think the documentation saying those fields must always be `Some` needs updating :D

PR #2437, opened 2023-12-28
Mesh shading example

Firestar99:

Changelog:
```markdown
### Additions
- A `mesh-shader` example.
````

### CODE REVIEW

**Comment by Firestar99** [comment]
Sure, I've squashed Rua's PR. Otherwise it was already rebased on Rua's PR, which was already rebased on current master.

**Comment by Rua** [comment]
But now this PR still contains the same changes as my PR, which makes reviewing it more difficult, as it's hard to separate your changes from mine.

**Comment by Firestar99** [comment]
Ohhhh.. my remote for vulkano didn't update for some reason, I didn't notice you've merged ext_mesh_shaders already

**Comment by Firestar99** [comment]
That's already how my setup works, it's just that Intellij Idea's "update project" button seems to only fetch the origin of the checked out branch, not all branches...

**Comment by Firestar99** [comment]
I'm not submitting this PR because it is ready, but because there is a weird struct layout issue here.

https://github.com/Firestar99/vulkano/blob/032f77709dcce49edd3fd487838c1473fbc5698a/examples/mesh-shader/shader.mesh#L47

First turn on LOAD_FROM_INSTANCE_BUFFER, which it is on the current latest commit, to use instance data loaded from buffers instead of generating all the geometry dynamically in the mesh shader.

https://github.com/vulkano-rs/vulkano/blob/032f77709dcce49edd3fd487838c1473fbc5698a/examples/mesh-shader/shader.mesh#L18-L22

On the GPU side this Instance struct uses STD140 layout, which causes it's alignment to be 16 bytes, but on the CPU it's aligned to 12 bytes. In the main.rs the InstanceData struct is currently defined manually, but even if I define it as

// also move the shaders here
type InstanceData = mesh::Instance;

it still is misalignend, producing garbage output. Manually using repr(C, alignn(16)) also does not help.

**Comment by Firestar99** [comment]
Wouldn't it be more consistent if we've used glslang's file endings?
```
$ glslang --help
[..]
'file' can end in .<stage> for auto-stage classification, where <stage> is:
    .conf   to provide a config file that replaces the default configuration
            (see -c option below for generating a template)
    .vert   for a vertex shader
    .tesc   for a tessellation control shader
    .tese   for a tessellation evaluation shader
    .geom   for a geometry shader
    .frag   for a fragment shader
    .comp   for a compute shader
    .mesh   for a mesh shader
    .task   for a task shader
    .rgen    for a ray generation shader
    .rint    for a ray intersection shader
    .rahit   for a ray any hit shader
    .rchit   for a ray closest hit shader
    .rmiss   for a ray miss shader
    .rcall   for a ray callable shader
    .glsl   for .vert.glsl, .tesc.glsl, ..., .comp.glsl compound suffixes
    .hlsl   for .vert.hlsl, .tesc.hlsl, ..., .comp.hlsl compound suffixes
``` The last two options could also be a good alternative.

File: examples/mesh-shader/shader.mesh [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+const uint n_instances = rows * cols;
+
+
+
+void main() {
+	vec2 position_offset;
+	float scale;
+
+	const bool LOAD_FROM_INSTANCE_BUFFER = true;
+	if (LOAD_FROM_INSTANCE_BUFFER) {
```
Firestar99, line 48: I'm also not that happy with this if being semi-hidden in the shader code. It's hopefully obvious what it does once you read the shader.

Rua: I suppose it could be made into a specialization constant?

Firestar99: In practice you'd make it a SpecConstant, but I feel like for an example it's overkill and only complicates it further. 
Instead I'd propose two possible solutions:
1. **uniform bool toggled with a keyboard press**: But that key would have to be known to the user, and the current state needs to be visualized. As we can't exactly render text easily, likely changing the color of the triangles.
2. **Keep it as is**: At the end of the day the if should switch between the two main use-cases of mesh shaders: reading vertex (& other) data manually and generating completely dynamic geometry. And those investigating mesh shaders will read the code and understand what the simple `const bool` does, and is likely less complicated than following where a uniform came from.

I'm currently leaning more on 2, especially with maybe some more docs.

Firestar99: I've implemented variant 2 as follows:

https://github.com/vulkano-rs/vulkano/blob/e89f419648cca71cffcd2c062325b99aaec9a7cf/examples/mesh-shader/mesh.glsl#L54-L61

https://github.com/vulkano-rs/vulkano/blob/e89f419648cca71cffcd2c062325b99aaec9a7cf/examples/mesh-shader/mesh.glsl#L67-L70

Firestar99:
I've also added an output variable to the mesh shader, simply passing the triangle color, so that pretty much everything you may want to do with mesh shaders is covered. Apart from task shaders, but those aren't super complicated to understand, and should easily be transferable from any other example or tutorial.

https://github.com/Firestar99/vulkano/blob/e89f419648cca71cffcd2c062325b99aaec9a7cf/examples/mesh-shader/mesh.glsl#L38-L39

PR #2448, opened 2024-01-05
vulkan-utils -> vulkan-tools README fix

HadrienG2:
There is no `vulkan-utils` package in current Ubuntu repos. But there used to be one in ubuntu 20.04, which was a compatibility alias for the `vulkan-tools` package that is still around. So I guess you meant `vulkan-tools`?


Rua:
Thanks!

PR #2460, opened 2024-02-01
Add runtime SPIR-V validation

Rua:
Changelog:
```markdown
### Additions
- Validation between shader code and device extensions, features and properties.
````

This adds validation for the VUIDs in the `RuntimeSpirv` section of the spec. Those VUIDs don't belong to any particular command, but in practice most of them apply during shader/pipeline creation. Not all of them could be validated, as some rely on knowing the inputs of particular commands; safety rules were added to the shader module for those instead.


maratik123:
Is it Ok, that such amount of code added without unit-testing validation?

AustinJ235 (approved):
👍

PR #2482, opened 2024-03-01
General shader improvements, specifically targeting rust-gpu

Firestar99:
This is a collection of a bunch of smaller shader-related bugs and issues I came across, including ones which only trigger when using rust-gpu. See changelog below, and these non-user facing or less unimportant changes:
* replaced `#[macro_use] extern crate XXX;` with normal imports
* added tests for shader! reflect() for both glsl and rust-gpu shaders
* fixed mesh shader output not being arrayed (as mesh shaders are not yet released, I wouldn't put it into the changelog)
* structs in shader in/out interface got an explanatory panic message that they are not yet supported, instead of "Type not found or invalid"

Changelog:
```markdown
### Breaking changes
Changes to `shader!` macro:
- Shaders included via `bytes: <path-to-spv>` **must** no longer specify a shader type, e.g. `ty: <vertex>`.

### Additions
Changes to `shader!` macro:
- added `generate_structs: true` option that may be used to disable rust structs from generating. Useful in e.g. rust-gpu contexts where such functionality is not needed.

### Bugs fixed
Changes to `shader!` macro:
- fixed shader struct names that are invalid rust idents from panicing the shader! macro. Rust-gpu emitted struct names such as `foo::bar::MyStruct` now work.
````

### CODE REVIEW

File: vulkano-shaders/src/codegen.rs [hunk 1]
```diff
@@ -634,4 +698,51 @@ mod tests {
...
+
+        let mut type_registry = TypeRegistry::default();
+        let (_shader_code, _structs) = reflect(
+            &MacroInput::empty(),
+            LitStr::new(
+                "descriptor_calculation_with_multiple_functions_shader",
+                Span::call_site(),
+            ),
+            String::new(),
+            &artifact.as_binary(),
```
Rua, line 714:
```suggestion
            artifact.as_binary(),
```

vulkano-shaders/src/structs.rs
```diff
@@ -685,7 +689,13 @@ impl TypeStruct {
             .names()
             .iter()
             .find_map(|instruction| match instruction {
-                Instruction::Name { name, .. } => Some(Ident::new(name, Span::call_site())),
+                Instruction::Name { name, .. } => {
+                    // rust-gpu uses fully qualified rust paths as names which contain `:`.
+                    // I sady don't know how to check against all kinds of invalid ident chars.
+                    let name = name.replace(':', "_");
```
Rua, line 695: How about something like this?
```rust
let mut name = name.replace(|c| !(c.is_ascii_alphanumeric() || c == '_'), "_");
if name.starts_with(char::is_ascii_numeric) {
    name.insert(0, '_');
}
```

stefnotch: Is this a general "convert any SpriV name to a Rust name" thing? Or is it fine if it sometimes breaks?

If it's intended to be a general one, then it's a bit more involved 
- Different names should map to different results. e.g. If the shader has `a_b` and `a:b`, then they should map to different Rust names.
- Rust has a bunch of reserved keywords

Basically what one would end up doing is implementing a name encoding scheme, where typical names don't get modified at all. And special characters would get encoded instead of just replaced. And names that would end up being Rust keywords get prefixed with something special.

Rua: It's not a perfect solution, but solving it perfectly without any name collisions is going to be quite difficult, and probably not really important for this PR.

stefnotch: For future PRs: Would a general solution be useful, or would it bring nearly nothing to the table?

Firestar99: I just want to point out that structs are already being deduplicated by their rust ident, see register_struct() and calls to it. I assume that is to allow two shaders to be declared within a single shader struct and only generate all structs once, allowing them to be shaded between the shaders. Thus currently, if you'd have a struct `bla_test` and `bla:test`, only one of them will get a rust struct generated and the other one "deduplicated", even though it may have a different definition.
And in case you do hit a keyword or some other invalid ident, the ident creation will Err and it will default to `UnnamedX` as if no name was present in the spv, whereas master just panics.
I personally don't see the value in investing too much effort in perfecting it. No sane developer should name their structs this similarly, and then they can often just change it, or worst case you can use rust's `type` to rename them manually. My goal was to just make it not panic on invalid idents and good enough to handle rust-gpu as a special case, though it really wouldn't need the struct generation at all.

Rua: Fair enough then.

Firestar99: I was gonna commit your suggestion before you merge :D

Firestar99: @Rua I was gonna commit your suggestion before you merge :D

Firestar99: @Rua there you go, your changes in another PR https://github.com/vulkano-rs/vulkano/pull/2484. I mean it'll only be a single commit on master, as you squash your commits.

vulkano-shaders/src/lib.rs
```diff
@@ -643,6 +656,10 @@ impl Parse for MacroInput {
                         ),
                     });
                 }
+                "generate_structs" => {
+                    let lit = input.parse::<LitBool>()?;
+                    generate_structs = lit.value;
```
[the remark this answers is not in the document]
Firestar99: https://github.com/vulkano-rs/vulkano/pull/2485

vulkano-shaders/src/structs.rs
stefnotch: @Firestar99 Lovely, thanks for the explanation. So there's a fallback when things are weird, which yep, should be good enough.

PR #2499, opened 2024-03-13
Add `Suballocator::suballocations`

marc0246:
This allows the user to display debug information about individual memory blocks internal to the allocator. `MemoryAllocator` isn't integrated into this yet. This unfortunately required me to change `Suballocator::allocate` and `Suballocator::deallocate` to take `&mut self`, which makes it a bit more inconvenient to work with in many cases. But alas, the alternative would have been to keep the internal mutability but change it to a `RefCell`, and I'm not a fan of internal synchronization if it can be helped. It just means that the user will have to put the `RefCell` externally if needed.

Changelog:
```markdown
### Breaking changes
Changes to memory allocation:
- `Suballocator::{allocate,deallocate}` now take `&mut self`.
- `Suballocator` has new required items `Suballocations` and `suballocations` for iterating over suballocations.
```

### DIFF
--- a/vulkano/src/memory/allocator/suballocator/buddy.rs
+++ b/vulkano/src/memory/allocator/suballocator/buddy.rs
@@ -1,4 +1,6 @@
-use super::{AllocationType, Region, Suballocation, Suballocator, SuballocatorError};
+use super::{
+    AllocationType, Region, Suballocation, SuballocationNode, Suballocator, SuballocatorError,
+};
 use crate::{
     memory::{
         allocator::{align_up, array_vec::ArrayVec, AllocationHandle, DeviceLayout},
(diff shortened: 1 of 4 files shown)


vulkano/src/memory/allocator/suballocator/buddy.rs
```diff
@@ -256,17 +255,14 @@ unsafe impl Suballocator for BuddyAllocator {
...
 
-#[derive(Debug)]
-struct BuddyAllocatorState {
-    // Every order has its own free-list for convenience, so that we don't have to traverse a tree.
-    // Each free-list is sorted by offset because we want to find the first-fit as this strategy
-    // minimizes external fragmentation.
-    free_list: ArrayVec<Vec<DeviceSize>, { BuddyAllocator::MAX_ORDERS }>,
+    #[inline]
+    fn suballocations(&self) -> Self::Suballocations<'_> {
+        todo!()
```
Rua, line 266: Is this intended?

PR #2509, opened 2024-03-29
Add details for error messages in validate_pipeline_descriptor_sets

ComfyFluffy:
Add details for `format` and `image_view_type` checks in `RecordingCommandBuffer::validate_pipeline_descriptor_sets` to make debugging easier.


Rua:
Thanks, I agree that this helps. Maybe the messages would be clearer still if the message said "pipeline requires {}, but image view type is {}"? Then there's no doubt on what pieces are referred to.

PR #2510, opened 2024-04-02
Upgrade to `ash 0.38`

MarijnS95:
https://github.com/ash-rs/ash/releases/tag/0.38.0

Code compiles, but there's still quite some more cleanup to do across autogenerated files.  A future change should more rigorously use the new builder pattern directly on Vulkan structs, so that the borrow checker can accurately track any lifetime and mutable-reference issues on all `PerXXX` structs that are now marked `'static`.

### TODO

- [x] Clean up unnecessary duplication in `autogen` scripts.
- [ ] Tidy up lifetime changes now that they're easy to spot in the diff; or do we postpone them to a separate PR?
- [ ] Update documentation to reflect any user-facing changes - in this repository.
- [ ] Make sure that the changes are covered by unit-tests.
- [x] Run `cargo clippy` on the changes.
- [x] Run `cargo +nightly fmt` on the changes.
- [ ] Please put changelog entries **in the description of this Pull Request**
- [ ] Describe in common words what is the purpose of this change, related

Changelog:
```markdown
### Public dependency updates
- [ash](https://crates.io/crates/ash) 0.38.0

### Breaking changes
Changes to `Surface`:
- Changed `from_win32()` argument types to take `HWND` and `HINSTANCE` as `isize`.

Renamed `PipelineStages::SUBPASS_SHADING` to PipelineStages::SUBPASS_SHADER`, following upstream Vulkan changes.
```


vulkano/src/sync/pipeline.rs
```diff
@@ -377,7 +377,8 @@ vulkan_bitflags_enum! {
     ]),
 
     /// Subpass shading shaders are executed.
-    SUBPASS_SHADING, SubpassShading = SUBPASS_SHADING_HUAWEI
+    // TODO: SUBPASS_SHADER?
+    SUBPASS_SHADING, SubpassShading = SUBPASS_SHADER_HUAWEI
```
MarijnS95, line 381:
```xml
                <enum bitpos="39" extends="VkPipelineStageFlagBits2"        name="VK_PIPELINE_STAGE_2_SUBPASS_SHADER_BIT_HUAWEI"/>
                <enum             extends="VkPipelineStageFlagBits2"        name="VK_PIPELINE_STAGE_2_SUBPASS_SHADING_BIT_HUAWEI" alias="VK_PIPELINE_STAGE_2_SUBPASS_SHADER_BIT_HUAWEI" deprecated="aliased"/>
```

Rua: Yeah this should be `SUBPASS_SHADER` now.

MarijnS95: Yeah the naming was fixed for pipeline stages, but nowhere else (probably because it makes less sense). Still unsure what to do about `subpass shading shader` :grimacing:

Rua: Apparently that's actually what Vulkan calls it!
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineStageFlagBits2.html#_description

Rua:
What still needs to be done for this PR?

vulkano/src/sync/pipeline.rs
Rua: I raised the question with the docs repo: https://github.com/KhronosGroup/Vulkan-Docs/issues/2343

MarijnS95:
I'm fairly certain to have written a TODO list via `gh pr create` when submitting this, but only the standard template entries remain (still need to be done anyway). Here are a few things:

- [ ] Clean up unnecessary duplication in `autogen` scripts;
- [ ] Tidy up lifetime changes now that they're easy to spot in the diff; or do we postpone them to a separate PR?

0xcaff:
> Regarding the change of `*mut c_void` to `HANDLE`, they don't seem to be the same type. In C it's a typedef to `void*`, which is equivalent to `*mut c_void` in Rust. Changing it to anything else would be a breaking change.

For what it is worth, raw-window-handle treated this as a breaking change. https://github.com/rust-windowing/raw-window-handle/commit/69d9eac14401e1a8868d14febe30583f9ea70a98#diff-ee9da70e8a36920d0c024abb7c25308cb435c3dcba46a3566abca2a11fd1cfffL36-R34

I think this is the right thing to do to maintain semantic versioning correctness.

Rua:
Sounds good.

MarijnS95:
Yes, we followed suit on the `isize` change.

vulkano/autogen/mod.rs
```diff
@@ -434,6 +437,30 @@ impl<'r> VkRegistryData<'r> {
...
+                            match c {
+                                Command::Alias { name, .. } => name.as_str(),
+                                Command::Definition(d) => d.proto.name.as_str(),
+                                _ => todo!(),
+                            },
+                            c,
+                        )
+                    }));
+                }
+                None
```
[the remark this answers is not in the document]
Rua: I plan to refactor a lot of the autogen stuff, so I will have a look at that as well then.

PR #2511, opened 2024-04-03
Handle more depends values in vk.xml

0xcaff:
## Overview

These changes are necessary to support the [vk.xml from the latest version of ash](https://github.com/KhronosGroup/Vulkan-Headers/blob/38899bf60605f0a2c50f7968e5c4c9826a824683/registry/vk.xml). To exercise these codepaths, cherry pick this commit https://github.com/0xcaff/vulkano/commit/f43f1188

* Adds a conjunctive normal form helper to express arbitrarily nested depends constrains. Can now handle strings like

  ```
  ((VK_KHR_get_physical_device_properties2,VK_VERSION_1_1)+VK_KHR_depth_stencil_resolve),VK_VERSION_1_2
  ```

* Add parsing for depends values like
  
  ```
  a+b,c
  ```

  This evaluates to

  ```
  (A AND B) OR C
  ```

  Previously, it would crash the parse_depends function because parens were missing and the grouping order was ambiguous.

## Merge Order
These changes add behavior which is used when the new vk.xml is used. This can be safely merged independently of https://github.com/vulkano-rs/vulkano/pull/2510 (either before or after).

I plan to make a PR on top of both these changes and https://github.com/vulkano-rs/vulkano/pull/2510 pulling in the new vk.xml.

## Details

1. It would be nice to have tests for parse_depends but given this is code in a buildscript, not sure how to set that up. Open to ideas here. Would like to do this in a separate PR as the priors lack test coverage.

2. This conjuctive normal form transformation is kinda complex. It is clever but not the most straightforward thing. I think it would be simpler to store the original boolean expression instead and evaluate against that. Left it as is as I don't completely understand the usages of `requires_all_of` enough to safely change them.

### DIFF
--- a/vulkano/autogen/extensions.rs
+++ b/vulkano/autogen/extensions.rs
@@ -1,4 +1,5 @@
 use super::{write_file, IndexMap, RequiresOneOf, VkRegistryData};
+use crate::autogen::conjunctive_normal_form::ConjunctiveNormalForm;
 use heck::ToSnakeCase;
 use nom::{
     branch::alt, bytes::complete::take_while1, character::complete, combinator::all_consuming,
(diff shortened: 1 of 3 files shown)


vulkano/autogen/extensions.rs
```diff
@@ -791,33 +830,15 @@ fn extensions_members(ty: &str, extensions: &IndexMap<&str, &Extension>) -> Vec<
             let requires_all_of = extension.depends.as_ref().map_or_else(Vec::new, |depends| {
                 let depends_panic = |err| -> ! {
                     panic!(
-                        "couldn't parse `depends=` attribute for extension `{}`: {}",
-                        extension.name, err
+                        "couldn't parse `depends={}` attribute for extension `{}`: {}",
```
Rua, line 833: Technically this should have " " around it because it's XML. Very silly nitpick though. :p

Rua:
Great work! Thank you very much!

vulkano/autogen/extensions.rs
0xcaff: changed it, just for you Rua :p

using the debug formatter which adds "" and escaping automatically

PR #2512, opened 2024-04-05
Handle arrays in autogen

0xcaff:
## Overview
In the vk.xml added in ash 0.38, the newly added `VkPhysicalDeviceHostImageCopyPropertiesEXT` struct has the following xml:

```
<type category="struct" name="VkPhysicalDeviceHostImageCopyPropertiesEXT" structextends="VkPhysicalDeviceProperties2">
    <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_IMAGE_COPY_PROPERTIES_EXT"><type>VkStructureType</type> <name>sType</name></member>
    <member optional="true"><type>void</type>*                                                        <name>pNext</name></member>
    <member optional="true" limittype="noauto"><type>uint32_t</type>                                  <name>copySrcLayoutCount</name></member>
    <member optional="true" limittype="noauto" len="copySrcLayoutCount"><type>VkImageLayout</type>*   <name>pCopySrcLayouts</name></member>
    <member optional="true" limittype="noauto"><type>uint32_t</type>                                  <name>copyDstLayoutCount</name></member>
    <member optional="true" limittype="noauto" len="copyDstLayoutCount"><type>VkImageLayout</type>*   <name>pCopyDstLayouts</name></member>
    <member optional="true" limittype="noauto"><type>uint8_t</type>                                   <name>optimalTilingLayoutUUID</name>[<enum>VK_UUID_SIZE</enum>]</member>
    <member limittype="bitmask"><type>VkBool32</type>                                                 <name>identicalMemoryTypeRequirements</name></member>
</type>
```

Currently, the `pCopySrcLayouts` and `pCopyDstLayouts` arrays get a type of `*mut T` (where T is defined by the `vulkano_type` function). This PR changes the autogen behavior to do the following.

1. Interpret the `*mut T` as a `&[T]` with a length specified by the `len` attribute.
2. Omit the field corresponding to the length attribute as its information is available in the array field.
3. Rename the field containing the array from `p_{field_name}` to `{field_name}`

## Merge Order
This is safe to merge today as this code adds functionality to handle the new vk.xml. The behavior for the current vk.xml will remain unchanged.

Related PRs:
* https://github.com/vulkano-rs/vulkano/pull/2511
* https://github.com/vulkano-rs/vulkano/pull/2510
* https://github.com/0xcaff/vulkano/pull/1

## Reviewing
I've broken up the changes int logical commits which have straightforward diffs to make it a bit easier to review.

### DIFF
--- a/vulkano/autogen/properties.rs
+++ b/vulkano/autogen/properties.rs
@@ -440,41 +440,46 @@ fn members(ty: &Type) -> Vec<Member<'_>> {
         ))(input)
     }
 
-    if let TypeSpec::Members(members) = &ty.spec {
-        members
-            .iter()
-            .filter_map(|member| {
-                if let TypeMember::Definition(def) = member {
-                    let name = def.markup.iter().find_map(|markup| match markup {
-                        TypeMemberMarkup::Name(name) => Some(name.as_str()),
-                        _ => None,
-                    });
-                    let ty = def.markup.iter().find_map(|markup| match markup {
-                        TypeMemberMarkup::Type(ty) => Some(ty.as_str()),
-                        _ => None,
-                    });
-                    let len = def
-                        .markup
-                        .iter()
-                        .find_map(|markup| match markup {
-                            TypeMemberMarkup::Enum(len) => Some(len.as_str()),
-                            _ => None,
-                        })
-                        .or_else(|| array_len(&def.code).map(|(_, len)| len).ok());
-                    if name != Some("sType") && name != Some("pNext") {
-                        return name.map(|name| Member {
-                            name,
-                            ty: ty.unwrap(),
-                            len,
-                        });
-                    }
-                }
-                None
+    let TypeSpec::Members(members) = &ty.spec else {
+        return vec![];
+    };
+
+    members
+        .iter()
+        .filter_map(|member| {
+            let TypeMember::Definition(def) = member else {
+                return None;
+            };
+
+            let name = def.markup.iter().find_map(|markup| match markup {
+                TypeMemberMarkup::Name(name) => Some(name.as_str()),
+                _ => None,
+            });
+            let ty = def.markup.iter().find_map(|markup| match markup {
+                TypeMemberMarkup::Type(ty) => Some(ty.as_str()),
+                _ => None,
+            });
+
+            let len = def
+                .markup
+                .iter()
+                .find_map(|markup| match markup {
+                    TypeMemberMarkup::Enum(len) => Some(len.as_str()),
+                    _ => None,
+                })
+                .or_else(|| array_len(&def.code).map(|(_, len)| len).ok());
+
+            if name == Some("sType") || name == Some("pNext") {
+                return None;
+            }
+
+            name.map(|name| Member {
+                name,
+                ty: ty.unwrap(),
+                len,
             })
-            .collect()
-    } else {
-        vec![]
-    }
+        })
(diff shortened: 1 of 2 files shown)


vulkano/autogen/properties.rs
```diff
@@ -80,8 +87,23 @@ fn properties_output(members: &[PropertiesMember]) -> TokenStream {
...
-                let ffi_members = ffi_members.iter().map(|(ffi_member, ffi_member_field)| {
-                    quote! { properties_ffi.#ffi_member.map(|s| s #ffi_member_field .#ffi_name) }
+                let ffi_members = ffi_members.iter().map(|FFIMember { ident: ffi_member, tokens: ffi_member_field, len_field_name }| {
+                    if let Some(len_field_name) = len_field_name {
+                        let len_field_name = Ident::new(len_field_name.as_str(), Span::call_site());
+
+                        quote! {
+                            properties_ffi.#ffi_member.map(|s|
+                                unsafe {
+                                    std::slice::from_raw_parts(
```
0xcaff, line 97: Maybe this should be a Vec::from_raw_parts as the ownership seems like it is transferred? https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts

0xcaff: Unlike other struct members, the values in arrays are not owned. Changed the blanket impl for FromVulkan to use references https://github.com/vulkano-rs/vulkano/pull/2512/commits/603aa84f7c4fc2d9f84adc019349d1020c5c4784

Rua: I don't think `from_raw_parts` is correct, because its documentation says the pointer must have been allocated by the Rust allocator. The data is not owned by Rust either; the returned pointer still references data inside the Vulkan implementation. Presumably, the lifetime of the pointer is only as long as the physical device exists.

vulkano/src/device/properties.rs
```diff
@@ -251,11 +251,11 @@ impl FromVulkan<ash::vk::SubgroupFeatureFlags> for SubgroupFeatures {
     }
 }
 
-impl<U: FromVulkan<T>, T: Clone> FromVulkan<&[T]> for Vec<U> {
+impl<U: for<'a> FromVulkan<&'a T>, T: Clone> FromVulkan<&[T]> for Vec<U> {
```
Rua, line 254: The `T: Clone` bound probably isn't needed anymore?

0xcaff: Yes, good catch

Rua:
I think, the way the trait is implemented now, it's not going to work. It expects the base type to implement `FromVulkan` on the reference, but all the existing `FromVulkan` implementations are on the base type directly. So it may be necessary to include the `clone` call after all.

0xcaff:
I was thinking of having a blanket impl for references which implement Copy.

```
impl <U: FromVulkan<T>, T: Copy> FromVulkan<&T> for U {
    fn from_vulkan(val: &T) -> Option<Self> {
        U::from_vulkan(*val)
    }
}
``` 

I was thinking of putting it in the next PR, given your question I've decided to add it to this one.

Rua:
Since this will only be dealing with Vulkan/C types, all of them are expected to be `Copy` anyway.

0xcaff:
I see, nice!

Anything else you want changed in this PR before it's ready to merge?

PR #2514, opened 2024-04-08
add memoryMapPlaced and memoryMapRangePlaced features of VK_EXT_map_memory_placed extension

0xcaff:
## Overview

Implements the memoryMapPlaced and memoryMapRangePlaced features of the `VK_EXT_map_memory_placed` extension ([docs](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_map_memory_placed.html), [proposal](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/proposals/VK_EXT_map_memory_placed.adoc)).

Adds a field, `MemoryMapInfo.placed_address` for use with `DeviceMemory::map` and `DeviceMemory::map_unchecked` specifying an address.

## Testing

As `VK_EXT_map_memory_placed` is currently not yet released in any implementation, I've tested this using a trunk revision of mesa (https://gitlab.freedesktop.org/mesa/mesa/-/commit/6b383ca810bb853f11c127d9e1d284dc4dcb0992). Mesa 24.1 coming out in the beginning of May is scheduled to include this extension.

Here are the instructions I used to run the tests locally. These will not make global system changes.

```
$ git clone git@gitlab.freedesktop.org:mesa/mesa.git
$ cd mesa
$ mkdir builddir
$ meson setup builddir/
$ meson compile -C builddir/
$ meson devenv -C builddir
[mesa] $ vkinfo # validate extension available on host
[mesa] $ cd ~/projects/vulkano
[mesa] $ cargo test
```

## Questions
* Do we care about the tests running in CI? It seems like currently `gfx_dev_and_queue` will early return, silently passing if a feature or extension requested is not available.
* <strike>Would adding a field to a struct (in this case `MemoryMapInfo.placed_address`) be considered a breaking change? Wasn't sure so I erred on the conservative side and am assuming yes.</strike> The non-exhaustive allows for non-breaking change.
* <strike>Am I using the right vuids? See the inline comments for specifics.</strike> Discussing in inline comments.

## Changelog
```markdown
### Additions
* Support for the `ext_map_memory_placed` extension.
```


vulkano/src/memory/device_memory.rs
```diff
@@ -1427,6 +1456,94 @@ impl MemoryMapInfo {
...
+
+            let features = memory.device.enabled_features();
+            if !features.memory_map_placed {
+                return Err(Box::new(ValidationError {
+                    context: "placed_address".into(),
+                    problem: "is not empty".into(),
+                    requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
+                        "memory_map_placed",
+                    )])]),
+                    vuids: &["VUID-VkMemoryMapInfoKHR-flags-09569"],
```
0xcaff, line 1479: Is this the right vuid to use here? https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkMemoryMapInfoKHR.html#VUID-VkMemoryMapInfoKHR-flags-09569

Its not exactly covering the same condition. Should I leave the VUID out instead?

vulkano/src/memory/device_memory.rs
```diff
@@ -1427,6 +1456,93 @@ impl MemoryMapInfo {
...
+                }
+
+                if size != DeviceSize::MAX {
+                    return Err(Box::new(ValidationError {
+                        context: "size".into(),
+                        problem: "must be VK_WHOLE_SIZE".into(),
+                        vuids: &["VUID-VkMemoryMapInfoKHR-flags-09572"],
+                        ..Default::default()
+                    }));
+                }
```
0xcaff, line 1542: This validation contradicts the previous validation

https://github.com/vulkano-rs/vulkano/blob/b27c9b5d2ad31200984cf0fc3af7f80eb4646ac4/vulkano/src/memory/device_memory.rs#L1449-L1457

How should this be resolved?

0xcaff: https://github.com/vulkano-rs/vulkano/pull/2514/commits/f3c33fbbeface3215ef08721e0a486e7c6bd7468

vulkano/src/memory/device_memory.rs
```diff
@@ -2271,4 +2390,64 @@ mod tests {
...
+                panic!("failed to map memory")
+            }
+
+            address
+        };
+
+        memory
+            .map(MemoryMapInfo {
+                offset: 0,
+                size: DeviceSize::MAX,
```
0xcaff, line 2447: I'm not sure how I feel about exposing this to callers, maybe worth wrapping it?

0xcaff: `DeviceSize::MAX`/`VK_WHOLE_SIZE` is the only valid parameter when `features.memory_map_range_placed` is not available. I've updated the size argument to be translated to `DeviceSize::MAX` when `features.memory_map_range_placed` is not available and the size is equal to the full allocation size to ensure full allocations can be mapped without changing the code, regardless of whether `features.memory_map_range_placed` is available.

vulkano/src/memory/device_memory.rs
```diff
@@ -463,12 +463,10 @@ impl DeviceMemory {
         let MemoryMapInfo {
             offset,
             size,
+            placed_address,
             _ne: _,
         } = map_info;
 
-        // Sanity check: this would lead to UB when calculating pointer offsets.
-        assert!(size <= isize::MAX.try_into().unwrap());
-
```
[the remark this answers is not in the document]
0xcaff: I see, I put it back.

0xcaff:
I've updated the interface to no longer use `DeviceSize::MAX`

vulkano/src/memory/device_memory.rs
Rua: It looks like the right VUID to me. What are you unsure about?

vulkano/src/memory/device_memory.rs
```diff
@@ -1388,6 +1419,18 @@ pub struct MemoryMapInfo {
...
+    ///
+    /// Requires [`DeviceExtensions::ext_map_memory_placed`] and
+    /// [`DeviceFeatures::memory_map_placed`] to be enabled.
+    ///
+    /// Must align with [`DeviceProperties::min_placed_memory_map_alignment`].
+    ///
+    /// [`DeviceExtensions::ext_map_memory_placed`]: crate::device::DeviceExtensions::ext_map_memory_placed
+    /// [`DeviceFeatures::memory_map_placed`]: crate::device::DeviceFeatures::memory_map_placed
+    /// [`DeviceProperties::min_placed_memory_map_alignment`]: crate::device::DeviceProperties::min_placed_memory_map_alignment
+    pub placed_address: Option<*mut c_void>,
```
Rua, line 1432:
```suggestion
    pub placed_address: Option<NonNull<c_void>>,
```

The pointer can't be null, and this allows using the niche for the `Option`.

vulkano/src/memory/device_memory.rs
Rua: If it's redundant, we don't need the flag in Vulkano, as long as we add it before calling Vulkan. Which is what's done here.

0xcaff: Yes, exactly. The VUID here is targeting the flag while the error check is regarding whether `placed_address` is provided or not. The presence of `placed_address` in vulkan implies the flag.

I was having a hard time coming up with an API which doesn't have the implicit behavior. I guess we could make a new fn with its own set of arguments instead? Let me know if this is a dealbreaker.

vulkano/src/memory/device_memory.rs
0xcaff: Please do link the issue when you create it!

Without `features.memory_map_range_placed`, the only valid range for mappings is the entire allocation. I'm curious, what would you expect to use to specify the entire allocation?
* 0 would conflict with other vulkan validations.
* Allocation size requires the caller to keep track of the allocation size.
* Maybe you're thinking omitting the fields, though I think that would force callers who want to allocate the full range to change their args based on whether the feature is available or not.

vulkano/src/memory/device_memory.rs
0xcaff: Good idea

0xcaff:
I've updated the PR but seems like the windows actions has crashed and its blocking the check from completing.

vulkano/src/memory/device_memory.rs
Rua: I'm not that attached to it, and you're right that following Vulkan to the letter makes Vulkano more robust against future additions.

vulkano/src/memory/device_memory.rs
0xcaff: I see

vulkano/src/memory/device_memory.rs
0xcaff: Great, I will update this in the next few hours.

0xcaff: I've updated the PR. One more possible interface to consider is instead of adding fields to `MemoryMapInfo` a new `map_placed` function with a separate `MemoryMapPlacedInfo` could be created which implies `MemoryMapFlags::PLACED_EXT`. This interface has less parameter combinations which work unexpectedly. For example, as is if there is no flags and a placed address specified, it will silently fail. I don't feel strongly either way, I will defer the decision to @marc0246 and @Rua 🙏

0xcaff: If we have more cycles of reviewing it might be worthwhile all of us finding a time and sitting down together to hammer out an ideal api interface here. I think it will be faster for all parties involved to meet synchronously for 15m than more review cycles.

It seems like we're all active in U.S. Central mornings, happy to coordinate a time if needed.

0xcaff: Ahh, I see. I've moved it.

vulkano/src/memory/device_memory.rs
```diff
@@ -1427,6 +1475,94 @@ impl MemoryMapInfo {
             }));
         }
 
+        if flags.contains(MemoryMapFlags::PLACED) {
+            let features = memory.device.enabled_features();
+            if !features.memory_map_placed {
+                return Err(Box::new(ValidationError {
+                    context: "flags".into(),
+                    problem: "is PLACED".into(),
```
Rua, line 1483:
```suggestion
                    problem: "contains `MemoryMapFlags::PLACED`".into(),
```

0xcaff:
ok, I've incorporated all the suggestions

0xcaff:
I agree, not a fan of the branch https://github.com/vulkano-rs/vulkano/blob/5b7618fc483b3b4150c6e3eba5a488c57092b71c/vulkano/src/memory/device_memory.rs#L480-L526 either.

That said, wouldn't what you proposed here https://github.com/vulkano-rs/vulkano/pull/2514#issuecomment-2052267894 change the behavior of callers which pass currently pass allocation size? I'm concerned this might break compatibility for something which relies on the alternative way (allows allocation size but not `VK_WHOLE_SIZE`). Are you sure this is never the case?

If we're not sure, maybe its prudent to go down the route of having a separate map_placed function for now where we only do this remapping.

0xcaff:
Updated to remove the extra branch. Kept the feature gate.

I did do

```
if size == self.allocation_size() {
    size = ash::vk::WHOLE_SIZE;
}
```

instead of

```
if offset + size == self.allocation_size() {
    size = ash::vk::WHOLE_SIZE;
}
```

as a non-zero `offset` is invalid when `features.memory_map_range_placed` is not available.

https://github.com/0xcaff/vulkano/blob/955dc3ebdf0abc0e9eb84dfd2b5caf41037df8ea/vulkano/src/memory/device_memory.rs#L1543-L1550

0xcaff:
Yes, I believe this is correct. Previously, placed_address implied the flag and `memory_map_placed`, now the flag implies `memory_map_placed` and `placed_address` implies nothing. The feature is a global thing and the flag is allocation specific.

vulkano/src/memory/device_memory.rs
```diff
@@ -1427,6 +1472,94 @@ impl MemoryMapInfo {
...
+                        context: "offset".into(),
+                        problem: "must be aligned to an integer multiple of the \
+                            `min_placed_memory_map_alignment` device property"
+                            .into(),
+                        vuids: &["VUID-VkMemoryMapInfoKHR-flags-09573"],
+                        ..Default::default()
+                    }));
+                }
+
+                if !is_aligned(size, min_placed_memory_map_alignment) {
```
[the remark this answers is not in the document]
0xcaff: I'm going off the validations defined here.

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkMemoryMapInfoKHR.html#VUID-VkMemoryMapInfoKHR-flags-09574

If `memory_map_range_placed` is enabled, it doesn't need to be the entire allocation size, it can be a subset. If `memory_map_range_placed` is disabled, the mapping must be the entire allocation.

As for the `VK_WHOLE_SIZE`, I left it out here because it seems like we concluded earlier we didn't want to expose `VK_WHOLE_SIZE` to callers. https://github.com/vulkano-rs/vulkano/pull/2514#discussion_r1557491787

We already don't allow passing VK_WHOLE_SIZE through validation here

https://github.com/vulkano-rs/vulkano/blob/00345348f69a0ffd187dd1f6fe6c1d79e36cee8a/vulkano/src/memory/device_memory.rs#L1465-L1473

0xcaff:
What if `map_memory_range_placed` is enabled and you want to pass `allocation_size` instead of `VK_WHOLE_SIZE`? I think its a good idea to avoid transforming this value unless needed. `allocation_size` might be sometimes valid for `map_memory_range_placed` but is never valid for `map_memory_range_placed` disabled.

0xcaff:
Ultimately I've concluded this decision is an opinion and I care more about merging than having a debate. Updated to exactly what you asked for.

If there are further minor changes, the branch has maintainer edits on, please feel free to edit directly.

Really curious what the docs ppl end up concluding.

0xcaff:
Ah hey yeah @marc0246 no worries, honestly I tried the copy way you suggested and was finding the performance to be acceptable. I decided to focus on correctness for now in my emulation project and worry about performance later. I was planning to come back and update this eventually but it kept getting pushed below the cutline.

If this is important to you I'm happy for you to take it over. If its not, I'll get to it eventually.

Rua:
Could you merge in the changes from the latest master? I'm encountering a bug that you fixed in another PR, and that's preventing me from testing this currently.

Rua:
Thank you! Sorry it took so long.

0xcaff:
No worries! Thank you!

PR #2515, opened 2024-04-09
Bindless Descriptor fixes and optimizations

Firestar99:
* added `DescriptorSet::invalidate()` to make vulkano forget about resources that bound to a descriptor_set, so they can be freed. As (normal) descriptors are validated when bound, this is entirely safe to do.
* disable validation for individual descriptors if the descriptor set is `UPDATE_AFTER_BIND` or `PARTIALLY_BOUND`, making them actually usable
  * I'm not sure these changes are entirely safe. For it to be, we'd need to validate the descriptor set is properly populated just before flushing any commands to the GPU. And I'm not sure if that's done already, or where it should be added. We could ofc also decide to just leave these two flags as unsafe, somehow.
* some small Descriptor optimizations:
  * More SmallVec use in DescriptorPool to improve variable descriptor count allocation. It only ever allocates one descriptor set at a time, i.e. `allocate_infos` only ever has one entry in that code path. Annoyed me and just had to fix :D
  * Descriptor updates with no writes or copies return early instead of calling into vulkan. Particularly useful when creating a new descriptor, writing nothing, and updating it properly later.

Changelog:
```markdown
### Additions
* added `DescriptorSet::invalidate()` to make vulkano forget about resources that bound to a descriptor_set, so they can be freed

### Bugs fixed
* fixed descriptor sets with `UPDATE_AFTER_BIND` or `PARTIALLY_BOUND` being wrongly validated on bind
```

### DIFF
--- a/vulkano/src/command_buffer/commands/pipeline.rs
+++ b/vulkano/src/command_buffer/commands/pipeline.rs
@@ -10,8 +10,8 @@ use crate::{
         DrawMeshTasksIndirectCommand, RecordingCommandBuffer, ResourceInCommand, SubpassContents,
     },
     descriptor_set::{
-        layout::DescriptorType, DescriptorBindingResources, DescriptorBufferInfo,
-        DescriptorImageViewInfo,
+        layout::{DescriptorBindingFlags, DescriptorType},
+        DescriptorBindingResources, DescriptorBufferInfo, DescriptorImageViewInfo,
     },
     device::{DeviceOwned, QueueFlags},
     format::{FormatFeatures, NumericType},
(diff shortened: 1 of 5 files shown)


Firestar99:
I've turned this PR into a draft, as I've added `DescriptorSet::invalidate()` to it and haven't gotten around to extensively test it

Rua:
If `UPDATE_AFTER_BIND` is enabled, then it's possible that the resources in the descriptor set, at the time the command is recorded, are not the same as those that will be bound when the command buffer is actually executed. So any record-time checks on the bound resources are probably superfluous?

Firestar99:
> So any record-time checks on the bound resources are probably superfluous?
Yes they are, so for now I've just disabled them. Which means they are indeed not validated at all, and honestly I'd keep it that way as I don't think modding that into the old sync has any benefit. Rather invest that time into the new sync.

Issue #2534, opened 2024-05-31
Push constants don't support first field offset decorations


chaynabors:
# Template

*   Version of vulkano: main
*   OS: windows
*   GPU (the selected PhysicalDevice): gtx 4090
*   GPU Driver: TODO
*   Upload of a reasonably minimal complete `main.rs` file that demonstrates the issue: TODO

# Issue
I have two definitions of push constants, one from my vertex shader, and one from my fragment shader. I'd like to use the types provided by the shader macro with each occupying different push constant memory. It's my understanding that to do this, I have to do something like the following:

```rust
  .push_constants(
      self.layout.clone(),
      0,
      SkinnedVertPushConstants { mvp },
  )?
  .push_constants(
      self.layout.clone(),
      std::mem::size_of::<SkinnedVertPushConstants>() as u32,
      DefaultFragPushConstants {
          fog_origin: [0.0, 0.0, 0.0, 0.0],
          fog_color: [1.0, 1.0, 1.0],
          fog_intensity: 1.0,
      },
  )?;
  
  ...
  
  let layout_create_info = PipelineDescriptorSetLayoutCreateInfo {
    flags: PipelineLayoutCreateFlags::empty(),
    set_layouts: vec![DescriptorSetLayoutCreateInfo {
        bindings: BTreeMap::from([(0, layout_binding)]),
        ..Default::default()
    }],
    push_constant_ranges: vec![
        PushConstantRange {
            stages: ShaderStages::VERTEX,
            offset: 0,
            size: std::mem::size_of::<SkinnedVertPushConstants>() as u32,
        },
        PushConstantRange {
            stages: ShaderStages::FRAGMENT,
            offset: std::mem::size_of::<SkinnedVertPushConstants>() as u32,
            size: std::mem::size_of::<DefaultFragPushConstants>() as u32,
        },
    ],
};
```

```
layout(push_constant) uniform PushConstants {
    layout(offset = 64) vec4 fog_origin;
    vec3 fog_color;
    float fog_intensity;
} constants;
```

note that the offset is defined for the zeroth field in this screenshot from https://docs.vulkan.org/guide/latest/push_constants.html#pc-offsets
![image](https://github.com/vulkano-rs/vulkano/assets/25893753/7b7bb76e-9ade-42d6-88a7-dcc999d84545)

relevant code is here
[...]

PR #2535, opened 2024-05-31
Allow memory interface blocks to have offset decorations on first member

chaynabors:
fixes https://github.com/vulkano-rs/vulkano/issues/2534

There is likely a better, much more involved solution here.

The offending struct only contains the following decorations per spirv
`[Decorate { target: Id(14), decoration: Block }]`

That doesn't look like an indication that this is a push constant declaration to me, so discriminating against this might mean passing additional data alongside the spirv from source. Since I don't know the spec well enough to know if this is actually useful or valid for non push-constant declarations, this is the best I can offer.

### DIFF
--- a/vulkano-shaders/src/structs.rs
+++ b/vulkano-shaders/src/structs.rs
@@ -844,11 +844,6 @@ impl TypeStruct {
                 if last.offset + last_size > offset {
                     bail!(shader.source, "struct members must not overlap");
                 }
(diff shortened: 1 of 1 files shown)


Rua:
I don't immediately see any problems that could occur due to this, so it may just be Vulkano being overly conservative on its sanity checks. We'll see if anything turns up later.

PR #2537, opened 2024-06-22
Fix typos and introduce a workflow

szepeviktor:
Found a few misspellings and prevent future ones.


Rua:
There's no point in fixing vk.xml, because that's generated by the Vulkan project itself. It would be overwritten as soon as Vulkano updates to a new Vulkan version.

szepeviktor:
Thank you. I didn't know that.

.typos.toml
```diff
@@ -0,0 +1,20 @@
+[files]
+extend-exclude = [
+    ".git/",
+    "vulkano/spirv.core.grammar.json",
+    "vulkano/vk.xml",
+]
+ignore-hidden = false
+
+[default]
```
Rua, line 9: Is this needed?

szepeviktor: Yes. That is the section for `extend-ignore-re`.

Rua: Is it not possible to tell it to ignore the whole file, rather than specific names within the file?

szepeviktor: Oh! I am sorry.
Those two name are in vk.xml.
Fixed.

szepeviktor:
Glad to contribute.

PR #2539, opened 2024-06-27
Optimize sequential draws

Firestar99:

Profiling based on my meshlet demo on sponza. Before at ~40fps and fully CPU limited:
image

With this PR I get ~550fps and am most likely CPU GPU sync limited, due not having a working frame in flight system.
image

And the best part is: It's absolutely free and still safe!*
*(if I didn't mess up)

Current master evaluates all accessed resources (buffers, images) if you call any draw or dispatch immediately. My change defers the evaluation specifically of descriptor sets for their resources to the end of the cmd buffer recording. This allows me to deduplicate draws and dispatches using the same pipeline, merging their resources and descriptor sets together, and then deduplicate the descriptor sets again before evaluating each unique one for their actual resources.

Changelog:
```markdown
### Additions
- Optimized performance of back to back draws/dispatches using the same pipeline significantly
```

### DIFF
--- a/vulkano/src/command_buffer/commands/pipeline.rs
+++ b/vulkano/src/command_buffer/commands/pipeline.rs
@@ -4,7 +4,9 @@ use crate::{
     acceleration_structure::AccelerationStructure,
     buffer::{view::BufferView, BufferUsage, Subbuffer},
     command_buffer::{
-        auto::{RenderPassState, RenderPassStateType, Resource, ResourceUseRef2},
+        auto::{
+            RenderPassState, RenderPassStateType, Resource, ResourceUseRef2, UsedResourcesDeferred,
+        },
         sys::RawRecordingCommandBuffer,
         DispatchIndirectCommand, DrawIndexedIndirectCommand, DrawIndirectCommand,
         DrawMeshTasksIndirectCommand, RecordingCommandBuffer, ResourceInCommand, SubpassContents,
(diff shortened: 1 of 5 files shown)


Firestar99:
I'm seeing some weird performance behavior in bistro I would like to investigate first before officially submitting this

vulkano/src/command_buffer/commands/pipeline.rs
```diff
@@ -3439,20 +3416,22 @@ impl RecordingCommandBuffer {
...
-            .builder_state
+    ) -> Option<DescriptorSetState> {
+        self.builder_state
             .descriptor_sets
             .get(&pipeline.bind_point())
-        {
-            Some(x) => x,
-            None => return,
-        };
+            .cloned()
```
Firestar99, line 3426: Another **theoretical** performance optimization that could be done: remove this clone. It clones the entire DescriptorSetState, which contains a HashMap, and if push descriptors are used each one of those contains another HashMap. Instead it could be some `Cow<Arc<_>>` so that if state does not change between draws it isn't cloned again. 
But **practically** it's insignificant, it's done once for every draw/dispatch and 103 draws take just 0.12ms.

Firestar99:
I'm uncertain how secondary cmd buffers are handled. I could imagine a usecase, when they contain many draws with always changing pipelines (which is generally frowned upon anyways) and with very few buffers used, that *could maybe* be slower with this PR. But I dunno, that may need some testing.

PR #2548, opened 2024-07-21
Task graph [3/10]: execution

marc0246:
Follow-up to #2545, implementing the task graph executor.


vulkano-taskgraph/src/graph/execute.rs
```diff
@@ -0,0 +1,1787 @@
...
+        unsafe { self.update_resource_state(&resource_map, 0..self.instructions.len()) };
+
+        resource_map
+            .resources
+            .try_advance_global_and_collect(&resource_map.guard);
+
+        res
+    }
+
+    unsafe fn acquire_images_khr(
```
Rua, line 150: Why the `_khr` suffix?

Rua: I don't think Vulkano itself ever uses those suffixes? Or is this different?

Rua: Oh, I see. You intend to have different versions of these with different suffixes.

PR #2560, opened 2024-08-25
Task graph [4/10]: compilation

marc0246:
Follow-up to #2548, implementing the task graph compiler.

I gave up on subresource-granularity sync because it was making everything complicated and I just didn't want to deal with it anymore. The only reason I did it in the first place was because the old sync did it. But that's also what we're trying to replace, so maybe it was just not meant to be. Few things come to mind that couldn't be worked around with whole-resource-granularity sync anyway. A nice side-effect is that this is a whole lot faster both when compiling and executing the task graph.

I converted the async-update example as a very very very (very) early test. Mainly because the kind of syncronization it does was the biggest thorn in my ass.

Changelog:
```markdown
### Additions
- Added `AccessFlags::{contains_reads,contains_writes}`.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,6 +114,7 @@ dependencies = [
  "rand",
  "vulkano",
  "vulkano-shaders",
+ "vulkano-taskgraph",
  "winit 0.29.15",
 ]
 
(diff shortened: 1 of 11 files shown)


Rua:
When I run the async example, I get this validation layer error whenever I press space:

> UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout(ERROR / SPEC): msgNum: 1303270965 - Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout ] Object 0: handle = 0x64f9f7847980, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xdd3a8a0000000015, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x4dae5635 | vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] command buffer VkCommandBuffer 0x64f9f7847980[] expects VkImage 0xdd3a8a0000000015[] (subresource: aspectMask 0x1 array layer 0, mip level 0) to be in layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL.
>     Objects: 2
>         [0] 0x64f9f7847980, type: 6, name: NULL
>         [1] 0xdd3a8a0000000015, type: 10, name: NULL

Is that ok/still todo?

PR #2561, opened 2024-09-08
macOS/iOS: Use `raw-window-metal` to get a `CAMetalLayer` from `raw-window-handle`

madsmtm:
    - I've done this in the `raw-window-metal` crate itself. Surface construction should be covered by existing examples.
4. [x] Run `cargo clippy` on the changes.
5. [x] Run `cargo +nightly fmt` on the changes.
6. [x] Please put changelog entries **in the description of this Pull Request**
7. [x] Describe in common words what is the purpose of this change, related

The way `raw-window-metal` works is by creating a layer, and inserting that as a sublayer, just like we did on iOS before. The bounds are then kept in-sync with an observer, ensuring smooth resizing.

This also fixes compilation errors on iOS, and adds preliminary support for tvOS.

The implementation now solely uses `VK_EXT_metal_surface`, which was added in 2018, and I've removed support for `VK_MVK_ios_surface` / `VK_MVK_macos_surface`, which are deprecated, and only available a year and a half earlier anyhow.

Changelog:
```markdown
### Breaking changes
Changes to `Surface`:
- `Surface::from_ios` and the corresponding `SurfaceApi::Ios` was removed, use `Surface::from_metal` instead.
- `Surface::from_mac_os` and the corresponding `SurfaceApi::MacOs` was removed, use `Surface::from_metal` instead.
- `Surface::update_ios_sublayer_on_resize` was removed as it is no longer necessary.

### Additions
- Add support for tvOS.

### Bugs fixed
- Make resizing smooth on macOS and iOS, and let it interoperate better with windowing libraries.
- Fix compiling on iOS.
```

Waiting on https://github.com/rust-windowing/raw-window-metal/pull/19, hence why this is still a draft.

See also corresponding `ash` PR: https://github.com/ash-rs/ash/pull/939


Rua:
I don't think support for extensions should be removed. Vulkano should work on any Vulkan-supporting device, including old ones with deprecated extensions.

madsmtm:
Thing is, the support that Vulkano previously used for these two included the somewhat implicit feature that you can pass `CAMetalLayer` directly. This feature was only added to the spec in 2020, see [the diff](https://github.com/KhronosGroup/Vulkan-Docs/commit/792c6d18f64b73463a2023cf7610a694c11e5bd2#diff-d255a0a7bf648ab09897dceb27aef05eb709a493b678ae7ab184c50c37f6f68e). So this PR is actually only _extending_ the support that Vulkano has for older versions.

But sure, I'll keep the extension code themselves in.

madsmtm:
I guess if we wanted to keep this free from breaking changes, we could keep the `update_ios_sublayer_on_resize` as a no-op, and deprecate it instead?

madsmtm:
About `vulkano-win`, I felt like if given you still have it in the `git` repo, then I might as well update it. I don't really care whether you end up releasing the changes or not.

hakolao:
My mac's so slow nowadays so I don't want to do much with it. I ran the triangle fine, anything else to test? @marc0246.

PR #2562, opened 2024-09-08
Add `PhysicalDevice::presentation_support`

marc0246:
This adds a function that abstracts away the platform-specific presentation support calls akin to `Surface::from_window`, allowing you to determine presentation support to the surface of any window of a given event loop. This addresses two issues:
- There are cases where you don't have a window to test for support with. For instance, ever since winit 0.30 you can only create a surface in the `Resumed` callback, which makes people think that you have to create the whole device in that callback, when that's not the case. This addition is going to make updating to winit 0.30 easier.
- There are cases where you have multiple windows, or you could be creating additional windows at the user's request. This too is impractical with `PhysicalDevice::surface_support`. This also fixes the issues that the multi-window example was having.

`PhysicalDevice::surface_support` is essentially only usable if you have a window since right after the creation of the event loop and when the window exists for the app's lifetime. Therefore, I replaced all `PhysicalDevice::surface_support` calls in the examples with `PhysicalDevice::presentation_support` to promote the better alternative (and for the future when we update to winit 0.30).

Unfortunately, this function was not implementable without adding dependencies for X11 libraries on X11 platforms. I added a default feature same as winit does, so the user can disable the dependencies if they e.g. only support Wayland.

Changelog:
```markdown
### Additions
- Added `PhysicalDevice::presentation_support` for determining presentation support to the surface of any window of a given event loop.
```

### DIFF
--- a/vulkano/Cargo.toml
+++ b/vulkano/Cargo.toml
@@ -34,6 +34,10 @@ vulkano-macros = { workspace = true, optional = true }
 objc = { workspace = true }
 core-graphics-types = { workspace = true }
 
+[target.'cfg(any(target_os = "dragonfly", target_os = "freebsd", target_os = "hurd", target_os = "illumos", target_os = "linux", target_os = "netbsd", target_os = "openbsd", target_os = "solaris"))'.dependencies]
+x11-dl = { workspace = true, optional = true }
+x11rb = { workspace = true, features = ["allow-unsafe-code"], optional = true }
+
 [build-dependencies]
 ahash = { workspace = true }
 heck = { workspace = true }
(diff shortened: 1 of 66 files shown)


vulkano/Cargo.toml
```diff
@@ -34,6 +34,10 @@ vulkano-macros = { workspace = true, optional = true }
 objc = { workspace = true }
 core-graphics-types = { workspace = true }
 
+[target.'cfg(any(target_os = "dragonfly", target_os = "freebsd", target_os = "hurd", target_os = "illumos", target_os = "linux", target_os = "netbsd", target_os = "openbsd", target_os = "solaris"))'.dependencies]
```
Rua, line 37: Wouldn't it be easier to write something like:
`cfg(all(unix, not(target_os = "foo"), not(target_os = "bar")))`

PR #2575, opened 2024-10-14
Fix SPIR-V validation of image ops with offsets

LDeakin:
Shader functions like `textureLodOffset` would incorrectly fail validation.

Checks like this would always evaluate to true (and fail validation) because the property wraps:
https://github.com/vulkano-rs/vulkano/blob/a2704689603bd614186edd17c73779fd1f401548/vulkano/src/pipeline/shader/validate_runtime.rs#L2540

Changelog:
```markdown
### Bugs fixed
- Fix SPIR-V validation of image ops with offsets
```

### DIFF
--- a/vulkano/src/pipeline/shader/validate_runtime.rs
+++ b/vulkano/src/pipeline/shader/validate_runtime.rs
@@ -6,8 +6,9 @@ use crate::{
     },
     shader::{
         reflect::{
-            get_constant, get_constant_composite, get_constant_composite_composite,
-            get_constant_float_composite, get_constant_maybe_composite, size_of_type,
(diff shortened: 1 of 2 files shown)


Rua:
Thanks! One remark: The `get_constant_*` functions are supposed to return the constants as unsigned integers. If they should return signed integers, they should be renamed to `get_constant_signed_*` so that they match `get_constant_float_*`

LDeakin:
I've renamed the offending functions. Note that now there are no unsigned variants of `get_constant_composite_composite` and `get_constant_maybe_composite`, but they would be dead code if I brought them back. Thoughts?

PR #2576, opened 2024-10-15
Revert #2425

marc0246:
This reverts #2425. I was cooking something, but plans have changed and we're going with a new crate for the new command buffer and deprecating the old one, so it doesn't make sense to break code for no reason now.

Changelog:
Remove these lines:
```markdown
### Breaking changes
Changes to command buffers:
- Merged `{Primary,Secondary}AutoCommandBuffer` into `CommandBuffer` and the corresponding `RecordingCommandBuffer::{primary,secondary}` into `new`.
- `RecordingCommandBuffer` no longer has a type parameter for the command buffer level.
```


Rua:
Should `RecordingCommandBuffer` not be renamed back to `AutoCommandBufferBuilder` as well?

Rua:
No, I just wasn't sure if you were planning to do them in stages.

Issue #2583, opened 2024-10-20
Dynamic viewport documentation and examples are wrong.


coolcatcoder:
Run triangle-util and resize the window. The triangle doesn't move. This is because the viewport is never updated.

I originally wrote the triangle-util example, so this is probably my fault, but I'd also like to drag Viewport's documentation down with me. Specifically the documentation of its field called extent. Extent's documentation is: "Dimensions in pixels of the viewport. The default value is `[1.0; 2]`, which you probably want to override if you are not using dynamic state.". Even with dynamic state you still need to override the default.

The documentation should be changed, and the example fixed. (Should also check to see if any other examples suffer from this.)

PR #2584, opened 2024-10-20
Fix 2583.

coolcatcoder:
Fixes [2583](https://github.com/vulkano-rs/vulkano/issues/2583). It seems that triangle-util is the only example that fell into this trap, but there are a lot of examples, so I can't say for certain.


coolcatcoder:
Cool, thanks!

PR #2590, opened 2024-10-22
Ignore `vk.xml` and `spirv.core.grammar.json` from editor search results

marc0246:
Having these files be searched when one is doing a project-wide search is rather counter-productive. This fixes that for any editor that uses ripgrep e.g. vscode, Helix and Neovim with the Telescope plugin.

### DIFF
--- a/.ignore
+++ b/.ignore
@@ -1,0 +1,2 @@
+vulkano/spirv.core.grammar.json
+vulkano/vk.xml


Rua:
Is there a way to un-ignore those files, if I do want to search in them?

Rua:
I do hope that this won't confuse and surprise people who edit/contribute to the codebase, but we'll see.

Issue #2360, opened 2023-10-17
Support sparse binding


Rua:
A long-standing missing feature. Some of the basic logic is currently there, but a lot more is needed. In particular:
- Buffer and image state tracking needs to track what memory is bound where at any given time. Not just for validation purposes, but also to keep the memory allocations alive.
- When submitting a command buffer, checks should be performed to ensure that all the accessed regions of buffers and images in that command buffer have memory bound to them. This needs to take into account sync ordering (sparse binds are not ordered unless synced with semaphores).
- How should sparse binding interact with subbuffers?
- How to deal with host access? Sparse memory is not contiguous, so the current `read` and `write` methods, which return a slice, don't work.

LDeakin:
I have been using sparse images with `vulkano` where I manually manage allocations and binding. I had to make [a few small changes](https://github.com/vulkano-rs/vulkano/compare/master...LDeakin:vulkano:sparse_images) to `vulkano` though:
- Uncomment `ImageCreateFlags::SPARSE_BINDING` etc
- Make `Image::from_raw()` public
- Make `RawImage::get_sparse_memory_requirements()` public

Would you accept a PR with such changes? I recognise that this does not actually help with adding safe sparse image support to `vulkano`, but it is enough for my use case.

PR #2608, opened 2024-11-28
Make `Image::from_raw` public

LDeakin:
Part 1 of a series of small changes for partial sparse image support. See #2360.

Changelog:
```markdown
### Additions
- Make `Image::from_raw` public
```

### DIFF
--- a/vulkano/src/image/mod.rs
+++ b/vulkano/src/image/mod.rs
@@ -174,7 +174,14 @@ impl Image {
         Ok(Arc::new(image))
     }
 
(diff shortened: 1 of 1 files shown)


Rua:
I have the code ready to implement the basics of sparse buffers and images, so this PR may be unnecessary. However, the reason I haven't submitted it yet is because there's a bunch of parts of the spec that are unclear. Maybe I should do it anyway and see where it ends up?

LDeakin:
> Maybe I should do it anyway and see where it ends up?

That would be fantastic @Rua

Rua:
See #2609 now.

PR #2564, opened 2024-09-10
Ray Tracing Pipeline (KHR)

ComfyFluffy:
1. [x] Update documentation to reflect any user-facing changes - in this repository.

2. [ ] Make sure that the changes are covered by unit-tests.

3. [x] Run `cargo clippy` on the changes.

4. [x] Run `cargo +nightly fmt` on the changes.

5. [x] Please put changelog entries **in the description of this Pull Request**
   if knowledge of this change could be valuable to users. No need to put the
   entries to the changelog directly, they will be transferred to the changelog
   file by maintainers right after the Pull Request merge.

## Changelog:
```markdown
### Additions
- Support for the `VK_KHR_ray_tracing_pipeline`.
```

## Changes

- New struct `RayTracingPipeline`, which corresponds to the `vkPipeline` object and construction methods and structs.
- `bind_pipeline_ray_tracing` & `trace_rays` (`vkCmdTraceRaysKHR`) in the old and new sync.
- `Device::get_ray_tracing_shader_group_handles`. Used in building the shader binding table.
- `ShaderGroupHandlesData`. Helper for the data from `get_ray_tracing_shader_group_handles`.
- `ShaderBindingTable`. Struct that holds the addresses and underlying buffer of shader binding table, which is used in  `trace_rays`. It contains a constructor that automatically build the SBT from pipeline groups.
- Examples: `triangle-raytracing-auto` for the old `AutoCommandBuffer`, and `triangle-raytracing` for the new `taskgraph` sync.

## Not implemented (yet) in this PR

- `DeferredOperationKHR`. Not planned.
- `RayTracingCaptureReplay`. Not planned.
- `RayTracingPipelineStackSize` (in dynamic stats). Maybe in this PR?

PR #2625, opened 2025-01-28
Ray tracing fixes

marc0246:
This fixes the inconsistencies from #2564. This is mainly just formatting/documentation/conventions:
- We use `.glsl` file extensions because they are the most widely supported.
- 4 space indentation and snake case for GLSL files.
- The examples weren't consistent with the rest of the examples.
- Error message/documentation/code style conventions.
- Some methods weren't documented.
- `trace_rays` wasn't taking an array like dispatches do.
- `get_ray_tracing_pipeline_shader_group_handles` was scoped wrong.

cc @ComfyFluffy


ComfyFluffy (approved):

ComfyFluffy:
Examples run correctly on my machine.

PR #2640, opened 2025-02-06
Remove the `ash::vk` stutter

marc0246:
`ash::vk` is a module because it's supposed to be imported. It's really hard to look at whatever this was.


Rua:
I don't see why you think it was supposed to be imported. But I do think this is a good change regardless.

PR #2641, opened 2025-02-07
Replace `NonZeroT` with `NonZero<T>`

Rua:
### Breaking changes
Miscellaneous:
- All occurrences of `NonZeroT` have been replaced with the generic `NonZero<T>`. The Vulkano-specific type aliases `NonZeroDeviceSize` and `NonNullDeviceAddress` have been replaced with `NonZero<DeviceSize>` and `NonZero<DeviceAddress>`, respectively.
```


Rua:
Add back only the type aliases, but not the places where they were formerly used?

PR #2649, opened 2025-02-09
Add `new` constructors to vulkano-taskgraph `Info` structs too

marc0246:


Rua:
For `ResourcesCreateInfo`, I think the actual decimal values for the default should also be given, just for the sake of clarity. Yes, many programmers know their powers of two, but not necessarily all of them.

PR #2657, opened 2025-02-24
Update concurrent-slotmap dependency and add a single-threaded slotmap

marc0246:
This updates our concurrent-slotmap dependency which has had many breaking changes. With the new version, we can finally fix resource destroying and make it safe next.

I also implemented a new single-threaded slotmap because the concurrent slotmap that was being used in places was always just a stand-in and there are many reasons not to use it in these places:
- Allocating pages can be expensive and so is not suited for something that might be transient in the game loop.
- It's not as efficient due to atomic operations in the immutable case.
- It's not as efficient memory-wise.
- Unneccessary unsafe code for things such as immutable getting and immutable iteration.
- Having to specify a max capacity is an unnecessary hassle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -297,8 +297,8 @@ dependencies = [
 
 [[package]]
 name = "concurrent-slotmap"
-version = "0.1.0-alpha.1"
-source = "git+https://github.com/vulkano-rs/concurrent-slotmap?rev=db6810ab699b99fd79052252c057851fa80117cf#db6810ab699b99fd79052252c057851fa80117cf"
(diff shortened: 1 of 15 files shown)


Rua:
The `deferred` example is panicking when I run it:
> Using device: AMD Radeon RX 7600 (RADV NAVI33) (type: DiscreteGpu)
>
> thread 'main' panicked at /home/rua/code/vulkano/vulkano-taskgraph/src/graph/execute.rs:89:9:
> you must wait on the fence for the current frame before submitting more work
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

PR #2661, opened 2025-03-07
Disabled validation for descriptor binding merge.

LukeP0WERS:
This is probably not the best solution, but it works.

Changelog:
```markdown
### Bugs fixed
- When using bindless the validation checks for DescriptorBindingRequirements would panic.
```

### DIFF
--- a/vulkano/src/shader/mod.rs
+++ b/vulkano/src/shader/mod.rs
@@ -1253,11 +1253,11 @@ impl DescriptorBindingRequirements {
             descriptor_types,
             descriptor_count,
             image_format,
+            image_multisampled: _,
             image_scalar_type,
             image_view_type,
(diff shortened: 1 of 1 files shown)


vulkano/src/shader/mod.rs
```diff
@@ -1253,11 +1253,11 @@ impl DescriptorBindingRequirements {
             descriptor_types,
             descriptor_count,
             image_format,
-            image_multisampled,
             image_scalar_type,
             image_view_type,
             stages,
             descriptors,
+            ..
```
Rua, line 1260:
```suggestion
            image_multisampled: _,
            image_scalar_type,
            image_view_type,
            stages,
            descriptors,
```

PR #2664, opened 2025-03-17
Add nix flake

marc0246:
Adds a nix flake in order to make our entire build process easily reproducible.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,9 @@
       in {
         devShells.default = with pkgs; mkShell rec {
           buildInputs = [
-            (rust-bin.stable.latest.default.override { extensions = [ "rust-analyzer" ]; })
+            (rust-bin.stable.latest.default.override {
+              extensions = [ "rust-analyzer" "rust-src" ];
+            })
 
             # Vulkan dependencies
             shaderc
--- a/flake.nix
+++ b/flake.nix
@@ -16,9 +16,10 @@
       in {
         devShells.default = with pkgs; mkShell rec {
           buildInputs = [
-            (rust-bin.stable.latest.default.override {
-              extensions = [ "rust-analyzer" "rust-src" ];
+            (rust-bin.stable.latest.minimal.override {
+              extensions = [ "clippy" "rust-analyzer" "rust-docs" "rust-src" ];
             })
+            (rust-bin.selectLatestNightlyWith (toolchain: toolchain.rustfmt))
 
             # Vulkan dependencies
             shaderc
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,47 @@
...
+            libxkbcommon
+            wayland
+            xorg.libX11
+            xorg.libXcursor
+            xorg.libXi
+            xorg.libXrandr
+          ];
+
+          LD_LIBRARY_PATH = "${lib.makeLibraryPath buildInputs}";
+          SHADERC_LIB_DIR = "${shaderc.lib}/lib";
```
nbraud, line 42:
```suggestion
          LD_LIBRARY_PATH = lib.makeLibraryPath buildInputs;
          SHADERC_LIB_DIR = lib.makeLibraryPath [ shaderc ];
```

flake.nix
```diff
@@ -0,0 +1,47 @@
...
+      system: let
+        overlays = [ (import rust-overlay) ];
+        pkgs = import nixpkgs { inherit system overlays; };
+      in {
+        devShells.default = with pkgs; mkShell rec {
+          buildInputs = [
+            (rust-bin.stable.latest.minimal.override {
+              extensions = [ "clippy" "rust-analyzer" "rust-docs" "rust-src" ];
+            })
+            (rust-bin.selectLatestNightlyWith (toolchain: toolchain.rustfmt))
```
nbraud, line 22: What's the rationaly for getting  the nightly `rustfmt` ?  Even if there's a good reason, it would be best to document it with a comment.

nbraud (commented):
Hi!  In addition to some slight comments & suggestions, I had one much more high-level question:  what's the goal here, and what are the next steps?  I read the stated rationale:

> make our entire build process easily reproducible

Presumably, this entails making various processes use the flake, for instance
1.interactive builds and development, incl. how Vulkano's devs work ;
2. automated builds and tools, such as CI and release processes:
  from experience, I can confirm having `nix` definitions of CI tasks, is highly useful for testing CI changes locally and/or debugging CI issues
3. downstreams users of Vulkano (seems unlikely)

It's much harder to evaluate a PR in a vacuum, so I'd suggest implementing (at least in part) one of those usecases; (2) seems to me like the most immediately-useful and simplest, so it might be a good idea.

nbraud:
As I said, “just adding a dev shell” without considering how it's meant to be used, make review difficult; you do you though, I only pitched in because @Rua asked me about this PR.

nbraud:
> how it's supposed to be used: you run `nix develop` and then work on the code base/run examples. That's all.

OK, so it's *not* meant to be used by downstream users?  *is confused, just goes curl up somewhere.*

axelkar:
I recommend not using the unstable channel, since packages change so much on there. Please use `nixpkgs-24.11` so there won't be so many duplicate packages. It also helps out people getting started with Nix.

PR #2669, opened 2025-03-25
Fix min swapchain image amount to allow variable MAX_FRAMES_IN_FLIGHT in examples

Letronix624:
This change allows the user to play with the example code and modify `MAX_FRAMES_IN_FLIGHT` in taskgraph examples without having to experience errors.

There is 1 extra frame breathing room in the swapchain as seen by the `+ 1`.

This one is merely an example update and has no changelog worthy changes.

### DIFF
--- a/examples/async-update/main.rs
+++ b/examples/async-update/main.rs
@@ -414,7 +414,9 @@ impl ApplicationHandler for App {
                     self.graphics_flight_id,
                     surface,
                     SwapchainCreateInfo {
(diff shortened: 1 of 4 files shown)


Letronix624:
A `MIN_SWAPCHAIN_IMAGES` would work for every single example. I could do that. Should the consts be declared at the top of the files or right before swapchain creation for sequential context in this case?

PR #2673, opened 2025-04-03
Add convenience types for projects using the task graph

LukeP0WERS:
I have been using the combination of `ResourceAccess` and "global trackers" to make things much easier to work with in my engine. `GlobalImageTracker` is especially useful since it allows you to store everything you need related to that image in one place.

Ex.
```rust
let example_image: GlobalImageTracker = access.create_global_image(
    Some(task_graph),// If this is `None` no virtual image will be created and managed.
    ImageCreateInfo {
        format: Format::R8G8B8A8_UNORM,
        usage: ImageUsage::STORAGE,
        extent: [window_size[0], window_size[1], 1],
        ..Default::default()
    },
    AllocationCreateInfo::default(),
    GlobalImageCreateInfo::storage_sampled(),// This determines whats added to the bindless system, in this case it adds both a `StorageImageId` and `SampledImageId` which can retrieved from the `GlobalImageTracker`
)
.unwrap();
```

Later on I am going to make a procedural macro and trait: `ResourceManager` which will allow you to keep all of your `GlobalImageTracker`s in a single struct, and it will automatically add all of your fields to a resource map and update their virtual images automatically when the task graph is recreated (currently I'm just using regular macros).

Currently this is far from ready to be a public API, I need to fix `ModifyImageViewCreateInfo` and document everything. But I wanted to see if y'all think this would even be worth adding to vulkano first. I don't know if similar things have already been planned for the taskgraph and I'm just making a shitty version of something yet to come, but considering the very high level and project specific nature of the other parts of vulkano-util I don't think it would be too out of place.

About the other stuff:
* `GlobalAccelerationStructureTracker` and `GlobalBufferTracker` may also have interactivity with `ResourceManager` in the future but right now they only exist for the convenience of having the related info in one place.
* The `buffer_from_data` and `buffer_from_slice` methods are also unrelated to tracking.
* I added `Allocators` since I use it a lot and I've seen the same thing used in other crates like egui_winit_vulkano, but its unrelated to the other functionality.
* The `load_module` and `create_compute_pipeline` functions are also mostly unrelated, but I find them really nice for reducing code duplication and adding specialization constants.


LukeP0WERS:
I'm actually super glad to hear these issues will be resolved in the future so I don't have to rely on some shitty interface I made everywhere in my code. I also thought I heard vulkano-util was going to be deprecated, and that makes sense considering how specific the code in there was. Also its good to get some feedback on the code I write since usually only I see it so thanks for checking it out.

PR #2677, opened 2025-04-09
D3D11 textures handle external memory import support 

shaltupssen:
…These are the handle variants that come from D3D textures on Windows.

5.  [X] Please put changelog entries **in the description of this Pull Request**

6.  [X] Describe in common words what is the purpose of this change, related

Adds support for importing D3D11Texture and D3D11TextureKmt handles. These are the handle variants that come from other D3D textures on Windows.

Changelog:
```
- Added support for importing D3D11Texture / D3D11TextureKmt memory handles.
```

### DIFF
--- a/vulkano/src/memory/device_memory.rs
+++ b/vulkano/src/memory/device_memory.rs
@@ -1215,6 +1215,27 @@ pub enum MemoryImportInfo {
         handle_type: ExternalMemoryHandleType,
         handle: vk::HANDLE,
     },
+
+    /// Import memory from a Windows handle to a D3D texture.
+    ///
+    /// `handle_type` must be [`ExternalMemoryHandleType::D3D11Texture`] or
+    /// [`ExternalMemoryHandleType::D3D11TextureKmt`].
+    ///
+    /// # Safety
+    ///
+    /// - `handle` must be a valid Windows handle to a D3D texture resource.
+    /// - Vulkan will not take ownership of `handle`.
+    /// - The handle must be created by the Vulkan API or an API compatible with Vulkan's external
+    ///   memory handle types for D3D textures.
+    /// - [`MemoryAllocateInfo::allocation_size`] and [`MemoryAllocateInfo::memory_type_index`]
+    ///   must match those of the original memory allocation.
+    /// - If the original memory allocation used [`MemoryAllocateInfo::dedicated_allocation`], the
+    ///   imported one must also use it, and the associated image must be defined identically to
+    ///   the original.
+    D3D {
+        handle_type: ExternalMemoryHandleType,
+        handle: vk::HANDLE,
+    },
 }
 
 impl MemoryImportInfo {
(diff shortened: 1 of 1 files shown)


vulkano/src/memory/device_memory.rs
```diff
@@ -1309,6 +1330,52 @@ impl MemoryImportInfo {
...
+                    }));
+                }
+
+                handle_type.validate_device(device).map_err(|err| {
+                    err.add_context("handle_type")
+                        .set_vuids(&["VUID-VkImportMemoryWin32HandleInfoKHR-handleType-parameter"])
+                })?;
+
+                match handle_type {
+                    ExternalMemoryHandleType::D3D11TextureKmt
```
zaiste-linganer, line 1356: If namespace is D3D, I would add D3D12 related resources here as well.

vulkano/src/memory/device_memory.rs
```diff
@@ -1353,6 +1421,15 @@ impl MemoryImportInfo {
...
                     .handle(handle),
             ),
+            MemoryImportInfo::D3D {
+                handle_type,
+                handle,
+            } => {
+                let mut import_info = vk::ImportMemoryWin32HandleInfoKHR::default()
+                    .handle_type(handle_type.into())
+                    .handle(handle);
+                MemoryImportInfoVk::Win32Handle(import_info)
```
zaiste-linganer, line 1431: Nit: for the sake of the style coherency, I'd suggest using the same semantics as Win32 above (without creating extra variable).

vulkano/src/memory/device_memory.rs
```diff
@@ -1215,6 +1215,27 @@ pub enum MemoryImportInfo {
...
+    /// `handle_type` must be [`ExternalMemoryHandleType::D3D11Texture`] or
+    /// [`ExternalMemoryHandleType::D3D11TextureKmt`].
+    ///
+    /// # Safety
+    ///
+    /// - `handle` must be a valid Windows handle to a D3D texture resource.
+    /// - Vulkan will not take ownership of `handle`.
+    /// - The handle must be created by the Vulkan API or an API compatible with Vulkan's external
+    ///   memory handle types for D3D textures.
+    /// - [`MemoryAllocateInfo::allocation_size`] and [`MemoryAllocateInfo::memory_type_index`]
```
zaiste-linganer, line 1230: For D3D11Textures and D3D12Resource (but not for D3D12Heap) imported to Vulkan, allocation_size is ignored (per https://registry.khronos.org/vulkan/specs/latest/man/html/VkMemoryAllocateInfo.html: 

> If the parameters define an import operation and the external handle type is VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT, VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_KMT_BIT, or VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE_BIT, allocationSize is ignored. The implementation must query the size of these allocations from the OS.
)

zaiste-linganer, line 1230: For D3D resources there is no memory_type_index available at the memory allocation time. One needs to follow https://registry.khronos.org/vulkan/specs/latest/man/html/VkMemoryAllocateInfo.html:

> If the parameters define an import operation and the external handle is an NT handle or a global share handle created outside of the Vulkan API, the value of memoryTypeIndex must be one of those returned by [vkGetMemoryWin32HandlePropertiesKHR](https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetMemoryWin32HandlePropertiesKHR.html)

PR #2678, opened 2025-04-09
Adds support for importing D3D11Texture and D3D11TextureKmt handles. …

shaltupssen:
…These are the handle variants that come from D3D textures on Windows.

5. [X] Please put changelog entries **in the description of this Pull Request**

6. [X] Describe in common words what is the purpose of this change, related

Changelog:
```
- Added support for importing D3D11Texture / D3D11TextureKmt memory handles.
```


vulkano/src/memory/device_memory.rs
```diff
@@ -1312,6 +1336,83 @@ impl MemoryImportInfo {
...
+                            vuids: &["VUID-VkImportMemoryWin32HandleInfoKHR-handleType-00660"],
+                            ..Default::default()
+                        }));
+                    }
+                }
+
+                // VUID-VkMemoryAllocateInfo-memoryTypeIndex-00645
+                let memory_type_bits: u32 = {
+                    let fns = device.fns();
+                    let func = fns
```
Rua, line 1383: This function should get its own public wrapper function on `Device`, like so:
```rust
pub fn acceleration_structure_build_sizes(
    &self,
    handle_type: ExternalMemoryHandleType,
) -> Result<Vec<MemoryWin32HandleProperties>, Validated<VulkanError>>

#[cfg_attr(not(feature = "document_unchecked"), doc(hidden))]
pub unsafe fn acceleration_structure_build_sizes_unchecked(
    &self,
    handle_type: ExternalMemoryHandleType,
) -> Result<Vec<MemoryWin32HandleProperties>, VulkanError>
```

shaltupssen: I have added this in a follow up commit

Rua: Thanks! Can you adjust this part of the code now, so that it uses the function you added?

shaltupssen: Done. Thanks for the review

Rua:
A lot of the documentation you added seems to be copied directly from the Vulkan documentation. Vulkano does not do that if it can be avoided. Copyright issues aside, the Vulkan documentation often just doesn't make sense for Vulkano because things are named/structured differently. Moreover, Vulkano does not try to fully replace all documentation for Vulkan; it's assumed that the user is somewhat familiar with the equivalent raw Vulkan API, or at least will read up on it if they want to know the details.

PR #2680, opened 2025-04-12
Lower global descriptor set defaults

marc0246:
llvmpipe uses decimal number limits for these (cringe) which results in me not being able to run the task graph examples using llvmpipe.


Rua:
Are the old limits the minimums required by the spec? And llvmpipe is being noncompliant here?

PR #2682, opened 2025-04-29
Use references as params instead of owned `Arc`s and collections [1/2]

marc0246:
This is the first part to eradicating owned `Arc`s, structs and collections in parameters (except for anything relating to old sync as it's getting deprecated). This part includes the instance, device, buffer, image, swapchain, acceleration structure, and memory, pool and allocator modules.

There are many reasons for doing this:
- Owned `Arc`s in parameters mean cloning on the user's end, which is unnecessary clutter.
- In some instances `Arc`s were used in `Info` structs even though they weren't stored, simply because there was no lifetime on the containing struct. This means incrementing and decrementing reference counts for nothing. It also forces the user into wrapping each of the objects in an `Arc`. An example could be seen here: https://github.com/vulkano-rs/vulkano/blob/846c373daa3101b1c0287318df71b91102a01c2c/vulkano-taskgraph/src/resource.rs#L110-L112 Here we are forced to store each semaphore in its own heap allocation for no reason in particular.
- Owned structs are similar to the `Arc`s. Reusing them meant cloning on the user's end, which is both bad ergonomics and forces a copy each time. This is why the original API uses pointers to structs so much, and is something to keep IMO.
- Owned collections in parameters mean that the user is forced to use the specific collection, and worse, give up ownership of it each time. Again the point regarding cloning and its ergonomics and performance implications. And even when reuse is not needed, there simply is no reason the user should be forced to allocate a collection. If they want to use a stack allocation, `ArrayVec`, `SmallVec`, a static allocation, or a different heap allocation, they should have the option to do so. The absolute worst offender here is `SmallVec`: the amount of `.into_iter().collect()` hurts just thinking about, and the fact that it's not reexported means no reuse because you can't name the type. `SmallVec` makes for an extremely poor choice for user-facing APIs.
- You may think that it's good to take an owned collection in cases where it's stored internally since there a copy would be needed internally otherwise. But this is actually also *a good thing*: any optimization we want to do to the data's layout is private as opposed to leaking these details as in the case of the aforementioned `SmallVec`. This also allows us to optimize much more aggressively (***foreshadowing***). Also the previous issues of ergonomics and flexibility on the user's end.

All in all, what I'm looking for most of all is the improved ergonomics on the user's end and flexibility on the user's end as well as in our internals. But of course, there is a downside, mainly for dynamic data, since this makes it so the user has to have a set of structs that borrow from another set of structs etc. in a hierarchy like in the original API. However, that's still much better ergonomics, since the biggest pain in the original API is the struct chaining.

Changelog:
```markdown
### Breaking changes

Changes to parameters:
- Where owned `Arc`s and owned structs, as well as owned collections such as `Vec`s, `SmallVec`s, `HashMap`s and `BTreeMap`s were previously taken as arguments, references and slices are used instead, respectively.

Changes to images:
- `FormatProperties` no longer has a `_ne` field and is now marked `#[non_exhaustive]` instead.

Changes to memory allocation:
- `StandardMemoryAllocator::new_default` was removed in favor of `StandardMemoryAllocator::new(&device, &Default::default())` fulfilling the same function instead.
- `MemoryRequirements` is now marked `#[non_exhaustive]`.

Changes to `GraphicsPipeline`:
- `RasterizationConservativeState` no longer has a `_ne` field.

Changes to external memory and external sync:
- Where `File`s were previously taken as arguments for importing and returned for exporting, the raw descriptor is now used instead.
```


Rua:
Looks good! A few points:
- The various caches in `PhysicalDevice` call `to_owned` before querying the cache, which is a bit unfortunate. I don't know if there's a way to avoid that.
- `AliasableBox` and `self_referential` have no documentation for why they exist and what they do.
- If those ^ belong together, then they might warrant a single module.

Rua:
Ah, that's a shame. Well, if you think of a brilliant way to do it later, you can always improve it further I suppose.

PR #2683, opened 2025-05-02
Use references as params instead of owned `Arc`s and collections [2/2]

### LINKED ISSUE #2672: Questionable public API choice wrt HashMap
I noticed that in ShaderModule context all HashMap use is using the foldhash type alias:

`pub type HashMap<K, V> = std::collections::HashMap<K, V, RandomState>;`

This could of course be seen as an implementation detail, if it wasn't for the fact that it is used in the public API contract, e.g. in `vulkano::shader::ShaderModule::specialize` etc.
This isn't clear when reading the rustdoc, as it just mentions `HashMap` making most of us think of `std::collections::HashMap`.
To me, the need of the extra performance properties of foldhash is not really evident in this use case.  Are there other reasons for this?
I question this design choice, just based on the principle of least surprise.   At least it struck me when I wanted to specialize a shader today.

Now I know changing this would constitute an API change, but I still think it's hygenic enough to warrant a change.
I could volunteer to do the changes needed, but I have no earlier experience from vulkano internals.  Not knowing the rationale for foldhash use in the first place, doesn't help either, but I guess it is warranted for more performance-critical areas, and it just happened to infect some public APIs.

Comments?

marc0246:
Follow-up to #2682. This part includes everything else (again except for anything relating to old sync as it's getting deprecated).

Fixes #2672.

Changelog:
```markdown
### Breaking changes
Changes to pipelines:
- `PipelineDescriptorSetLayoutCreateInfo` was replaced by `PipelineLayout::from_stages` and `pipeline::layout::push_constant_ranges_from_stages`.

Changes to render passes:
- `Subpass::from` was renamed to `Subpass::new`.
```


Rua:
I'm getting a validation error from Vulkano when I run the async example:

>      Running `/home/rua/code/vulkano/target/debug/async-update`
> Using device: AMD Radeon RX 7600 (RADV NAVI33) (type: DiscreteGpu)
> Using queue family 0 for graphics and queue family 1 for transfers
> 
> Press space to update part of the texture
> 
> thread 'main' panicked at examples/async-update/main.rs:549:10:
> called `Result::unwrap()` on an `Err` value: a validation error occurred
> 
> Caused by:
>     create_info: the pipeline is not being created with fragment output state, or `subpass` has a depth/stencil attachment, but `depth_stencil_state` is `None`
> 
> Vulkan VUIDs:
>     VUID-VkGraphicsPipelineCreateInfo-renderPass-06590
>     VUID-VkGraphicsPipelineCreateInfo-renderPass-06043
>     VUID-VkGraphicsPipelineCreateInfo-renderPass-06053

PR #2686, opened 2025-05-08
Reexport `ash`

hazelwiss:
I did not see this type being exposed despite it being used inside of a public field. So I assume that was by mistake.

Changelog:
```markdown
### Additions
- Exposing the `ObjectType` from ash inside of `instance/debug.rs`.
```

### DIFF
--- a/vulkano/src/instance/debug.rs
+++ b/vulkano/src/instance/debug.rs
@@ -428,13 +428,11 @@ impl<'a> Iterator for DebugUtilsMessengerCallbackLabelIter<'a> {
     }
 }
 
-pub type ObjectType = vk::ObjectType;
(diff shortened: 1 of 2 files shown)


Rua:
After thinking about this, I wonder if it makes more sense to re-export all of `ash::vk`?

hazelwiss:
I'm all for the idea of just re-exporting all of ash. I guess a clean way could be either putting it inside of an `export` module or renaming it to something like `raw` so it communicates to people unfamiliar with ash it's the raw bindings?

hazelwiss:
Yeah I mean this is great to me I just didn't know we wanted this lol. But this is good!

PR #2688, opened 2025-05-18
Remove ranges from public APIs

marc0246:
Besides this being more consistent with the original API and other parts of our own code base, ranges are also still unergonomic because they are unfortunately still not fixed in edition 2024 in the sense that they don't implement `Copy`. Now `ImageSubresourceRange` and the like finally implement `Copy`. I believe having a size/count like in the original API is also more ergonomic because that's what you have most of the time, resulting in a whole lot of `offset..offset + size` ugliness. It's also not uncommon to want to leave the offset/base at 0, which you can now leave the default for. Separate fields also provide us with the future possibility of making the size/count an `Option` for `None` to mean that it's unbounded.

I haven't *completely* removed ranges from public APIs because the APIs that do slicing don't have a Vulkan counterpart and ranges are well established for this purpose in Rust.

Changelog:
```markdown
### Breaking changes
Global changes:
- Where `Range` and `RangeInclusive` were previously used in parameters, two separate parameters for the offset/base and size/count are now used instead to match Vulkan.
```

### DIFF
--- a/vulkano/src/image/mod.rs
+++ b/vulkano/src/image/mod.rs
@@ -441,8 +441,15 @@ impl Image {
         subresource_range: ImageSubresourceRange,
     ) -> SubresourceRangeIterator {
         assert!(self.format().aspects().contains(subresource_range.aspects));
-        assert!(subresource_range.mip_levels.end <= self.mip_levels());
-        assert!(subresource_range.array_layers.end <= self.array_layers());
+        assert!(subresource_range.base_mip_level < self.mip_levels());
+        assert!(
+            subresource_range.level_count <= self.mip_levels() - subresource_range.base_mip_level,
+        );
+        assert!(subresource_range.base_array_layer < self.array_layers());
+        assert!(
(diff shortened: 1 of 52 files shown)


vulkano/src/image/mod.rs
```diff
@@ -893,16 +906,17 @@ impl SubresourceRangeIterator {
...
+        let mip_levels = subresource_range.base_mip_level
+            ..subresource_range.base_mip_level + subresource_range.level_count;
+        let array_layers = subresource_range.base_array_layer
+            ..subresource_range.base_array_layer + subresource_range.layer_count;
 
-        let next_fn = if subresource_range.array_layers.start != 0
-            || subresource_range.array_layers.end != image_array_layers
-        {
+        assert!(!mip_levels.is_empty());
+        assert!(!array_layers.is_empty());
```
Rua, line 915: This can probably be done more easily by asserting the count is not 0?

PR #2689, opened 2025-05-18
Pass `World` to the task node's clear color implementation

CptPotato:
Provides access to a task node's `World` in the `clear_color()` implementation. I ran into this when trying to dynamically change the clear color of an image at runtime.

Changelog:
```markdown
### Breaking changes
Changes to `Task` trait:
- Added a `&Self::World` parameter to `clear_values()`.
```

### DIFF
--- a/examples/async-update/main.rs
+++ b/examples/async-update/main.rs
@@ -717,7 +717,7 @@ struct RenderTask {
 impl Task for RenderTask {
     type World = RenderContext;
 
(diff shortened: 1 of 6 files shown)


CptPotato:
I'll check if it actually works in my project and remove the draft status afterwards.

CptPotato:
I tried modulating the clear color per frame and it seems to work as expected in my project.

PR #2691, opened 2025-06-02
Switch to new CI and use Nix in workflows

marc0246:
This moves us to the new and shiny CI we've been setting up. The main benefit is reliability due to the amazing new hardware. We also set up Nix for reproducibility: no more Clippy errors in CI that don't happen locally because CI has updated Rust. That means that I also removed the separate Clippy jobs because the only reason I made them separate was so that we can ignore them precisely because of the aforementioned issue.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,72 +8,61 @@ on:
     paths-ignore:
       - '**.md'
 
-env:
-  CARGO_TERM_COLOR: always
-
 jobs:
   build:
     strategy:
       matrix:
         include:
         - name: windows-latest
-          runs-on: [self-hosted, Windows]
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -8,72 +8,55 @@ on:
...
-  CARGO_TERM_COLOR: always
-
 jobs:
   build:
     strategy:
       matrix:
         include:
         - name: windows-latest
-          runs-on: [self-hosted, Windows]
+          runs-on: [self-hosted, 25.05]
```
AustinJ235, line 17:
```suggestion
          runs-on: [self-hosted, Windows, 25.05]
```
When the linux self-hosted runner is added we'll need the OS tag also.

AustinJ235 (approved):

PR #2693, opened 2025-06-07
Update shaderc to 0.9.1

Josh194:
Only breaking changes from 0.8.x should be the constructor return types of `Compiler`/`CompileOptions` changing from `Option`s to `Result`s; changes have been made accordingly (I searched for every instance of the above types and reviewed them).
Could use `map_err(|_| x)` instead of `or(Err(x))` obviously, I just personally think the latter looks cleaner.

`shaderc` 0.8.x does not build on my system (and thus probably others) whereas 0.9.x does, hence this PR.
There are further issues preventing me from running this repo's tests locally (existed before this change), but I've run them via Github actions in my fork successfully (except for windows which was never picked up by a runner; probably because it uses a self-hosted target).

Changelog:
```markdown
### Public dependency updates
- [shaderc](https://crates.io/crates/shaderc) 0.9.1
```


Rua:
Thanks!

PR #2695, opened 2025-06-12
Use not mutable reference for get_additional_image_view

mrdev023:
Mutable reference is useless to just get the additional image view and add rust complexity, I think it's just a mistake


Rua:
Thanks! I agree, there's not much reason to have `mut` here.

PR #2696, opened 2025-06-12
Clippy fixes

Kotexander:
Reintroduced some clippy workspace lints in `Cargo.toml`

clippy lints re-added:
`clippy.needless_borrowed_reference`
`clippy.nonminimal_bool`

These changes should increase code clarity.


Rua:
The `needless_borrowed_reference` fix seems good, but I'm not sure if `nonminimal_bool` increases clarity for every case. For validation checking required API versions and such, the inner expression always states what is _needed_:
- `if !(api_version >= Version::V1_1 && device.enabled_extensions().khr_foo)`
- `if !(device.enabled_extensions().khr_bar && device.enabled_features().baz)`
- `if !device.enabled_extensions().khr_foo`
- `if !(api_version >= Version::V1_1)`

I feel like rewriting that last one as `api_version < Version::V1_1` makes it less readable and less consistent with the others.

Kotexander:
Thanks for the quick replies, I undid the `nonminimal_bool` changes which also undid any autogen-out changes.

PR #2701, opened 2025-06-29
Increase MSRV to 1.82

Rua:
Changelog:
```markdown
### Public dependency updates
- Rust version: 1.82.0
```


Rua:
> May I know why you think we should be using `&raw (const|mut)` instead of `&[mut]`?

Not me, Clippy.

Rua:
`borrow_as_ptr`

PR #2705, opened 2025-06-30
Add support for `null_descriptor` device feature

ComfyFluffy:
DOCS TODO

- **option types for null descriptor**
- **update definition**
- **update validation & taskgraph**

3. [ ] Run `cargo clippy` on the changes.

4. [ ] Run `cargo +nightly fmt` on the changes.

5. [ ] Please put changelog entries **in the description of this Pull Request**

6. [ ] Describe in common words what is the purpose of this change, related

Changelog:
```markdown
### Public dependency updates
- [some_crate](https://crates.io/crates/some_crate) 1.0

### Breaking changes
Changes to `Foo`:
- Renamed to `Bar`.

### Additions
- Support for the `khr_foobar` extension.

### Bugs fixed
- `bar` panics when calling `foo`.
```

### DIFF
--- a/vulkano/src/command_buffer/commands/bind_push.rs
+++ b/vulkano/src/command_buffer/commands/bind_push.rs
@@ -175,6 +175,7 @@ impl<L> AutoCommandBufferBuilder<L> {
 
                         if let Some(buffer_info) = element {
                             let Some(buffer_info) = buffer_info else {
+                                // TODO(Comfy): Duplicated?
                                 if !self.device().enabled_features().null_descriptor {
                                     return Err(Box::new(ValidationError {
                                         problem: format!(
@@ -184,7 +185,7 @@ impl<L> AutoCommandBufferBuilder<L> {
                                             descriptor_sets_index, set_num, binding_num, index,
                                         )
                                         .into(),
-                                        vuids: &[todo!()],
+                                        // vuids?
                                         ..Default::default()
                                     }));
                                 }
--- a/vulkano/src/command_buffer/commands/bind_push.rs
+++ b/vulkano/src/command_buffer/commands/bind_push.rs
@@ -174,6 +174,23 @@ impl<L> AutoCommandBufferBuilder<L> {
                         }
 
                         if let Some(buffer_info) = element {
+                            let Some(buffer_info) = buffer_info else {
+                                if !self.device().enabled_features().null_descriptor {
+                                    return Err(Box::new(ValidationError {
+                                        problem: format!(
+                                            "the buffer of `descriptor_sets[{}]` \
+                                            (for set number {}) for binding {} index {} is \
+                                            `None`, but `null_descriptor` is not enabled",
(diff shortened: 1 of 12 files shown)


ComfyFluffy:
The primary reason I want to add null descriptor support is to use null objects in shader. For example I may provide a null acceleration structure when there is nothing to render and it just calls the miss shader. But I don't think we can currently just pass Id:INVALID to the shader?

vulkano/src/command_buffer/commands/bind_push.rs
```diff
@@ -174,6 +174,10 @@ impl<L> AutoCommandBufferBuilder<L> {
                         }
 
                         if let Some(buffer_info) = element {
+                            let Some(buffer_info) = buffer_info else {
```
Rua, line 177: You can just use `if let Some(Some(buffer_info))`.

vulkano/src/descriptor_set/mod.rs
```diff
@@ -85,10 +85,10 @@ use crate::{
         DescriptorBindingFlags, DescriptorSetLayoutCreateFlags, DescriptorType,
     },
     device::{Device, DeviceOwned},
-    image::{sampler::Sampler, ImageLayout},
+    image::ImageLayout,
     Validated, ValidationError, VulkanError, VulkanObject,
 };
-use ash::vk;
+use ash::vk::{self};
```
Rua, line 91:
```suggestion
use ash::vk;
```

vulkano/src/descriptor_set/mod.rs
```diff
@@ -505,14 +499,11 @@ impl DescriptorSetResources {
 /// The resources that are bound to a single descriptor set binding.
 #[derive(Clone, Debug)]
 pub enum DescriptorBindingResources {
-    None(Elements<()>),
-    Buffer(Elements<DescriptorBufferInfo>),
-    BufferView(Elements<Arc<BufferView>>),
-    ImageView(Elements<DescriptorImageViewInfo>),
-    ImageViewSampler(Elements<(DescriptorImageViewInfo, Arc<Sampler>)>),
-    Sampler(Elements<Arc<Sampler>>),
+    Image(Elements<DescriptorImageInfo>),
```
Rua, line 502: If `DescriptorImageInfo` is not wrapped in an `Option`, why is `DescriptorBufferInfo` wrapped in an `Option`? Seems inconsistent.

ComfyFluffy: Since `DescriptorImageInfo`'s fields `sampler` and `image_view` are both wrapped in `Option`, `DescriptorImageInfo::default()` effectively acts as null value, while `DescriptorBufferInfo` holds a `Subbuffer` which is not optional.

However, after looking into `DescriptorBufferInfo`, it seems redundant to hold `Subbuffer` + `offset` + `range` in it, as we can just get `offset` and `range` from `Subbuffer`. So we maybe can just replace Option<DescriptorBufferInfo> with `Option<Subbuffer>`?

PR #2708, opened 2025-07-05
Fix swapchain image IDs no worky in `ResourceAccesses::image_unchecked`

marc0246:


Rua:
What made them not work? (Just for my own understanding)

PR #2714, opened 2025-08-04
Fixed the gl-interop example

andesyv:
Hi!,
I was experimenting and learning about Vulkan <-> OpenGL interoperability and I was sad to see that the gl-interop example has been broken for a while now due to some outdated transitive dependencies (See #2571 and 1da590f). So I've taken it upon myself to try to fix this by updating the `glium` crate.

I've tested the example with Wayland, running on a NVIDIA GPU.

 - [x] Run `cargo clippy` on the changes.
 - [x] Run `cargo +nightly fmt` on the changes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -30,6 +30,9 @@
             vulkan-tools-lunarg
             vulkan-validation-layers
 
+            # OpenGL dependencies (needed for the gl-interop example)
+            libGL
+
             # winit dependencies
             libxkbcommon
             wayland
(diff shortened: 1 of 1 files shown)


andesyv:
Thanks for taking the time to review my PR!

andesyv:
I think this might be a NixOS or `glutin` specific problem. On Linux, `glutin` defaults to using EGL for initializing the window surface, which `glium` later binds its context onto. And from what I can tell from this error message, the only place where this error could be tossed would be from `glutin::api::egl::display::Display::new` where it tries to dynamically load the `libEGL.so.1` or `libEGL.so` file and returns this `NotFound` error if it couldn't find or load them.

https://github.com/rust-windowing/glutin/issues/1650 seems to mention this same issue.

PR #2766, opened 2026-01-31
Fix format aspects validation for DRM format modifier images

cadubentzen:

   When calling `subresource_layout()` on a `RawImage` that uses DRM format modifiers (`ImageTiling::DrmFormatModifier`), the validation would fall through into the format-aspects checks after correctly validating the `MemoryPlane` aspect. This caused spurious validation errors because DRM format modifier images use `MemoryPlane` aspects, not the usual format aspects (`Color`, `Depth`, `Stencil`, `Plane0`, etc.).

   The fix adds an extra `self.tiling != ImageTiling::DrmFormatModifier` around the validation which was previously failing.

Changelog:
```markdown
### Bugs fixed
- `RawImage::subresource_layout` validation incorrectly rejected valid `MemoryPlane` aspects for DRM format modifier images.
```

### DIFF
--- a/vulkano/src/image/sys.rs
+++ b/vulkano/src/image/sys.rs
@@ -1409,7 +1409,8 @@ impl RawImage {
         // VUID-vkGetImageSubresourceLayout-format-04464
         // VUID-vkGetImageSubresourceLayout-format-01581
         // VUID-vkGetImageSubresourceLayout-format-01582
-        if !format_aspects.contains(aspect.into()) {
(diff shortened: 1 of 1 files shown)


marc0246:
Thank you! However, an early return in validation is something we don't do. And I wouldn't want to start doing it because it's inevitably going to be overlooked in the future. There even is commented-out validation later down pertaining to DRM format modifier. So instead, please put the other validation later down in an `if self.tiling != ImageTiling::DrmFormatModifier` block instead.

cadubentzen:
Thanks for the review! Done.

marc0246:
Thank you for the fix!

PR #2772, opened 2026-02-17
Fix invalid virtual ID check

Matusson:

5. [X] Please put changelog entries **in the description of this Pull Request**

6. [X] Describe in common words what is the purpose of this change, related

This PR addresses a minor bug/confusion point which can arise when invalid virtual image ID is inserted into the resource map. As `insert_image()` asserts the object type before `virtual_resources.get()`, if the virtual ID happens to be invalid, then it will trigger the `unreachable!()` condition in the match statement in `object_type()`. This is a bit confusing. I fixed this by moving the assertion after `virtual_resources.get()`. In addition I added error documentation and an additional assert in `object_type()` to hopefully clear this up in the future. Let me know if this is right.

Changelog:
```markdown
### Bugs fixed
- ResourceMap::insert_image() now correctly returns an error with invalid virtual_id instead of panicking
```

### DIFF
--- a/vulkano-taskgraph/src/graph/execute.rs
+++ b/vulkano-taskgraph/src/graph/execute.rs
@@ -2492,6 +2492,10 @@ impl<'a> ResourceMap<'a> {
     ///
     /// - Panics if the physical resource doesn't match the virtual resource.
     /// - Panics if the physical resource already has a mapping from another virtual resource.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if either `virtual_id` or `physical_id` is invalid.
     #[inline]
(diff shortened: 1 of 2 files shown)


marc0246:
I had already fixed the underlying issue in a local branch of mine. That being the fact that `Id<Buffer>`, `Id<Image>` and `Id<Swapchain>` are separate types. They will be replaced by `ResourceId`. So you can wait a little bit, or if you don't want to, I can merge this (with some changes).

Matusson:
Nah, no need. I PR'd this since it caused me some confusion, if there's a better solution coming then it's unnecessary.

PR #2773, opened 2026-02-22
Emit type alias for runtime array element types in generated shader structs

nshyrei:
This PR acts as a proposal for a maintainability issue when working with vulkano-generated shader structs containing runtime arrays.
##
Issue:

When creating GPU buffers for shader structs, the padding information differs between singular types and array elements due to std430 layout rules.
For example, given this GLSL:
```
struct SurfaceProperty {
      <fields>...
};

layout(set = 0, std430, binding = 0) buffer SurfaceProperties {
      SurfaceProperty surface_properties[];
};
```
Vulkano generates:
```
pub struct SurfaceProperty {
      <fields>...
}

pub struct SurfaceProperties {
      pub surface_properties: [Padded<SurfaceProperty, 12>],
}
```
When allocating a buffer with `vulkano::buffer::Buffer::new_slice`, using an array of `SurfaceProperty` directly produces incorrect std430 stride, while the array wrapper type has the correct stride. However, the correctly-padded element type `Padded<SurfaceProperty, 12>` is anonymous — to use it, I must inspect the generated code with `cargo expand` and recreate the padding myself in my Rust code.
##
Proposal:

Emit a type alias for the padded element type alongside each struct containing a runtime array:
```
pub type SurfacePropertiesElement = Padded<SurfaceProperty, 12>;
```
This allows users to write `Buffer<[SurfacePropertiesElement]>` with correct std430 layout without hardcoding padding or inspecting generated code.

### DIFF
--- a/vulkano-shaders/src/structs.rs
+++ b/vulkano-shaders/src/structs.rs
@@ -132,7 +132,21 @@ pub(super) fn write_structs(
             #[derive(::vulkano::buffer::BufferContents #(, #custom_derives )* )]
             #[repr(C)]
             #struct_ser
-        })
+        });
+
+        // Emit type aliases for runtime array element types
+        for member in &struct_ty.members {
+            if let Type::Array(ty) = &member.ty {
+                let alias_ident = format_ident!("{}Element", struct_ty.ident);
+                let element_type = array_element_type(&ty, input);
+
+                if let None = ty.length.map(NonZeroUsize::get) {
+                    structs.extend(quote! {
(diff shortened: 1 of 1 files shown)


marc0246:
Thank you for the PR. However, you're supposed to use `Buffer::new_unsized::<SurfaceProperties>`. Your solution also doesn't work for structs such as this, which makes it feel too ad-hoc:
```glsl
buffer Foo {
    float x;
    float y;
    uint array[];
};
```

PR #2775, opened 2026-03-13
Make raw-window-handle an optional dependency

Nikitf777:

   * Make `raw-window-handle` an optional dependency

This PR makes the `raw-window-handle` dependency optional. This is useful because there are a few use cases where this dependency is simply not needed:
* Compute shaders
* Off-screen rendering
* Device info querying

The dependency is now hidden begind a corresponding feature called "raw_window_handle". This feature is enabled by deafult, so it shouldn't break anything.
Without this feature, and the following functions are not included:
* `PhysicalDevice::presentation_support()`,
* `Surface::required_extensions()`,
* `Surface::from_window()`,
* `Surface::from_window_ref()`,
* `SurfaceApi::try_from()`;
and the whole `FromWindowError` enum.

Changelog:
```markdown

### Feature list updates:
- A new feature is introduced: `raw_window_handle`.
```

### DIFF
--- a/vulkano/src/lib.rs
+++ b/vulkano/src/lib.rs
@@ -21,7 +21,7 @@
 //!    example, if you use a `winit` version that uses a different version from the one vulkano
 //!    uses, you can add one of the [features](https://docs.rs/crate/winit/latest/features) that
 //!    starts with `rwh` to `winit`. Currently, vulkano is compatible with `rwh_06`.
-//!    
+//!
 //! 4. [Enumerate the physical devices] that are available on the `Instance`, and choose one that
 //!    is suitable for your program. A [`PhysicalDevice`] represents a Vulkan-capable device that
 //!    is available on the system, such as a graphics card, a software implementation, etc.
@@ -96,6 +96,7 @@
 //! | `x11`                | Support for X11 platforms. Enabled by default.                 |
 //! | `document_unchecked` | Include `_unchecked` functions in the generated documentation. |
 //! | `serde`              | Enables (de)serialization of certain types using [`serde`].    |
+//! | `raw_window_handle`  | Include the raw-window-handle dependency and its dependents.   |
 //!
 //! [`Instance`]: instance::Instance
 //! [`Surface`]: swapchain::Surface
--- a/vulkano/src/lib.rs
+++ b/vulkano/src/lib.rs
@@ -90,13 +90,13 @@
 //!
 //! # Cargo features
 //!
-//! | Feature              | Description                                                    |
-//! |----------------------|----------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


vulkano/src/lib.rs
```diff
@@ -95,7 +95,8 @@
 //! | `macros`             | Include reexports from [`vulkano-macros`]. Enabled by default. |
 //! | `x11`                | Support for X11 platforms. Enabled by default.                 |
 //! | `document_unchecked` | Include `_unchecked` functions in the generated documentation. |
-//! | `serde`              | Enables (de)serialization of certain types using [`serde`].    |
+//! | `serde`
+//! | `raw_window_handle`  | Include the raw-window-handle dependency and its dependents.   |              | Enables (de)serialization of certain types using [`serde`].    |
```
CptPotato: Looks like the serde row got broken here.

Nikitf777: You're right. I'm gonna fix this.

marc0246:
Thank you for the PR, and I'm sorry it has taken so long for you to get a review. Maintainer bandwidth isn't great at the moment. I fixed my nitpicks myself, and this LGTM.

The reason there wasn't a feature for this previously is that raw-window-handle is a micro-crate, and I didn't think it mattered. But if it matters to you, I see no reason not to have the feature. But note that there is no way to make previously ungated code gated behind a feature without a breaking change. A downstream crate can use `default-features = false`, and this doesn't opt out of semver AFAIK. This is especially evident if you consider downstream library crates; these have to use `default-features = false`, otherwise the user would be forced into the default features.

Nikitf777:
The reason I did that is because I submitted a PR into another project where I needed vulkano to only query hardware information, without rendering or computing anything. And the PR was rejected because the author didn't want to add too many dependencies for such a simple task. I could use `ash`, but I requires using `unsafe` blocks, and inserting these blocks into mostly safe code is, in my opinion, not the best idea. You should properly wrap them in a different place, but at this point you're just reimplementing vulkano. So I wanted to try to make the library as lightweight as possible for such simple use cases. Does this make sense? And is it even possible? I mean, how many other crates can become optional?

marc0246:
That makes perfect sense then, if this is just the lowest hanging fruit in a greater goal. And I respect it.

I have my own ideas for how to entirely remove &ndash; not feature-gate &ndash; many different dependencies in different ways. In particular, `smallvec` can be replaced with `ArrayVec` in many cases (which is extremely easy to implement in-house, and we sort of already did) and in other cases with a different solution that is also &ndash; like `ArrayVec` &ndash; more optimal, which I also want to develop myself. But this will require *a lot* of internal changes, and I urge you to leave this one to me. You should probably eye `raw-window-metal` next (this one will require a feature gate). Also, you should join our Discord for easier coordination.

PR #2778, opened 2026-04-11
Add protected support to AutoCommandBufferBuilder

nathan-google:

Changelog:
```markdown
### Additions
- Support for AutoCommandBuffer protected Vulkan workflows.
```

Hi reviewers. I have been using Vulkano for an Android protected Vulkan pipeline workflow. The following commits unblock use of protected Vulkan and add unit tests including an end-to-end test which passes on a dev-mode Lenovo Chromebook Plus 14". All unit tests also pass on a device with no `protected_memory` vulkan support.

There are some design decisions that I made that are worth pointing out.

1. Modifying StandardCommandBufferAllocator is not strictly necessary, but to avoid changing it, users need to re-implement a lot of code. The second commit lets the user specify the pool flags when using the StandardCommandBufferAllocator wtihout many changes.

2. In the third commit, I am adding a way to, at SubmitInfo time determine if any CommandBuffers were protected. Another approach would be to add a "use protected" metadata in the SubmitInfo structs. The user would have to do additional tracking to handle that. The most visible side effect of the is the addition of an `is_protected()` function to the `PrimaryCommandBufferAbstract`. Maybe there is a better way to handle getting the protected status at submission time.

3. I split up the changes into 3 commits for ease of review, but I am happy to let them squash. :)

Thanks for taking a look!

### DIFF
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -363,13 +363,11 @@ impl RawBuffer {
                         the `property_flags` of the memory type of `allocation.device_memory()` \
                         does not contain `MemoryPropertyFlags::PROTECTED`"
                         .into(),
-                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01899"],
+                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01898"],
                     ..Default::default()
                 }));
             }
-        }
-
-        if !self.flags.intersects(BufferCreateFlags::PROTECTED) {
+        } else {
             if memory_type
                 .property_flags
                 .intersects(MemoryPropertyFlags::PROTECTED)
@@ -379,7 +377,7 @@ impl RawBuffer {
                         the `property_flags` of the memory type of `allocation.device_memory()` \
                         contains `MemoryPropertyFlags::PROTECTED`"
                         .into(),
-                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01900"],
+                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01899"],
                     ..Default::default()
                 }));
             }
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -353,17 +353,36 @@ impl RawBuffer {
             }
         }
 
-        if memory_type
-            .property_flags
-            .intersects(MemoryPropertyFlags::PROTECTED)
-        {
-            return Err(Box::new(ValidationError {
-                problem: "the `property_flags` of the memory type of \
-                    `allocation.device_memory()` contains `MemoryPropertyFlags::PROTECTED`"
-                    .into(),
-                vuids: &["VUID-VkBindBufferMemoryInfo-None-01899"],
-                ..Default::default()
-            }));
+        if self.flags.intersects(BufferCreateFlags::PROTECTED) {
+            if !memory_type
+                .property_flags
+                .intersects(MemoryPropertyFlags::PROTECTED)
+            {
+                return Err(Box::new(ValidationError {
+                    problem: "the buffer was created with `BufferCreateFlags::PROTECTED`, but \
+                        the `property_flags` of the memory type of `allocation.device_memory()` \
+                        does not contain `MemoryPropertyFlags::PROTECTED`"
+                        .into(),
+                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01899"],
+                    ..Default::default()
+                }));
+            }
+        }
+
+        if !self.flags.intersects(BufferCreateFlags::PROTECTED) {
+            if memory_type
+                .property_flags
+                .intersects(MemoryPropertyFlags::PROTECTED)
+            {
+                return Err(Box::new(ValidationError {
+                    problem: "the buffer was not created with `BufferCreateFlags::PROTECTED`, but \
+                        the `property_flags` of the memory type of `allocation.device_memory()` \
+                        contains `MemoryPropertyFlags::PROTECTED`"
+                        .into(),
+                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01900"],
+                    ..Default::default()
+                }));
+            }
         }
 
         if !memory.export_handle_types().is_empty() {
--- a/vulkano/src/buffer/sys.rs
+++ b/vulkano/src/buffer/sys.rs
@@ -829,6 +829,57 @@ impl<'a> BufferCreateInfo<'a> {
             }
         }
 
+        if flags.intersects(BufferCreateFlags::PROTECTED) {
+            if !device.enabled_features().protected_memory {
+                return Err(Box::new(ValidationError {
+                    context: "flags".into(),
+                    problem: "contains `BufferCreateFlags::PROTECTED`".into(),
+                    requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
+                        "protected_memory",
+                    )])]),
+                    vuids: &["VUID-VkBufferCreateInfo-flags-01887"],
+                }));
+            }
+
+            // If VK_BUFFER_CREATE_SPARSE_ALIASED_BIT support is added, then it should be added to
+            // this validation.
+            if flags
+                .intersects(BufferCreateFlags::SPARSE_BINDING | BufferCreateFlags::SPARSE_RESIDENCY)
+            {
+                return Err(Box::new(ValidationError {
+                    context: "flags".into(),
+                    problem: "contains `BufferCreateFlags::PROTECTED`, but also \
+                        contains one of the sparse memory flags"
+                        .into(),
+                    vuids: &["VUID-VkBufferCreateInfo-None-01888"],
+                    ..Default::default()
+                }));
+            }
+
+            // If VK_BUFFER_USAGE_2_VIDEO_DECODE_SRC_BIT_KHR,
+            // VK_BUFFER_USAGE_2_VIDEO_ENCODE_DST_BIT_KHR,
+            // or VK_BUFFER_USAGE_2_DESCRIPTOR_HEAP_BIT_EXT get added, then they should be added to
+            // the allowed list.
+            if !(usage
+                - (BufferUsage::TRANSFER_SRC
+                    | BufferUsage::TRANSFER_DST
+                    | BufferUsage::UNIFORM_TEXEL_BUFFER
+                    | BufferUsage::STORAGE_TEXEL_BUFFER
+                    | BufferUsage::UNIFORM_BUFFER
+                    | BufferUsage::STORAGE_BUFFER
+                    | BufferUsage::SHADER_DEVICE_ADDRESS))
+                .is_empty()
+            {
+                return Err(Box::new(ValidationError {
+                    problem: "contains `BufferCreateFlags::PROTECTED`, but also contains \
+                        usages other than [allowed list]"
+                        .into(),
+                    vuids: &["VUID-VkBufferCreateInfo-flags-09641"],
+                    ..Default::default()
+                }));
+            }
+        }
+
         if flags.intersects(BufferCreateFlags::SPARSE_BINDING) {
             if !device.enabled_features().sparse_binding {
                 return Err(Box::new(ValidationError {
@@ -958,8 +1009,8 @@ pub(crate) struct BufferCreateInfoExtensionsVk {
 
 #[cfg(test)]
 mod tests {
-    use super::{BufferCreateInfo, BufferUsage, RawBuffer};
-    use crate::device::DeviceOwned;
+    use super::{BufferCreateFlags, BufferCreateInfo, BufferUsage, RawBuffer};
+    use crate::{device::DeviceOwned, Validated};
 
     #[test]
     fn create() {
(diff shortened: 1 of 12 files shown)


nathan-google:
Hello. I am hoping to see if there is a maintainer that is available for reviewing this pull request, or who can share an ETA for when they would be available to review it. Thanks!

marc0246:
I'll review within the next 3 days. Your PR is going to take me much longer than the others' because I have to read up on protected memory.

nathan-google:
Thanks, marc0246.

In order to get the protected pipeline working for me, these were the following differences from a "normal" pipeline:

1. There needs to be a physical device with `protectedMemory` set to true.
2. Queue is created with `VK_DEVICE_QUEUE_CREATE_PROTECTED_BIT`  set and `VK_QUEUE_PROTECTED_BIT` set.
3. Command buffer pool is created with `VK_COMMAND_POOL_CREATE_PROTECTED_BIT` set.
4. Vulkan image created with `VK_IMAGE_CREATE_PROTECTED_BIT`, or  Vulkan buffers created with `VK_BUFFER_CREATE_PROTECTED_BIT` set.
5. Command buffers are submitted with `protectedSubmit` set to true in the `VkProtectedSubmitInfo` structure which is linked in the `pNext`  linked list of VkSubmitInfo`.

These were sufficient for me and the device that I am testing with (a Lenovo Chromebook Plus 14). I am not a Vulkan expert, though, so your review will be more than welcome.

marc0246:
Thanks, but I don't have any devices to test this with anyway :/ And I doubt we have anyone else, but I'll ask. So we'll most likely have to trust your testing alone.

vulkano/src/buffer/sys.rs
```diff
@@ -353,17 +353,36 @@ impl RawBuffer {
...
+            if !memory_type
+                .property_flags
+                .intersects(MemoryPropertyFlags::PROTECTED)
+            {
+                return Err(Box::new(ValidationError {
+                    problem: "the buffer was created with `BufferCreateFlags::PROTECTED`, but \
+                        the `property_flags` of the memory type of `allocation.device_memory()` \
+                        does not contain `MemoryPropertyFlags::PROTECTED`"
+                        .into(),
+                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01899"],
```
marc0246:
```suggestion
                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01898"],
```

vulkano/src/buffer/sys.rs
```diff
@@ -353,17 +353,36 @@ impl RawBuffer {
...
+                    ..Default::default()
+                }));
+            }
+        }
+
+        if !self.flags.intersects(BufferCreateFlags::PROTECTED) {
```
marc0246: nit:
```suggestion
        } else {
```

vulkano/src/buffer/sys.rs
```diff
@@ -353,17 +353,36 @@ impl RawBuffer {
...
+            if memory_type
+                .property_flags
+                .intersects(MemoryPropertyFlags::PROTECTED)
+            {
+                return Err(Box::new(ValidationError {
+                    problem: "the buffer was not created with `BufferCreateFlags::PROTECTED`, but \
+                        the `property_flags` of the memory type of `allocation.device_memory()` \
+                        contains `MemoryPropertyFlags::PROTECTED`"
+                        .into(),
+                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01900"],
```
marc0246:
```suggestion
                    vuids: &["VUID-VkBindBufferMemoryInfo-None-01899"],
```

vulkano/src/command_buffer/allocator.rs
```diff
@@ -178,6 +178,7 @@ pub struct StandardCommandBufferAllocator {
     // Each queue family index points directly to its entry.
     pools: ThreadLocal<SmallVec<[UnsafeCell<Option<Entry>>; 8]>>,
     buffer_count: [usize; 2],
+    _command_pool_create_flags: CommandPoolCreateFlags,
```
marc0246: Fields/variables starting with an underscore are meant for ones that aren't read, but this one is.

vulkano/src/command_buffer/traits.rs
```diff
@@ -38,6 +38,9 @@ pub unsafe trait PrimaryCommandBufferAbstract:
     /// Returns the usage of this command buffer.
     fn usage(&self) -> CommandBufferUsage;
 
+    /// Returns true if this command buffer was created in a protected pool.
+    fn is_protected(&self) -> bool;
```
marc0246: This seems to duplicate the functionality of `.as_raw().is_protected()`, and as it's a breaking change, I would rather not have it.

marc0246 (commented):
You can commit my suggestions directly from the GitHub interface; there's no need to rebase, and you can create as many commits as you like. We squash-merge anyway.

marc0246:
Additionally, there's a broader concern I have in that we like to stay as close to the raw API as possible so that we are as forwards-compatible as possible as well as less surprising for people coming from raw Vulkan or vice versa, and we also generally don't like implicit behavior. To that end, I would prefer if you added `SubmitInfo::protected_submit` and validated it instead. Other than the above, your PR looks good :+1:

vulkano/src/command_buffer/traits.rs
nathan-google: Ah. Gotcha. Dropping it. For some reason, I thought that it was needed for checking the `is_protected()` status when creating `vk::SubmitInfo`, but I was wrong. Thanks for pointing is out.

nathan-google:
Quick question regarding `SubmitInfo::protected_submit`.

Would you prefer a struct like

```
pub struct ProtectedSubmitInfo {
    pub protected_submit: bool,
    pub _ne: crate::NonExhaustive<'static>,
}
```

or just a bool. The struct would be closer to the spec, but also may be unnecessarily complex. Thanks!

marc0246:
Instead of chaining structs as a form of extension, we extend the main struct with more fields. So one `pNext` chain becomes one nonexhaustive struct. When one struct holds a pointer to another struct (i.e. not part of the `pNext` chain), that's when we also have a reference to another struct (which can itself be nonexhaustive if the original has a `pNext` field). That's one of the differences of our API compared to the original. The other differences obviously being that we use idiomatic Rust types and have safe versions of functions. I won't go into all of the details here because it's not relevant. I need to finally write that contribution guide where everything is going to be explained in depth.

vulkano/src/buffer/sys.rs
nathan-google: Done.

vulkano/src/buffer/sys.rs
nathan-google: Done.

vulkano/src/buffer/sys.rs
nathan-google: Done.

vulkano/src/command_buffer/allocator.rs
nathan-google: Whoops. Sorry about that. Done.

nathan-google:
Thanks for the reviews. I added a structure ProtectedSubmitInfo to SubmitInfo (and OldSubmitInfo) and added validations like you suggested. If a user directly creates a SubmitInfo, nothing automatic happens with inferring whether to create the extension structure.

The only place where the value is inferred is in `CommandBufferExecFuture::build_submission_impl()`, which I think is only used by the automatic builder.

Let me know if this is not what you were thinking!

nathan-google:
Is a maintainer available to look at the latest commits in the PR? I _think_ this addresses marc0246's requests, but I could have misunderstood something.

marc0246:
Don't worry, I didn't forget about you :slightly_smiling_face: I will make time to review again tomorrow. I have to do a more thorough review because I was in a hurry last time.

marc0246:
> I added a structure ProtectedSubmitInfo to SubmitInfo

Sorry, I explained the reasoning, but neglected to actually answer the question; that's on me :sweat_smile: You should not add a struct for this. The reason is that the original API doesn't have a struct for this either (it's part of the `pNext` chain, and as such an extension to the original struct). It should be a `bool` field.

Additionally, on closer inspection, there is missing validation for the following VUIDs:
- [VUID-VkBindImageMemoryInfo-None-01901](https://docs.vulkan.org/refpages/latest/refpages/source/VkBindImageMemoryInfo.html#VUID-VkBindImageMemoryInfo-None-01901) and [VUID-VkBindImageMemoryInfo-None-01902](https://docs.vulkan.org/refpages/latest/refpages/source/VkBindImageMemoryInfo.html#VUID-VkBindImageMemoryInfo-None-01902): the same adjustment you did for buffers needs to be done for images as well. That is, in `RawImage::bind_memory`.
- [VUID-VkBufferCreateInfo-flags-01887](https://docs.vulkan.org/refpages/latest/refpages/source/VkBufferCreateInfo.html#VUID-VkBufferCreateInfo-flags-01887), [VUID-VkBufferCreateInfo-None-01888](https://docs.vulkan.org/refpages/latest/refpages/source/VkBufferCreateInfo.html#VUID-VkBufferCreateInfo-None-01888) and [VUID-VkBufferCreateInfo-flags-09641](https://docs.vulkan.org/refpages/latest/refpages/source/VkBufferCreateInfo.html#VUID-VkBufferCreateInfo-flags-09641): these need to be added in `BufferCreateInfo::validate`.
- [VUID-VkImageCreateInfo-flags-01890](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageCreateInfo.html#VUID-VkImageCreateInfo-flags-01890) and [VUID-VkImageCreateInfo-None-01891](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageCreateInfo.html#VUID-VkImageCreateInfo-None-01891): these need to be added in `ImageCreateInfo::validate`.

nathan-google:
Thanks! The explanation about why pNext makes more sense now. Luckily switching back to a bool is a fast change.

Speaking of speed, I am working through adding the validations and _slowly_ adding unit tests for them. I'll let you know when the changes are ready for review again.

nathan-google:
Hi marc0246. Sorry about the delay . I added commits to address the bool data type suggestion and for each of the `RawImage::bind_memory`, `BufferCreateInfo::validate`, and `ImageCreateInfo::validate` functions.

I also added unit tests for most of the new validations. I didn't add them for [VUID-VkBufferCreateInfo-flags-01887](https://docs.vulkan.org/refpages/latest/refpages/source/VkBufferCreateInfo.html#VUID-VkBufferCreateInfo-flags-01887) or [VUID-VkImageCreateInfo-flags-01890](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageCreateInfo.html#VUID-VkImageCreateInfo-flags-01890) since opening a device without a feature is a little cumbersome. Let me know if you want me to add these.

I also re-ran the format and clippy checks.

vulkano/src/command_buffer/mod.rs
```diff
@@ -1013,6 +1020,27 @@ impl SubmitInfo<'_> {
...
                 .validate(device)
                 .map_err(|err| err.add_context(format!("command_buffers[{}]", index)))?;
+
+            if !protected_submit && command_buffer_submit_info.command_buffer.is_protected() {
+                return Err(Box::new(ValidationError {
+                    context: format!("command_buffer[{}]", index).into(),
+                    problem: "is a protected command_buffer, but the protected structure has protected_submit set to false.".into(),
+                requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
+                    "protected_memory",
+                )])]),
```
marc0246:
```suggestion
                    problem: "is a protected command buffer, but `protected_submit` is `false`"
                        .into(),
                    requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
                        "protected_memory",
                    )])]),
```

vulkano/src/command_buffer/mod.rs
```diff
@@ -1013,6 +1020,27 @@ impl SubmitInfo<'_> {
...
+                    vuids: &["VUID-VkSubmitInfo-pNext-04120"],
+                }));
+            } else if protected_submit && !command_buffer_submit_info.command_buffer.is_protected()
+            {
+                return Err(Box::new(ValidationError {
+                    context: format!("command_buffer[{}]", index).into(),
+                    problem: "is an unprotected command_buffer, but the protected structure has protected_submit set to true.".into(),
+                requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
+                    "protected_memory",
+                )])]),
```
marc0246:
```suggestion
                    problem: "is an unprotected command buffer, but `protected_submit` is `true`"
                        .into(),
                    requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
                        "protected_memory",
                    )])]),
```

marc0246 (commented):
You can commit suggestions directly from the GitHub interface. It saves you time.

marc0246:
(You also have merge conflicts.)

nathan-google:
Thanks for the suggestions

> (You also have merge conflicts.)

Fixing these now. I pushed a change before finishing testing, so there is going to be some churn for the next few minutes. :(

I'll let you know when it is ready again.

nathan-google:
Aaaaand. I think the PR is in good shape again. I rebased onto the latest master commit, fixed the conflicts (which were isolated to `vulkano/src/buffer/sys.rs`. I also fixed some unit tests after the change which introduced `try_` function variants for getting Validated return values.

I re-ran cargo unit tests, `cargo +nightly fmt`, and `cargo clippy`.

marc0246:
Yeehaw! Thank you for the PR and your patience :heart:

Issue #2622, opened 2025-01-15
vulkano-shaders: Support compiling Slang source


Molive-0:
Currently vulkano-shaders only supports what `shaderc` can compile. It would be nice if it could also compile Slang shaders from either inline source or a path to a file.

nshyrei:
`slangc` is a part of Vulkan SDK, would adding support for a Slang language in `vulkano-shaders` be a good feature to work on? @Rua , @rhythm1705

marc0246:
Yes. You should start by adding a `lang` field to the macro as can be seen [here](https://github.com/vulkano-rs/vulkano/pull/2568/changes/075e9812d89c7384d7cc31a2284755dfc95122a0).

nshyrei:
@marc0246 I didn't understand, should I use the branch you shared? I see that the last commit was done previous year, it looks abandoned to me.

marc0246:
You can use that that specific commit or copy the changes. Whatever works for you.

PR #2791, opened 2026-04-30
Add lang option to vulkano_shaders::shader! macro for source language selection

nshyrei:
Adds `source_language` field to `MacroInput` to support compiling shaders in multiple languages (e.g. HLSL, Slang), motivated by https://github.com/vulkano-rs/vulkano/issues/2622. Changes adapted from https://github.com/vulkano-rs/vulkano/pull/2568 as suggested by @marc0246.

Currently adds HLSL as the first supported language variant, which is passed to `shaderc` during shader compilation.

Closes #2568

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -1,7 +1,4 @@
-use crate::{
-    structs::{self, TypeRegistry},
-    MacroInput,
-};
+use crate::{structs::{self, TypeRegistry}, MacroInput, SourceLanguage};
 use heck::ToSnakeCase;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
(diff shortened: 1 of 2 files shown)


CptPotato:
Side note: In order to support bindless for HLSL we'd need an equivalent of [`vulkano.glsl`](https://github.com/vulkano-rs/vulkano/blob/0551bb2582501eef1529ef7f2ebca19311e656a7/vulkano-shaders/include/vulkano.glsl). I think this can be tackled separately, though.

nshyrei:
@CptPotato  I can remove the call to `shaderc` with HLSL if it produces unsupported features.

marc0246:
No no, it's fine. `vulkano.hlsl` should be a separate concern, not one for this PR.

marc0246:
You need to run `cargo +nightly fmt`.

nshyrei:
@marc0246 done.

marc0246:
Thank you for the PR :heart: Next, you can add the `slang` lang option. You could use libslang like we use libshaderc, but I think it's best to move away from this. We don't need to link against anything because this is a proc-macro... It doesn't end up in the binary. And a program invocation is much easier to work with and less set up for the user.

nshyrei:
@marc0246 You mean you want to call `libshaderc` and `libslang` as a subprocess in Rust directly instead of using a `shaderc` crate (and some equivalent for slang) ?

marc0246:
I mean not using libshaderc or libslang. Because those are libraries that we need to link against, which is more set up for the user, and not necessary because it doesn't end up in the binary anyway. Instead, I would like us to use `Command::new("glslc")` /  `Command::new("slangc")`.

PR #2793, opened 2026-05-07
Use actual alerts for notes

marc0246:
This adds custom CSS for rustdoc in order to make our notes look like actual alerts. The [GitHub ones](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) look good, so I decided to make ours look just like them except that the headings, font, and colors match the rustdoc theme in use.

Right now, our notes look like this:
Our old alerts

They're barely noticeable, and the point of notes is after all to be well, noticeable. This is a problem worth solving.

Unfortunately, rustdoc doesn't have a native solution to this, so I went with custom rustdoc CSS. When they do add something like GitHub's alerts, we can always switch to that.

This is how our new alerts look (ayu theme):
Our new alerts

And this is how GitHub's look (dark theme):
GitHub's alerts

As you can see, they look pretty much the same, just that ours fit in with the rest of the rustdoc theme that's in use.

If you want to verify that it works, you need to run `cargo rustdoc -- --extend-css assets/rustdoc.css` from within the crate's directory (i.e. the `vulkano`, `vulkano-taskgraph`, etc. directory). Alternatively, you can run `cargo rustdoc -p <package name> -- --extend-css assets/rustdoc.css` from the root of the repo. It doesn't *break* without the custom rustdoc CSS, it just looks like a regular blockquote.


AustinJ235 (approved):
Looks very nice!

PR #2794, opened 2026-05-07
Update triangle example to use the task graph

CptPotato:
Updates the `triangle` example to use the task graph.

The changes follow the `deferred` example. Colors were adjusted, as discussed on Discord.


CptPotato:
I kept the previous comments where applicable, but haven't added any new ones. So this still needs another pass in terms of documentation.

But other than comments, this one is ready for review.

examples/triangle/main.rs
```diff
@@ -58,20 +60,242 @@ struct App {
...
     recreate_swapchain: bool,
-    previous_frame_end: Option<Box<dyn GpuFuture>>,
+    task_graph: ExecutableTaskGraph<Self>,
+    virtual_swapchain_id: Id<Swapchain>,
+}
+
+struct TriangleTask {
+    pipeline: Option<Arc<GraphicsPipeline>>,
+    vertex_buffer_id: Id<Buffer>,
+    swapchain_id: Id<Swapchain>,
```
CptPotato: I first tried to use `target: Id<Image>` here instead of directly referring to the swapchain. This seemed cleaner, since the triangle task technically doesn't need to know that the target is a swapchain.

But didn't get it to work after some trial and error. Not sure if this is possible atm.

marc0246: `<Id<Swapchain>>::current_image_id` just casts the type parameter if you look at the implementation. This is but one of the reasons that the type parameter doesn't make sense.

marc0246:
Do you think rgb(251, 113, 133) looks the best for the trongle? :thinking:

marc0246:
First thing that jumps out is that the `TriangleTask` struct and its impls should come after all the `App` and `RenderContext` structs and impls. We always order free-standing functions as well as structs and their impls from outermost to innermost. I suggest you fix this ASAP because it matters a lot for when you write the comments.

CptPotato:
This took longer than anticipated, I'm not much of a writer I guess :upside_down_face:. But I'm mostly happy with the example now. I'll give it another read soon in case I left unfinished passages in or something doesn't make sense, then it should be ready.

> Do you think rgb(251, 113, 133) looks the best for the trongle? 🤔

I liked this the most out of the options, but I'm actually on the fence whether or not this should be the classic RGB Triangle.

examples/triangle/main.rs
marc0246: I know you didn't write this, but it has been long out of date. I don't even know what "helper functions" it's referring to. The examples all look the same in terms of code, just that this example is the only one that comments every step.

marc0246, line 138: This too is out of date. We don't check for presentation support for a specific surface anymore, rather presentation support for any surface. So maybe "presenting images to *a* surface."

marc0246: Again, you didn't write this, but might as well while on it (the error here is this being a comma splice).
```suggestion
            // application. However, not every device is equal; some are preferred over others.
```

marc0246: ("Here we only use one queue" is an ordinate clause followed by a subordinate clause.)
```suggestion
                // The list of queues that we are going to use. Here we only use one queue from
                // the previously chosen queue family.
```

examples/triangle/main.rs
```diff
@@ -207,51 +205,50 @@ impl App {
...
-        let memory_allocator = Arc::new(StandardMemoryAllocator::new(&device, &Default::default()));
-
-        // Before we can start creating and recording command buffers, we need a way of allocating
-        // them. Vulkano provides a command buffer allocator, which manages raw Vulkan command
-        // pools underneath and provides a safe interface for them.
-        let command_buffer_allocator = Arc::new(StandardCommandBufferAllocator::new(
-            &device,
-            &Default::default(),
-        ));
+        // We will use Vulkano's "task graph", which is available through the `vulkano_taskgraph`
```
marc0246:
```suggestion
        // We will use vulkano's "task graph", which is available through the `vulkano_taskgraph`
```

examples/triangle/main.rs
```diff
@@ -207,51 +205,50 @@ impl App {
...
+        // crate.
+        //
+        // The task graph is an optional abstraction built on top of Vulkano. It simplifies parts
```
marc0246:
```suggestion
        // The task graph is an optional abstraction built on top of vulkano. It simplifies parts
```

examples/triangle/main.rs
```diff
@@ -207,51 +205,50 @@ impl App {
...
+        // of the Vulkan API by providing a modular, node based approach to structure and execute
+        // GPU work.
+        //
+        // In order to use the task graph, we need to create a `Resources` instance. This will be
```
marc0246:
```suggestion
        // In order to use the task graph, we need to create a `Resources` collection. This will be
```

examples/triangle/main.rs
```diff
@@ -207,51 +205,50 @@ impl App {
...
+        // This is where the concept of "pipelining" comes into play. Rather than waiting for the
+        // GPU to finish drawing each individual frame, it's preferred to start preparing the next
+        // frame right way. This allows us to overlap CPU and GPU work, thus maximizing throughput
+        // at the cost of some latency. Flights are the task graph's mechanism to do just that.
+        //
+        // The number of frames in flight puts a hard limit on how far the CPU side is allowed to
+        // advance ahead of the GPU's execution before it needs to wait. Higher numbers increase
+        // the size of this buffer, while a value of 1 effectively disables pipelining.
+        //
+        // We choose to use 2 frames in flight, enabling pipelining without excessive latency.
```
marc0246:
```suggestion
        // We choose to use 2 frames in flight, which is the go-to for desktops. For mobile
        // devices, Arm recommends 3 frames in flight.
```

examples/triangle/main.rs
```diff
@@ -207,51 +205,50 @@ impl App {
...
+        //
+        // One reason to choose a higher number is making the application more resilient to spikes
+        // in frame time. A longer buffer allows the application to make up for missed frames, but
+        // may increase latency noticeably. Memory usage is also increased, since data transferred
+        // to the GPU needs to be available for longer.
```
marc0246: ("since" acts as a subordinating conjunction here, which generally don't have commas before them.)
```suggestion
        // may increase latency noticeably. Memory usage is also increased since data transferred
        // to the GPU needs to be available for longer.
```

examples/triangle/main.rs
```diff
@@ -207,51 +205,50 @@ impl App {
...
+        //
+        // On the other hand, a single frame in flight might be attractive for applications that
+        // have very low workloads and thus don't benefit much from pipelining.
+        let flight_id = resources.create_flight(MAX_FRAMES_IN_FLIGHT).unwrap();
 
+        // The "render context" is left uninitialized for now. In order to set it up we need a
+        // window and swapchain first, which can be created once winit's event loop has started.
```
marc0246: ("In order to set it up" is a subordinate clause followed by an ordinate clause.)
```suggestion
        // The "render context" is left uninitialized for now. In order to set it up, we need a
        // window and swapchain first, which can be created once winit's event loop has started.
```

examples/triangle/main.rs
```diff
@@ -260,11 +257,11 @@ impl App {
...
         // The objective of this example is to draw a triangle on a window. To do so, we first need
-        // to create the window. We use the `WindowBuilder` from the `winit` crate to do that here.
+        // to create the window.
         //
-        // Before we can render to a window, we must first create a `vulkano::swapchain::Surface`
-        // object from it, which represents the drawable surface of a window. For that we must wrap
-        // the `winit::window::Window` in an `Arc`.
+        // Before we can render to a window, we must first create a `Surface` object from it, which
+        // represents the drawable surface of a window. For that we must wrap the `Window` in an
+        // `Arc`.
```
marc0246: ("For that" is a subordinate clause followed by an ordinate clause.)
```suggestion
        // represents the drawable surface of a window. For that, we must wrap the `Window` in an
        // `Arc`.
```

examples/triangle/main.rs
```diff
@@ -273,10 +270,12 @@ impl ApplicationHandler for App {
         let surface = Surface::from_window(&self.instance, &window).unwrap();
         let window_size = window.inner_size();
 
-        // Before we can draw on the surface, we have to create what is called a swapchain.
+        // In order to draw on a surface, we need to create a "swapchain".
+        //
         // Creating a swapchain allocates the color buffers that will contain the image that will
         // ultimately be visible on the screen. These images are returned alongside the swapchain.
```
marc0246: This should be updated too IMO because using the term "buffer" for images is prehistoric terminology. Additionally, the images are not returned alongside the swapchain now.
```suggestion
        // Creating a swapchain allocates the swapchain images that will contain the image that
        // will ultimately be visible on the screen.
```

examples/triangle/main.rs
```diff
@@ -273,10 +270,12 @@ impl ApplicationHandler for App {
...
-        let (swapchain, images) = {
+        let swapchain_format;
+        let swapchain_id = {
             // Querying the capabilities of the surface. When we create the swapchain we can only
             // pass values that are allowed by the capabilities.
```
marc0246: ("When we create the swapchain" is a subordinate clause followed by an ordinate clause.)
```suggestion
            // Querying the capabilities of the surface. When we create the swapchain, we can only
            // pass values that are allowed by the capabilities.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
-                        .next()
-                        .unwrap(),
+            self.resources
+                .create_swapchain(
+                    &surface,
+                    &SwapchainCreateInfo {
+                        // We choose the lowest image count that the surface supports in order to
+                        // to minimize latency. However, we also take our minimum value of
+                        // `MAX_FRAMES_IN_FLIGHT + 1` into account. The latter will ensure that
+                        // enough images are available such that all frames in flight can be used.
```
marc0246: This seems to confuse frames in flight and swapchain images. That's precisely why we must document this with great care: a lot of people get confused by the difference.
```suggestion
                        // We choose the lowest image count that the surface supports in order to
                        // minimize memory usage. However, we also take our minimum value of
                        // `MAX_FRAMES_IN_FLIGHT + 1` into account. We need at least as many images
                        // as frames in flight, otherwise not all frames would actually be in
                        // flight since there wouldn't be enough images for the device to work on
                        // at the same time. At least, that would be the case if only the host and
                        // device were involved. When it comes to presentation, there is a third
                        // party involved: the "Presentation Engine". It can be working on one
                        // swapchain image at a time, and its work is not necessarily in sync with
                        // the device, so we need one more swapchain image. With fewer swapchain
                        // images, we could be blocking on the host while acquiring the next image.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                        min_image_count: surface_capabilities
+                            .min_image_count
+                            .max(MIN_SWAPCHAIN_IMAGES),
+                        image_format: swapchain_format,
```
marc0246, line 315:
```suggestion

                        image_format: swapchain_format,
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+
+                        // The size of the window, only used to initially setup the swapchain.
+                        //
+                        // NOTE:
+                        // On some drivers the swapchain extent is specified by
+                        // `surface_capabilities.current_extent` and the swapchain size must use
+                        // this extent. This extent is always the same as the window size.
```
marc0246: ("On some drivers" is a subordinate clause followed by an ordinate clause, and "and" acts as a coordinating conjunction here, which generally have commas before them.)
```suggestion
                        // On some drivers, the swapchain extent is specified by
                        // `surface_capabilities.current_extent`, and the swapchain size must use
                        // this extent. This extent is always the same as the window size.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                        //
+                        // However, other drivers don't specify a value, i.e.
+                        // `surface_capabilities.current_extent` is `None`. These drivers will allow
+                        // anything, but the only sensible value is the window size.
+                        //
+                        // Both of these cases need the swapchain to use the window size, so we just
+                        // use that.
+                        image_extent: window_size.into(),
+                        image_usage: ImageUsage::COLOR_ATTACHMENT,
```
marc0246, line 332:
```suggestion

                        image_usage: ImageUsage::COLOR_ATTACHMENT,
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+
+                        // The alpha mode indicates how the alpha value of the final image will
+                        // behave. For example, you can choose whether the window will be opaque or
+                        // transparent.
+                        composite_alpha: surface_capabilities
+                            .supported_composite_alpha
+                            .into_iter()
+                            .next()
+                            .unwrap(),
+                        ..Default::default()
```
marc0246, line 343:
```suggestion

                        ..Default::default()
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+            offset: [0.0, 0.0],
+            extent: window_size.into(),
+            min_depth: 0.0,
+            max_depth: 1.0,
+        };
+
+        // Now let's define what we want the GPU to do each frame by creating a task graph.
+        //
+        // This graph lets us structure our GPU work, while taking care of resource synchronization
+        // among other things.
```
marc0246: ("while" acts as a subordinating conjunction here.) The list of things the task graph is responsible for is not that long, so it should be exhaustive IMO.
```suggestion
        // This graph lets us structure our GPU work while taking care of resource synchronization
        // and resource cleanup.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+        let mut task_graph = TaskGraph::new(&self.resources);
+
+        // Here we add a "virtual" swapchain.
+        //
+        // Virtual resources allow us to declare resources ahead of time and reference them in the
+        // task graph. These can be thought of as placeholders that are filled in once the graph
+        // is executed.
+        //
+        // In the case of our swapchain, this means that we can recreate the swapchain whenever
+        // the window is resized, without having to recreate and recompile the entire task graph.
```
marc0246: (IMO, "without ..." is a restrictive clause here.)
```suggestion
        // In the case of our swapchain, this means that we can recreate the swapchain whenever
        // the window is resized without having to recreate and recompile the entire task graph.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+        let virtual_swapchain_id = task_graph.add_swapchain(&SwapchainCreateInfo {
+            image_format: swapchain_format,
+            ..Default::default()
+        });
+
+        // We also create a virtual framebuffer that we will use to render into the swapchain
+        // images.
```
marc0246: A bit more explanation is in order. (I don't blame you if you didn't know the these internals.)
```suggestion
        // We also create a virtual framebuffer, which, unlike virtual resources, is not a stand-in
        // for a physical framebuffer. For one, the task graph creates its framebuffers and render
        // passes internally, so you don't need to specify these yourself. Also, a single virtual
        // framebuffer doesn't necessarily correspond to a single physical framebuffer. Instead,
        // all a virtual framebuffer is for is a way to inform the task graph that different nodes'
        // attachments have the same dimensions. This allows the task graph to combine different
        // nodes into the same render pass, or even the same subpass. Therefore, you should use the
        // same virtual framebuffer across nodes that share the same framebuffer dimensions.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+        let virtual_framebuffer_id = task_graph.add_framebuffer();
+
+        // Next, we instantiate our `TriangleTask`.
+        //
+        // A "task" defines some work that we want the GPU to perform. We insert it into the graph
+        // by creating a node.
+        //
+        // A "task node" is an instance of a task. It is a unit of work that can be independently
+        // scheduled and synchronized by the task graph.
```
marc0246: (Technicalities.)
```suggestion
        // A "task" defines some work that we want the GPU to perform. A task is only useful when
        // it is inserted into the task graph as part of a task node.
        //
        // A "task node" contains a task alongside information about how it needs to be
        // synchronized. It is a unit of work that can be independently scheduled and synchronized
        // by the task graph.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                TriangleTask::new(self, virtual_swapchain_id),
+            )
+            // We bind the framebuffer that we want to use...
+            .framebuffer(virtual_framebuffer_id)
+            // ...and set the color attachment to the current swapchain image.
+            .color_attachment(
+                // `current_image_id()` means that this color attachment will always use the
+                // currently acquired swapchain image.
+                virtual_swapchain_id.current_image_id(),
+                AccessTypes::COLOR_ATTACHMENT_READ | AccessTypes::COLOR_ATTACHMENT_WRITE,
```
marc0246: Access types demand explanation.
```suggestion
                // We only need `COLOR_ATTACHMENT_WRITE` for the color attachment because our
                // graphics pipeline has color blending disabled, which would otherwise be a read
                // as well.
                AccessTypes::COLOR_ATTACHMENT_WRITE,
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+        // possible.
+        let mut task_graph = unsafe {
+            task_graph.compile(&CompileInfo {
+                // We need to provide all queues that we want to use for executing the graph. These
+                // must be compatible with the queue family types that were specified in the task
+                // nodes.
+                //
+                // In this example, we only have a single graphics queue.
+                queues: &[&self.queue],
+                // We use the same queue for presentation.
```
marc0246:
```suggestion
                // We use the same queue for presentation. You must specify a present queue if your
                // task graph uses any swapchains.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+            .task_mut()
+            .downcast_mut::<TriangleTask>()
             .unwrap()
-        };
+            .create_pipeline(self, &subpass);
+
+        // In some situations, the swapchain will become invalid by itself. This includes for
+        // example when the window is resized (as the images of the swapchain will no longer match
+        // the window's) or, on Android, when the application went to the background and goes back
+        // to the foreground.
```
marc0246: Again, I know you didn't write this, but I believe this is wrong. On Android, it is UB to use a surface after the application was suspended. That's why you must dispose of the window when receiving the `ApplicationHandler::suspended` event, and this is one of the reasons winit restructured itself to use the `ApplicationHandler` trait. However, it still doesn't prevent the UB. Just as pointless as the breakage done to raw-window-handle. Yes, this does mean that all of our examples have UB on Android.
```suggestion
        // the window's).
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                // Do not draw the frame when the screen size is zero. On Windows, this can occur
+                // when minimizing the application. In Vulkan, it is not allowed to create images
+                // that have a width, height (or depth) of zero.
+                if window_size.width == 0 || window_size.height == 0 {
+                    return;
+                }
+
+                // Whenever the window resizes we need to recreate everything dependent on the
+                // window size. In this example that includes the swapchain, the framebuffers and
+                // the dynamic state viewport.
```
marc0246: ("Whenever the window resizes" and "in this example" are subordinate clauses followed by ordinate clauses.) The framebuffers part is out of date.
```suggestion
                // Whenever the window resizes, we need to recreate everything dependent on the
                // window size. In this example, that includes the swapchain and the viewport.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+
+                // Finally, it is time to execute the graph.
+                match unsafe {
+                    rcx.task_graph
+                        .execute(resource_map, rcx, || rcx.window.pre_present_notify())
+                } {
+                    Ok(()) => {}
+                    // Since the task graph also handles presenting to the swapchain, it may return
+                    // a swapchain error. When the swapchain is "out of date" we set a flag to
+                    // recreate it during the next frame.
```
marc0246, line 571: ("When the swapchain is "out of date"" is a subordinate clause followed by an ordinate clause.)
```suggestion
                    // Since the task graph also handles presenting to the swapchain, it may return
                    // a swapchain error. When the swapchain is "out of date", we set a flag to
                    // recreate it during the next frame.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                position: [0.0, 0.5],
+            },
+            MyVertex {
+                position: [0.25, -0.1],
+            },
+        ];
+
+        // Allocate the Vulkan buffer that will hold the vertices.
+        //
+        // Since we are using Vulkano's task graph, the buffer is created using the `Resources`
```
marc0246:
```suggestion
        // Since we are using vulkano's task graph, the buffer is created using the `Resources`
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                    memory_type_filter: MemoryTypeFilter::PREFER_DEVICE
+                        | MemoryTypeFilter::HOST_SEQUENTIAL_WRITE,
+                    ..Default::default()
+                },
+                // The device layout determines the size and alignment of the buffer.
+                DeviceLayout::for_value(vertices.as_slice()).unwrap(),
+            )
+            .unwrap();
+
+        app.resources.flight(app.flight_id).wait(None).unwrap();
```
marc0246:
```suggestion
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+                },
+                [(vertex_buffer_id, HostAccessType::Write)],
+                [],
+                [],
+            )
+        }
+        .unwrap();
+
+        // As mentioned earlier, the pipeline depends on the subpass which is only created once the
+        // task graph is compiled. The pipeline field is initialized below.
```
marc0246: (IMO, "which ..." is a nonrestrictive clause here.)
```suggestion
        // As mentioned earlier, the pipeline depends on the subpass, which is only created once
        // the task graph is compiled. The pipeline field is initialized below.
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
+
+    pub fn create_pipeline(&mut self, app: &App, subpass: &Subpass) {
         // The next step is to create the shaders.
         //
-        // The raw shader creation API provided by the vulkano library is unsafe for various
-        // reasons, so The `shader!` macro provides a way to generate a Rust module from GLSL
-        // source - in the example below, the source is provided as a string input directly to the
+        // The raw shader creation API provided by the Vulkano library is unsafe for various
+        // reasons, so the `shader!` macro provides a way to generate a Rust module from GLSL
+        // source. In the example below, the source is provided as a string input directly to the
```
marc0246:
```suggestion
        // The raw shader creation API provided by the vulkano library is unsafe for various
        // reasons, so the `shader!` macro provides a way to generate a Rust module from shader
        // source. In the example below, the source is provided as a string input directly to the
```

examples/triangle/main.rs
```diff
@@ -286,68 +285,374 @@ impl ApplicationHandler for App {
...
         // shader, but a path to a source file can be provided as well. Note that the user must
         // specify the type of shader (e.g. "vertex", "fragment", etc.) using the `ty` option of
         // the macro.
         //
         // The items generated by the `shader!` macro include a `load` function which loads the
-        // shader using an input logical device. The module also includes type definitions for
-        // layout structures defined in the shader source, for example uniforms and push constants.
+        // shader using a logical device. The module also includes type definitions for layout
+        // structures defined in the shader source, for example uniforms and push constants.
```
marc0246: ("for example ..." is an ordinate clause preceded by an ordinate clause, so this is a comma splice. There's also a missing comma after "for example". The suggested formulation with "such as" is IMO a nonrestrictive clause.)
```suggestion
        // shader using a logical device. The module also includes structs compatible with the ones
        // defined in the shader source, such as uniforms and push constants for example.
```

examples/triangle/main.rs
```diff
@@ -375,69 +680,27 @@ impl ApplicationHandler for App {
...
-        // how a GPU operation is to be performed. It is similar to an OpenGL program, but it also
-        // contains many settings for customization, all baked into a single object. For drawing,
-        // we create a **graphics** pipeline, but there are also other types of pipeline.
+        // Before we draw, we have to create what is called a "pipeline". A pipeline describes how
+        // a GPU operation is to be performed. It is similar to an OpenGL program, but it also
+        // contains many settings for customization, all baked into a single object. For drawing
+        // triangles, we create a graphics pipeline, but there are also other types of pipelines.
         let pipeline = {
-            // First, we load the shaders that the pipeline will use: the vertex shader and the
+            // First, we load the shaders that the pipeline will use: The vertex shader and the
```
marc0246:
```suggestion
            // First, we load the shaders that the pipeline will use: the vertex shader and the
```

examples/triangle/main.rs
marc0246: ("that" acts as a subordinating conjunction here, which generally don't have commas before them.)
```suggestion
            // Automatically generate a vertex input state from the vertex shader's input
            // interface that takes a single vertex buffer containing `Vertex` structs.
```

marc0246: ("We use a resizable viewport" is an ordinate clause followed by a subordinate clause.)
```suggestion
                    // How primitives are transformed and clipped to fit the framebuffer. We use a
                    // resizable viewport set to draw over the entire window.
```

marc0246: ("The default value overwrites the old value with the new one" is an ordinate clause followed by a subordinate clause.)
```suggestion
                    // framebuffer. The default value overwrites the old value with the new one
                    // without any blending.
```

examples/triangle/main.rs
```diff
@@ -503,272 +762,67 @@ impl ApplicationHandler for App {
...
+        _tcx: &mut TaskContext<'_>,
+        rcx: &Self::World,
+    ) -> vulkano_taskgraph::TaskResult {
+        // This method is called when the task graph executes the task node. Here, we record all
+        // GPU commands to execute as part of this task.
 
-                    rcx.recreate_swapchain = false;
-                }
+        // Update the dynamic viewport, which is set to the current window and swapchain size.
+        cbf.try_set_viewport(0, slice::from_ref(&rcx.viewport))?;
```
marc0246:
```suggestion
        cbf.set_viewport(0, slice::from_ref(&rcx.viewport));
```

examples/triangle/main.rs
```diff
@@ -503,272 +762,67 @@ impl ApplicationHandler for App {
...
-                // `acquire_next_image` can be successful, but suboptimal. This means that the
-                // swapchain image will still work, but it may not display correctly. With some
-                // drivers this can be when the window resizes, but it may not cause the swapchain
-                // to become out of date.
-                if suboptimal {
-                    rcx.recreate_swapchain = true;
-                }
+        // Bind the graphics pipeline and vertex buffer.
+        cbf.try_bind_pipeline_graphics(self.pipeline.as_ref().unwrap())?;
+        cbf.try_bind_vertex_buffers(0, &[self.vertex_buffer_id], &[0], &[], &[])?;
```
marc0246:
```suggestion
        cbf.bind_pipeline_graphics(self.pipeline.as_ref().unwrap());
        cbf.bind_vertex_buffers(0, &[self.vertex_buffer_id], &[0], &[], &[]);
```

examples/triangle/main.rs
```diff
@@ -503,272 +762,67 @@ impl ApplicationHandler for App {
...
-                // Note that we have to pass a queue family when we create the command buffer. The
-                // command buffer will only be executable on that given queue family.
-                let mut builder = AutoCommandBufferBuilder::primary(
-                    self.command_buffer_allocator.clone(),
-                    self.queue.queue_family_index(),
-                    CommandBufferUsage::OneTimeSubmit,
-                )
-                .unwrap();
+        // Draw the triangle using one instance of our three vertices.
+        unsafe { cbf.try_draw(3, 1, 0, 0) }?;
```
marc0246:
```suggestion
        unsafe { cbf.draw(3, 1, 0, 0) };
```

examples/triangle/main.rs
```diff
@@ -503,272 +762,67 @@ impl ApplicationHandler for App {
...
-                        // previous_frame_end = Some(sync::now(&device).boxed());
-                    }
-                }
-            }
-            _ => {}
-        }
-    }
+        // If you are familiar with Vulkan, you will notice that we have performed no manual
+        // synchronization here. This is handled entirely by the task graph, as long as we have
+        // specified all resources that we want to access when creating the task node.
```
marc0246: ("as long as" acts as a subordinating conjunction here, which generally don't have commas before them.)
```suggestion
        // synchronization here. This is handled entirely by the task graph as long as we have
        // specified all resources that we want to access when creating the task node.
```

examples/triangle/main.rs
```diff
@@ -503,272 +762,67 @@ impl ApplicationHandler for App {
...
 
-    fn about_to_wait(&mut self, _event_loop: &ActiveEventLoop) {
-        let rcx = self.rcx.as_mut().unwrap();
-        rcx.window.request_redraw();
+        Ok(())
     }
 }
 
 // We use `#[repr(C)]` here to force rustc to use a defined layout for our data, as the default
 // representation has *no guarantees*.
```
marc0246: ("as" acts as a subordinating conjunction here, which generally don't have commas before them.)
```suggestion
// We use `#[repr(C)]` here to force rustc to use a defined layout for our data as the default
// representation has *no guarantees*.
```

examples/triangle/main.rs
CptPotato: I was actually going to ask about this. I got a bit confused that it didn't have a counterpart in the resource map 🙃

examples/triangle/main.rs
CptPotato: > I don't even know what "helper functions" it's referring to.

I think it might refer to `vulkano_util`... but that's just used in `triangle-util`, `interactive-fractal` and `multi-window-game-of-life` 🤔.

What's the plan for this section? Should I just get rid of the paragraph, or replace it by a quick rundown of what this example covers?

CptPotato (commented):
Thanks for taking the time to correct and explain everything

examples/triangle/main.rs
CptPotato: GitHub failed to auto resolve the rest of the suggestions. I'll double check the changes and resolve manually from here.

examples/triangle/main.rs
marc0246: vulkano-util predates this paragraph by many years. I think it should just be removed.

marc0246:
You did very well. Clear and easy to understand explanations. I disagree that you aren't much of a writer (save for the comma placement :melting_face:). If anything, you are great for this job. If it makes you feel any better, it takes me hours to write a few paragraphs :laughing:

PR #2795, opened 2026-05-09
Remove shaderc from vulkano-shaders and lay the groundwork for slang shaders compilation

nshyrei:
A continuation of (https://github.com/vulkano-rs/vulkano/pull/2791).
Removes the `shaderc` Rust crate dependency from `vulkano-shaders` and replaces shader compilation with direct invocations of the `glslc` CLI tool. This eliminates a heavy C++ build dependency (`libshaderc`) that required CMake, Ninja, and Python to compile from source and also adds the possibility to directly support other shader languages like Slang by calling an appropriate compiler (like `slangc`).

### Implementation

  - glslc is invoked via `std::process::Command`
  - #include directives are now resolved natively by `glslc` via `-I` flags and `current_dir`, with `vulkano.glsl` is written to a temp directory and prepended as the first include path.
  - A list of resolved #include directives is written by `glslc` during compilation phase into an `-MF` file.
  - Necessary Enum types are copied from `shaderc` and recreated in crate
 
Changelog:
```md
### Public dependency updates
- Removed the [`shaderc`](https://crates.io/crates/shaderc) dependency. `glslc` must now be installed and available on `PATH` (bundled with the Vulkan SDK).

### Breaking changes
- The `shaderc-build-from-source` Cargo feature has been removed.
- Shader compilation now requires `glslc` to be present on `PATH` at build time instead of `libshaderc` having to be supplied via pkg-config, the `SHADERC_LIB_DIR` environment variable, or the `VULKAN_SDK` environment variable. This hopefully makes setup easier.
```


marc0246:
`glslc` supports passing `-` as the input and output path, making it use stdin and stdout, respectively. Please no file writing and reading.

You can ignore the typos lints. I'll deal with it.

nshyrei:
@marc0246 updated to use stdin/out in the latest commit.

marc0246:
I think the preprocessing is unnecessary, and I would like to avoid it because it's very hard to do correctly. For one, you would have to use an *actual* parser because the primitive way is going to cause trouble for the user. It's not fun. Also, anything to do with parsing a shader ourselves is a no-go because it would be per-language, which is just a maintenance nightmare. And you would have to run the preprocessor on the source first, so a second process invocation per shader. I would really rather not add much more overhead compared to what we have now. The alternative is to use an actual C preprocessor rewritten in Rust. That's *much, much* less fun than the parser.

What I was thinking is to have shaderc do everything for us in one invocation as it already implements all the functionality. With `#include` preprocessing, we wouldn't need shaderc and could use glslang instead.

First, some background:
1. We have the `root_path` variable that was previously passed to `codegen::compile`, which is the path we should be searching relative to for relative includes in the source.
2. You can set the working directory of a spawned process with [`Command::current_dir`](https://doc.rust-lang.org/std/process/struct.Command.html#method.current_dir).
3. shaderc has the `-I` argument, which can be used to add directory paths for include searching.

Now, for the implementation:
- With (1) and (2), you can set shaderc's working directory to the `root_path`. When the source is piped in, relative includes in that source are relative to the working directory. Nested relative includes, so a file included in the source includes a file itself, is relative to the file that's including. So relative includes Just Work™️ and work like before, except they work *correctly*. C relative includes are supposed to search relative first, but if that fails, standard include. shaderc of course implements this right, unlike our `include_callback`.
- Standard includes are relative to include directories, so you can use (3) for that.
- The only wart is `vulkano.glsl` (and `vulkano.hlsl` and `vulkano.slang` in the future). Cargo does have an uncompressed vulkano-shaders *somewhere*, but there's no guarantee it's in the file system or where in that file system it resides, so anything like that is a no-go. I think the only solution we have is to create a temp directory, write `vulkano.glsl` in it, and add that temp directory as **the first** include path. It's really unfortunate, but at least this only needs to happen once for all shaders being compiled. In other words, this should happen before the `for` loop in `shader_inner`.

With all that, you can remove `preprocess_includes`, `include_callback`, and their call sites.

vulkano-shaders/src/codegen.rs
```diff
@@ -1,14 +1,20 @@
...
-use shaderc::{CompileOptions, Compiler, EnvVersion, TargetEnv};
+pub enum IncludeType {
+    Relative,
+    Standard,
+}
+
+pub struct ResolvedInclude {
+    pub resolved_name: String,
+    pub content: String,
+}
```
marc0246: Please don't put items in-between `use` statements.

vulkano-shaders/src/codegen.rs
```diff
@@ -144,75 +148,274 @@ fn include_callback(
...
+
+fn spv_to_words(data: &[u8]) -> Vec<u32> {
+    data.chunks(4)
+        .map(|c| u32::from_le_bytes([c[0], c[1], c[2], c[3]]))
+        .collect()
+}
+
+struct Compiler {
+    command: std::process::Command,
+}
```
marc0246: I like that you added `CompileOptions` as it does make the code cleaner. However, `Compiler` doesn't make sense. It made sense in the shaderc crate because you can reuse the `Compiler`, but we always create and immediately consume one, which is an anti-pattern. `Compiler::compile_into_spirv` should be a free-standing function.

vulkano-shaders/src/lib.rs
```diff
@@ -405,14 +413,105 @@ enum SourceKind {
...
+            "raygen" => Ok(ShaderKind::RayGeneration),
+            "anyhit" => Ok(ShaderKind::AnyHit),
+            "closesthit" => Ok(ShaderKind::ClosestHit),
+            "miss" => Ok(ShaderKind::Miss),
+            "intersection" => Ok(ShaderKind::Intersection),
+            "callable" => Ok(ShaderKind::Callable),
+            _ => Err(format!(
+                "unknown shader kind: {}. Expected `vertex`, `tess_ctrl`, `tess_eval`, `geometry`, `task`, \
+                `mesh`, `fragment`, `compute`, `raygen`, `anyhit`, `closesthit`, \
+                `miss`, `intersection` or `callable`", s
```
marc0246: Rust errors don't contain periods. Also, formatting.
```suggestion
                "unknown shader kind: {s}; expected `vertex`, `tess_ctrl`, `tess_eval`, \
                `geometry`, `task`, `mesh`, `fragment`, `compute`, `raygen`, `anyhit`, \
                `closesthit`, `miss`, `intersection` or `callable`",
```

marc0246:
Please put `EnvVersion` and `SpirvVersion` in the lib module together with the other enums you copied from shaderc.

vulkano-shaders/src/codegen.rs
```diff
@@ -144,75 +148,274 @@ fn include_callback(
...
+#[allow(dead_code)]
+pub(crate) enum EnvVersion {
+    Vulkan1_0,
+    Vulkan1_1,
+    Vulkan1_2,
+    Vulkan1_3,
+    Vulkan1_4,
+    OpenGL4_5,
+    WebGPU,
+}
```
marc0246: You should not allow this lint. It's a code smell.

vulkano-shaders/src/lib.rs
```diff
@@ -405,14 +413,105 @@ enum SourceKind {
     Bytes(LitStr),
 }
 
+#[derive(Copy, Clone)]
+#[allow(dead_code, clippy::upper_case_acronyms)]
+enum SourceLanguage {
+    GLSL,
+    HLSL,
+    Slang,
+}
```
marc0246: You should not allow either of these lints. If you don't need the `Slang` variant yet then there's no point adding it now. Also, the shaderc crate breaks Rust naming conventions, and that's not something to copy.

vulkano-shaders/src/codegen.rs
```diff
@@ -144,75 +148,274 @@ fn include_callback(
...
+            debug: false,
+        }
+    }
+}
+
+fn spv_to_words(data: &[u8]) -> Vec<u32> {
+    data.chunks(4)
+        .map(|c| u32::from_le_bytes([c[0], c[1], c[2], c[3]]))
+        .collect()
+}
```
marc0246: vulkano has this function, but a bit more correct and optimized: `vulkano::shader::spirv::bytes_to_words`.

vulkano-shaders/src/codegen.rs
nshyrei: Having a `Compiler` struct is a sort of a design for the future. I thought that if we are gonna have a Slang language it would require a different compiler binary + maybe potentially different setup options, this is why I created this struct. If this is not needed/not a good design I see no issues removing it.

marc0246: What's an anti-pattern is that we construct a struct and immediately consume it. The struct is never going to be reused as there's no point to it. And a combo of `new()` immediately followed by `method()` can always be reduced to that method as a free-standing function. The function can do everything that combo can do.

vulkano-shaders/src/lib.rs
marc0246: I used tabs by mistake. I updated the suggestion.

nshyrei:
@marc0246 I have removed preprocessing and added a new way of dependency handling the way you outlined in the latest commit. With the way dependencies are handled now I had to go back to my first idea (not present in the initial commits) to use the dependency file (`-MF`) written by `glslc` to return the list of shader includes.

marc0246:
Huh, I see. You needed that argument for the reflection. To be honest, I haven't thought about that at all :sweat_smile: But you came up with a nice solution :+1:

marc0246:
Maybe you decided against this because such an argument may not exist for all compilers? That would be a legitimate concern, but I think at least Slang supports it.

marc0246:
It seems I was wrong about a couple things in my big-ass comment. The `root_path` is not to be included relative to for includes in an inline shader. It would only accept absolute paths for relative includes then. And this PR has the same behavior, so that's good. Also,  relative includes in the `include_callback` did work correctly before. shaderc will retry a failed relative path as a standard path itself. But anyway, I tested your changes and everything seems to work like before :+1:

nshyrei:
@marc0246 This is great news, do I need to do anything about Rust typos failure or you will fix it?

marc0246:
No no, I'll fix it myself. I'm still reviewing though.

vulkano-shaders/src/codegen.rs
```diff
@@ -22,197 +22,193 @@ pub struct Shader {
...
+        "vulkano_shaders_{}_{}",
+        std::process::id(),
+        id,
+    )));
+    let vulkano_dir = &vulkano_dir.0;
+
+    // Write vulkano.glsl to a temp directory and prepend it as the first include path so glslc
+    // can resolve `#include "vulkano.glsl"` or `#include <vulkano.glsl>`.
+    fs::create_dir_all(vulkano_dir)
+        .map_err(|e| format!("Failed to create vulkano include dir: {e}"))?;
```
marc0246: Rust errors aren't capitalized.
```suggestion
        .map_err(|e| format!("failed to create vulkano include dir: {e}"))?;
```

vulkano-shaders/src/codegen.rs
```diff
@@ -22,197 +22,193 @@ pub struct Shader {
...
+    fs::write(
+        vulkano_dir.join("vulkano.glsl"),
+        include_str!("../include/vulkano.glsl"),
+    )
+    .map_err(|e| format!("Failed to write vulkano.glsl: {e}"))?;
```
marc0246:
```suggestion
    .map_err(|e| format!("failed to write vulkano.glsl: {e}"))?;
```

vulkano-shaders/src/codegen.rs
```diff
@@ -22,197 +22,193 @@ pub struct Shader {
...
+        .arg("-o")
+        .arg("-")
+        .arg("-")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped());
+
+    let mut child = cmd
+        .spawn()
+        .map_err(|e| format!("Failed to call glslc: {e}"))?;
```
marc0246:
```suggestion
        .map_err(|e| format!("failed to call glslc: {e}"))?;
```

vulkano-shaders/src/codegen.rs
```diff
@@ -22,197 +22,193 @@ pub struct Shader {
...
+
+    child
+        .stdin
+        .take()
+        .ok_or("Failed to open glslc stdin")?
```
marc0246:
```suggestion
        .ok_or("failed to open glslc stdin")?
```

vulkano-shaders/src/codegen.rs
```diff
@@ -22,197 +22,193 @@ pub struct Shader {
...
+        .write_all(source.as_bytes())
+        .map_err(|e| format!("Failed to write to glslc stdin: {e}"))?;
```
marc0246:
```suggestion
        .map_err(|e| format!("failed to write to glslc stdin: {e}"))?;
```

vulkano-shaders/src/codegen.rs
```diff
@@ -22,197 +22,193 @@ pub struct Shader {
...
+
+    let output = child
+        .wait_with_output()
+        .map_err(|e| format!("Failed to wait for glslc: {e}"))?;
```
marc0246:
```suggestion
        .map_err(|e| format!("failed to wait for glslc: {e}"))?;
```

marc0246 (commented):
You can commit suggestions directly from the GitHub interface, which saves you time.

marc0246:
Thank you for this! It's been sitting on my todo list (currently at 140 entries) forever, and it's such a great thing for UX.

nshyrei:
@marc0246 Happy to help with the rest of your backlog if you have it written somewhere.

marc0246:
That's very sweet of you. Come join our discord then :slightly_smiling_face: That's where we coordinate. But Slang support in vulkano-shaders is on the list, so that would be a good continuation for you :slightly_smiling_face: Also related is having all of GLSL, HLSL and Slang for every example that uses shaders. Then we should place all the example's shaders in their own `shaders` subdirectory in the example's directory.

PR #2796, opened 2026-05-11
Remove `Subbuffer` from `BufferView`

CptPotato:
This removes `Subbuffer` from `BufferView` for the same reason as #2746 and instead specifies the range of the buffer in `BufferViewCreateInfo`.

Changelog:
```markdown
### Breaking changes
Changes to `BufferView` and `BufferViewCreateInfo`:
- `BufferView` is now created from a `Buffer` instead of a `Subbuffer`. Use `offset` and `range` of `BufferViewCreateInfo` to specify a sub range within the buffer.
- `BufferView::buffer()` now returns the underlying `Buffer`.
```

### DIFF
--- a/vulkano/src/buffer/view.rs
+++ b/vulkano/src/buffer/view.rs
@@ -123,7 +123,8 @@ impl BufferView {
 
         if offset >= buffer.size() {
             return Err(Box::new(ValidationError {
-                problem: "`create_info.offset` must be less than `buffer.size()`".into(),
+                context: "offset".into(),
+                problem: "must be less than `buffer.size()`".into(),
                 vuids: &["VUID-VkBufferViewCreateInfo-offset-00925"],
                 ..Default::default()
             }));
@@ -175,7 +176,8 @@ impl BufferView {
         if let Some(range) = range {
             if range == 0 {
                 return Err(Box::new(ValidationError {
-                    problem: "`create_info.range` was specified but must be greater than 0".into(),
+                    context: "range".into(),
+                    problem: "is zero".into(),
                     vuids: &["VUID-VkBufferViewCreateInfo-range-00928"],
                     ..Default::default()
                 }));
@@ -183,9 +185,8 @@ impl BufferView {
 
             if !range.is_multiple_of(block_size) {
                 return Err(Box::new(ValidationError {
-                    problem: "`create_info.range` was specified but is not a multiple of \
-                        `create_info.format.block_size()`"
-                        .into(),
+                    context: "range".into(),
+                    problem: "is not a multiple of `create_info.format.block_size()`".into(),
                     vuids: &["VUID-VkBufferViewCreateInfo-range-00929"],
                     ..Default::default()
                 }));
@@ -195,7 +196,7 @@ impl BufferView {
                 > properties.max_texel_buffer_elements
             {
                 return Err(Box::new(ValidationError {
-                    problem: "`buffer.size / create_info.format.block_size() * \
+                    problem: "`buffer.size() / create_info.format.block_size() * \
                         create_info.format.texels_per_block()` is greater than the \
                         `max_texel_buffer_elements` limit"
                         .into(),
@@ -204,7 +205,7 @@ impl BufferView {
                 }));
             }
 
-            if range.saturating_add(offset) > buffer.size() {
+            if range > buffer.size() - offset {
                 return Err(Box::new(ValidationError {
                     problem: "`create_info.offset + create_info.range` must be less than or \
                         equal to `buffer.size()`"
@@ -457,11 +458,15 @@ pub struct BufferViewCreateInfo<'a> {
     pub format: Format,
 
     /// The offset in bytes.
+    ///
+    /// The default value is `0`.
     pub offset: DeviceSize,
 
     /// The size in bytes.
     ///
-    /// The default value `None` refers to the remaining length of the buffer.
+    /// If set to `None`, the size until the end of the buffer will be used.
+    ///
+    /// The default value is `None`.
     pub range: Option<DeviceSize>,
 
     pub _ne: crate::NonExhaustive<'a>,
--- a/vulkano/src/buffer/view.rs
+++ b/vulkano/src/buffer/view.rs
@@ -11,9 +11,9 @@
 //! ```
 //! # use std::sync::Arc;
 //! use vulkano::{
-//!     buffer::{view::{BufferView, BufferViewCreateInfo}, Buffer, BufferContents, BufferCreateInfo, BufferUsage},
+//!     buffer::{view::{BufferView, BufferViewCreateInfo}, Buffer, BufferCreateInfo, BufferUsage},
 //!     format::Format,
-//!     memory::allocator::AllocationCreateInfo,
+//!     memory::allocator::{AllocationCreateInfo, DeviceLayout},
 //! };
 //!
 //! # let queue: Arc<vulkano::device::Queue> = return;
@@ -26,7 +26,7 @@
 //!         ..Default::default()
 //!     },
 //!     &AllocationCreateInfo::default(),
-//!     u32::LAYOUT.layout_for_len(128).unwrap(),
+//!     DeviceLayout::new_unsized::<[u32]>(128).unwrap(),
 //! )
 //! .unwrap();
 //!
@@ -528,9 +528,9 @@ impl BufferViewCreateInfo<'_> {
 mod tests {
     use super::{BufferView, BufferViewCreateInfo};
     use crate::{
-        buffer::{Buffer, BufferContents, BufferCreateInfo, BufferUsage},
+        buffer::{Buffer, BufferCreateInfo, BufferUsage},
         format::Format,
-        memory::allocator::{AllocationCreateInfo, StandardMemoryAllocator},
+        memory::allocator::{AllocationCreateInfo, DeviceLayout, StandardMemoryAllocator},
     };
     use std::sync::Arc;
 
@@ -547,7 +547,7 @@ mod tests {
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            <[u8; 4]>::LAYOUT.layout_for_len(128).unwrap(),
+            DeviceLayout::new_unsized::<[[u8; 4]]>(128).unwrap(),
         )
         .unwrap();
 
@@ -574,7 +574,7 @@ mod tests {
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            <[u8; 4]>::LAYOUT.layout_for_len(128).unwrap(),
+            DeviceLayout::new_unsized::<[[u8; 4]]>(128).unwrap(),
         )
         .unwrap();
         BufferView::new(
@@ -600,7 +600,7 @@ mod tests {
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            u32::LAYOUT.layout_for_len(128).unwrap(),
+            DeviceLayout::new_unsized::<[u32]>(128).unwrap(),
         )
         .unwrap();
         BufferView::new(
@@ -626,7 +626,7 @@ mod tests {
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            <[u8; 4]>::LAYOUT.layout_for_len(128).unwrap(),
+            DeviceLayout::new_unsized::<[[u8; 4]]>(128).unwrap(),
         )
         .unwrap();
 
@@ -654,7 +654,7 @@ mod tests {
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            <[f64; 4]>::LAYOUT.layout_for_len(128).unwrap(),
+            DeviceLayout::new_unsized::<[[f64; 4]]>(128).unwrap(),
         )
         .unwrap();
 
--- a/vulkano/src/buffer/view.rs
+++ b/vulkano/src/buffer/view.rs
@@ -31,7 +31,7 @@
 //! .unwrap();
 //!
 //! let view = BufferView::new(
-//!     &buffer,
+//!     &buffer.buffer(),
 //!     &BufferViewCreateInfo {
 //!         format: Format::R32_UINT,
 //!         ..Default::default()
@@ -40,8 +40,9 @@
 //! .unwrap();
 //! ```
 
-use super::{BufferUsage, Subbuffer};
+use super::BufferUsage;
 use crate::{
+    buffer::Buffer,
     device::{Device, DeviceOwned},
     format::{Format, FormatFeatures},
     macros::impl_id_counter,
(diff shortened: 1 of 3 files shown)


vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -3680,7 +3680,7 @@ impl<L> AutoCommandBufferBuilder<L> {
                 DescriptorBindingResources::BufferView(elements) => {
                     for (index, element) in elements.iter().enumerate() {
                         if let Some(Some(buffer_view)) = element {
-                            let buffer = buffer_view.buffer();
+                            let buffer = Subbuffer::new(buffer_view.buffer().clone());
```
CptPotato, line 3683: This is a bit unfortunate, but trying to use `Buffer` directly here is way too big of an impact.

vulkano/src/buffer/view.rs
```diff
@@ -405,6 +435,14 @@ pub struct BufferViewCreateInfo<'a> {
     /// The default value is `Format::UNDEFINED`.
     pub format: Format,
 
+    /// The offset in bytes.
+    pub offset: DeviceSize,
+
+    /// The size in bytes.
+    ///
+    /// The default value `None` refers to the remaining length of the buffer.
+    pub range: Option<DeviceSize>,
```
CptPotato, line 477: I called this `size` first, but the spec refers to it as `range` so I went with that.

Also, taking a peek at the changelog, this is an `Option` to cover the `VK_WHOLE_SIZE` case.

> Where the Vulkan API accepts VK_WHOLE_SIZE, VK_REMAINING_* and LOD_CLAMP_NONE, vulkano now accepts an Option for the end of the range with None meaning that it is unbounded.

vulkano/src/buffer/view.rs
```diff
@@ -306,44 +335,49 @@ impl BufferView {
...
 
         Arc::new(BufferView {
             handle,
-            subbuffer: subbuffer.clone().into_bytes(),
+            buffer: buffer.clone(),
             id: Self::next_id(),
             format,
             format_features,
-            range: 0..size,
+            range: offset..(offset + size),
```
CptPotato: Please double check that I didn't misunderstand what range is supposed to be here.

vulkano/src/buffer/view.rs
```diff
@@ -139,27 +151,47 @@ impl BufferView {
...
+                    problem: "`buffer.size / create_info.format.block_size() * \
+                        create_info.format.texels_per_block()` is greater than the \
+                        `max_texel_buffer_elements` limit"
+                        .into(),
+                    vuids: &["VUID-VkBufferViewCreateInfo-range-00930"],
+                    ..Default::default()
+                }));
+            }
+
+            if range.saturating_add(offset) > buffer.size() {
```
marc0246: We always check if an `offset` is less than the total size, and you did too, so there is no need for overflow checks anymore.
```suggestion
            if range > buffer.size() - offset {
```

vulkano/src/buffer/view.rs
```diff
@@ -68,40 +69,51 @@ impl BufferView {
...
-        let buffer = subbuffer.buffer();
         let properties = device.physical_device().properties();
 
         let format_properties =
             unsafe { device.physical_device().format_properties_unchecked(format) };
         let format_features = format_properties.buffer_features;
 
+        if offset >= buffer.size() {
+            return Err(Box::new(ValidationError {
+                problem: "`create_info.offset` must be less than `buffer.size()`".into(),
```
marc0246:
```suggestion
                context: "offset".into(),
                problem: "must be less than `buffer.size()`".into(),
```

vulkano/src/buffer/view.rs
```diff
@@ -139,27 +151,47 @@ impl BufferView {
...
-                    `create_info.format.block_size()`"
-                    .into(),
-                vuids: &["VUID-VkBufferViewCreateInfo-range-00929"],
-                ..Default::default()
-            }));
-        }
+        if let Some(range) = range {
+            if range == 0 {
+                return Err(Box::new(ValidationError {
+                    problem: "`create_info.range` was specified but must be greater than 0".into(),
```
marc0246:
```suggestion
                    context: "range".into(),
                    problem: "is zero".into(),
```

vulkano/src/buffer/view.rs
```diff
@@ -139,27 +151,47 @@ impl BufferView {
...
-                    `max_texel_buffer_elements` limit"
-                    .into(),
-                vuids: &["VUID-VkBufferViewCreateInfo-range-00930"],
-                ..Default::default()
-            }));
+            if !range.is_multiple_of(block_size) {
+                return Err(Box::new(ValidationError {
+                    problem: "`create_info.range` was specified but is not a multiple of \
+                        `create_info.format.block_size()`"
+                        .into(),
```
marc0246:
```suggestion
                    context: "range".into(),
                    problem: "is not a multiple of `create_info.format.block_size()`".into(),
```

vulkano/src/buffer/view.rs
```diff
@@ -139,27 +151,47 @@ impl BufferView {
...
+            }
+
+            if ((range / block_size) * texels_per_block as DeviceSize) as u32
+                > properties.max_texel_buffer_elements
+            {
+                return Err(Box::new(ValidationError {
+                    problem: "`buffer.size / create_info.format.block_size() * \
+                        create_info.format.texels_per_block()` is greater than the \
+                        `max_texel_buffer_elements` limit"
+                        .into(),
```
marc0246:
```suggestion
                    problem: "`buffer.size() / create_info.format.block_size() * \
                        create_info.format.texels_per_block()` is greater than the \
                        `max_texel_buffer_elements` limit"
                        .into(),
```

vulkano/src/buffer/view.rs
marc0246: We always check if an `offset` is less than the total size, and you did too, so there is no need for overflow checks anymore.
```suggestion
            if range > buffer.size() - offset {
```

vulkano/src/buffer/view.rs
```diff
@@ -405,6 +435,14 @@ pub struct BufferViewCreateInfo<'a> {
     /// The default value is `Format::UNDEFINED`.
     pub format: Format,
 
+    /// The offset in bytes.
```
marc0246, line 467:
```suggestion
    /// The offset in bytes.
    ///
    /// The default value is `0`.
```

vulkano/src/buffer/view.rs
```diff
@@ -405,6 +435,14 @@ pub struct BufferViewCreateInfo<'a> {
...
+    pub offset: DeviceSize,
+
+    /// The size in bytes.
+    ///
+    /// The default value `None` refers to the remaining length of the buffer.
```
marc0246:
```suggestion
    /// The size in bytes.
    ///
    /// If set to `None`, the size until the end of the buffer will be used.
    ///
    /// The default value is `None`.
```

vulkano/src/command_buffer/auto/commands/pipeline.rs
marc0246: It's not more overhead than before; you just need to remove the `.clone()` below as it's superfluous.

marc0246:
As always, you can commit suggestions directly from the GitHub interface to save you time. You also have merge conflicts. And one more thing: please remove mentions of `Subbuffer` and `Buffer::new_*` from the example and tests in `buffer::view`.

CptPotato:
Rebased and addressed points from the review. Should be good now.

vulkano/src/buffer/view.rs
```diff
@@ -11,22 +11,22 @@
 //! ```
 //! # use std::sync::Arc;
 //! use vulkano::{
-//!     buffer::{view::{BufferView, BufferViewCreateInfo}, Buffer, BufferCreateInfo, BufferUsage},
+//!     buffer::{view::{BufferView, BufferViewCreateInfo}, Buffer, BufferContents, BufferCreateInfo, BufferUsage},
 //!     format::Format,
 //!     memory::allocator::AllocationCreateInfo,
 //! };
```
marc0246: Changing this because of the below, but also, this was longer than 99 columns and as such would render with a horizontal scrollbar, and the imports weren't sorted. If only rustfmt worked :pensive: It's configured to format this.
```suggestion
//!     buffer::{view::{BufferView, BufferViewCreateInfo}, Buffer, BufferCreateInfo, BufferUsage},
//!     format::Format,
//!     memory::allocator::{AllocationCreateInfo, DeviceLayout},
//! };
```

vulkano/src/buffer/view.rs
```diff
@@ -11,22 +11,22 @@
...
-//! let buffer = Buffer::new_slice::<u32>(
+//! let buffer = Buffer::new(
 //!     &memory_allocator,
 //!     &BufferCreateInfo {
 //!         usage: BufferUsage::STORAGE_TEXEL_BUFFER,
 //!         ..Default::default()
 //!     },
 //!     &AllocationCreateInfo::default(),
-//!     128,
+//!     u32::LAYOUT.layout_for_len(128).unwrap(),
```
marc0246: `BufferContentsLayout::layout_for_len` discards the length for sized types. Also, `DeviceLayout` is meant to be constructed with its constructors.
```suggestion
//!     DeviceLayout::new_unsized::<[u32]>(128).unwrap(),
```

vulkano/src/buffer/view.rs
```diff
@@ -457,23 +507,28 @@ impl BufferViewCreateInfo<'_> {
...
 
 #[cfg(test)]
 mod tests {
     use super::{BufferView, BufferViewCreateInfo};
     use crate::{
-        buffer::{Buffer, BufferCreateInfo, BufferUsage},
+        buffer::{Buffer, BufferContents, BufferCreateInfo, BufferUsage},
         format::Format,
         memory::allocator::{AllocationCreateInfo, StandardMemoryAllocator},
     };
```
marc0246: (See below.)
```suggestion
        buffer::{Buffer, BufferCreateInfo, BufferUsage},
        format::Format,
        memory::allocator::{AllocationCreateInfo, DeviceLayout, StandardMemoryAllocator},
    };
```

vulkano/src/buffer/view.rs
```diff
@@ -485,14 +540,14 @@ mod tests {
...
-        let buffer = Buffer::new_slice::<[u8; 4]>(
+        let buffer = Buffer::new(
             &memory_allocator,
             &BufferCreateInfo {
                 usage: BufferUsage::UNIFORM_TEXEL_BUFFER,
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            128,
+            <[u8; 4]>::LAYOUT.layout_for_len(128).unwrap(),
```
marc0246:
```suggestion
            DeviceLayout::new_unsized::<[[u8; 4]]>(128).unwrap(),
```

vulkano/src/buffer/view.rs
```diff
@@ -512,14 +567,14 @@ mod tests {
...
-        let buffer = Buffer::new_slice::<[u8; 4]>(
+        let buffer = Buffer::new(
             &memory_allocator,
             &BufferCreateInfo {
                 usage: BufferUsage::STORAGE_TEXEL_BUFFER,
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            128,
+            <[u8; 4]>::LAYOUT.layout_for_len(128).unwrap(),
```
marc0246:
```suggestion
            DeviceLayout::new_unsized::<[[u8; 4]]>(128).unwrap(),
```

vulkano/src/buffer/view.rs
```diff
@@ -538,14 +593,14 @@ mod tests {
...
-        let buffer = Buffer::new_slice::<u32>(
+        let buffer = Buffer::new(
             &memory_allocator,
             &BufferCreateInfo {
                 usage: BufferUsage::STORAGE_TEXEL_BUFFER,
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            128,
+            u32::LAYOUT.layout_for_len(128).unwrap(),
```
marc0246:
```suggestion
            DeviceLayout::new_unsized::<[u32]>(128).unwrap(),
```

vulkano/src/buffer/view.rs
```diff
@@ -564,14 +619,14 @@ mod tests {
...
-        let buffer = Buffer::new_slice::<[u8; 4]>(
+        let buffer = Buffer::new(
             &memory_allocator,
             &BufferCreateInfo {
                 usage: BufferUsage::TRANSFER_DST, // Dummy value
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            128,
+            <[u8; 4]>::LAYOUT.layout_for_len(128).unwrap(),
```
marc0246:
```suggestion
            DeviceLayout::new_unsized::<[[u8; 4]]>(128).unwrap(),
```

vulkano/src/buffer/view.rs
```diff
@@ -592,14 +647,14 @@ mod tests {
...
-        let buffer = Buffer::new_slice::<[f64; 4]>(
+        let buffer = Buffer::new(
             &memory_allocator,
             &BufferCreateInfo {
                 usage: BufferUsage::UNIFORM_TEXEL_BUFFER | BufferUsage::STORAGE_TEXEL_BUFFER,
                 ..Default::default()
             },
             &AllocationCreateInfo::default(),
-            128,
+            <[f64; 4]>::LAYOUT.layout_for_len(128).unwrap(),
```
marc0246:
```suggestion
            DeviceLayout::new_unsized::<[[f64; 4]]>(128).unwrap(),
```

marc0246:
I forgor, since the parameters are now `offset` and `range`, we should have the same getters. That also means replacing the `BufferView::range` field with those two fields. You can first LSP-rename the current `BufferView::range` method to `as_range` or something. It doesn't really matter; it should be private anyway.

vulkano/src/buffer/view.rs
CptPotato: Ah I see, I mistraced the generics in the previous call hierarchy (`new_slice<T>` calls `new_unsized<[T]>`, not `new_unsized<T>` which is where I lifted the `T::LAYOUT` from).

I'll look through the remaining changes soon.

CptPotato:
> I forgor, since the parameters are now `offset` and `range`, we should have the same getters. That also means replacing the `BufferView::range` field with those two fields. You can first LSP-rename the current `BufferView::range` method to `as_range` or something. It doesn't really matter; it should be private anyway.

Done. I didn't add an `as_range` since there is only a single use of it and it's in the `AutoCommandBufferBuilder`.

vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -3680,14 +3680,16 @@ impl<L> AutoCommandBufferBuilder<L> {
...
                             let (use_ref, memory_access) = use_iter(index as u32);
+                            let offset = buffer_view.offset();
+                            let size = buffer_view.range().unwrap_or(buffer.size());
```
marc0246:
```suggestion
                            let size = buffer_view.range().unwrap_or(buffer.size() - offset);
```

PR #2799, opened 2026-05-14
Make relative includes work in embedded shader source, and make the `path` and `bytes` search relative to the file with the `shader!` invocation

marc0246:
Previously, `shader!` relative includes in embedded source would only accept absolute paths, or fall back to standard inclusion. Getting this to work required hacks such as adding the `include: ["path/to/rust/mod"]` option such that standard includes would search relative to the Rust file that embeds the shader source. This is ugly and unintuitive. I always expected relative includes to Just Work™️ and always search relative to the file where the shader source resides in. The same goes for `path` and `bytes` options being relative to the `Cargo.toml`; they should be relative to the Rust file with the macro invocation. This PR implements just that. It did necessitate raising the MSRV, but it's just one version anyway.

Changelog:
```markdown
### Public dependency updates
- Rust version: 1.88

### Breaking changes
Changes to vulkano-shaders:
- Files specified with the `path` or `bytes` argument are now relative to the Rust file with the `shader!` invocation instead of being relative to your `Cargo.toml`. The old behavior can be achieved using the `root_path_env: "CARGO_MANIFEST_DIR"` macro option.

### Additions
- Vulkano-shaders: relative includes (`#include "..."`) now work in shader source embedded in Rust, and they are relative to the file in which that source is embedded.
```


LukeP0WERS:
Compilation is still failing on my machine for the shader-include example. Looks like the same whitespace problem as before:
```
couldn't read `C:\Users\lukep\Documents\Rust`: The system cannot find the file specified. (os error 2)
couldn't read `examples\shader-include\Projects\vulkano\examples\shader-include\standard-shaders\common.glsl`: The system cannot find the path specified. (os error 3)
couldn't read `examples\shader-include\Projects\vulkano\examples\shader-include\standard-shaders\standard-include.glsl`: The system cannot find the path specified. (os error 3)
couldn't read `examples\shader-include\Projects\vulkano\examples\shader-include\standard-shaders\..\relative-shaders\relative-include.glsl`: The system cannot find the path specified. (os error 3)
```

marc0246:
This PR isn't connected the issue you submitted, but I have been working on your issue.

AustinJ235 (approved):

PR #2800, opened 2026-05-15
Work around Makefile limitations when parsing the dependencies file we get from shaderc

### LINKED ISSUE #2798: Shader inclusion is broken when compiling with `glslc`
### Version of vulkano

master

### OS (incl. version)

Windows

### Selected physical device

irrelevant

### Driver (incl. version)

irrelevant

### Issue description

#2795 broke both standard and relative shader inclusion for me. The `bloom` and `shader-include` examples fail to compile as a result of this.

### Minimal reproducible example

`bloom.rs` fails to compile with the following errors:
```
error: couldn't read `C:\Users\lukep\Documents\Rust`: The system cannot find the file specified. (os error 2)                                                                                                             
   --> examples\bloom\bloom.rs:160:5
    |
160 | /     vulkano_shaders::shader! {
161 | |         ty: "compute",
162 | |         path: "upsample.glsl",
163 | |         include: ["."],
164 | |     }
    | |_____^

error: couldn't read `examples\bloom\Projects\vulkano\examples\bloom\shared_exponent.glsl`: The system cannot find the path specified. (os error 3)                                                                       
   --> examples\bloom\bloom.rs:160:5
    |
160 | /     vulkano_shaders::shader! {
161 | |         ty: "compute",
162 | |         path: "upsample.glsl",
163 | |         include: ["."],
164 | |     }
    | |_____^
```
Just read through the errors again and I think I see the potential source of the issue. There isn't a folder on my computer called: `C:\Users\lukep\Documents\Rust`, but there is `C:\Users\lukep\Documents\Rust Projects`. So it looks like the space in the name is breaking some parsing added in #2795. Also it looks like the second half of the filename snuck its way into the second error lmao.

marc0246:
The Makefile format doesn't support escaping paths, and shaderc simply emits the paths unquoted. This breaks our reflection when the user has directory and/or file names with spaces. This PR implements an ugly hack, but at least it covers all but the most unlikely name(s): consisting of a single space. Naturally, I added a regression test to make sure this doesn't happen again. I also added proper error handling and error messages to the parser to make debugging easier.

Fixes #2798

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -140,45 +140,149 @@ fn compile_into_spirv(
         ));
     }
 
-    let includes = fs::read_to_string(&dependencies_file)
-        .ok()
-        .map(|content| parse_deps_file(&content, vulkano_dir))
-        .unwrap_or_default();
+    let content = &fs::read_to_string(&dependencies_file)
+        .map_err(|e| format!("failed to read dependencies file: {e}"))?;
+    let includes = parse_deps_file(content, vulkano_dir).map_err(|e| {
+        let content = content
+            .lines()
+            .flat_map(|line| ["    ", line])
+            .collect::<String>();
+
+        format!("failed to parse dependencies file: {e}\nfile content:\n{content}")
+    })?;
 
     let words = vulkano::shader::spirv::bytes_to_words(&output.stdout)
         .map(|w| w.into_owned())
-        .map_err(|e| format!("Malformed SPIR-V: {e}"))?;
+        .map_err(|e| format!("malformed SPIR-V: {e}"))?;
 
     Ok((words, includes))
 }
 
 /// Parses a Makefile-format dependency file produced by glslc `-MF`, returning the list of
 /// included file paths. The format is `target: source dep1 dep2 ...`.
-fn parse_deps_file(content: &str, vulkano_dir: &Path) -> Vec<String> {
-    // Join continuation lines (lines ending with `\` followed by newline).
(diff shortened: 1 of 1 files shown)


LukeP0WERS:
Fixes the issue for the vulkano examples. It breaks existing code that includes files relative to the GLSL file, but I assume that's intentional and inevitable when compiling with `glslc`?

LukeP0WERS:
Actually I think I'm just seeing issues that will be fixed by the other PR. The whitespace problem is fixed by this one.

marc0246:
You could have always included relative to a GLSL file. Without the other PR too. The other PR adds relative includes to *embedded* shader code.

LukeP0WERS:
Embedded shader source is inline right? If so the relative shader inclusion is still broken for me on my egui_winit_vulkano fork. I'm not at my computer rn so I don't have the error details but the vs and fs files can't find _push_constants.glsl with relative shader inclusion.

AustinJ235 (approved):

marc0246:
> Embedded shader source is inline right?

Yes, it's the shader code that's inline in the macro call site.

> If so the relative shader inclusion is still broken for me

We have tests for relative includes, as well as the the shader-include example.
thonk

marc0246:
Ah, I see. Relative includes have relative paths in the Makefile :sob: So relative paths are going to get concatenated with the new parsing. I did also think of a solution that would allow relative paths, but decided against it because the current solution covers more names. But we have no choice if this is the only thing shaderc gives us.

marc0246:
I implemented the other solution: #2801. Please test again.

PR #2801, opened 2026-05-16
Fix relative includes borked by #2800 and #2795

marc0246:
This fixes relative includes that were borked by #2800. This solution restricts the file and directory names more, but at least it works, and the restricted file and directory names are still pretty unlikely. Additionally, this fixes the relative includes that were borked by #2795.


LukeP0WERS:
I'm still getting this error:
```couldn't read `src\_push_constants.glsl`: The system cannot find the file specified. (os error 2)```
I'm testing with a slightly modified version of https://github.com/hakolao/egui_winit_vulkano/pull/74
The `shader-include` example is working though and I'm not sure what the difference between them is.

**Comment by marc0246** [comment]
Thank you. What I fixed is not related to that message. I believe that was borked in #2795. I need to see what's actually going on, so I'll add a temporary example with the same directory structure as yours that fails.

**Comment by marc0246** [comment]
Now we have more useful information:
```
failed to parse dependencies file: expected `standard-shaders/../relative-shaders/relative-include.glsl` to be a file existing on the file system
file content:
    -: <stdin> standard-shaders/../relative-shaders/relative-include.glsl standard-shaders/standard-include.glsl standard-shaders/common.glsl
```
The paths are relative to `src/shaders/`, but rustc is opening them relative to `src/`. I think this should be an easy fix.

marc0246:
Please test again. It's superb that you found all these regressions so quickly.

LukeP0WERS:
Lets go, that fixed everything that I'm able to test with atm.

marc0246:
:pray:

AustinJ235 (approved):

PR #2802, opened 2026-05-17
Remove subbuffer from `RecordingCommandBuffer`

CptPotato:
Removed `Subbuffer` from `RecordingCommandBuffer`.

Changelog:

```markdown
### Breaking changes

Changes to `RecordingCommandBuffer`:
- `build_acceleration_structure_indirect`, `try_build_acceleration_structure_indirect`, `validate_build_acceleration_structure_indirect` and `build_acceleration_structure_indirect_unchecked` use an `&Arc<Buffer>` as input instead of `&Subbuffer<[u8]>`.
- `copy_query_pool_results`, `try_copy_query_pool_results`, `validate_copy_query_pool_results` and `copy_query_pool_results_unchecked` use an `&Arc<Buffer>` and `offset` instead of `&Subbuffer<[u8]>`.
```

### DIFF
--- a/vulkano/src/command_buffer/auto/commands/acceleration_structure.rs
+++ b/vulkano/src/command_buffer/auto/commands/acceleration_structure.rs
@@ -7,7 +7,7 @@ use crate::{
         BuildAccelerationStructureMode, CopyAccelerationStructureInfo,
         CopyAccelerationStructureToMemoryInfo, CopyMemoryToAccelerationStructureInfo,
     },
-    buffer::Subbuffer,
+    buffer::{BufferUsage, Subbuffer},
     command_buffer::{
         auto::{Resource, ResourceUseRef2},
         sys::RecordingCommandBuffer,
@@ -17,6 +17,7 @@ use crate::{
     sync::PipelineStageAccessFlags,
     ValidationError,
 };
+use ash::vk::DeviceSize;
 use smallvec::SmallVec;
 use std::sync::Arc;
 
@@ -212,9 +213,11 @@ impl<L> AutoCommandBufferBuilder<L> {
         stride: u32,
         max_primitive_counts: &[u32],
     ) -> Result<(), Box<ValidationError>> {
+        let indirect_buffer = indirect_buffer.buffer();
+
         self.inner.validate_build_acceleration_structure_indirect(
             info,
-            indirect_buffer.buffer(),
+            indirect_buffer.device_address(),
             stride,
             max_primitive_counts,
         )?;
@@ -226,6 +229,27 @@ impl<L> AutoCommandBufferBuilder<L> {
                 vuids: &["VUID-vkCmdBuildAccelerationStructuresIndirectKHR-renderpass"],
                 ..Default::default()
             }));
+        }
+
+        if info.geometries.len() as DeviceSize * stride as DeviceSize > indirect_buffer.size() {
+            return Err(Box::new(ValidationError {
+                 problem: "`info.geometries.len()` * `stride` is greater than the size of \
+                     `indirect_buffer`".into(),
+                 vuids: &["VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pIndirectDeviceAddresses-03646"],
+                 ..Default::default()
+             }));
+        }
+
+        if !indirect_buffer
+            .usage()
+            .intersects(BufferUsage::INDIRECT_BUFFER)
+        {
+            return Err(Box::new(ValidationError {
+                 context: "indirect_buffer".into(),
+                 problem: "the buffer was not created with the `BufferUsage::INDIRECT_BUFFER` usage".into(),
+                 vuids: &["VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pIndirectDeviceAddresses-03647"],
+                 ..Default::default()
+             }));
         }
 
         Ok(())
@@ -250,7 +274,7 @@ impl<L> AutoCommandBufferBuilder<L> {
                 unsafe {
                     out.build_acceleration_structure_indirect_unchecked(
                         &info,
-                        indirect_buffer.buffer(),
+                        indirect_buffer.buffer().device_address(),
                         stride,
                         &max_primitive_counts,
                     )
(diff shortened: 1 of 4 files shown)


marc0246:
The indirect acceleration structure commands are supposed to have an `indirect_device_address` field instead of the `indirect_buffer`. If you only have an `&Arc<Buffer>` argument, there's no way to specify an offset. We could add an `indirect_buffer_offset` argument in addition, but that would be inconsistent with the spec, and take flexibility away. You will notice that there is validation for that indirect buffer, but you can simply rehome that to the auto command buffer's validation. There was never supposed to be validation for this in the first place. These APIs are all already unsafe anyway. So the new way is device addresses for acceleration structures.

Also, for the query commands, you correctly added an `destination_offset` argument; however, they're missing the `stride` argument (which was not there because the `Subbuffer`'s type parameter prevented it), and the fields are supposed to be named like they are in the spec (so `dst_*`).

vulkano/src/command_buffer/auto/commands/acceleration_structure.rs
```diff
@@ -228,6 +231,27 @@ impl<L> AutoCommandBufferBuilder<L> {
...
+            .intersects(BufferUsage::INDIRECT_BUFFER)
+        {
+            return Err(Box::new(ValidationError {
+                 context: "indirect_buffer".into(),
+                 problem: "the buffer was not created with the `BufferUsage::INDIRECT_BUFFER` usage".into(),
+                 vuids: &["VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pIndirectDeviceAddresses-03647"],
+                 ..Default::default()
+             }));
+        }
+
```
CptPotato, line 253: > you can simply rehome that to the auto command buffer's validation

I moved the two validation checks for `indirect_buffer` here. Is that what you had in mind?

CptPotato (commented):
I think I got it the way you've outlined.

I added `stride` with validation for the query commands and switched `build_acceleration_structure` to use a device address (`NonZero<DeviceAddress>`).

marc0246:
I'm aware that you addressed my review, but I haven't been able to make time to rereview yet.

CptPotato:
> I'm aware that you addressed my review, but I haven't been able to make time to rereview yet.

No worries, take your time. I just wanted to signal that I'm done with the changes I was planning to make, so there's no guessing whether im finished or not.

vulkano/src/command_buffer/auto/commands/acceleration_structure.rs
```diff
@@ -212,9 +213,11 @@ impl<L> AutoCommandBufferBuilder<L> {
         stride: u32,
         max_primitive_counts: &[u32],
     ) -> Result<(), Box<ValidationError>> {
+        let indirect_buffer = indirect_buffer.buffer();
```
marc0246: This discards the offset. `Subbuffer::device_address` returns the device address of the subbuffer, not the buffer.

vulkano/src/command_buffer/auto/commands/acceleration_structure.rs
marc0246: Yes.

vulkano/src/command_buffer/auto/commands/acceleration_structure.rs
```diff
@@ -250,7 +274,7 @@ impl<L> AutoCommandBufferBuilder<L> {
                 unsafe {
                     out.build_acceleration_structure_indirect_unchecked(
                         &info,
-                        &indirect_buffer,
+                        indirect_buffer.buffer().device_address(),
```
marc0246:
```suggestion
                        indirect_buffer.device_address(),
```

vulkano/src/command_buffer/commands/query.rs
```diff
@@ -628,14 +641,12 @@ impl RecordingCommandBuffer {
...
-        assert_eq!(device, destination.buffer().device());
+        assert_eq!(device, dst_buffer.device());
         assert_eq!(device, query_pool.device());
 
         // VUID-vkCmdCopyQueryPoolResults-flags-00822
         // VUID-vkCmdCopyQueryPoolResults-flags-00823
-        debug_assert!(destination
-            .offset()
-            .is_multiple_of(size_of::<T>() as DeviceSize));
+        debug_assert!(dst_offset.is_multiple_of(size_of::<T>() as DeviceSize));
```
marc0246: These VUIDs need to be validated. `Subbuffer` ensured this by construction before.

vulkano/src/command_buffer/commands/query.rs
```diff
@@ -655,22 +666,18 @@ impl RecordingCommandBuffer {
         let per_query_len = query_pool.result_len(flags);
         let required_len = per_query_len * query_count as DeviceSize;
 
-        if destination.len() < required_len {
+        if dst_buffer.size() < required_len {
```
marc0246: This doesn't take the offset into account, and `Buffer::size` is in bytes, not elements.

vulkano/src/command_buffer/commands/query.rs
```diff
@@ -689,6 +696,29 @@ impl RecordingCommandBuffer {
             }));
         }
 
+        if query_count > 1 {
+            if stride == 0 {
+                return Err(Box::new(ValidationError {
+                    problem: "`query_count` is greater than 1 but `stride` is zero".into(),
```
marc0246:
```suggestion
                    problem: "`query_count` is greater than 1, but `stride` is zero".into(),
```

vulkano/src/command_buffer/commands/query.rs
```diff
@@ -689,6 +696,29 @@ impl RecordingCommandBuffer {
...
+                    vuids: &["VUID-vkCmdCopyQueryPoolResults-queryCount-09438"],
+                    ..Default::default()
+                }));
+            }
+
+            if !stride.is_multiple_of(std::mem::size_of::<T>() as DeviceSize) {
```
marc0246: `size_of[_val]` and `align_of[_val]` are in the prelude as of Rust 1.80.
```suggestion
            if !stride.is_multiple_of(size_of::<T>() as DeviceSize) {
```

vulkano/src/command_buffer/commands/query.rs
```diff
@@ -689,6 +696,29 @@ impl RecordingCommandBuffer {
...
+                return Err(Box::new(ValidationError {
+                    problem: "`query_count` is greater than 1 but `stride` is not a multiple of \
```
marc0246:
```suggestion
                    problem: "`query_count` is greater than 1, but `stride` is not a multiple of \
```

marc0246:
`NonZero` should not be an argument. It's too much syntactic overhead to construct. `NonZero` *should* be a return type for the same reason.

PR #2803, opened 2026-05-17
Replace `todo!()` in `suballocations` method for `BuddyAllocator`

LukeP0WERS:
I added the missing `suballocations` logic for `BuddyAllocator`. Since `BuddyAllocator` only stores free nodes, we can't determine the `SuballocationType` for allocated regions without storing additional information in the allocator so I set the `SuballocationType` to `Unknown` for those regions. I don't need the type for my purposes, but if yall think it would be useful it should be possible to track without too much added overhead.

### DIFF
--- a/vulkano/src/memory/allocator/suballocator/buddy.rs
+++ b/vulkano/src/memory/allocator/suballocator/buddy.rs
@@ -2,9 +2,14 @@ use super::{
     AllocationType, Region, Suballocation, SuballocationNode, Suballocator, SuballocatorError,
 };
 use crate::{
-    DeviceSize, memory::{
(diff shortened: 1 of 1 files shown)


marc0246:
This is an interesting solution, and would certainly work as a stand-in. However, as you surely know, I don't accept half-baked solutions :pensive: An iterator isn't supposed to allocate; the whole point is after all that it iterates over an existing data structure. I left this out because the proper solution entails implementing a binary tree (which is the structure that a buddy allocator forms) and I didn't think anyone would *actually use* a buddy allocator. But since you seem to be in need of it, I'll implement it real quick.

LukeP0WERS:
Thanks. The buddy allocator is very useful for ^2 sized data.

PR #2808, opened 2026-05-29
Fix `BuddyAllocator` region offset not having to be aligned to `BuddyAllocator::MIN_NODE_SIZE`

marc0246:
#2807 added this as part of a comment:

> The division can't discard any bits because offsets and sizes are aligned to `BuddyAllocator::MIN_NODE_SIZE`.

However, that's not true. There was no validation to ensure that a region's offset is aligned to `BuddyAllocator::MIN_NODE_SIZE`, which means that the offsets could have had bits set in those alignment bits, and those would be discarded.


marc0246:
This is nonsense. Why would you open this?!

Issue #1530, opened 2021-04-01
vulkano-shaders generate different types with the same name in one namespace


p0lunin:
For the following code:
```glsl
struct Foo { uint a; };
layout(std140, set = 0, binding = 0) readonly buffer A{
    Foo foo[];
};
layout(set = 1, binding = 0) readonly buffer B {
    Foo foo[];
};
void main() {}
```
`vulkano-shaders` generates 2 types with name `Foo`. I think this is because shader uses the same structure with different memory layouts `std140` and `shared` (uses by default).

Rua:
Clearly these need to have different names, but what should they be?

p0lunin:
Maybe macro will create namespaces with memory layout as name? e.g. `types::std140::Foo` and `types::packed::Foo`.

Rua:
What about the struct names `A` and `B`? How come those aren't used by Vulkano-shaders?

Rua:
I accidentally stumbled upon this old issue. I'll elaborate a bit more for future reference. Given this definition in GLSL:
```glsl
struct Foo {
    float bar;
};

layout(push_constant) uniform PushConstants {
    Foo x;
} pc;

layout(binding = 0) uniform Uniforms {
    Foo x;
} uni;
```

The shader compiler will generate this SPIR-V output:
```
               OpName %Foo "Foo"
        %Foo = OpTypeStruct %float
%PushConstants = OpTypeStruct %Foo
%_ptr_PushConstant_PushConstants = OpTypePointer PushConstant %PushConstants
         %pc = OpVariable %_ptr_PushConstant_PushConstants PushConstant
```
```
               OpName %Foo_0 "Foo"
      %Foo_0 = OpTypeStruct %float
   %Uniforms = OpTypeStruct %Foo_0
%_ptr_Uniform_Uniforms = OpTypePointer Uniform %Uniforms
        %uni = OpVariable %_ptr_Uniform_Uniforms Uniform
```

Two different structures with the name `Foo` are defined in the SPIR-V; one for the push constants and another copy for the uniform buffer. The same thing happens with a uniform buffer combined with any other kind of input variable, such as a storage buffer, as long as they use the same GLSL struct in their definition. This happens because uniform buffers use the extended alignment (std140) by default, instead of the base alignment (std430) used by everything else. The difference in layout becomes more clear when `bar` is turned into an array `float bar[2]`:

```
               OpDecorate %_arr_float_uint_2 ArrayStride 4
[...]

PR #2810, opened 2026-06-09
Add Slang shader language

### LINKED ISSUE #2622: vulkano-shaders: Support compiling Slang source

nshyrei:
Adds Slang shader language support to `vulkano-shaders` and `vulkano-macros`, including SPIR-V compilation and automatic Rust struct derivation from Slang source.

Fixes #2622

While working on this PR I found the following Slang-related issues worth discussing:

1) Duplicate struct names from StructuredBuffer

The recommended way to declare GPU buffers in Slang is via the StructuredBuffer type which looks like this:
```
[vk::binding(1, 0)] StructuredBuffer<uint> BufferA;
[vk::binding(2, 0)] StructuredBuffer<float> BufferB;
```
The resulting SPIR-V code produced by Slang doesn't contain discriminants in `OpName` instructions and just outputs `StructuredBuffer` for both definitions which in turn triggers duplicate type detection code in `vulkano-macros`. To solve this I append the name of the struct with an underlying generic type which acts as a discriminant.

GLSLC doesn't have this problem because it allows explicit buffer naming like:
```
layout(set = 0, std430, binding = 1) buffer BufferA {
     float data[];
};

layout(set = 0, std430, binding = 2) buffer BufferB {
    uint data[];
};
```

2) Memory layout suffix in generated struct names

Slang appends the name of the memory layout qualifier to a struct name, resulting in a generated Rust structs having unexpected names like `StructuredBuffer_std430` instead of `StructuredBuffer`. Minimal reproducible example:
```
 struct UintValue { uint data; };

  [vk::binding(0, 0)] RWStructuredBuffer<float> output;
  [vk::binding(1, 0)] StructuredBuffer<UintValue> buffer_1;

  [numthreads(1, 1, 1)]
  void main(uint3 dispatchThreadID : SV_DispatchThreadID) {
      output[dispatchThreadID.x] = (float)buffer_1[dispatchThreadID.x].data;
  }
```
Currently nothing is done about this behavior, but I do feel that we should strip the memory qualifier suffix.

3) Path resolution in `#include` directives.

Slang resolves .. in include paths to their canonical form before writing them to the depfile, so `include_dir_a/../include_dir_b/target_b.glsl` appears as `include_dir_b/target_b.glsl`. GLSLC preserves the unresolved form.
Currently nothing is done about this behavior.

4) Spaces in `#include` directives.

 Slang strips spaces from include filenames when using angle-bracket includes (`#include <foo bar.glsl>` becomes a lookup for `foobar.glsl`), so files with spaces in their names cannot be resolved this way. GLSLC resolves both space includes correctly. For now I skipped includes with spaces in unit tests for Slang.

5) Path resolution in depfiles on Windows

  Slang's depfile escaping double-escapes Windows paths — `C\\:` becomes `C\:`. The depfile parser has to replace `\\:` back to a singular `:` to recover a valid Windows path. GLSLC doesn't escape colons, so Windows drive letters pass through unmodified.

### DIFF
--- a/vulkano-shaders/src/codegen.rs
+++ b/vulkano-shaders/src/codegen.rs
@@ -400,13 +400,13 @@ fn normalize_str(path: impl AsRef<Path>) -> String {
 }
 
 pub(super) fn compile(
-    source_language: SourceLanguage,
     input: &MacroInput,
     source: &str,
     working_dir: &Path,
     shader_kind: ShaderKind,
     macro_defines: &[(String, String)],
 ) -> Result<(Vec<u32>, Vec<String>), String> {
+    let source_language = input.source_language.unwrap_or(SourceLanguage::Glsl);
     let mut compile_options = CompileOptions::new();
 
     compile_options.source_language = source_language;
@@ -523,20 +523,12 @@ mod tests {
     use vulkano::shader::reflect;
 
     fn compile_inline(
-        source_language: SourceLanguage,
         input: &MacroInput,
         source: &str,
         shader_kind: ShaderKind,
         macro_defines: &[(String, String)],
     ) -> Result<(Vec<u32>, Vec<String>), String> {
-        compile(
-            source_language,
-            input,
-            source,
-            Path::new("."),
-            shader_kind,
-            macro_defines,
-        )
+        compile(input, source, Path::new("."), shader_kind, macro_defines)
     }
 
     fn convert_paths(root_path: &Path, paths: &[PathBuf]) -> HashSet<String> {
@@ -580,15 +572,22 @@ mod tests {
         let root_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("tests");
 
         let (_spirv, includes) = compile_inline(
-            source_language,
             &MacroInput {
+                source_language: Some(source_language),
                 include_directories: vec![
                     root_path.join("include_dir_a"),
                     root_path.join("include_dir_b"),
                 ],
                 ..MacroInput::empty()
             },
-            &format!("{shader_prefix}\n#include <target_a.glsl>\n#include <target_b.glsl>\n{shader_suffix}"),
+            &format!(
+                "
+                    {shader_prefix}
+                    #include <target_a.glsl>
+                    #include <target_b.glsl>
+                    {shader_suffix}
+                ",
+            ),
             ShaderKind::Vertex,
             &[],
         )
@@ -624,12 +623,18 @@ mod tests {
     fn include_resolution_dotdot_glslc() {
         let root_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("tests");
 
-        let (_spirv2, includes2) = compile_inline(SourceLanguage::Glsl,
+        let (_spirv2, includes2) = compile_inline(
             &MacroInput {
+                source_language: Some(SourceLanguage::Glsl),
                 include_directories: vec![root_path.join("include_dir_a")],
                 ..MacroInput::empty()
             },
-            "#version 450\n#include <target_a.glsl>\n#include <../include_dir_b/target_b.glsl>\nvoid main() {}",
+            r#"
+                #version 450
+                #include <target_a.glsl>
+                #include <../include_dir_b/target_b.glsl>
+                void main() {}
+            "#,
             ShaderKind::Vertex,
             &[],
         )
@@ -652,9 +657,17 @@ mod tests {
             .expect("cannot run tests in a folder with non unicode characters")
             .replace('\\', "/");
         let (_spirv3, includes3) = compile_inline(
-            SourceLanguage::Glsl,
-            &MacroInput::empty(),
-            &format!("#version 450\n#include \"{absolute_path_str}\"\nvoid main() {{}}"),
+            &MacroInput {
+                source_language: Some(SourceLanguage::Glsl),
+                ..MacroInput::empty()
+            },
+            &format!(
+                "
+                    #version 450
+                    #include \"{absolute_path_str}\"
+                    void main() {{}}
+                ",
+            ),
             ShaderKind::Vertex,
             &[],
         )
@@ -669,15 +682,19 @@ mod tests {
         );
 
         let (_spirv4, includes4) = compile_inline(
-            SourceLanguage::Glsl,
             &MacroInput {
+                source_language: Some(SourceLanguage::Glsl),
                 include_directories: vec![
                     root_path.join("include_dir_b"),
                     root_path.join("include_dir_c"),
                 ],
                 ..MacroInput::empty()
             },
-            "#version 450\n#include <target_c.glsl>\nvoid main() {}",
+            r#"
+                #version 450
+                #include <target_c.glsl>
+                void main() {}
+            "#,
             ShaderKind::Vertex,
             &[],
         )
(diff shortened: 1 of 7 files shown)


vulkano-shaders/src/structs.rs
```diff
@@ -853,6 +878,16 @@ impl TypeStruct {
             members.push(Member { ident, ty, offset });
         }
 
+        let ident =
```
nshyrei: This is where I append generic type of a StructuredBuffer (`element_type`) to a Rust struct identifier.

vulkano-shaders/src/codegen.rs
```diff
@@ -242,10 +384,13 @@ fn parse_deps_file(
         path.push_str(token);
 
         if has_extension {
-            let normalized = if Path::new(&path).is_absolute() {
-                normalize_str(path)
-            } else {
-                normalize_str(PathBuf::from_iter([working_dir, &PathBuf::from(path)]))
+
+            let normalized = {
+                let mut result = normalize_str(path.replace("\\:", ":"));
```
nshyrei, line 348: Fix double-escaped Windows drive letter paths in Slang depfiles.

nshyrei:
@marc0246, @CptPotato, @Rua Please take a look, I have marked the PR as draft because I would like to have your input about how to proceed with the issues that I have found.

marc0246:
1. That honestly seems like a slangc bug. `StructuredBuffer<T>` should at least be given an `OpName` like `StructuredBuffer_T`. So your solution seems fine.
2. This is actually a good thing, and it's what I'm planning to add to glslc's `OpName`s. Reason being that, ironically, glslc generates identical names for the same struct used in `std140` vs `std430` blocks. It's the same struct in GLSL, but it generates 2 different SPIR-V structs because of the differing layout. See #1530. Now, I'm warning you that this will require almost a rewrite of the reflection in vulkano-shaders because we don't just need reflection but also a degree of *decompilation* to recover the alignment. It's nowhere to be found in SPIR-V. I'll do this myself &ndash; I already know how I want to do it. Just make sure that when it comes to `StructuredBuffer<T>`, it's `StructuredBuffer_T_stdX`.
3. I don't think this matters, right? We're normalizing them either way.
4. Nothing we can do about it.
5. Your solution seems fine.

As for the code itself, what immediately jumps out is the use of a type parameter `C` in `codegen::compile`. The reason being that type parameters (but not lifetime parameters) need to be monomorphized for each instantiation of the parameter (or even worse, Cartesian product of parameter instantiations in the case of multiple). This puts unnecessary strain on the compiler, resulting in worse compile times and bigger binaries. That's why you shouldn't use type parameters unless it's absolutely necessary, especially in a library. And when they are necessary, you should manually monomorphize as much as possible such that the amount the compiler has to monomorphize is minimal.

Also, instead of `C::compile_into_spirv`, you can simply use a match on the `source_language`. However, please make sure not to use a fallback match arm `_ =>`. The reason is that we want this to break when a new variant is added. So like this:
```rs
match source_langauge {
    SourceLanguage::Glsl | SourceLanguage::Hlsl => compile_into_spirv_glslc(...),
    SourceLanguage::Slang => compile_into_spirv_slangc(...),
}
```

nshyrei:
@marc0246 removed generics in the latest commit, please review.

marc0246:
You need to run `cargo +nightly fmt` to fix that CI job, and undo the changes to `codegen::reflect` (I suspect this is a rebase oopsie woopsie). I'll review more thoroughly in a bit.

marc0246:
You should not update the CI workflow. We use Nix for CI; all you need to do is add `shader-slang` before (every) `shaderc` line in `flake.nix`.

vulkano-shaders/src/codegen.rs
```diff
@@ -1153,4 +1530,104 @@ mod tests {
...
+                  [vk::binding(2, 0)] StructuredBuffer<uint> buffer_2;
+
+                  [numthreads(1, 1, 1)]
+                  void main(uint3 dispatchThreadID : SV_DispatchThreadID) {
+                    output[dispatchThreadID.x] = (float)buffer_1[dispatchThreadID.x] + buffer_2[dispatchThreadID.x];
+                  }
+            "#,
+            ShaderKind::Compute,
+            &[],
+        ).expect("slangc failed to compile shader");
```
marc0246: Here for example, the way the string was formatted prevented everything else in the statement from being formatted.
```suggestion
        let (words, _includes) = compile_inline(
            SourceLanguage::Slang,
            &MacroInput::empty(),
            r#"
                [vk::binding(0, 0)] RWStructuredBuffer<float> output;
                [vk::binding(1, 0)] StructuredBuffer<float> buffer_1;
                [vk::binding(2, 0)] StructuredBuffer<uint> buffer_2;

                [numthreads(1, 1, 1)]
                void main(uint3 dispatchThreadID : SV_DispatchThreadID) {
                    output[dispatchThreadID.x] = (float)buffer_1[dispatchThreadID.x]
                        + buffer_2[dispatchThreadID.x];
                }
            "#,
            ShaderKind::Compute,
            &[],
        )
        .expect("slangc failed to compile shader");
```

vulkano-shaders/src/codegen.rs
```diff
@@ -854,48 +1203,75 @@ mod tests {
...
+            SourceLanguage::Slang,
+            "",
+            "float4 main() : SV_Position { return float4(0, 0, 0, 1); }",
+        );
+    }
+
+    fn macros(source_language: SourceLanguage, shader_prefix: &str, shader_suffix: &str) {
+        let need_defines = format!(
+            "{shader_prefix}\n#ifndef NAME1\n#error NAME1 must be defined\n#endif\n#if NAME2 <= 29\n#error NAME2 must be greater than 29\n#endif\n{shader_suffix}"
+        );
```
marc0246, line 1408: Here's an example of how you can make this more readable.
```suggestion
        let need_defines = format!(
            "
                {shader_prefix}
                #ifndef NAME1
                #error NAME1 must be defined
                #endif
                #if NAME2 <= 29
                #error NAME2 must be greater than 29
                #endif
                {shader_suffix}
            ",
        );
```

marc0246 (commented):
Is there a reason you converted line breaks in strings to `\n`? That's really hard to read :sweat_smile: There are also strings that exceed 100 columns. Please don't do that. Apart from being unconventional, it also makes rustfmt give up on the whole statement.

vulkano-shaders/src/lib.rs
```diff
@@ -296,7 +296,9 @@ fn shader_inner(mut input: MacroInput) -> Result<TokenStream> {
     {
         let (code, types) = match source_kind {
             SourceKind::Src(source) => {
+                let source_language = input.source_language.unwrap_or(SourceLanguage::Glsl);
```
marc0246: I think doing this here is unnecessary duplication, and the `compile` function should do this internally like it did before.

nshyrei: fixed in [commit](https://github.com/vulkano-rs/vulkano/pull/2810/commits/db09c1a1722b6dca478efd7430ff7c2af63113dd)

vulkano-shaders/src/codegen.rs
nshyrei: fixed in [commit](https://github.com/vulkano-rs/vulkano/pull/2810/commits/db09c1a1722b6dca478efd7430ff7c2af63113dd)

vulkano-shaders/src/codegen.rs
nshyrei: fixed in [commit](https://github.com/vulkano-rs/vulkano/pull/2810/commits/db09c1a1722b6dca478efd7430ff7c2af63113dd)

nshyrei:
@marc0246 Fixed line breaks in the latest commit. I ran some find-replace command in my IDE that replaced line breaks in shader code, funny that the tests still worked.

marc0246:
> I ran some find-replace command in my IDE that replaced line breaks in shader code

Makes sense.

marc0246:
Yeehaw! We finally have Slang support in vulkano-shaders :tada:

PR #2813, opened 2026-06-16
Add support for the `ray_tracing_pipeline_trace_rays_indirect` device feature

LukeP0WERS:
I've added the `trace_rays_indirect` and related functions to `RecordingCommandBuffer`. Its a bit strange that `cmd_trace_rays_indirect_khr` only takes a device address for the indirect command buffer and no offset, but everything else was pretty straightforward. Let me know if I missed anything in the docs or validation and I'll fix it.

### DIFF
--- a/vulkano-taskgraph/src/command_buffer/commands/pipeline.rs
+++ b/vulkano-taskgraph/src/command_buffer/commands/pipeline.rs
@@ -1226,11 +1226,11 @@ impl RecordingCommandBuffer<'_> {
         self
     }
 
-    /// Performs multiple ray tracing operations using a ray tracing pipeline, panicking on a
-    /// validation error.
+    /// Performs a single ray tracing operation using a ray tracing pipeline, reading the
+    /// ray trace query dimensions from a separate buffer.
     ///
-    /// One ray tracing operation is performed for each `TraceRaysIndirectCommand` struct that
-    /// is read from `buffer`.
+    /// A single ray tracing operation is performed for the `TraceRaysIndirectCommand` struct that
+    /// is read from `buffer` starting at `offset`.
     ///
     /// A ray tracing pipeline must have been bound using [`bind_pipeline_ray_tracing`]. Any
     /// resources used by the ray tracing pipeline, such as descriptor sets, must have been set
@@ -1256,11 +1256,13 @@ impl RecordingCommandBuffer<'_> {
         &mut self,
         shader_binding_table_addresses: &ShaderBindingTableAddresses,
         buffer: Id<Buffer>,
+        offset: DeviceSize,
     ) -> &mut Self {
-        unsafe { self.try_trace_rays_indirect(shader_binding_table_addresses, buffer) }.unwrap()
+        unsafe { self.try_trace_rays_indirect(shader_binding_table_addresses, buffer, offset) }
+            .unwrap()
     }
 
-    /// Performs multiple ray tracing operations using a ray tracing pipeline.
+    /// Performs multiple ray tracing operations using a ray tracing pipeline
     ///
     /// A ray tracing pipeline must have been bound using [`bind_pipeline_ray_tracing`]. Any
     /// resources used by the ray tracing pipeline, such as descriptor sets, must have been set
@@ -1278,14 +1280,18 @@ impl RecordingCommandBuffer<'_> {
         &mut self,
         shader_binding_table_addresses: &ShaderBindingTableAddresses,
         buffer: Id<Buffer>,
+        offset: DeviceSize,
     ) -> Result<&mut Self> {
-        Ok(unsafe { self.trace_rays_indirect_unchecked(shader_binding_table_addresses, buffer) })
+        Ok(unsafe {
+            self.trace_rays_indirect_unchecked(shader_binding_table_addresses, buffer, offset)
+        })
     }
 
     pub unsafe fn trace_rays_indirect_unchecked(
         &mut self,
         shader_binding_table_addresses: &ShaderBindingTableAddresses,
         buffer: Id<Buffer>,
+        offset: DeviceSize,
     ) -> &mut Self {
         let buffer = unsafe { self.accesses.buffer_unchecked(buffer) };
 
@@ -1302,7 +1308,7 @@ impl RecordingCommandBuffer<'_> {
                 &miss,
                 &hit,
                 &callable,
-                buffer.device_address().into(),
+                buffer.device_address().get() + offset,
             )
         };
 
--- a/vulkano-taskgraph/src/command_buffer/commands/pipeline.rs
+++ b/vulkano-taskgraph/src/command_buffer/commands/pipeline.rs
@@ -1226,11 +1226,9 @@ impl RecordingCommandBuffer<'_> {
         self
     }
 
-    /// Performs a single ray tracing operation using a ray tracing pipeline, reading the
-    /// ray trace query dimensions from a separate buffer.
-    ///
-    /// A single ray tracing operation is performed for the `TraceRaysIndirectCommand` struct that
-    /// is read from `buffer` starting at `offset`.
+    /// Performs a single ray tracing operation using a ray tracing pipeline, panicking on a  
+    /// validation error. One ray tracing operation is performed for the  
+    /// [`TraceRaysIndirectCommand`] struct that is read from `indirect_device_address`.
     ///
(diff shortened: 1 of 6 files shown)


vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -1701,6 +1702,70 @@ impl<L> AutoCommandBufferBuilder<L> {
...
+    /// - The [safety requirements for `TraceRaysIndirectCommand`] apply.
+    ///
+    /// # Panics
+    ///
+    /// - Panics if [`try_trace_rays_indirect`] returns a [`ValidationError`].
+    ///
+    /// [`bind_pipeline_ray_tracing`]: Self::bind_pipeline_ray_tracing
+    /// [shader safety requirements]: vulkano::shader#safety
+    /// [safety requirements for `TraceRaysIndirectCommand`]: TraceRaysIndirectCommand#safety
+    /// [`try_trace_rays_indirect`]: Self::try_trace_rays_indirect
```
CptPotato: I think you copied the doc comment from your implementation over at `RecordingCommandBuffer::trace_rays_indirect`. Some of it does not apply here.

To my knowledge, `AutoCommandBufferBuilder` is going to be deprecated. So I'm not sure if new additions are worth it at this point.

LukeP0WERS: It isn't deprecated yet, its only a few lines of code anyways. Docs should be fixed

LukeP0WERS:
It's very annoying to use this without an offset parameter, so I'm going to add it to the API. Pretty sure you can just add the offset to the device address, but let me know if that's wrong.

LukeP0WERS:
Also just realized that all the docs I wrote are wrong since it actually only performs a single ray tracing operation based on the dispatch you provide.

marc0246:
Both `RecordingCommandBuffer`s should only have an `indirect_device_address` for the indirect buffer. The rule is to mirror the underlying API. `AutoCommandBufferBuilder` can stay as it is since it's not possible to JIT synchronize the indirect buffer otherwise.

LukeP0WERS:
How would we provide validation for the buffer usage, size limits are exceeded, ect.?

marc0246:
Those can be moved to the `AutoCommandBufferBuilder`'s validation. We simply won't validate them otherwise.

vulkano-taskgraph/src/command_buffer/commands/pipeline.rs
```diff
@@ -1225,4 +1225,93 @@ impl RecordingCommandBuffer<'_> {
 
         self
     }
+
+    /// Performs a single ray tracing operation using a ray tracing pipeline, reading the
+    /// ray trace query dimensions from a separate buffer.
+    ///
+    /// A single ray tracing operation is performed for the `TraceRaysIndirectCommand` struct that
+    /// is read from `buffer` starting at `offset`.
```
marc0246:
```suggestion
    /// Performs a single ray tracing operation using a ray tracing pipeline, panicking on a
    /// validation error. One ray tracing operation is performed for the
    /// [`TraceRaysIndirectCommand`] struct that is read from `indirect_device_address`.
```

vulkano-taskgraph/src/command_buffer/commands/pipeline.rs
```diff
@@ -1225,4 +1225,93 @@ impl RecordingCommandBuffer<'_> {
...
+        shader_binding_table_addresses: &ShaderBindingTableAddresses,
+        indirect_device_address: DeviceAddress,
+    ) -> &mut Self {
+        unsafe {
+            self.try_trace_rays_indirect(shader_binding_table_addresses, indirect_device_address)
+        }
+        .unwrap()
+    }
+
+    /// Performs multiple ray tracing operations using a ray tracing pipeline
```
marc0246:
```suggestion
    /// Performs a single ray tracing operation using a ray tracing pipeline. One ray tracing
    /// operation is performed for the [`TraceRaysIndirectCommand`] struct that is read from
    /// `indirect_device_address`.
```

vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -1701,6 +1702,115 @@ impl<L> AutoCommandBufferBuilder<L> {
         self
     }
 
+    /// Performs a single ray tracing operation using a ray tracing pipeline, reading the
+    /// ray trace query dimensions from a separate buffer.
+    ///
+    /// A single ray tracing operation is performed for the `TraceRaysIndirectCommand` struct that
+    /// is read from `indirect_buffer`.
```
marc0246:
```suggestion
    /// Performs a single ray tracing operation using a ray tracing pipeline. One ray tracing
    /// operation is performed for the [`TraceRaysIndirectCommand`] struct that is read from
    /// `indirect_buffer`.
```

vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -1701,6 +1702,115 @@ impl<L> AutoCommandBufferBuilder<L> {
...
+        let indirect_device_address = buffer.device_address().get() + offset;
+
+        self.inner.validate_trace_rays_indirect(
+            shader_binding_table_addresses,
+            indirect_device_address,
+        )?;
+
+        if !buffer.usage().intersects(BufferUsage::INDIRECT_BUFFER) {
+            return Err(Box::new(ValidationError {
+                context: "buffer.usage()".into(),
```
marc0246:
```suggestion
                context: "indirect_buffer.usage()".into(),
```

vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -1701,6 +1702,115 @@ impl<L> AutoCommandBufferBuilder<L> {
...
+        }
+
+        if !offset.is_multiple_of(4) {
+            return Err(Box::new(ValidationError {
+                context: "offset".into(),
+                problem: "is not a multiple of 4".into(),
+                vuids: &["VUID-vkCmdTraceRaysIndirectKHR-indirectDeviceAddress-03634"],
+                ..Default::default()
+            }));
+        }
```
marc0246: This VUID should be validated by `RecordingCommandBuffer` instead. Additionally, this is guaranteed by `Subbuffer`.

vulkano/src/command_buffer/auto/commands/pipeline.rs
```diff
@@ -1701,6 +1702,115 @@ impl<L> AutoCommandBufferBuilder<L> {
...
+            .is_none_or(|end| end > buffer.size())
+        {
+            return Err(Box::new(ValidationError {
+                problem: "`offset + size_of::<TraceRaysIndirectCommand>()` is greater than \
+                    `buffer.size()`"
+                    .into(),
+                vuids: &["VUID-vkCmdTraceRaysIndirectKHR-indirectDeviceAddress-03636"],
+                ..Default::default()
+            }));
+        }
```
marc0246: This is guaranteed by `Subbuffer`.

vulkano/src/command_buffer/commands/pipeline.rs
```diff
@@ -1674,4 +1675,105 @@ impl RecordingCommandBuffer {
...
+    ) -> Result<(), Box<ValidationError>> {
+        if !self.device().enabled_features().ray_tracing_pipeline {
+            return Err(Box::new(ValidationError {
+                requires_one_of: RequiresOneOf(&[RequiresAllOf(&[Requires::DeviceFeature(
+                    "ray_tracing_pipeline",
+                )])]),
+                ..Default::default()
+            }));
+        }
+
```
marc0246: This is implied by the feature check below.
```suggestion
```

vulkano/src/command_buffer/auto/commands/pipeline.rs
marc0246: This VUID is still missing from `RecordingCommandBuffer`'s validation.

LukeP0WERS:
> Those can be moved to the `AutoCommandBufferBuilder`'s validation. We simply won't validate them otherwise.

Are there long term plans for task graph validation, or is it impossible for some reason?

marc0246:
It's not *impossible*, but when it comes to a performant solution then yes. There would be overhead regardless of whether you use the validation of not. Validation is not a good enough reason to not have what the original API gives us, especially when it comes to something so incredibly unsafe as running shaders.

PR #2818, opened 2026-06-30
Support for the `ext_memory_budget` device extension

LukeP0WERS:
Added `memory_budget` and related functions to `PhysicalDevice` that return the `MemoryBudget` struct added inside `vulkano::memory`.

### DIFF
--- a/vulkano/src/device/physical.rs
+++ b/vulkano/src/device/physical.rs
@@ -11,7 +11,7 @@ use crate::{
     },
     instance::{Instance, InstanceOwned},
     macros::{impl_id_counter, vulkan_bitflags, vulkan_enum},
-    memory::{ExternalMemoryHandleType, MemoryProperties},
+    memory::{ExternalMemoryHandleType, MemoryBudget, MemoryProperties},
     swapchain::{
         ColorSpace, FullScreenExclusive, PresentMode, Surface, SurfaceApi, SurfaceCapabilities,
         SurfaceInfo, SurfaceInfo2ExtensionsVk,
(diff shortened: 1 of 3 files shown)


vulkano/src/device/physical.rs
```diff
@@ -497,6 +497,79 @@ impl PhysicalDevice {
...
+    /// [`ext_memory_budget`]: crate::device::DeviceExtensions::ext_memory_budget
+    #[inline]
+    pub fn try_memory_budget(&self) -> Result<MemoryBudget, Box<ValidationError>> {
+        self.validate_memory_budget()?;
+
+        Ok(unsafe { self.memory_budget_unchecked() })
+    }
+
+    fn validate_memory_budget(&self) -> Result<(), Box<ValidationError>> {
+        if !self.supported_extensions().ext_memory_budget {
```
marc0246, line 538: The extension must be *enabled*, not just supported.

marc0246: It's good that you changed the docs, but you must change this to `.enabled_extensions()` as well.

LukeP0WERS: There isn't an enabled device extensions list for the physical device. Other similar methods only check the supported extensions like `validate_tool_properties` for example.

marc0246: Right. I somehow missed that this is a physical device method :clueless: And in that case, the extension indeed only needs to be supported. So what you had initially was correct. Sorry about that 😅

PR #2826, opened 2026-07-27
Fix `Resources::wait_idle` not cleaning up old swapchains

marc0246:


vulkano-taskgraph/src/resource/state.rs
```diff
@@ -762,6 +777,36 @@ impl SwapchainState {
...
+            for swapchain_id in garbage.swapchains {
+                unsafe { resources.remove_invalidated_swapchain_unchecked(swapchain_id) };
+            }
+
+            for semaphore in garbage.semaphores {
+                sync_state.deallocate_semaphore(semaphore);
+            }
+
+            garbage_index += 1;
+        }
```
AustinJ235: `while garbage_index < sync_state.garbage_queue.len()` seems weird, is this correct? The garbage queue may not always fully empty unless this is intended.

Why not:
```rust
while let Some(garbage) = sync_state.garbage_queue.pop_front() {
```

marc0246: `while let` would indeed be better. I just found a bug though: I thought I fixed the loop in `collect`, but I actually made it potentially skip over present operations because there is a `VecDequeue::remove` in the middle :facepalm: So I'll un"fix" that.

AustinJ235 (approved):

PR #2827, opened 2026-08-07
use crate_ident in derive_buffer_contents hack

vxpm:
fix a very small mistake in the `derive_buffer_contents` macro that made it incompatible with using `vulkano` under another name.


marc0246:
I'm sorry you had to wait so long, and thank you for the fix!

PR #2828, opened 2026-08-21
Allow using custom create functions for Device and Instance

Schmarni-Dev:
adds new_custom functions that take a closure which is in charge of actually calling the raw vulkan create function, the reasoning being that in some cases you might want to call a wrapper instead, like in the case of OpenXR with the XR_KHR_vulkan_enable2 ext, where you have to make an OpenXR call to create the vulkan instance and device.

this seems like a cleaner approach than trying to somehow publicly expose the to_vk methods.

Changelog:
```markdown
### Breaking changes
Changes to devices and queues:
- `Device::new` and friends now return a `Vec` of queues rather than an iterator.

### Additions
- Added `new_with` constructor functions to `Device` and `Instance`, which can be used to manually invoke create functions while using vulkano's create infos.
```


marc0246:
Just to confirm, you're aware that you can use `Instance::from_handle` and `Device::from_handle` with OpenXR, right? I take it, then, that the intent is to not duplicate the extensions and features and/or to not have to write that ugly ash code yourself.

I was never intending to make the vulkano -> ash conversions public anyway because those are internal details, and they're only going to get interner. So something callback-based seems to be the only solution, really. And this is indeed cleaner as well like you said.

Schmarni-Dev:
yeah i know about `Instance/Device::from_handle` (have done some very cursed stuff with that in the past) but having to use the normal vulkan/ash feature selection is so annoying that its one of the main reasons for me picking vulkano

at first i was thinking about exposing the ash conversion directly but i then realized how annoying the lifetimes of that would be, and the callbacks just fit pretty much perfectly imo

marc0246:
I added my own changes because conveying exactly what I wanted from this PR would be very inefficient, so to summarize:
- Ordering, naming, and formatting: you should make sure that when you open a PR, it matches the rest of the codebase. This includes the way functions are ordered, named, and formatted as well as the way comments (especially doc comments) are written and formatted. Contributions generally go smoother that way :slightly_smiling_face: That's just a heads up; it's not a big deal.
- I suffixed the functions with `_with` because that's the convention for Rust functions that take a closure in addition/instead of parameter(s). Prior art:
  - [`iter::repeat`](https://doc.rust-lang.org/std/iter/fn.repeat.html) vs [`iter::repeat_with`](https://doc.rust-lang.org/std/iter/fn.repeat_with.html)
  - [`Vec::resize`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize) vs [`Vec::resize_with`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize_with)
  - [`Option::get_or_insert`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert) vs [`Option::get_or_insert_with`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with)
  - [`Entry::or_insert`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert) vs [`Entry::or_insert_with`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with)
  - etc.
- I noticed you most likely used `&mut dyn FnMut` because of the compiler limitation that `use<...>` bounds must name all type parameters, so it would have to (wrongly) name an `impl FnOnce`. It's good that you wanted to avoid a breaking change; however, in this case, we wanted to get rid of the `impl ExactSizeIterator` return types anyway because they're an anti-pattern, so I replaced them with `Vec`s.
- I used `impl FnOnce` since it gives the caller the most freedom and has the best ergonomics. However, the other thing a `&mut dyn FnMut` provided is that it avoided parametric polymorphism (perhaps by accident). This is something I'm very anal about actually &ndash; I try to keep the amount of monomorphization in the codebase as low as possible while maintaining good ergonomics. Thankfully, it's always possible to manually monomorphize a generic function parameter, even an `impl FnOnce`. Unfortunately, it involves some unsafe in the latter case. This way, the old functions don't need monomorphization (like before) and the new ones only need the bare minimum.

So the only thing left to do is for you to test the changes with your use case, and unless there are any objections, I'll merge this.

Schmarni-Dev:
sorry it took me so long to respond. i somehow didn't know about the `_with` suffix convention, i agree that's a better name. i tested these changes with a minimal app of mine, i can't easily test with my main usecase as i want that to be based off of a release for now, but it should work just the same

let me know if i should resolve the merge conflicts

PR #2834, opened 2026-09-10
Fix `FreeListAllocator` ignoring buffer-image granularity against the next suballocation

angli66:
   Nothing to document: no API change, only which offsets the allocator hands out.

   `free_list_allocator_respects_granularity_backwards` fails on master and passes
   with the change. It needs no device.

5. Changelog:

```markdown
### Bugs fixed
- `FreeListAllocator` could place a suballocation on the same buffer-image granularity page as the suballocation following it, aliasing a linear resource with a non-linear one on devices where `bufferImageGranularity` is greater than 1.
```

6. Description

`FreeListAllocator::allocate` aligns the candidate offset up when the suballocation
*before* a free node holds a conflicting allocation type on the same granularity
page, and then commits the node as soon as the suballocation fits. It never looks at
the suballocation *after* the node, so a non-linear suballocation can be placed in a
hole whose upper neighbour is linear and on the same page, and the other way round.
That is the aliasing `VkPhysicalDeviceLimits::bufferImageGranularity` forbids.

Allocating strictly forward never reaches this, which is probably why it has gone
unnoticed: the hole has to be opened by a `deallocate` below an existing
suballocation first. We ran into it in a test suite that builds and drops resources
from several threads on one device.

What it does on hardware: on an RTX 4090 (`bufferImageGranularity` is 1024 there,
driver 595.84) an 8x8 optimal-tiled depth image was placed at the start of a page
whose upper half held the beginning of an acceleration-structure buffer. Writing the
image overwrote the first 128 bytes of the acceleration structure, and tracing
against the corrupted structure either lost the geometry for a frame or wandered
until the driver killed the context with a context-switch timeout (`Xid 109`). The
suite hung in 5 runs out of 100; with this change it ran 300 times with no hang and
no failure.

The fix mirrors the existing `prev` check. Once the suballocation is known to fit,
compare it against the node's `next` and move on to the next free node if the two
would share a page with conflicting types. Skipping the node is enough: `offset` is
already the lowest the node can give, and moving it up only brings the end closer to
the neighbour, so no position in that node avoids the conflict.

The other two suballocators are not affected. `BumpAllocator` only ever appends, so
it can only meet a neighbour below, which it already checks. `BuddyAllocator` rounds
non-linear allocations up to whole granularity pages, so nothing can share a page
with one.


marc0246:
Thank you so much for finding this and fixing it!

PR #2837, opened 2026-09-10
Simplify the C string -> `String` conversions

marc0246:
The way we converted the strings we get from Vulkan was less than ideal because the code was:
- repeated over and over, which this PR addresses by adding a utility function;
- manually calculating the length when `CStr` exists for that purpose, and better yet, ash has getters that do both the cast and return a `&CStr`;
- checking for UTF-8 validity when Vulkan guarantees UTF-8 strings.


AustinJ235 (commented):
I am not sure about this one. Though vulkan may guarantee UTF-8 strings, what happens when driver returns a malformed string? Feel like this may be outside of the bounds of where unchecked methods should be used.

marc0246:
There are thousands of places where if the driver does something slightly wrong, much worse UB happens. We could make an effort to mitigate that here, but I don't see the point when we have always been and always will be at the mercy of the driver for soundess.

AustinJ235 (approved):
Fair point. There is always the `debug_assert!`.

PR #2841, opened 2026-09-13
Refactor vulkano-shaders

### LINKED ISSUE #2822: `vulkano_shaders::shader!` hardcodes "main" as the entry point
### Version of vulkano

master

### OS (incl. version)

NixOS 26.11 (Zokor) x86_64

### Selected physical device

Radeon RX 7900 XT

### Driver (incl. version)

Mesa 26.1.2

### Issue description

It seems that the slang shader macro hardcodes the entry point as `"main"`.
The language doesn't seem to matter.
This is caused by [this line](https://github.com/vulkano-rs/vulkano/blob/master/vulkano-shaders/src/codegen.rs#L426)

### Minimal reproducible example

I loaded a simple vertex shader with `vulkano_shaders::shader!`:
```rust
mod vs {
    vulkano_shaders::shader! {
        ty: "vertex",
        lang: "slang",
        path: "shaders/vertex.slang"
    }
}
```
The contents of it are as follows:
```slang
struct VSInput
{
    float3 position;
    float2 uv;
};

struct VSOutput
{
    float4 position : SV_Position;
    float2 uv;
};

VSOutput vertexMain(VSInput input)
{
    VSOutput output;
    output.uv = input.uv;
    output.position = float4(input.position.xyz, 1.0);
    return output;
}
```
I tried setting it's entry point to something other than "main":
```rust
let vs = unsafe { vs::load(&app.device) }
    .unwrap()
    .entry_point("vertexMain")
    .unwrap();
```
And got this error:
```
error: slangc failed:
       error[E30015]: undefined identifier
       error[E38000]: no function found matching entry point name 'main'
        --> <stdin>:1:1
         |
       1 | struct VSInput
         | ^^^^^^ no function found matching entry point name 'main'
       --'

   --> src/main.rs:723:15
    |
723 |         path: "shaders/vertex.slang"
    |               ^^^^^^^^^^^^^^^^^^^^^^
```

marc0246:
I just wanted to implement the `entry_point` macro option, but the amount of cruft that vulkano-shaders has built up over the years is so bad that it's neigh impossible to make a change at this point, so here I am refactoring almost everything.

Here are some highlights:
- Use of empty strings as sentinel values. The worst of this was the hash map, where an empty string key meant that that's the only entry. Since what's being represented is either just one value or a map of key-value pairs, the ideal solution is an enum of precisely those two.
- Other data that wasn't represented well, like `SourceKind`, and `MacroInput` not being split into two.
- State being stored in many different stack variables, making it very hard to split things into smaller functions.
- Functions being too big to understand.
- Functions being split into functions where they shouldn't have been.
- Code duplication.
- Inconsistency.

Apart from that, I also made the diagnostics a bit more helpful and consistent, and improved the docs. In particular, the code in the headings made links to those headings really bad, making linking to other parts of the same docs unpleasant. Also, the new docs have more room to properly show the syntax of each option.

I suggest reviewing one commit at a time.

Fixes #2822

Changelog:
```markdown
### Additions
- Vulkano-shaders: Added a `compiler` macro option, which allows you to specify the shader compiler to use.
- Vulkano-shaders: Added an `entry_point` macro option, which allows you to specify the entry point of the shader source to compile.
```


vulkano-shaders/src/lib.rs
```diff
@@ -450,111 +781,815 @@ fn root_path_from_env_var(root_path_env: &LitStr) -> Result<(PathBuf, String)> {
...
-        f.write_str(<&str>::from(*self))
+    #[cfg(target_endian = "little")]
+    if bytes.as_ptr().addr().is_multiple_of(align_of::<u32>()) {
+        let mut bytes = ManuallyDrop::new(bytes);
+        let len = bytes.len() / size_of::<u32>();
+        let cap = bytes.capacity();
+        let ptr = bytes.as_mut_ptr().cast::<u32>();
+
+        // SAFETY: We checked that the pointer is 4-byte aligned and that the size divides evenly.
+        return Ok(unsafe { Vec::from_raw_parts(ptr, len, cap) });
```
AustinJ235: I think this may be UB.

This works for slices, but since `Vec` takes ownership, it deallocate it with the wrong layout. `Vec` docs state:

>`T` needs to have the same alignment as what `ptr` was allocated with. (`T` having a less strict alignment is not sufficient, the alignment really needs to be equal to satisfy the `dealloc` requirement that memory must be allocated and deallocated with the same layout.)

vulkano-shaders/src/lib.rs
```diff
@@ -52,8 +41,8 @@
 //!   each shader. They are named based on the provided names, `load_first`, `load_second` etc.
 //!   `try_` and `_unchecked` variants are also generated for each shader.
 //! - A Rust struct translated from each struct contained in the shader data. By default, each
-//!   structure has a `Clone` and a `Copy` implementation. This behavior could be customized
-//!   through the `custom_derives` macro option (see below for details). Each struct also has an
+//!   structure has a `Clone` and a `Copy` implementation. This behavior an be customized through
```
AustinJ235:
```suggestion
//!   structure has a `Clone` and a `Copy` implementation. This behavior can be customized through
```

vulkano-shaders/src/lib.rs
```diff
@@ -450,111 +781,815 @@ fn root_path_from_env_var(root_path_env: &LitStr) -> Result<(PathBuf, String)> {
...
+        let lit = input.parse::<LitStr>()?;
+
+        if self.global_compiler.is_some() {
+            bail!(lit, "field `compiler` is already defined");
+        }
+
+        self.global_compiler = Some(match lit.value().as_str() {
+            "shaderc" => Compiler::Shaderc,
+            "slangc" => Compiler::Slangc,
+            compiler => bail!(lit, "expected `shaderc` or `slang`, found `{compiler}`"),
```
AustinJ235:
```suggestion
            compiler => bail!(lit, "expected `shaderc` or `slangc, found `{compiler}`"),
```

vulkano-shaders/src/lib.rs
```diff
@@ -450,111 +781,815 @@ fn root_path_from_env_var(root_path_env: &LitStr) -> Result<(PathBuf, String)> {
...
+            bail!(
+                lit,
+                "fields `compiler` and `bytes` cannot be defined in the same shader entry",
+            );
+        }
+
+        self.compiler = Some(match lit.value().as_str() {
+            "shaderc" => Compiler::Shaderc,
+            "slangc" => Compiler::Slangc,
+            compiler => bail!(lit, "expected `shaderc` or `slang`, found `{compiler}`"),
```
AustinJ235:
```suggestion
            compiler => bail!(lit, "expected `shaderc` or `slangc`, found `{compiler}`"),
```

AustinJ235 (changes requested):

vulkano-shaders/src/lib.rs
marc0246: Oh Jesus, I must have been extremely sleep deprived. Good thing you caught it.

AustinJ235 (approved):

PR #2844, opened 2026-09-20
Add a `TlsfAllocator`

marc0246:
I should have implemented this allocator years ago, but I kept putting it off. Better late than never.

This allocator is really cute because there are only 2 unpredictable branches in both the allocate and deallocate path, everything else being either the definition of a predictable branch or a conditional move. And there are no loops, making the time complexity *O*(1). Even better, it has almost ideal fragmentation, almost as good as `FreeListAllocator`'s.

I went all the way making sure I benchmarked this as scientifically as possible: a pre-generated large synthetic workload, not just a micro-benchmark, to really stress the performance and fragmentation, and a pinned thread with maximum priority to take the OS's thread scheduler out of the equation. Each allocator runs the same synthetic workload, and the minimum, average and maximum times per iteration are gathered, as well as the maximum region memory that it used up, which is a measure of fragmentation. I chose to only use `Suballocator::allocate_buffer` because that's where performance really matters in the end, and to give `FreeListAllocator` a fighting chance. The results:

```
                             min time            avg time            max time          max memory
TlsfAllocator        14.600401ms/iter    16.045816ms/iter    23.504215ms/iter             5542MiB
FreeListAllocator   229.613494ms/iter   234.086887ms/iter   347.655521ms/iter             5389MiB
BuddyAllocator       35.166483ms/iter    36.145468ms/iter    75.032304ms/iter             7145MiB
```
The exact numbers vary based on the specific workload that was (randomly) generated, but the trends are always the same:
- `TlsfAllocator` outperforms the other allocators across the board;
- `TlsfAllocator` has only a 3% worse fragmentation than `FreeListAllocator`;
- `BuddyAllocator` has a 30% worse fragmentation than `FreeListAllocator`.

All of which is exactly as expected. What isn't expected is that `FreeListAllocator` is *that* much slower than the rest, which seems extremely fixable.

<details>
<summary>Program used to run the benchmark</summary>
Needs to be run with priviledge because the program 100%s CPU core 0 at all times, essentially having it all to itself.

```toml
[dependencies]
core_affinity = "0.8"
rand = "0.8"
thread-priority = "3.0"
```

```rs
use rand::Rng;
use std::{
    cmp,
    time::{Duration, Instant},
};
use vulkano::{
    memory::allocator::{
        suballocator::{Region, TlsfAllocator},
        AllocationHandle, AllocationType, BuddyAllocator, DeviceLayout, FreeListAllocator,
        Suballocation, Suballocator,
    },
    DeviceSize,
};

const ITERATIONS: usize = 10_000;
const OPERATIONS: usize = 500_000;
const REGION_SIZE: DeviceSize = 1 << 35;
const MIN_ALLOCATION_SIZE: DeviceSize = 16;
const MAX_ALLOCATION_SIZE: DeviceSize = 1024 * 1024;

enum Operation {
    Allocate { index: u32, size: DeviceSize },
    Deallocate { index: u32 },
}

#[rustfmt::skip]
fn main() {
    let core_ids = core_affinity::get_core_ids().unwrap();
    assert!(core_affinity::set_for_current(core_ids[0]));

    thread_priority::set_current_thread_priority(thread_priority::ThreadPriority::Max).unwrap();

    let workload = generate_workload();
    println!("                             min time            avg time            max time          max memory");
    let (min_time, avg_time, max_time, max_memory) = bench::<TlsfAllocator>(&workload);
    println!("TlsfAllocator    {min_time:>15?}/iter{avg_time:>15?}/iter{max_time:>15?}/iter    {max_memory:>16}");
    let (min_time, avg_time, max_time, max_memory) = bench::<FreeListAllocator>(&workload);
    println!("FreeListAllocator{min_time:>15?}/iter{avg_time:>15?}/iter{max_time:>15?}/iter    {max_memory:>16}");
    let (min_time, avg_time, max_time, max_memory) = bench::<BuddyAllocator>(&workload);
    println!("BuddyAllocator   {min_time:>15?}/iter{avg_time:>15?}/iter{max_time:>15?}/iter    {max_memory:>16}");
}

fn bench<S: Suballocator>(
    workload: &(Vec<Operation>, usize),
) -> (Duration, Duration, Duration, String) {
    let (operations, max_simultaneous_allocations) = workload;
    let dummy_suballocation = Suballocation {
        offset: 0,
        size: 0,
        allocation_type: AllocationType::Unknown,
        handle: AllocationHandle::null(),
    };
    let mut allocations = vec![dummy_suballocation; *max_simultaneous_allocations];
    let mut min_time = Duration::MAX;
    let mut avg_time = Duration::ZERO;
    let mut max_time = Duration::ZERO;
    let mut max_memory = 0;

    for _ in 0..ITERATIONS {
        let mut allocator = S::new(Region::new(0, REGION_SIZE).unwrap());
        let now = Instant::now();

        for operation in operations {
            match operation {
                Operation::Allocate { index, size } => {
                    let allocation = unsafe { allocations.get_unchecked_mut(*index as usize) };
                    let layout = unsafe { DeviceLayout::from_size_alignment_unchecked(*size, 1) };
                    *allocation = allocator.allocate_buffer(layout).unwrap();
                    max_memory = cmp::max(max_memory, allocation.offset + allocation.size);
                }
                Operation::Deallocate { index } => {
                    let allocation = *unsafe { allocations.get_unchecked(*index as usize) };
                    unsafe { allocator.deallocate(allocation) };
                }
            }
        }

        let elapsed = now.elapsed();
        min_time = cmp::min(min_time, elapsed);
        avg_time += elapsed;
        max_time = cmp::max(max_time, elapsed);
    }

    let avg_time = avg_time / ITERATIONS as u32;
    let max_memory = format!("{}MiB", max_memory / 1024 / 1024);

    (min_time, avg_time, max_time, max_memory)
}

fn generate_workload() -> (Vec<Operation>, usize) {
    let mut allocated = Vec::new();
    let mut allocated_indices = Vec::new();
    let mut rng = rand::thread_rng();
    let mut operations = Vec::with_capacity(OPERATIONS);
    let mut operation_count = 0;

    'outer: loop {
        let count_upper_bound_magnitude = rng.gen_range(0usize..=8);
        let count_upper_bound = 1usize << count_upper_bound_magnitude;
        let count = rng.gen_range(1..=count_upper_bound);
        let allocate = rng.r#gen::<bool>();

        for _ in 0..count {
            if allocate {
                let index = if let Some(index) = allocated.iter().position(|b| !b) {
                    index
                } else {
                    let index = allocated.len();
                    allocated.push(false);

                    index
                }
                .try_into()
                .unwrap();
                allocated[index as usize] = true;
                allocated_indices.push(index);
                let size = rng.gen_range(MIN_ALLOCATION_SIZE..=MAX_ALLOCATION_SIZE);
                operations.push(Operation::Allocate { index, size });
            } else {
                if allocated_indices.is_empty() {
                    break;
                }

                let indices_index = rng.gen_range(0..allocated_indices.len());
                let index = allocated_indices.remove(indices_index);
                allocated[index as usize] = false;
                operations.push(Operation::Deallocate { index });
            }

            operation_count += 1;

            if operation_count == OPERATIONS {
                break 'outer;
            }
        }
    }

    (operations, allocated.len())
}
```

</details>

I also rewrote the docs of the other suballocators, discouraging their use more, and talking about fragmentation in absolute terms. Maybe this addresses @LukeP0WERS's concerns about the docs.

Changelog:
```markdown
### Additions
- Added `TlsfAllocator`, the new go-to suballocator.
```


vulkano/src/memory/allocator/suballocator/buddy.rs
```diff
@@ -22,15 +22,15 @@ use std::{
...
-/// fragmentation completely.
+/// You generally shouldn't use this allocator. It has some of the worst [fragmentation] overall,
+/// worse than both [`TlsfAllocator`] and [`FreeListAllocator`]. This is because it can only
+/// allocate sizes that are powers of two. For example, if you need an allocation size of 70MiB,
+/// you would use a whole 128MiB instead, wasting 45% of the memory. It is recommended that you use
+/// `TlsfAllocator` instead. However, this allocator does have its niche: when all your allocations
+/// are dynamically- but power-of-two-sized, then there is no [internal fragmentation], and since
+/// this algorithm somewhat reduces [external fragmentation], the overall fragmentation can beat
+/// `TlsfAllocator` and `FreeListAllocator`. The algorithm is also fast in practice. Please make
+/// sure to test and profile to make sure that this is actually better.
```
CptPotato: nit: "make sure to ... to make sure" sounds a bit odd

vulkano/src/memory/allocator/suballocator/tlsf.rs
```diff
@@ -0,0 +1,1110 @@
...
+#[derive(Debug)]
+pub struct TlsfAllocator {
+    first_level: FirstLevel,
+    free_size: DeviceSize,
+}
+
+unsafe impl Suballocator for TlsfAllocator {
+    type Suballocations<'a> = Suballocations<'a>;
+
+    /// Creates a new `BuddyAllocator` for the given [region].
```
CptPotato: This was taken from `BuddyAllocator` and needs to be updated.

CptPotato (commented):
I was curious and read through the docs. It's an interesting read.

Found two nits, but otherwise it's very well documented.

vulkano/src/memory/allocator/suballocator/free_list.rs
```diff
@@ -13,14 +13,9 @@ use std::{cmp, hint, iter::FusedIterator, marker::PhantomData, ptr::NonNull};
...
 ///
-/// The strength of this allocator is that it can create and free allocations completely
-/// dynamically, which means they can be any size and created/freed in any order. The downside is
-/// that this always leads to horrific [external fragmentation] the more such dynamic allocations
-/// are made. Therefore, this allocator is best suited for long-lived allocations. If you need
-/// to create allocations of various sizes, but can't afford this fragmentation, then the
-/// [`BuddyAllocator`] is your best buddy. If you need to create allocations which share a similar
-/// size, consider an allocation pool. Lastly, if you need to allocate very often, then
-/// [`BumpAllocator`] is best suited.
+/// You generally shouldn't use this allocator as [`TlsfAllocator`] is better in almost every way.
```
LukeP0WERS: Not saying I disagree with this since I'm might not know all the benefits, but my understanding is that `TlsfAllocator` achieves significantly better performance at the cost of a negligible increase in fragmentation. Reading "better in almost every way" would imply to me that it achieves better memory usage, and although it is explained in more detail below someone working quickly might not read the text below. I would phrase it as "better for most applications" since that conveys the opinion without being contestable.

